PR #140, opened 2014-04-10
Syntax: object literal

hhugo:
imported from https://github.com/astrada/pa_extjs cc @astrada
for the upcoming release or later

example 
```
let obj = {:
  p1 = Js._true;
  p2 = Js.string "string property";
:}
```
is infered to 
```
< p1 : < set : bool Js.t -> unit; .. > Js.gen_prop;
  p2 : < set : Js.js_string Js.t -> unit; .. > Js.gen_prop;
  .. >  Js.t
```

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -101,7 +101,7 @@ js_parser.mli: js_parser.mly javascript.cmi js_token.cmi
 	menhir --infer --external-tokens Js_token --explain $<
 
 %.ml: %.mll
-	ocamllex $<
+	ocamllex -q $<
 clean:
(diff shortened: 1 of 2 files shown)


hhugo:
I think it make sense to add the syntax `jsobject end`.
no sure about `{: x = 3 :}` one ? any thought @Drup, @vouillon

Drup:
Either we drop it, or we add something based on the record syntax in the ppx. I don't have ideas to make this lightweight in ppx, though.

hhugo:
I've kept `jsobject` syntax only.

Drup:
I'm afraid to tell you that you will have to change it. I changed the generated code for object literals. ;)

hhugo:
I'll do :)

PR #180, opened 2014-06-29
Basic nat.js implementation

percontation:
I was doing a js_of_ocaml thing that ended up needing a simple nat.js implementation (part of ocaml's nums.cma).
It's not high quality; my only claim is that it worked for me :3. Feel free to take it if you'd like.

[I probably missed some Makefiles that need to be changed, and someone who knows actually knows ocaml/js_of_ocaml should find a proper fix to my Obj.size hack.]


hhugo:
The tests doesn't seems to pass (anymore?)
https://github.com/ocsigen/js_of_ocaml/issues/320

PR #239, opened 2014-10-18
Ppx for js_of_ocaml

Drup:
Should not be merged, the syntax is not the final one (waiting for ocaml here).

It's feature-par with the camlp4 one, plus object literals with optional fields.

 - `obj##.field` for field access
 - `obj##.field := x` for field modification
 - `obj##meth arg1 arg2 arg3` for method calls
 - `obj##meth` for method calls without arguments
 - `new%js constructor` for prototype instanciation.
 - `object%js ... end` for literal objects. `[@@optional]` to make the field optional.

What remains to do:

 - Automatic generation of literal object's type (especially wrt to optional fields).
 - (DONE) Use the mutable annotation in object literals to choose between `readonly_prop` and `prop`.
 - Clean up and factorize the code some more.
 - Document.
 - Test error locations.
 - Solve the "fresh name" issue. I'm not fond of the fixed seed prng solution.

cc @whitequark @alainfrisch

### DIFF
--- a/lib/ppx/ppx_js.cppo.ml
+++ b/lib/ppx/ppx_js.cppo.ml
@@ -185,11 +185,9 @@ let format_meth self_id body =
     ] [@metaloc body.pexp_loc]
   | _ -> body
 
-let is_optional attrs =
-  List.exists (fun ({txt},_) -> txt = "optional") attrs
 
-let preprocess_literal_object ?(optional=false) self_id fields =
-  let is_opt e = is_optional e.pcf_attributes in
+
+let preprocess_literal_object self_id fields =
 
   let check_name id names =
     if S.mem id.txt names then
@@ -204,10 +202,10 @@ let preprocess_literal_object ?(optional=false) self_id fields =
   let f (names, fields) exp = match exp.pcf_desc with
     | Pcf_val (id, mut, Cfk_concrete (bang, body)) ->
       let names = check_name id names in
-      names, (`Val (id, mut, bang, body, optional || is_opt exp) :: fields)
+      names, (`Val (id, mut, bang, body) :: fields)
     | Pcf_method (id, priv, Cfk_concrete (bang, body)) ->
       let names = check_name id names in
-      names, (`Meth (id, priv, bang, format_meth self_id body, optional || is_opt exp) :: fields)
+      names, (`Meth (id, priv, bang, format_meth self_id body) :: fields)
     | _ ->
       Location.raise_errorf ~loc:exp.pcf_loc
         "This field is not valid inside a js literal object."
@@ -223,10 +221,10 @@ let literal_object ?loc fields =
 
   let fields =
     List.map (function
-      | `Val (n, mut, bang, body, is_opt) ->
+      | `Val (n, mut, bang, body) ->
         let ty = fresh_type n.loc in
-        `Val (n, mut, bang, body, ty, is_opt)
-      | `Meth (n, priv, bang, body, is_opt) ->
+        `Val (n, mut, bang, body, ty)
+      | `Meth (n, priv, bang, body) ->
         let rec create_meth_ty exp = match exp.pexp_desc with
           | Pexp_fun (label,_,_,body) ->
             (label, fresh_type exp.pexp_loc) :: create_meth_ty body
@@ -234,16 +232,16 @@ let literal_object ?loc fields =
         in
         let ret_ty = fresh_type body.pexp_loc in
         let fun_ty = create_meth_ty body in
-        `Meth (n, priv, bang, body, (fun_ty, ret_ty), is_opt)
+        `Meth (n, priv, bang, body, (fun_ty, ret_ty))
     )
       fields
   in
 
   let set_field = function
-    | `Val (id, _, _, body, ty, _) ->
+    | `Val (id, _, _, body, ty) ->
       Js.unsafe "set" [ evar obj_lid ; Js.string id.txt ; Exp.constraint_ body ty ]
         [@metaloc body.pexp_loc]
-    | `Meth (id, _, _, body, (fun_ty, ret_ty), _) ->
+    | `Meth (id, _, _, body, (fun_ty, ret_ty)) ->
       Js.unsafe "set" [
         evar obj_lid ;
         Js.string id.txt ;
@@ -263,31 +261,19 @@ let literal_object ?loc fields =
     List.fold_right
       (fun field e -> match field with
          | `Val _ -> e
-         | `Meth (n, _, _, body, ty, _) -> meth_def n body ty e)
+         | `Meth (n, _, _, body, ty) -> meth_def n body ty e)
       fields
       assignments in
 
   let obj_field_meth_type = function
-    | `Val  (id, _, _, _body, ty, _) ->
+    | `Val  (id, _, _, _body, ty) ->
       (id.txt, [], Js.type_ "prop" [ty])
-    | `Meth (id, _, _, _body, (fun_ty, ret_ty), _) ->
+    | `Meth (id, _, _, _body, (fun_ty, ret_ty)) ->
       (id.txt, [], arrows (List.tl fun_ty) (Js.type_ "meth" [ret_ty]))
   in
 
-  let ty_rows = List.map obj_field_meth_type fields in
-
-  let optional_part_ty = Typ.object_ ty_rows Open in
   let obj_ty =
-    (* We deliberately leave optional properties/methods out
-       so that we don't need to cast the object manually. *)
-    let l =
-      ("__optional__", [], optional_part_ty) ::
-      (List.map obj_field_meth_type
-        ( List.filter
-            (function `Val  (_, _, _, _, _, is_opt) |
-                      `Meth (_, _, _, _, _, is_opt) -> not is_opt)
-            fields )
-      )
+    let l = List.map obj_field_meth_type fields
     in Typ.object_ l Closed
   in
   [%expr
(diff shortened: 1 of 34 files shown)


hhugo:
why `##.` rather than the initial `#.` proposal  ?
@Drup can you update the description to fit the current state of the PR

Drup:
> why ##. rather than the initial #. proposal ?

`#.` is not a ppx reserved operator, `##.` is. I favor using ppx reserved operator when possible. In this case, the cost is literally insignificant.

The object literal part of the patch is bugged, I know it, so don't use it just yet. :)

I have yet to add the ghost locations (it helps merlin) and check scopes.

Drup:
hum, ok, cool !

Drup:
I was not happy about the object literal design (that was inherited from @mfp's one). The issue were related to the fact that it was cluttering the type with a rather big `__optional__` field. It didn't bring much in term of type safety and it was rather inconvenient to use. 

After discussion with @vouillon, we identified two different use cases for literate javascript objects:
 - Building an object to expose it to the javascript side, as a reversed FFI.
 - regular FFI: javascript libraries that take an object as argument to emulate optional arguments.

We decided it was best to provide two different solutions for these different problems:

 - A literal object tailored for reversed FFI (that appropriately wrap methods in meth_callback, and so on). It would not say anything about optional fields. It would look like that: 
```ocaml
object%js (self) (* Can expose ocaml-friendly equivalent of this *)
  var x = 3 (* readonly_prop *)
  var mutable y = 4 (* read/write prop *)
  meth foo = ... (* appropriately wrapped with meth_callback *)
end 
```

 - A deriver would provide the ability to define "object constructors": 
```ocaml
type t = < foo : int prop ; bar : int Js.optdef prop > Js.t
[@@deriving jsoo.constructor]
``` 
would define a function of type:
```ocaml
?bar:int -> ~foo:int -> unit -> t 
```

opinions ?

hhugo:
:+1: :+1: :+1:

Drup:
Ok, first part is done. It seems to work pretty well. @hhugo : can you check I'm not doing anything silly in the code gen ?

Also, this seems to work:
```ocaml
let res =
  object%js (self)
    val mutable foo = 3
    method bar x =
      object%js (self2)
        method boum = self ##. foo := x
      end
  end
```
:D

lib/ppx/ppx_js.cppo.ml
```diff
@@ -0,0 +1,386 @@
...
+
+  let obj_ty =
+    let l = List.map obj_field_meth_type fields
+    in Typ.object_ l Closed
+  in
+  [%expr
+    let ([%p pvar obj_lid] : [%t Typ.alias (Js.type_ "t" [obj_ty]) self_ty]) =
+      let [%p pvar constr_lid] : [%t Js.type_ "constr" [Typ.var self_ty]] =
+        [%e Js.unsafe "variable" [str "Object"]] in
+      let [%p pvar obj_lid] = [%e Js.unsafe "new_obj" [ evar constr_lid ; Exp.array []]] in
```
hhugo: The generated js will be smaller and easier to read if you use the form: 

```
Js.Unsafe.obj [| "x", Js.Unsafe.inject 10;"y", (Js.Unsafe.inject (wrap_meth_callback ..)) |]
```
It should generate
```
{ x : 5, y : wrap_meth_callback ..}
```
compared to
```
let o = new Object;
o.x = 5;
o.y = wrap_meth_callback  ..;
```

Drup:
You are looking at the old version, yes.

hhugo:
I've synced the camlp4 syntax https://github.com/ocsigen/js_of_ocaml/pull/140.
it currently allows 2 new syntax: 
```
let o1 = {: x = int ; y = string :}
let o2 =
  jsobject (self)
    val mutable foo = 3
    method bar x =
      jsobject (self2)
        method boum = self ## foo <- x
      end
  end
```

lib/ppx/ppx_js.cppo.ml
```diff
@@ -0,0 +1,412 @@
...
+
+  let create_value = function
+    | `Val (id, _, _, body, ty) ->
+      (id.txt, Exp.constraint_ body ty)
+      [@metaloc body.pexp_loc]
+
+    | `Meth (id, _, _, body, (fun_ty, ret_ty)) ->
+      (id.txt,
+       Exp.constraint_
+         (Js.fun_ "wrap_meth_callback" [ body ])
```
hhugo: Use `wrap_callback` when `self` is not bounded.

lib/ppx/ppx_js.cppo.ml
```diff
@@ -0,0 +1,412 @@
...
+open Asttypes
+open Parsetree
+
+open Ast_convenience
+
+(** Check if an expression is an identifier and returns it.
+    Raise a Location.error if it's not.
+*)
+let exp_to_string = function
+  | {pexp_desc= Pexp_ident {txt = Longident.Lident s}} -> s
```
hhugo:
```
| {pexp_desc= Pexp_construct ({txt = Longident.Lident s}, None)} -> s
```
so we can support 
```
let date_constr = Unsafe.global##.Date
```

hhugo:
note the following is not allowed
```
new%js (Js.Unsafe.global##.A)
``` 
and has to be replaced by
```
let a = Js.Unsafe.global##.A in
new%js a
```

lib/ppx/ppx_js.cppo.ml
hhugo: maybe not a good move after all.

Drup: It seemed like a good idea, what is the issue ?

Drup:
> note the following is not allowed
> ```
> new%js (Js.Unsafe.global##.A)
> ```

Indeed, I wonder if we could get the restriction relaxed in the parser.

lib/ppx/ppx_js.cppo.ml
hhugo: `let f x = x##A = 0`
has type
`val f : < A : int Js.meth; .. > Js.t -> 'jsoo_bcb1e9dd`.
But this type cannot be reread by ocaml because method name cannot start with capital letter

Drup: "oups" indeed. Maybe just for attributes ? it's probably not really worth it then.

hhugo: another solution is to make it have type
```
val f : < _A : int Js.meth; .. > Js.t -> bool.
```

hhugo:
@drup, do you want to add more to this PR ? I had a look and it seems fine.
From the initial todo: only remains documentation and test error locations.

you can check the result of rewriting js_of_ocaml lib with ppx https://github.com/ocsigen/js_of_ocaml/commit/ca4e43620046dd1d5b1f2d0bdb6d153a9e9e5b66

lib/ppx/ppx_js.cppo.ml
Drup: I think it's a good solution, since it will be compatible with the current convention out of the box.

Drup:
After another discussion with @vouillon, I modified the literal object code to allow better error reports. I also fixed a good amount of error locations and I removed all the generated names that were not necessary (replacing them with stuff like `jsoo_self`, `jsoo_ret` and so on). 

There are still some edge cases like
```
let r =
  object%js (self)
    method bar = self##baz 1.
    method baz x = x + 1
  end
```
but I know how to fix it, tomorrow. 
I have ghost locations to add in the relevant places too.

Does inheriting from an object make sens in javascript ?

Drup:
Also, @frisch proposed to allow method application like that : `obj#(meth arg)` to allow easy chaining `obj#(meth foo)#(meth bar`. I will implement that too.

Drup:
Do I add {js|...|js} for Js.string literals ? I don't know how useful it would be.

Drup:
@hhugo I would like to add tests where a file is typed (-i) with the ppx, and the output diffed with some expected output to check the error locations/messages. My makefile-fu is rather low, do you know how to do that cleanly ?

hhugo:
don't add `{js|...|js}`

hhugo:
https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/Makefile#L79

Drup:
Do I change `new%js` in `[%js.new]` ? It would allow non-identifier constructors, at the cost of a less nice syntax. are non-identifier constructor common enough ?

hhugo:
I would keep `new%js`. non-identifier constructor appear mostly with  `Js.Unsafe.global##`

Drup:
@hhugo I tried to replace `wrap_meth_callback` by `"wrap_callback` when self is not specified, but I don't know what to do if the method doesn't have any arguments. Currently it makes `self` (or `_` when not present) the only argument. Do I replace it by `()` ?

If so, it would mean I have to check if the method has no argument. I can do it but it makes the code more complicated, is it really important ?

hhugo:
It is not. I did not do it for camlp4 syntax because of this.

Drup:
Ok, then I'm done. The locations are tested and the documentation is written. :)

Drup:
@hhugo Ideally, I would like the last line of the ppx_js.cppo.ml to be in a separated file (that generates a binary) and export the `Ppx_js` module as a library, can you do the build system incantation for me ? :D

hhugo:
What is the use case for having a separate lib ? toplevel ?

Drup:
Selective loading of the syntax extension. Only in client sections, for example.

Drup:
might be useful for toplevel too, yes.

Drup:
I added some tests for the error locations and the type errors!
@vouillon I didn't save the file we did last time (it was in my /tmp ...) so I created a new set. Can you check if I forget something ?

hhugo:
@drup, I've added the lib for ppx_js and added tests for camlp4 syntax (but didn't commit the expected result yet).

Drup:
@hhugo I added a META entry.

Drup:
@hhugo I think you broke camlp4 on previous compiler version. At least CI complains a lot.

hhugo:
I'll fix that tomorrow. (it comes from my attempt to highlighting errors in js_of_ocaml toplevel)

PR #251, opened 2015-02-12
Refactor JavaScript parser

keleshev:
Changes made:
 * Factoring out operator definitions, instead of handling them inline:
    * `comparison_or_logical_or_bit_operator`
    * `comparison_or_logical_or_bit_operator_except_in`
    * `arithmetic_or_shift_operator`
    * `prefix_operator`
    * `postfix_operator`
 * Factoring out some productions like `parenthesised`
 * Removal of redundant ocamlyacc-style `/* */` comments
 * Using more conventional `*` and `?` operators instead of `list` and `option` productions
 * Renaming many single-letter variables
 * Using pattern-matching instead of `fst` and `snd` on tuples

### DIFF
--- a/compiler/js_parser.mly
+++ b/compiler/js_parser.mly
@@ -598,15 +598,15 @@ element_list_rev:
  |            assignment_expression { [Some $1] }
  | element_list_rev elison assignment_expression { (Some $3) :: (List.rev_append $2 $1) }
 
-separated_nonempty_list2(sep,X):
+separated_or_terminated_list(sep, X):
  | x=X { [x] }
  | x=X; sep { [x] }
- | x=X; sep; xs=separated_nonempty_list2(sep, X) { x :: xs }
+ | x=X; sep; xs=separated_or_terminated_list(sep, X) { x :: xs }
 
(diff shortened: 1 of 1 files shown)


vouillon:
Thanks a lot! I'll have a look as soon as possible.

hhugo:
What about using `%inline` for `postfix_operator` and `assignment_operator` ?

keleshev:
@hhugo [Menhir reference](http://cristal.inria.fr/~fpottier/menhir/manual.pdf) describes why you need to `%inline` infix operators with different precedence. So it is necessary for `comparison_or_logical_or_bit_operator` (which have different precedence), but not necessary for `assignment_operator` which have same (unspecified) precedence. The same goes for prefix/postfix operators.

In the end, my guiding principle for `%inline`ing is the following. When I factor out one production from another, I check if number of shift/reduce conflicts changed:
* if number of conflicts grows—that means this transformation introduced more ambiguity, so I `%inline` it (which effectively undoes the transformation), or
* if number of conflicts stays same—I do not `%inline` it.

PR #256, opened 2015-03-07
Add support for 'blob' and 'arraybuffer' response types for XHRs.

puppeh:
This is meant more as an "RFC" than as code that I consider ready to be merged!

I just started playing with js_of_ocaml, and currently only text (or XML) XMLHttpRequests seemed to be supported, but I wanted to load some binary data from the server, which appears to invoke character-encoding demons. So this is an attempt to implement (what I understand to be!) the javascript-y way of doing that: setting the responseType for an XMLHttpRequest to "blob" or "arraybuffer". The new APIs I added are kind of grotty and incomplete, but seem to work (...for one simple test anyway), and ought to be mildly backwards-compatible, I think.

### DIFF
--- a/lib/xmlHttpRequest.ml
+++ b/lib/xmlHttpRequest.ml
@@ -23,6 +23,8 @@ open Js
 
 type readyState = UNSENT | OPENED | HEADERS_RECEIVED | LOADING | DONE
 
+type xmlHttpRequestResponseType = ArrayBuffer | Blob | Document | JSON | Text
+
 class type xmlHttpRequest = object ('self)
   method onreadystatechange : (unit -> unit) Js.callback Js.writeonly_prop
   method readyState : readyState readonly_prop
@@ -41,8 +43,10 @@ class type xmlHttpRequest = object ('self)
   method statusText : js_string t readonly_prop
   method getResponseHeader : js_string t -> js_string t opt meth
   method getAllResponseHeaders : js_string t meth
+  method response : File.file_any readonly_prop
   method responseText : js_string t readonly_prop
   method responseXML : Dom.element Dom.document t opt readonly_prop
+  method responseType : js_string t prop
 
   inherit File.progressEventTarget
   method ontimeout : ('self t, 'self File.progressEvent t) Dom.event_listener writeonly_prop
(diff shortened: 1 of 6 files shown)


Drup:
One way you could improve the API is to turn `xmlHttpRequestResponseType` into a GADT parametrize by the type of the answer. Something along the lines of:

```ocaml
type _ response = 
  | ArrayBuffer : Typed_array.arrayBuffer t response 
  | Blob : #File.blob t response
  ...
``` 

and instead of the various ` perform_raw_url_*` functions, you have one, with a mandatory `response_type` argument (no more optional). The name could be simply `perform_raw`. The signature would be:
```ocaml
val perform_raw :
 ...
-> ~response_type:('a response)
-> string
-> 'a http_frame Lwt.t
```

Note that the argument can't be optional here (OCaml type system doesn't handle "optional polymorphism").

You still keep `perform_raw_url` specialized on json for backward compatibility (and maybe add a deprecation warning, that remains to be decided).

puppeh:
Thanks, yeah, I was wondering if something like that might be possible. I'll look into it some time soonish!

Cheers,

Julian

puppeh:
I've had a go at implementing Drup's idea, and it seems to work OK (thanks to ggole & companion_cube on #ocaml for GADT hints!). So this might be getting closer to being useful -- though there's still no documentation for the new API, nor am I quite sure that an (OCaml) 'string' response for each of Document, JSON and Text is the most sensible type to use (someone more familiar with Javascript would have to say).

HTH!

lib/xmlHttpRequest.ml
```diff
@@ -149,7 +158,47 @@ let extract_get_param url =
...
+  }
+
+let blob_response url code headers req =
+  { url = url;
+    code = code;
+    content =
+      begin match Js.Opt.to_option (File.CoerceTo.blob (req##response)) with
+        None -> assert false
+      | Some blob -> blob
+      end;
```
Drup: This can be simplified with `Js.Opt.get`.

lib/xmlHttpRequest.ml
```diff
@@ -149,7 +158,47 @@ let extract_get_param url =
...
+    content_xml = (fun () -> assert false);
```
Drup: I think returning `None` is better.

Drup:
Seems nice to me. @vouillon ?

puppeh:
Here's some more bits -- more sensible types for JSON, Document and Text responses and use of JavaScript option types for returned values (in case of type mismatches, errors, etc.).

Thanks,

Julian

vouillon:
Thanks!

PR #260, opened 2015-03-17
Batch4

hhugo:


Drup:
Yay for cppo !

Drup:
I think you forgot the rule for META.

PR #271, opened 2015-04-14
Bring float_of_string more in line with the OCaml implementation.

jonludlam:
This adds the ability to parse infinities of various capitalisation
and sign. The following now works in the JS toplevel:

    float_of_string (string_of_float infinity)

~~It's more permissive than standard ocaml, accepting, for example,
"infini" whereas that would be rejected by the normal toplevel.~~

Signed-off-by: Jon Ludlam <jonathan.ludlam@citrix.com>

### DIFF
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -454,6 +454,8 @@ function caml_float_of_string(s) {
     res = +s.substring(0,pidx);
     return res * Math.pow(2,exp);
   }
+  if(/^\+?inf(inity)?$/i.test(s)) return Infinity;
+  if(/^-inf(inity)?$/i.test(s)) return -Infinity;
   caml_failwith("float_of_string");
 }
 


runtime/stdlib.js
```diff
@@ -454,6 +454,8 @@ function caml_float_of_string(s) {
     res = +s.substring(0,pidx);
     return res * Math.pow(2,exp);
   }
+  if(/^[+]?inf/i.test(s)) return Infinity;
+  if(/-inf/i.test(s)) return -Infinity;
```
hhugo: should probably `/^-inf/`

hhugo:
I would rather like something less permissive. like `/^-inf(inity)?$/` and `/^inf(inity)?$/`

jonludlam:
that looks good to me. I'll update the PR.

jonludlam:
I've also got rid of the redundant square braces around the plus.

hhugo:
Thanks for the fix.

Issue #278, opened 2015-04-22
Identity function not inlined


alainfrisch:
I don't know much about js_of_ocaml inlining strategy, but it seems it doesn't even inline the identity function if used several times:

````
type t
external set: t -> string -> t -> unit = "caml_js_set"
external variable: string -> t = "caml_js_var"

let (id :t -> t)= fun x1  -> x1

let () =
  let elt = variable "foo" in
  set (id elt) "xxxx" elt;
  set (id elt) "xxxx" elt
````

produces:

````
    function _aR_(_a_){return _a_}
    var _N_=foo;
    _aR_(_N_).xxxx=_N_;
    _aR_(_N_).xxxx=_N_;
````

Removing one of the two calls to id has the effect of removing the identity function in JS.

Is this behavior expected?

vouillon:
Indeed, at the moment, the compiler only inlines functions that are used once, as this is easy and result clearly in a gain in size.

PR #280, opened 2015-04-23
Inline identity function.

alainfrisch:
Fix https://github.com/ocsigen/js_of_ocaml/issues/278 .


alainfrisch:
We might want to do something to preserve tail-call (if the call to the identity is in tail position).

hhugo:
thanks. I've added code to preserve tail-call.

PR #283, opened 2015-04-27
Also inline simple functions which simply call another function or primitive

alainfrisch:
This eliminates some simple calls to functions which immediately call another function or primitive.

It also adds some very naive logic to short-circuit calls to function which jump to another function.  This helps the inliner apply the elimination mentioned above.

### DIFF
--- a/compiler/inline.ml
+++ b/compiler/inline.ml
@@ -141,7 +141,7 @@ let simple blocks clos_pc clos_args clos_params f_args =
     begin match exp with
       | Const _ -> `Exp exp
       | Constant (Float _ | Int64 _ | Int _ | IString _) -> `Exp exp
-      | Apply (f, args, true) -> `Exp (Apply (f, List.map map_var args, true))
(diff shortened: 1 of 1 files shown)


compiler/inline.ml
```diff
@@ -114,23 +124,31 @@ let rec find_mapping x src trg =
...
+  in
+  try match clos with
+  | {handler = _; body = []; branch = Return ret} ->
+      `Alias (map_var ret)
+  | {handler = _; body = [Let (x, exp)]; branch = Return ret}
+    when Code.Var.compare ret x = 0 ->
     begin match exp with
       | Const _ -> `Exp exp
       | Constant (Float _ | Int64 _ | Int _ | IString _) -> `Exp exp
+      | Apply (f, args, true) -> `Exp (Apply (f, List.map map_var args, true))
```
vouillon: That should be `map_var f`, not just `f`.

alainfrisch: Thanks, fixed!

PR #287, opened 2015-05-01
Add an optional parameter to set the method for a XHR request.

slegrand45:
Useful for PUT or DELETE request.

### DIFF
--- a/lib/xmlHttpRequest.mli
+++ b/lib/xmlHttpRequest.mli
@@ -110,7 +110,7 @@ val perform_raw :
   -> ?progress:(int -> int -> unit)
   -> ?upload_progress:(int -> int -> unit)
   -> ?override_mime_type:string
-  -> ?override_method:string
+  -> ?override_method:[< `Get | `Post | `Head | `Put | `Delete | `Options | `Patch > `Get `Post]
   -> response_type:('a response)
   -> string
   -> 'a generic_http_frame Lwt.t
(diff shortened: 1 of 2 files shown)


Drup:
Why a string ? It should clearly be a sum type or at least a polymorphic variant.

What should be the behavior on invalid combinations? (GET + post parameters, for example)

slegrand45:
For the first point, i do completely agree. I propose a subset of the methods defined in Ocsigen (in http/ocsigen_http_frame.mli): 
```ocaml
  type http_method = GET | POST | HEAD | PUT | DELETE | OPTIONS | PATCH
```

Can i also then modify the _open method? Its first parameter is the http method and its type is "js_string t":
```ocaml
  method _open : js_string t -> js_string t -> bool t -> unit meth
```
so it would become:
```ocaml
  method _open : http_method -> js_string t -> bool t -> unit meth
```

For the second point, i would simply add a note in the documentation to specify that if the request method is overriden and there is some post_args or form_arg, those data will be in the request body. No matter the overriden method would be.

In fact, RFC 7231 (http://tools.ietf.org/html/rfc7231, section 4.3) seems to explicitely forbid the message body only for the TRACE method. For the GET, HEAD and DELETE methods the RFC says:
> "A payload within a GET [HEAD/DELETE] request message has no defined semantics; sending a payload body on a GET [HEAD/DELETE] request might cause some existing implementations to reject the request".

For the OPTIONS request, the RFC says:
> "A client that generates an OPTIONS request containing a payload body MUST send a valid Content-Type header field describing the representation media type. Although this specification does not define any use for such a payload, future extensions to HTTP might use the OPTIONS body to make more detailed queries about the target resource".

Drup:
I think you should just put polymorphic variants, no need to define a new data type for that.

slegrand45:
I have modify the changes to use polymorphic variants.

vouillon:
You cannot change the type of the `_open` method. It's a method of a JavaScript object, and we don't have control over what this object expects.

slegrand45:
Indeed. The proposed patch doesn't change the type of the _open method.

slegrand45:
The type is now [`Get | `Post | `Head | `Put | `Delete | `Options | `Patch].

rgrinberg:
This is just a consideration, but it might make sense to uppercase the method names like cohttp for consistency. https://github.com/mirage/ocaml-cohttp/blob/master/lib/code.mli#L10

Drup:
Ah, perfect, thanks for the link @rgrinberg. Being consistent with cohttp was exactly the goal. :)

vouillon:
Thanks a lot!

SGrondin:
If the goal is to be consistent with Cohttp, then it might need an `Other too?

https://github.com/mirage/ocaml-cohttp/blob/83498a0e511ba5578410c70534bc2d6cc9f21503/lib/code.mli#L10

Drup:
I'm not sure what's the point of the `Other, tbh. If new verbs are going to be added, we will see it coming from quite far away, and adding new verb is reasonably backward compatible with subtyping.

rgrinberg:
I was never very convinced about cohttp adding `Other either. It definitely doesn't seem necessary here.

SGrondin:
I work at Mashape (we do API tooling, marketplace, analytics, management, etc) and it's surprisingly frequent to see users make up their own methods, to the point where library implementations that don't support custom verbs are considered "broken" by API devs.

lib/xmlHttpRequest.mli
```diff
@@ -128,14 +129,17 @@ val perform_raw_url :
...
+  (** [perform_raw_url] makes an asynchronous request to the specified [url] with
       specified options. The result is a cancelable thread returning
-      an HTTP frame. If [post_args] and [form_arg] are [None], a GET request is
-      used. If [post_args] or [form_arg] is [Some _] (even [Some []]) then a POST
-      request is made. The [check_headers] argument is run as soon as the answer
+      an HTTP frame. By default, if [post_args] and [form_arg] are [None], a GET 
+      request is used. If [post_args] or [form_arg] is [Some _] (even [Some []]) then a POST
+      request is made. But if [override_method] is set, the request method is forced, 
+      no matter the [post_args] or [form_arg] value. For example, with [override_method]
+      set to [`Put] and [form_arg] set to [Some _] a PUT request including the form data
```
rgrinberg: should be ````PUT```

Drup:
Hum, I'm quite surprised by that, but if you think it's necessary, why not!

lib/xmlHttpRequest.mli
slegrand45: My bad, sorry. I sent a patch to fix this. Thank you.

PR #296, opened 2015-05-07
Update vendors polyfills

slegrand45:
As stated in the current json.js (2011-02-23), this file has been superceded by json2.js. Use this newer polyfill (2015-05-03) from https://github.com/douglascrockford/JSON-js.

Update also the classlist polyfill with the latest version (2015-03-12) from https://github.com/eligrey/classList.js/.


vouillon:
Thanks!

PR #297, opened 2015-05-10
Documentation for Tyxml_js

slegrand45:
Add references to the Tyxml project and give an example of use.


Drup:
Oh, nice, thanks! Two remarks:
- I think you should use local open `T.( ... )` to avoid typing all these T. 
- you can remove `~a:[]` (at least to show it can be removed).

Drup:
Thanks!

PR #299, opened 2015-05-13
XmlHttpRequest: send blob objects

pveber:
The method `send` accepts blob objects [1]

[1] https://developer.mozilla.org/en/docs/Web/API/XMLHttpRequest#send%28%29


vouillon:
Thanks!

PR #304, opened 2015-05-19
Runtime: initial support for Core/Async kernel

hhugo:


avsm:
This is excellent! But is the plan to split out pieces into smaller chunks (e.g. `bigstring.js`) rather than a monolithic `core_kernel.js`?

hhugo:
Possibly.
Note that the only effect for the end-user is the set of files it needs to pass to js_of_ocaml.
In any case, js_of_ocaml will only link needed primitives and their dependencies.

```
js_of_ocaml +core_kernel.js main.byte
```
vs
```
js_of_ocaml +bigstring.js +core_kernel.js +async_kernel.js +bin_prot.js main.byte
```

hhugo:
We could also introduce dependencies between files so that an aync_kernel user only need
```
js_of_ocaml +async_kernel.js main.byte
```

PR #312, opened 2015-05-27
Lib: add getElementsByClassName

cago:


vouillon:
I believe this should be in `Dom_html`. There is no concept of classes in XML.

PR #313, opened 2015-05-30
Js_of_ocaml can now compile cmo and cma

hhugo:
Example:
Compile the stdlib and an ml file (test_cmo.ml)
```
js_of_ocaml `ocamlc -where`/stdlib.cma -o stdlib.js
ocamlc -c test_cmo.ml
js_of_ocaml test_cmo.cmo
```
Run with 
```
    <script type="text/javascript">
      var joo_global_object=this
    </script>
    <script type="text/javascript" src="runtime/runtime.js"></script>
    <script type="text/javascript" src="runtime/predefined_exceptions.js"></script>
    <script type="text/javascript" src="stdlib.js"></script>
    <script type="text/javascript" src="test_cmo.js"></script>

```


Drup:
Wow, Great work! I think that is going to be very useful!

cc @andrewray Pretty sure you can simplify iokerneljs with that.

andrewray:
We can currently dynamically load libraries but that tends to lead to stack overflows.  Will need to test if this improves matters.  If so, then this is very interesting!

@avsm Its possible we could set up a build service for ocaml libraries with this.

avsm:
@andrewray I'll happily set up the bulk builds to expose this via opam.ocaml.org as a HTTP service! Just ping me whenever there's a released version of js_of_ocaml and I'll make sure that the right build artefacts are made available from the bulk engine. /cc @dsheets @samoht

hhugo:
@andrewray do you know what's causing the stack overflows you're talking about ?

bobzhang:
This looks cool, but I got an error when trying this feature
```
play_js> js_of_ocaml t.cmo                                                                                                                                                    
js_of_ocaml: Error: Not an ocaml bytecode file                                                                                                                                 
js_of_ocaml: Error: Invalid magic number "d\027\n\139\214\204\220@@@@@" 
```

bobzhang:
I am using the master branch, so it should include this thing

hhugo:
a cmo should start with `Caml1999O010`.
what's the output of `head -c 12 t.cmo`

ejgallego:
@hhugo , we are also getting stack overflows with dynamic compilation.

I'm trying to debug them, meanwhile you can see an example using firefox and loading ssreflect as instructed in (https://x80.org/rhino-coq/).

Chrome will also SO if loading ssr in the standard way:
````
Require Import Srreflect.ssreflect.
````

Thanks,
E.

hhugo:
there is now an example of separate compilation https://github.com/ocsigen/js_of_ocaml/commit/9b5958074a948cd7914b79b860b9cd4c5c4a646e

PR #322, opened 2015-06-19
Quick patch, which allows programs using COW to be compiled with js_of_ocaml

chrismamo1:
I made a couple of the primitives in runtime/bigstring.js accessible. They were already implemented, I just had to rename them and add them to the runtime Makefile (I'm not sure why they weren't named properly or set to be included in the runtime, there may be something to that). This was necessary to get a project using [COW](https://github.com/mirage/ocaml-cow) to compile, and appears to work.

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -1,4 +1,4 @@
-FILES= 	mlString.js ieee_754.js int64.js md5.js marshal.js \
+FILES= 	bigstring.js mlString.js ieee_754.js int64.js md5.js marshal.js \
 	lexing.js parsing.js bigarray.js unix.js \
 	stdlib.js io.js fs.js jslib.js jslib_js_of_ocaml.js \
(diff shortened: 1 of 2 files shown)


hhugo:
The names are for https://github.com/janestreet/core_kernel/blob/master/src/bigstring_stubs.c

You can create aliases instead.
```
//Provides: caml_blit_bigstring_to_string
//Requires: bigstring_blit_bigstring_string_stub
var caml_blit_bigstring_to_string = bigstring_blit_bigstring_string_stub
```

`bigstring.js` is not included in the default runtime, the same apply to other files : `core_kernel.js`, `bin_prot.js`.
I'm not sure it should go in the default runtime. (@vouillon ?)

if you list the file in https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/option.ml#L24,
the compiler will give hints when it sees missing primitives.

hhugo:
partial fix merged

PR #344, opened 2015-07-22
Syntax: synchronize ppx and camlp4

hhugo:
the two syntax now generate the same OCaml code (modulo http://caml.inria.fr/mantis/view.php?id=6936)

### DIFF
--- a/compiler/flow.ml
+++ b/compiler/flow.ml
@@ -237,7 +237,7 @@ let expr_escape st _x e =
                 Array.iter
                   (fun x ->
                      begin match st.defs.(Var.idx x) with
-                       | Expr (Block (_, [|k; v|])) ->
+                       | Expr (Block (_, [|_k; v|])) ->
                          block_escape st v
                        | _ ->
                          block_escape st x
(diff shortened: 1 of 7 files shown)


Drup:
Oh, nice. Give me tomorrow to review.

hhugo:
I've used https://github.com/janestreet/camlp4-to-ppx to perform automatic conversion and make sure the two syntax where equivalent.
I'll push my rewriter to https://github.com/janestreet/camlp4-to-ppx this week

Drup:
This seems good to me.

Do you have an example of bug caused by the non-defined ordering (both in OCaml and in jsoo's ppx prior to this patch) ?

hhugo:
The undefined ordering only affect side effects aka identifier generation.
It makes diffing post processed ocaml source file with both camlp4 and ppx more painful.

gasche:
Undefined map ordering is typically an issue when people implement `fold` (which has an expected iteration order) on top of `map` using mutable state. (This is why the traversals-generation function of Camlp4 can generate the three of `Map`, `Fold` and `FoldMap`.)

Drup:
@hhugo You used `@@` in the camlp4 code, which should compile on 4.00

hhugo:
fixed

hhugo:
see https://github.com/janestreet/camlp4-to-ppx/commit/d2b55dc16fb3fab8fe9cf33c3c81cb33795f4c61 for automatic rewrite to ppx

PR #345, opened 2015-07-22
Lib: fix #333

### LINKED ISSUE #333: `Tyxml_js.R.filter_attrib`
@hhugo Can you implement this one in jsoo, similar to the one you did in eliom ? 

I wonder what it would cost us to have that directly in tyxml ...

hhugo:
@Drup

### DIFF
--- a/lib/tyxml/tyxml_js.ml
+++ b/lib/tyxml/tyxml_js.ml
@@ -78,6 +78,7 @@ module Xml = struct
       let n = Js.string n in
         match att with
         | Attr a ->
+          (* Note that once we have weak pointers working, we'll need to React.S.retain *) 
           let _ : unit React.S.t = React.S.map (function
           | Some v -> ignore(e##setAttribute(n, v))
           | None ->   ignore(e##removeAttribute(n))) a
@@ -192,6 +193,7 @@ module Util = struct
 
   let update_children (dom : Dom.node Js.t) (nodes : Dom.node Js.t t) =
     removeChildren dom;
+    (* Note that once we have weak pointers working, we'll need to React.S.retain *) 
     let _s : unit React.S.t = fold (fun () msg -> merge_msg dom msg) nodes ()
     in ()
 end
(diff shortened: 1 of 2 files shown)


lib/tyxml/tyxml_js.ml
```diff
@@ -42,13 +42,10 @@ module Xml = struct
   type keyboard_event_handler = Dom_html.keyboardEvent Js.t -> bool
   type attrib_k =
     | Event of biggest_event_handler
-    | Attr of Dom.attr Js.t
+    | Attr of Js.js_string Js.t option React.S.t
```
Drup: Would it make sense (from an efficiency perspective) to keep a constant attribute constructor, with the new signal attribute ?

lib/tyxml/tyxml_js.ml
```diff
@@ -78,9 +75,14 @@ module Xml = struct
 
   let attach_attribs e l =
     List.iter (fun (n,att) ->
+      let n = Js.string n in
         match att with
-        | Attr a -> ignore(e##setAttributeNode(a))
-        | Event h -> Js.Unsafe.set e (Js.string n) (fun ev -> Js.bool (h ev))
+        | Attr a ->
+          let _ : unit React.S.t = React.S.map (function
```
Drup: Assuming that GC of signals on javascript might be one day fixed, shouldn't this use `Lwt_react.S.keep` ?

Drup:
Great, thanks!

Drup:
cc @slegrand45

lib/tyxml/tyxml_js.ml
hhugo: React already has fast path for constant signals. I don't think you'll gain any efficiency adding an extra constructor.

lib/tyxml/tyxml_js.ml
hhugo: I'll just add a comment about it.

lib/tyxml/tyxml_js.ml
Drup: Ok.

PR #352, opened 2015-08-12
Allow to customize 'in_channel' with specific 'refiller'

OCamlPro-Henry:
Trivial usage illustrated in the demo toplevel, where reads to stdin calls 'window.prompt'.


hhugo:
thank you very much

PR #362, opened 2015-09-29
Add Tyxml_js.register

Drup:
This is a request for comment on the interface.
I think this is the good set of functions we want :
 - It allows to define everything in pure tyxml *and* to rely on a statically writen html
 - It's rather safe
 - It works well with reactive nodes (I think, cc @hhugo).

@talex5 See https://github.com/Drup/js-skeleton/blob/master/main.ml#L101 for a use-case.

Next remaining step to be able to define only-tyxml webpages is `Lwt_js_events`.

### DIFF
--- a/lib/tyxml/tyxml_js.mli
+++ b/lib/tyxml/tyxml_js.mli
@@ -60,6 +60,40 @@ module Svg : Svg_sigs.Make(Xml).T
 
 module Html5 : Html5_sigs.Make(Xml)(Svg).T
 
+module Register : sig
+
+  val html :
+    ?head:Html5_types.head Html5.elt ->
+    Html5_types.body Html5.elt -> unit
+  (** [Register.html head body] uses the given head and body elements
+      as document. It replaces the previous body and head.
+
+      [head] and [body] can be reactive.
+  *)
+
+  val body : ?keep:bool -> [< Html5_types.body_content] Html5.elt list -> unit
+  (** [Register.body elements] add [elements] as children of [body].
+      If [keep] is false (default is true), the children of the body are
+      removed before adding the new elements.
+  *)
+
+  val head : ?keep:bool -> [< Html5_types.head_content] Html5.elt list -> unit
+  (** [Register.head elements] add [elements] as children of [body].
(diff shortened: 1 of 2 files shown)


lib/tyxml/tyxml_js.mli
```diff
@@ -60,6 +60,18 @@ module Svg : Svg_sigs.Make(Xml).T
 
 module Html5 : Html5_sigs.Make(Xml)(Svg).T
 
+val register : ?keep:bool -> Html5_types.html Html5.elt -> unit
+(** [register html] will use the given html element as document.
+
+    If [keep] is true (the default) then the content is added
+    and do not replace the previous body and head.
```
talex5: s/do not/does not/

talex5:
Looks useful - thanks!

vasilisp:
I would export `register_body` (at least); a static head suffices more often than not.

`register` could be more permissive, e.g., by accepting `[body]`.

You could put them all under a `Register` module.

Drup:
I realized my version was actually broken with reactive node and the semantic was stupid, so I redid it. It's much nicer now. Technically, I could be able to add more function if there are other ways to query a *unique* node, but I think the spec only guarantee that for head, body and lookup by id.

hhugo:
@Drup , do you want to merge this before the release ?

Drup:
No, let's postpone.

hhugo:
ok

hhugo:
@Drup do you want to merge this as it is now ?

Drup:
@hhugo I just looked at it again. It seems good. Could you check that nothing is going to go terribly wrong with reactive nodes ?

vasilisp:
Ping on this. It will make using `Tyxml_js` a tad easier for absolute beginners, and we can use it in the planned tutorial.

vasilisp:
For reactive nodes, I can only imagine trouble arising if we use this API to add children to nodes that have been produced by `R`. But no big deal, you can go out of your way to cause the same trouble otherwise, e.g., by directly using the DOM APIs.

PR #371, opened 2015-10-20
PPX: Add hints for Merlin

def-lkb:
This PR adjusts the code generated to help Merlin do its work, which suffered from the switch away from camlp4 (the-lambda-church/merlin#449).

The PPX generates multiple nodes at the same location in the Typedtree to improve the error messages.
But this confuses Merlin which is no longer able to map from cursor location to a node.

These patches add attributes and "ghostness" to help Melin navigates the tree.
Also, empty methods are ignored: these could be produced because of incomplete user input and caused an assertion to fail.

### DIFF
--- a/lib/ppx/ppx_js.cppo.ml
+++ b/lib/ppx/ppx_js.cppo.ml
@@ -118,9 +118,14 @@ let constrain_types ?loc obj res res_typ meth meth_typ args =
 
   let e_x, p_x = mk_id ~loc:obj.pexp_loc "x" in
   (* [($x$#meth : $meth_typ$)] *)
+  let loc = match loc with
+    | None -> None
+    | Some l -> Some {l with Location.loc_ghost = true}
+  in
+  let interesting = (Location.mknoloc "merlin.teresting", PStr []) in
   let body =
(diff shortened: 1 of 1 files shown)


Drup:
@hhugo Do we need to change the camlp4 extension to please camlp4-to-ppx too ?

lib/ppx/ppx_js.cppo.ml
```diff
@@ -152,6 +158,9 @@ let method_call ~loc obj meth args =
   let ret_type = fresh_type obj.pexp_loc in
   let method_type =
     arrows (List.map (fun (_,_,_,x) -> x) args) (Js.type_ "meth" [ret_type]) in
+  let method_type =
+    {method_type with ptyp_loc = {method_type.ptyp_loc with Location.loc_ghost = true}}
+  in
```
Drup: I think you can do that with `[@metaloc .. ]` in less code.

hhugo:
Drup, I don't think there is anything to do to please camlp4-to-ppx. We might want to update the camlp4 syntax anyway to easing maintenance.

PR #381, opened 2015-11-16
Add custom_header ability to jsoo, override default message

fxfactorial:
This would be very useful to me because then I could add a `#!/usr/bin/env node` to the beginning and therefore create command line tools. I tested this and it worked as expected.


fxfactorial:
@hhugo Does no one care to comment on this PR or at least close it if its irrelevant to the project?

compiler/commonArg.ml
```diff
@@ -63,10 +64,13 @@ let noinline =
 let quiet =
   let doc = "suppress non-error messages." in
   Arg.(value & flag & info ["quiet";"q"] ~doc)
-  
+
+let custom_header =
+  let doc = "Provide a custom header for generated JavaScript file" in
```
hhugo: The documentation could explain the use case for this option.

compiler/driver.ml
```diff
@@ -158,14 +158,16 @@ let generate d ?toplevel (p,live_vars) =
...
     let version = match Compiler_version.git_version with
       | "" -> Compiler_version.s
       | v  -> Printf.sprintf "%s+git-%s"Compiler_version.s v in
 
-    Pretty_print.string formatter
-      ("// Generated by js_of_ocaml " ^ version);
+    (match custom_header with
+     | None -> Pretty_print.string formatter
+                 ("// Generated by js_of_ocaml " ^ version)
+     | Some c -> Pretty_print.string formatter c);
```
hhugo: What about always keeping the header with js_of_ocaml version ?
````
 #!/usr/bin/env node
// Generated by js_of_ocaml blah
```

fxfactorial: @hhugo K, amended this and the other PR. thank you.

PR #388, opened 2015-12-08
Geolocation API: http://www.w3.org/TR/geolocation-API/

slegrand45:
This is a proposal for a binding to Geolocation API. Regarding the code:

- May be the `geolocation` object should be added to the current `navigator` object attributes?
- The last two parameters (for the error function and for the options) of `getCurrentPosition` and `watchPosition` should be optional. I don't know how to specify this?

### DIFF
--- a/lib/geolocation.ml
+++ b/lib/geolocation.ml
@@ -40,9 +40,9 @@ class type position = object
 end
 
 class type positionOptions = object
-  method enableHighAccuracy : bool Js.prop
-  method timeout : int Js.prop
-  method maximumAge : int Js.prop
+  method enableHighAccuracy : bool Js.writeonly_prop
+  method timeout : int Js.writeonly_prop
+  method maximumAge : int Js.writeonly_prop
 end
 
 class type positionError = object
--- a/lib/geolocation.ml
+++ b/lib/geolocation.ml
@@ -51,12 +51,12 @@ class type positionError = object
 end
 
 class type geolocation = object
-  method getCurrentPosition : (position Js.t -> unit)
-    -> (positionError Js.t -> unit)
+  method getCurrentPosition : (position Js.t -> unit) Js.callback
+    -> (positionError Js.t -> unit) Js.callback
     -> positionOptions Js.t
     -> unit Js.meth
-  method watchPosition : (position Js.t -> unit)
-    -> (positionError Js.t -> unit)
+  method watchPosition : (position Js.t -> unit) Js.callback
+    -> (positionError Js.t -> unit) Js.callback
     -> positionOptions Js.t
     -> watchId Js.meth
   method clearWatch : watchId -> unit Js.meth
--- a/lib/geolocation.ml
+++ b/lib/geolocation.ml
@@ -42,8 +42,8 @@ class type positionOptions = object
 end
 
 class type positionError = object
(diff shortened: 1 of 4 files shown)


lib/geolocation.mli
```diff
@@ -0,0 +1,91 @@
...
+      match code with
+      | Geolocation.TIMEOUT -> Firebug.console##debug(msg)
+      | _ -> ()
+    in
+    geo##getCurrentPosition(f_success, f_error, options) *)
+
+type positionErrorCode =
+    PERMISSION_DENIED
+  | POSITION_UNAVAILABLE
+  | TIMEOUT
```
hhugo: This doesn't seems correct. The spec says.
```
  [NoInterfaceObject]
  interface PositionError {
    const unsigned short PERMISSION_DENIED = 1;
    const unsigned short POSITION_UNAVAILABLE = 2;
    const unsigned short TIMEOUT = 3;
    readonly attribute unsigned short code;
    readonly attribute DOMString message;
  };
```

lib/geolocation.mli
```diff
@@ -0,0 +1,91 @@
...
+  method heading : float Js.opt Js.readonly_prop
+  method speed : float Js.opt Js.readonly_prop
+end
+
+class type position = object
+  method coords : coordinates Js.t Js.readonly_prop
+  method timestamp : Js.date Js.readonly_prop
+end
+
+class type positionOptions = object
```
hhugo: optional options should either be `Js.writeonly_prop` or `Js.optdef Js.readonly_prop`

lib/geolocation.ml
```diff
@@ -0,0 +1,67 @@
...
+    -> (positionError Js.t -> unit)
+    -> positionOptions Js.t
+    -> unit Js.meth
+  method watchPosition : (position Js.t -> unit)
+    -> (positionError Js.t -> unit)
+    -> positionOptions Js.t
+    -> int Js.meth
+  method clearWatch : int -> unit Js.meth
+end
+
```
hhugo: please introduce an abstract type for `watchId` and use it for `clearWatch` and `watchPosition`

lib/geolocation.ml
```diff
@@ -0,0 +1,67 @@
...
+class type positionError = object
+  method code : positionErrorCode Js.readonly_prop
+  method message : Js.js_string Js.t Js.readonly_prop
+end
+
+class type geolocation = object
+  method getCurrentPosition : (position Js.t -> unit)
+    -> (positionError Js.t -> unit)
```
hhugo: You should use Js.callback : `(position Js.t -> unit) Js.callback`

slegrand45:
Should be better now

lib/geolocation.ml
```diff
@@ -0,0 +1,68 @@
...
+end
+
+class type positionOptions = object
+  method enableHighAccuracy : bool Js.writeonly_prop
+  method timeout : int Js.writeonly_prop
+  method maximumAge : int Js.writeonly_prop
+end
+
+class type positionError = object
+  method _PERMISSION_DENIED : positionErrorCode Js.readonly_prop
```
hhugo: should be `_PERMISSION_DENIED_`.
The syntax extension remove the first `_` if it exists and everything after the last `_`

slegrand45:
Underscore added

hhugo:
thank you

Issue #250, opened 2015-02-06
Add mutation observers in Lwt_js_events


balat:

Drup:
Hum, What is it ? Why ? How ? ^^'

balat:
https://developer.mozilla.org/fr/docs/Web/API/MutationObserver

johnlepikhin:
Mutation events ( https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events ) are marked as deprecated. This patch adds support of replacement. Looks like it will be required in the nearest future.

vasilisp:
@johnlepikhin, is there a patch somewhere?

@balat, while it looks useful generally, do we have a concrete immediate application?

Also, would a `React` layer over this be of service?

slegrand45:
An example of practical use is with Material Design Lite ( http://www.getmdl.io ). When you use this framework with dynamic DOM, you have to register new elements ( http://www.getmdl.io/started/index.html#dynamic and http://quaintous.com/2015/07/09/react-components-with-mdl/ ). A trick to do that is to use MutationObserver to automagically call the MDL javascript function `upgradeDom` ( https://github.com/google/material-design-lite/issues/917 ) :

```javascript
var observer = new MutationObserver(function(mutations) {
        var upgrade = false;

        for (var i = 0; i < mutations.length; i++) {
            if (mutations[i].addedNodes.length > 0) {
                upgrade = true;
                break;
            } 
        }
        if (upgrade) {
            // If there is at least a new element, upgrade the DOM.
            // Note: upgrading elements one by one seems to insert bugs in MDL 
            window.componentHandler.upgradeDom();
        }
    });
observer.observe(document, {
    childList : true,
    subtree : true
});
```

balat:
I'm using this for example with a reactive list of comments, when I want to keep it scrolled to the bottom.

```ocaml
Bs_lib.content_changes %d (fun mutations -> Bs_lib.scroll_to_bottom %d);
```

Another use: If you want to implement custom scrollbars (which is useful as braowser scrollbars are not customizable and horrible), you want the scrollbar position to get updated automatically  when the contents change.

A basic implmentation of this is ```Lwt_js_events``` would be something like:
```ocaml
(* Content change observer: *)
let content_changes elt f =
  let observer =
    jsnew (Js.Unsafe.global##_MutationObserver)(Js.wrap_callback f)
  in
  observer##observe(To_dom.of_element elt,
[...]

PR #391, opened 2015-12-15
Mutation observers API: https://dom.spec.whatwg.org/#mutation-observers

slegrand45:
A proposal for a binding to the mutation observers API. See also https://github.com/ocsigen/js_of_ocaml/issues/250

### DIFF
--- a/doc/api/index
+++ b/doc/api/index
@@ -24,6 +24,7 @@ WebGL
 WebSockets
 Geolocation
 Jstable
+MutationObserver
 }
 
 {1 Ppx - API Reference}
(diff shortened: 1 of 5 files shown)


vasilisp:
Thanks!

Could you rebase d212aebba6931231b65e1a8b02bee6b0aa5be405 on top of master to keep the history cleaner?

The API is a bit too raw JS for my taste. You can expose an OCaml-friendlier function that accepts the callback plus optional fields for the `_MutationObserverInit` fields, and calls the constructor internally. But that's just my opinion; maybe the binding is low-level on purpose.

slegrand45:
I added a higher level OCaml function and the rebase is done.

vouillon:
For the class type, I would use `mutationObserver` rather than `_MutationObserver`.

slegrand45:
_MutationObserver renamed to mutationObserver.

vouillon:
Great! Thanks!

PR #396, opened 2016-01-06
Use ocamlfind to find camlp4 everywhere

FlorentBecker:
This is needed (at least on nixos) because with caml >= 4.02, camlp4 might not live
in the compiler's directory. See http://caml.inria.fr/mantis/view.php?id=6605


Drup:
Can you re-push to trigger jenkins ? It should be fixed now.

PR #398, opened 2016-01-13
Add the Jstable module

Drup:
This module is backported from Eliom where it lived a very quiet life since its original commit 4 years ago.

For now, It's exactly the same as the one that was in eliom, but I think it would be interesting to see if there is a faster solution now.
@hhugo, I believe you did several such benchmark, any insights ?

I also wonder if we should change the API a little bit, so that to (partially) respect Hashtbl.S, but I'll wait for the above question to be answered first.

### DIFF
--- a/doc/api/index
+++ b/doc/api/index
@@ -23,6 +23,7 @@ Typed_array
 WebGL
 WebSockets
 Geolocation
+Jstable
 }
 
 {1 Ppx - API Reference}
--- a/lib/.depend
+++ b/lib/.depend
@@ -16,6 +16,8 @@ firebug.cmo : js.cmi dom.cmi firebug.cmi
 firebug.cmx : js.cmx dom.cmx firebug.cmi
 form.cmo : js.cmi file.cmi dom_html.cmi form.cmi
 form.cmx : js.cmx file.cmx dom_html.cmx form.cmi
+geolocation.cmo : js.cmi geolocation.cmi
+geolocation.cmx : js.cmx geolocation.cmi
 js.cmo : js.cmi
 js.cmx : js.cmi
 json.cmo : js.cmi json.cmi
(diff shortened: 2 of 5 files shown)


vasilisp:
Good discovery :). I had never seen this. It definitely needs to move out of Eliom.

The implementation is weird from an OCaml point of view, but it seems like JS objects were used as tables at one point in time. What about [maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

Drup:
@vasilisp Considering the support on IE, that's not a great idea.

Drup:
However, according to [this](http://kangax.github.io/compat-table/es5/#es5shim), it's probably fine to use `Object.create(null)` and get rid of the underscores.

vasilisp:
OK. I guess you can commit what you have now for the move to happen, and we can improve it later.

In a related note, most of `Eliommod_dom` should be fused into the DOM-related JSOO modules, right?

Drup:
I haven't looked at dom stuff at all.

zakki:
Don't `Jstable.find` need `hasOwnProperty` check?

vasilisp:
@zakki, I think it is OK. It returns `'a Js.Optdef.t`, so the caller is obliged to check whether the result is defined or not, e.g., via `Js.Optdef.test`.

zakki:
@vasilisp I think it has some corner cases. 
```ocaml
let _ =
  let a = Jstable.create() in
  let v = Jstable.find a (Js.string "__proto_") in (* expected? *)
  Firebug.console##log(v)
```

Drup:
That is more-or-less expected, yes. We could solve it by using `Obj.create(null)` as mentioned earlier.

PR #399, opened 2016-01-18
store/access javascript error (with backtrace) within ocaml exceptions. 

hhugo:
optionally store javascript [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) in ocaml exception.
This Error contains the call stack at creation point. It can later be outputted in the console (with locations handled by sourcemap).

### DIFF
--- a/compiler/generate.ml
+++ b/compiler/generate.ml
@@ -211,9 +211,9 @@ module Share = struct
       gen s
 
 
-  let get_apply gen n t =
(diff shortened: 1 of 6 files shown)


Drup:
How does it work ? Is it automatic ? Please document it ....

hhugo:
Does not work well on constant exception

Issue #384, opened 2015-11-24
Check OCaml compiler change : is it compatible with Js_of_ocaml  ?


hhugo:
following https://github.com/ocsigen/js_of_ocaml/issues/374

New bytecode pattern introduce by @trefis https://github.com/ocaml/ocaml/pull/305
might break important invariant in the js_of_ocaml compiler.

We should figure out if js_of_ocaml can support this change before the feature makes its way into trunk. @vouillon, would you have time to look at it ?

alainfrisch:
Is the invarariant broken by ocaml/ocaml#305 exactly the same as the one causing #374?

hhugo:
It believe not but I didn't spend much time trying to figure this out.
```
$ compiler/js_of_ocaml test.cmo --no-runtime -disable check-magic-number
[Warning] long options with a single '-' are now deprecated. Please use '--' for the following options: -disable
compiler/js_of_ocaml: You found a bug. Please report it at https://github.com/ocsigen/js_of_ocaml/issues :
Error: File "generate.ml", line 1401, characters 8-14: Assertion failed
Raised at file "generate.ml", line 1401, characters 8-53
Called from file "generate.ml", line 1631, characters 4-62
Called from file "generate.ml", line 1509, characters 10-128
Called from file "generate.ml", line 1631, characters 4-62
Called from file "generate.ml", line 1509, characters 10-128
Called from file "generate.ml", line 1794, characters 4-88
Called from file "generate.ml", line 1826, characters 12-45
Called from file "generate.ml", line 1836, characters 10-32
Called from file "driver.ml", line 83, characters 26-31
Called from file "driver.ml", line 83, characters 26-31
Called from file "driver.ml", line 83, characters 26-31
Called from file "driver.ml", line 83, characters 26-31
Called from file "driver.ml", line 83, characters 26-31
Called from file "driver.ml", line 83, characters 26-31
Called from file "jsoo_compile.ml", line 39, characters 9-13
Re-raised at file "jsoo_compile.ml", line 39, characters 44-45
Re-raised at file "jsoo_compile.ml", line 45, characters 10-13
Called from file "jsoo_compile.ml", line 112, characters 6-300
```

the other invariant breakage was multiple poptrap for a single pushtrap.
and was failing in one of the two locations: 
https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/parse_bytecode.ml#L1678
[...]

PR #402, opened 2016-01-22
Compiler: attempt to fix match_with_exn

hhugo:
This is just a quick-fix that we should fix before/when we address https://github.com/ocsigen/js_of_ocaml/issues/384

### DIFF
--- a/compiler/generate.ml
+++ b/compiler/generate.ml
@@ -587,7 +587,6 @@ let rec build_graph st pc anc =
     Hashtbl.add st.all_succs pc s;
     let backs = AddrSet.inter s anc in
     Hashtbl.add st.backs pc backs;
-
     let s = fold_children st.blocks pc (fun x l -> x :: l) [] in
     let succs = List.filter (fun pc -> not (AddrSet.mem pc anc)) s in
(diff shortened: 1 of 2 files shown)


vouillon:
Oui, c'est quelque chose comme cela qu'il faut faire.
Cela ne doit pas suffire pour OCaml 4.03 mais avec les `assert` que tu as ajouté, cela va devrait crasher plutôt que de produire du code incorrect.

hhugo:
For 4.03:
- the hack in https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/parse_bytecode.ml#L561 should handle the specific case of https://github.com/ocaml/ocaml/commit/aaeda7763bca8402cb2e53bdb57a456a04ada04b
- https://github.com/ocaml/ocaml/pull/305 has been reverted and it not an issue anymore.

I'm not aware of any other incompatible change.

If you're happy with this (hopefully temporary) fix. I'll merge it and release a new version

vouillon:
I was thinking of ocaml/ocaml#305. So indeed, there should not be any problem with OCaml 4.03.

Thanks!

PR #407, opened 2016-01-25
Convert documentation to ppx.

Drup:
I think I converted all of it.
The internals and the examples are still camlp4.

@hhugo @vouillon Should I convert the examples ? In this case, the ppx
should be loaded in the toplevel, instead of the camlp4 extension.

### DIFF
--- a/doc/manual/src/bindings.wiki
+++ b/doc/manual/src/bindings.wiki
@@ -4,11 +4,11 @@
 Write in .ml:
 
 <<code language="ocaml"|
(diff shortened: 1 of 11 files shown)


hhugo:
A toplevel with ppx extensions will require some more work. I don't have plan for this yet.

Drup:
It's unfortunate, but ok. Did you have a look at https://github.com/ocaml/ocaml/pull/386 ?

hhugo:
Yes, I'm aware of it but didn't spend much time thinking about it.

PR #411, opened 2016-01-26
Dom: move `click` from `InputElement` to `HTMLElement`

OCamlPro-Henry:
This where it belongs according to:

  https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click


hhugo:
thanks

PR #415, opened 2016-02-01
add origin method to Dom_html.location

zoggy:


hhugo:
It seems this property doesn't have good support on IE. Can you make it `Optdef.t` instead ? and eventually provide an helpers function along the line of  http://tosbourn.com/a-fix-for-window-location-origin-in-internet-explorer/.
Thanks

zoggy:
Ok for the Optdef.t and I will add an origin_safe method too.

hhugo:
Manually merged with small update/fix.
location_origin_safe has been rename into location_origin.
Thanks

PR #422, opened 2016-02-22
Fix `Js.string` for deserialized strings

vasilisp:
The implementation of `Js.to_string` uses `caml_js_from_string`, which is implemented in terms of the `toString` method. This is problematic if the string has been produced by deserializing, in which case we do not have our own `toString` method, and we get back JS strings that look like this: `[object Object]`.

I think it is safer to call the `caml_to_js_string` primitive.

For context, this came up while experimenting with the serialization [provided by IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) (ocsigen/ocaml-indexeddb@ac2715fe1ecedbcedc43babda001d7a42334b2db). This should cover most objects we care to store (including strings I think), without any explicit conversions.

### DIFF
--- a/compiler/generate.ml
+++ b/compiler/generate.ml
@@ -914,8 +914,6 @@ let _ =
   register_un_prim "caml_js_from_bool" `Pure
     (fun cx _ -> J.EUn (J.Not, J.EUn (J.Not, cx)));
   register_un_prim "caml_js_to_bool" `Pure (fun cx _ -> to_int cx);
-  register_un_prim "caml_js_from_string" `Mutable
-    (fun cx loc -> J.ECall (J.EDot (cx, "toString"), [], loc));
   register_tern_prim "caml_js_set"
     (fun cx cy cz _ -> J.EBin (J.Eq, J.EAccess (cx, cy), cz));
   register_bin_prim "caml_js_get" `Mutable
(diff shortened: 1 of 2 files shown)


vasilisp:
Jenkins didn't fire up, and I have no clue why.

vouillon:
I think you could completely remove `caml_js_from_string` and use `caml_to_js_string` instead.

vouillon:
If I understand correctly, you get OCaml strings that are not an instance of `MlString`. This will break all the places where we are using `instanceof MlString` (for instance, polymorphic comparison).

vasilisp:
> I think you could completely remove caml_js_from_string and use caml_to_js_string instead.

I agree. I am not sure I understand the need for the wrappers in `jslib_js_of_ocaml.js`.

> If I understand correctly, you get OCaml strings that are not an instance of MlString. This will break all the places where we are using instanceof MlString (for instance, polymorphic comparison).

Right, I get objects that have the `t`, `l`, and `c` fields, but which haven't been produced by the `MlString` constructor. I missed the other uses of `MlString`.

It may be possible to post-process the de-serialized things to get true `MlString`'s, or to adapt the runtime to deal with these almost-MlStrings. I clearly need to think more about this, I know almost nothing about the runtime :).

PR #425, opened 2016-02-25
Prevent inlining to loop on a block that immediately branches to itself

### LINKED ISSUE #424: Stack Overflow in the inlining pass when compiling an immediately recursive function
Trying to compile the following code makes js_of_ocaml stack overflow during the inlining pass:

```
let rec f () = f ()
let _ = f ()
```

(`Called from file "inline.ml", line 52, characters 49-78`)

Armael:
E.g a block

```
=== 2 () ===
  branch 2 ()
```

would cause the `follow_branch` function of `inline.ml` to loop.

Is this the correct way to fix #424 ?

### DIFF
--- a/compiler/inline.ml
+++ b/compiler/inline.ml
@@ -46,14 +46,18 @@ let optimizable blocks pc _ =
           | _ -> true
         ) b.body )  pc blocks true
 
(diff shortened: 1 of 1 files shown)


hhugo:
Thanks

PR #432, opened 2016-03-08
Dom_html: Add selection attributes to textarea.

j0sh:
Readonly attribs for now. According to https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement there is also an setSelectionRange method to set these attributes, but unimplemented at the moment.

### DIFF
--- a/lib/dom_html.ml
+++ b/lib/dom_html.ml
@@ -576,6 +576,9 @@ class type textAreaElement = object ('self)
   method name : js_string t readonly_prop
   method readOnly : bool t prop
   method rows : int prop
+  method selectionDirection : js_string t prop
+  method selectionEnd : int prop
+  method selectionStart : int prop
   method tabIndex : int prop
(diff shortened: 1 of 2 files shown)


Drup:
Thanks! selectionEnd and selectionStart are not described as readonly in the spec, though, why did you put them that way ?

j0sh:
Huh yeah, I didn't actually test if they were settable in plain JS. Seems they are, so patch updated, 'readonly' attributes removed.

j0sh:
Not directly relevant to this patch, but apparently these selection attributes also apply to the input element, but only for certain input types (text, password, etc). Not sure how those attributes could be included in the input element safely within js_of_ocaml. Here is a list of attributes and the input types they apply to: https://html.spec.whatwg.org/multipage/forms.html#do-not-apply

hhugo:
I don't have a good answer for input elements. Can you open an issue for this.
Thanks.

PR #435, opened 2016-03-10
Alternative interface for JsooTop.

OCamlPro-Henry:
This is part of the code developped for the OCaml MOOC.

A new submodule `Wrapped` provides an alternative implementation of the
`execute` function (and of others functions from 'Toploop') where error
and warnings are not printed on stderr but kept as first class value.

This is a first PR in a serie in order to allow the toplevel code to run in a WebWorker.


OCamlPro-Henry:
Some part of this code requires ocaml-4.02.2. We used `optcomp` in our project, but what would prefer ?

Drup:
@OCamlPro-Henry js_of_ocaml already uses cppo.

OCamlPro-Henry:
@Drup Thanks, I updated the PR.

OCamlPro-Henry:
I just updated the PR with a WebWorker-based minimal toplevel

hhugo:
Thanks, I'll try to take a look at this this week.

OCamlPro-Henry:
@hhugo Thanks. While the part `lib/toplevel` should be OK, the example in `toplevel` is rather a quick hack. It does not include any UI that would allow to interrupt the WebWorker after a while.

OCamlPro-Henry:
I just rebased the PR.

Issue #428, opened 2016-03-01
JavaScript literal object containing field `type`


lostman:
I was trying to create an object like this:
```
var input = {
  x: [1, 2, 3, 4],
  y: [10, 15, 13, 17],
  type: 'scatter'
};
```
Using object syntax and `ppx_js`:
```
let input = object%js
    val x    = [|1; 2; 3; 4|]
    val y    = [|10; 15; 13; 17|]
    val type = "scatter"
end
```
But this doesn't work because `type` is a restricted keyword.

Would it be possible to specify a mapping between JS and OCaml using a `ppx` annotation?
```
object%js
    val type_ [@property "type"] = "scatter"
end
```

hhugo:
this is already handled 
```
let input = object%js
    val x    = [|1; 2; 3; 4|]
    val y    = [|10; 15; 13; 17|]
    val type_ = "scatter"
end
```
see http://ocsigen.org/js_of_ocaml/2.7/manual/library -> "Method name and underscore"

lostman:
I tried variations with underscores (string literals need to be converted, right?):
```
let input = object%js
    val type_ = Js.string "scatter"
    val x    = Array.of_list xs
    val y    = Array.of_list ys
    val mode = Js.string "lines+markers"
end
console##log input;
```
And the result object is:
```
Object {type_: "scatter", x: Array[23], y: Array[23], mode: "lines+markers"}
```
```
❯ js_of_ocaml --version
2.7
```

Drup:
@hhugo It works for method calls, but I'm afraid I didn't implemented it for object creation. 
@lostman I'll try to fix it when I have time.

lostman:
@Drup Thanks!

hhugo:
solved by https://github.com/ocsigen/js_of_ocaml/pull/443

PR #443, opened 2016-03-21
Syntax: object litteral, preprocess prop and meth name

hhugo:
@Drup: do you want to do more than this small fix ? https://github.com/ocsigen/js_of_ocaml/issues/428


Drup:
That looks sufficient to me. Can you add a test ?

hhugo:
I've fixed detection of duplicated meth name / prop name and added some tests.

Drup:
@hhugo Perfect!

PR #364, opened 2015-10-05
Ppx deriving

vasilisp:
This is a PPX version of our JSON deriving plug-in.

The generated code is somewhat different from the Camlp4 version. We follow the conventions of `ppx_deriving`, as opposed to the functorial style of (Camlp4) `deriving`.

The serialization format is the same, i.e., in principle you can dump `x` with Camlp4, parse the result with PPX, and get back `x`. Any divergence is a bug to be fixed.

I provide a few tests that demonstrate the syntax [as a Gist](https://gist.github.com/vasilisp/187ea176382511db3146).

PR #449, opened 2016-03-28
Deriving OCaml 4.03

hhugo:


hhugo:
We need to add some tests: https://gist.github.com/vasilisp/187ea176382511db3146 (https://github.com/ocsigen/js_of_ocaml/pull/364)

vasilisp:
Thanks! Looks good from a first glance. I will test and try to integrate the tests in the build system.

vasilisp:
This has a runtime dependency on `deriving`. I have a [messy patch](https://github.com/ocsigen/deriving/tree/4.03) to compile `deriving` on 4.03, but it breaks older versions. (I am still trying to figure out how to preserve compatibility.)

vasilisp:
I pushed a few fixes for inline records. To my knowledge, it works now.

hhugo:
thanks

hhugo:
I've added some tests.. merging

PR #451, opened 2016-04-02
jsoo_mktop: flushing stderr

Kakadu:
The output with -verbose option was a little bit misleading in case of errors. (Error was displayed before appropriate command). We are flushing stdout now to prevent that.


hhugo:
Thanks

PR #452, opened 2016-04-04
Fixed MutationObserver.{added,removed}Nodes type.

sagotch:
The extra `Js.t` make you get a `Dom.node Js.t Js.t` when accessing to an element of the list.


hhugo:
thanks

PR #456, opened 2016-04-08
Add Js.object_keys to retrieve jsarray of keys

### LINKED ISSUE #455: Provide Object.keys or other method for iteration
There is currently no way to get all keys of arbitrary javascript object or iterate through them, but this functionality come handy from time to time. One can use `Js.Unsafe.meth_call`, but not all browsers have this functionality. It seems, that good it would be useful to have `keys` available in jsoo runtime (with appropriate  [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Polyfill)).

What do you think?

little-arhat:
It uses native Object.keys under the hood or provided polyfill.

(fixes #455)


hhugo:
This has been merged manually. Thanks

PR #457, opened 2016-04-10
Include `ppx_js` in opam installation.

datkin:
I have no idea if this is the idiomatic way to do this.


hhugo:
It is already installed in
```
$(opam config var lib)/js_of_ocaml/ppx_js
```

hhugo:
reopen is needed

PR #459, opened 2016-04-12
Blob and URL

sagotch:
While working with blobs and canvas, I needed some extra methods in js_of_ocaml. 

https://developer.mozilla.org/en-US/docs/Web/API/Window/URL


hhugo:
thanks

Issue #414, opened 2016-02-01
Preserve as much as possible the initial OCaml variable names


hhugo:
extracted from https://github.com/ocsigen/js_of_ocaml/issues/291
We could try to preserve as much as possible the initial OCaml variable names (appending a suffix only if necessary), and prefix the JavaScript variable that are introduced by the compiler so that they are clearly separated.

hhugo:
https://github.com/ocaml/ocaml/pull/455

PR #464, opened 2016-04-20
Compiler: Preserve ocaml names

hhugo:


hhugo:
fix https://github.com/ocsigen/js_of_ocaml/issues/414

PR #475, opened 2016-05-06
Small changes wrt compiler-libs.

Drup:
Tiny cleanup that makes
- the code a bit cleaner
- js_of_ocaml plays more nicely with my WIP fork :p

This compiles under 4.00.1

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -88,6 +88,8 @@ clean:
 	rm -f js_lexer.ml js_parser.ml js_parser.mli
 	rm -f util.ml
 	rm -Rf man
+	$(MAKE) -C lib clean
+	$(MAKE) -C findlib_support clean
 
 .PHONY: depend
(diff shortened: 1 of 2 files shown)


vasilisp:
Looks good. Your fork of what? Eliom?

Drup:
Yes. :3

Drup:
In general, treating `Ident.t` as an abstract datatype is a good idea.

hhugo:
There are more places that should be cleaned in we start linking compilier-libs [1]. 
I don't know the reason for not having compiler-libs linked already.. 

@drup What does your eliom fork have to do with this change ? 

[1] https://github.com/ocsigen/js_of_ocaml/search?utf8=%E2%9C%93&q=%22copied+from%22+parse_bytecode.ml&type=Code

Drup:
@hhugo Among some other changes, I made `Ident.t` abstract (for various reasons that are a bit long to detail here). I will take the time to do the other changes.

hhugo:
If anyone's wondering, this is to play nicely with an ocaml compiler fork tuned for eliom

Drup:
@hhugo Thanks!

PR #476, opened 2016-05-07
Compat with Tyxml 4.0

Drup:
Most of the work was done by @vasilisp


Drup:
Tyxml is being released! @hhugo you can merge!

vasilisp:
Can we merge this and the corresponding branches for Eliom et al.?

I have quite a few Eliom branches that do not merge cleanly, and maintaining them really gets in the way...

hhugo:
Should we drop compat with ocaml 4.00 and 4.01 entirely ?
@vouillon, @balat ?

Drup:
I personally think it's fine, it would allow to migrate jsoo's core to ppx too, which would be nice.

PR #486, opened 2016-06-22
Add optional `prefix` arg to Jsonp.call

little-arhat:
This allows to avoid collisions, when integrating to other sites.


little-arhat:
@hhugo What do you think of this?

hhugo:
Thanks

PR #487, opened 2016-06-29
Textarea selection

sagotch:


sagotch:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextareaElement

hhugo:
thanks

lib/dom_html.ml
```diff
@@ -586,6 +586,8 @@ class type textAreaElement = object ('self)
   method select : unit meth
   method required : bool t writeonly_prop
   method placeholder : js_string t writeonly_prop
+  method selectionStart : int prop
+  method selectionEnd : int prop
```
sagotch: Oops.. sorry about that. It was not in inputElement and I did not checked carefully enough textarea...
Also, I can see I could add method `selectionDirection : js_string t prop` for inputElement as well... (I'll make a PR)

PR #488, opened 2016-06-29
Ppx: ppx_driver compat

hhugo:


Drup:
The deriver is not included ?

(Also, I'm almost surprised that jsoo's ppx does not trigger any of ppx_driver's warnings :p)

hhugo:
> The deriver is not included ?

I'm not sure what you mean by that..?

This has not been extensively tested - It's apparently enough for the upcoming janestreet release.
cc @trefis

trefis:
> It's apparently enough for the upcoming janestreet release

Yes the current thing works as I would expect it too.

> I'm almost surprised that jsoo's ppx does not trigger any of ppx_driver's warnings

We have been using it internally with ppx_driver for a while and it indeed does not produce any warning. :)

Drup:
@hhugo I meant `js_of_ocaml.deriving.ppx`.
@trefis Nice, glad to hear that.

PR #495, opened 2016-07-17
Add is_null function

fxfactorial:


Drup:
Why do you need this ? Something that is potentially nullable should always end up being an `Opt.t` when exposed on the OCaml side.

If you need that function, then you used the wrong types earlier in the chain.

fxfactorial:
Because sometimes I get a `'a` and I want to check it at runtime.

Drup:
If you get an `'a`, you can make it an `_ Js.t Opt.t` easily. Can you show a concrete use case ?

fxfactorial:
No, I don't want to play with Opt in this case, I just want `'a -> bool`. Can we at least start from the assumption that I have a valid use case enough to open a PR? In related items, I want to make a utility library with these kinds of functions. https://github.com/ocsigen/js_of_ocaml/issues/497

vasilisp:
> Can we at least start from the assumption that I have a valid use case enough to open a PR?

No, sorry. Given `Opt`, we can't assume the existence of a new valid use case. That would enable a bad style where you pass around random possibly-null things, and only occasionally check for `null`. That's not the OCaml we know.

fxfactorial:
Having used `js_of_ocaml` for a while now and in many applied environments, I'm speaking from experience. Useful escape hatches are always available that whether or not its bad style is up to me. (There are times when its a hassle to introduce the Opt when I don't want to introduce the associated typing, when something from a JS api, etc etc, or that I know 99% of time is correct...basically making the argument for exceptions, and OCaml we all know does raise exceptions)

Yes I understand this is for the compiler codebase itself, but that's why we can add this to a utility library. Otherwise I'll make a separate utility library and given the increase in OCaml popularity and hopefully js_of_ocaml as well I would assume other people would do the same, then we end up with silly competing libraries like JS land has, that would be silly too.

hhugo:
I also don't think it's a good idea to encourage people to write unsafe code. Unsafe code should be explicit as much as possible. I would rather inline the function every it's needed.

fxfactorial:
Nevermind. I've realized that js_of_ocaml is not the right community for my OCaml compiled to JavaScript needs. I've spent a considerable amount of energy with js_of_ocaml, using it, learning it, promoting it, but now I see that bucklescript is the future for me and probably many others as well.

dsheets:
For what it's worth, I agree that there are cases where I'd like to check whether a JavaScript object is null without changing the type to accommodate that. JavaScript interfaces aren't reknown for adhering to a nice type discipline and there are certainly cases where "almost always" a value is non-null but, due to nonsense in a JavaScript component, you find that in some weird circumstances a null is passed and you don't want to change the type.

With that said, this doesn't seem like a huge function so it may be best to introduce this type system evasion on a per-need basis. I'm sure there are other useful utility functions that the `js_of_ocaml` maintainers would be happy to have.

PR #500, opened 2016-07-22
Added document##.activeElement method.

sagotch:
https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

> Returns the currently focused element, that is, the element that will get keystroke events if the user types any. This attribute is read only.

> When there is no selection, the active element is the page's <body> or null.


hhugo:
Thanks

PR #502, opened 2016-07-23
Import new ppx scheme from let-def

hhugo:
It should provide better merlin support.

### DIFF
--- a/lib/ppx/ppx_js.cppo.ml
+++ b/lib/ppx/ppx_js.cppo.ml
@@ -205,10 +205,9 @@ let new_object constr args =
   let invoker = invoker
       (fun targs tres -> arrows targs (Js.type_ "t" [tres]))
       (fun targs tres ->
-         let lbl, tobj = List.hd targs and targs = List.tl targs in
-         arrows
-           ((lbl, Js.type_ "constr" [tobj]) :: targs)
-           (Js.type_ "t" [tres]))
+         let tres = Js.type_ "t" [tres] in
+         let arrow = arrows targs tres in
+         arrows [(Js.nolabel, Js.type_ "constr" [arrow])] arrow)
       (function
         | (constr :: args) ->
           Js.unsafe "new_obj" [constr; inject_args args]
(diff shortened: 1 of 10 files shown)


Drup:
Ah, finally! :) 
I'll review next week.

hhugo:
I've tuned locations a little bit. I'm quite happy with `merlin-type-enclosing` now.

Drup:
Hum, the location in the error messages are not really the right one, the one that were there before seems better.

hhugo:
I'll update the PR later this week.

hhugo:
@Drup, I've updated the location.

hhugo:
@Drup, do you have time for review ?

lib/ppx/ppx_js.cppo.ml
```diff
@@ -23,43 +23,35 @@ let exp_to_string = function
...
 
 let inside_Js = lazy
   (try
      Filename.basename (Filename.chop_extension !Location.input_name) = "js"
    with Invalid_argument _ -> false)
 
+let merlin_noloc = { txt = "merlin.loc"; loc = Location.none }, PStr []
+
+(* let to_js_of_ocaml = lazy (Some "Js_of_ocaml") *)
+let to_js_of_ocaml = lazy None
```
Drup: What is this for ? Packing jsoo like you do in janestreet ? (Could you add a comment ?)

Drup: Actually, we could directly add a ppx option, and pass the name of the module path prefix on the cli. I guess that would avoid issues for you.

hhugo: It's indeed used inside Janestreet. I'll remove this diff as it does not belong in the PR.

Drup:
Can you give a detailed explanation of what `invoker` is supposed to be doing ? 

If I understand things correctly, we now use locally abstract types (with functions) instead of modules, and `invoker` generates the general code pattern that is used for all type constraints. The unifying aspect is good, but it means `invoker` should really be explained.

Also, could you give a quick idea of how/when ghost location and `{@merlin.loc]` are used ?

PR #505, opened 2016-07-28
Fix ppx_deriving_json for types with both manifest and definition

vasilisp:
Consider a type definition like

```ocaml
type a = b = A | ...
```

Our `ppx_deriving_json` plugin used to rely on the deriver for `b`, which does not necessarily exist (e.g., if `b` comes from a lower-level library that does not provide derivers). This patch produces a new deriver instead.

This came up while translating Eliom to PPX.


vasilisp:
@hhugo any objection to merging this?

hhugo:
go ahead

PR #512, opened 2016-08-03
Build deriving_[Jj]son.cm* even without deriving package

### LINKED ISSUE #509: deriving.ppx sublib and META file
I think there is an issue with the current META file: if js_of_ocaml is installed without deriving, the `js_of_ocaml.deriving` sublib doesn't exist, even if `ppx_deriving_json.cma` exists (which should enable `js_of_ocaml.deriving.ppx`

vasilisp:
Our runtime `deriving` components (`deriving_[Jj]son*` files, `js_of_ocaml.deriving` Findlib package) do not technically depend on the `deriving` package. They are used by both the `deriving` and `ppx_deriving` plugins. This PR updates the  build system to install the runtime components by default, so that they will be there for PPX.

We could make installation conditional on `deriving | ppx_deriving`, but it is not a lot of code, and theoretically somebody may want to use only the runtime part (say, for JSON-aware API but without directly using any deriver).

@Drup, would that fix #509 ?

### DIFF
--- a/Makefile.filelist
+++ b/Makefile.filelist
@@ -7,16 +7,18 @@ endif
 INTF := lib/*.cmi
 
 IMPL := lib/$(LIBNAME).cma   \
-        lib/syntax/pa_js.cmo \
         lib/dll$(LIBNAME)$(DLLEXT) \
         lib/lib$(LIBNAME)$(LIBEXT)
 
 INTF += lib/log/*.cmi
 IMPL += lib/log/logger.cma
 
+ifeq "${WITH_CAMLP4}" "YES"
+IMPL += lib/syntax/pa_js.cmo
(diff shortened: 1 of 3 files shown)


Drup:
I think that solves it, yes. In theory, that's a breaking change, though.

vasilisp:
@Drup, why is this a breaking change? We just provide an extra module by default.

Drup:
Because `js_of_ocaml.deriving` used to enable the camlp4 syntax, and with this it doesn't.

vasilisp:
OK, you are right. But really, continued maintenance of Camlp4-based things will involve way more trouble than enabling an extra Ocamlfind package :).

Drup:
I agree, it was just a remark. :)

PR #514, opened 2016-08-15
Fix keyboard events

Strikeskids:
Add wrapper modules that extract useful keyboard information from key events. Performs `keyCode` to `code` normalization.

### DIFF
--- a/lib/dom_html.ml
+++ b/lib/dom_html.ml
@@ -166,13 +166,19 @@ end
 
 and keyboardEvent = object
   inherit event
-  method charCode : int optdef readonly_prop
-  method keyCode : int readonly_prop
-  method keyIdentifier : js_string t optdef readonly_prop
   method altKey : bool t readonly_prop
   method shiftKey : bool t readonly_prop
   method ctrlKey : bool t readonly_prop
   method metaKey : bool t readonly_prop
+  method location : int readonly_prop
+
+  method key : js_string t optdef readonly_prop
+  method code : js_string t optdef readonly_prop
+
+  method which : int optdef readonly_prop
+  method charCode : int optdef readonly_prop
+  method keyCode : int optdef readonly_prop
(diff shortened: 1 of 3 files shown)


hhugo:
Thank you

lib/dom_html.ml
```diff
@@ -166,13 +166,19 @@ end
...
   method altKey : bool t readonly_prop
   method shiftKey : bool t readonly_prop
   method ctrlKey : bool t readonly_prop
   method metaKey : bool t readonly_prop
+  method location : int readonly_prop
+
+  method key : js_string t optdef readonly_prop
+  method code : js_string t optdef readonly_prop
+
+  method which : int optdef readonly_prop
```
vasilisp: This breaks Eliom. We need [multiple inheritance on `event`, `mouseEvent`, and `keyboardEvent`](https://github.com/ocsigen/eliom/blob/d24b99e13bb957de4e83a31636dffec72490dd83/src/lib/eliom_runtime.shared.mli#L70-L74). The new `which` method renders them incompatible.

hhugo: It this multiple inheritance actually needed ?

hhugo: I'm already looking at the fix for ocaml 4.01

vasilisp: The multiple inheritance is needed to have a uniform internal representation of event handlers. The resulting "biggest event" is used in contravariant position in handlers, so we can use any of the events in its place.

I will see if we can separate them by providing multiple constructors in place of [`CE_client_closure`](https://github.com/ocsigen/eliom/blob/d24b99e13bb957de4e83a31636dffec72490dd83/src/lib/eliom_runtime.shared.mli#L60).

vasilisp: Fixed by ocsigen/eliom@4edaf22 .

PR #517, opened 2016-08-18
Depend on uchar

### LINKED ISSUE #516: Uchar dependency < 4.03.0
The last PR #514 supposed we are using OCaml 4.03.0 due to the module Uchar.
Uchar must be installed (https://github.com/ocaml/uchar) and added as dependency in the Makefile if OCaml < 4.03.0.

It could be useful to add it in the opam file.

vasilisp:
Fixes #516 . I may have missed something.

### DIFF
--- a/examples/Makefile.common
+++ b/examples/Makefile.common
@@ -4,7 +4,7 @@ include ../../Makefile.conf
 
 COMP=../../compiler/$(COMPILER)
 JSFILES=../../runtime/runtime.js
-OCAMLC=ocamlfind ocamlc -g -package lwt -pp "camlp4o ../../lib/syntax/pa_js.cmo" -I ../../lib
+OCAMLC=ocamlfind ocamlc -g -package lwt,uchar -pp "camlp4o ../../lib/syntax/pa_js.cmo" -I ../../lib
 STDLIB=$(LIBNAME).cma
 
 $(NAME).js: $(NAME).byte $(COMP) $(JSFILES)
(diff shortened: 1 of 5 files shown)


examples/Makefile.common
```diff
@@ -4,7 +4,7 @@ include ../../Makefile.conf
 
 COMP=../../compiler/$(COMPILER)
 JSFILES=../../runtime/runtime.js
-OCAMLC=ocamlfind ocamlc -g -package lwt -pp "camlp4o ../../lib/syntax/pa_js.cmo" -I ../../lib
+OCAMLC=ocamlfind ocamlc -g -package lwt,uchar -pp "camlp4o ../../lib/syntax/pa_js.cmo" -I ../../lib
```
hhugo: Is this actually needed ?

lib/Makefile
```diff
@@ -240,6 +240,12 @@ ppx/ppx_deriving_json.cmxa: ppx/ppx_deriving_json.cmx
 ppx/ppx_deriving_json.cmxs: ppx/ppx_deriving_json.cmx
 	$(OCAMLOPT) -shared -o $@ $^
 
+dom_html.cmo: dom_html.ml syntax/pa_js.cmo
+	$(OCAMLC) -pp "camlp4o syntax/pa_js.cmo" $(SAFESTRING) -package lwt,uchar -c -g dom_html.ml
+
+dom_html.cmi: dom_html.mli
+	$(OCAMLC) -package lwt,uchar -I deriving_json $(SAFESTRING) -c dom_html.mli
+
```
hhugo: I think it would be better to add uchar to the rules ` %.cmo: %.ml` and ` %.cmi: %.mli` instead of creating rules just for dom_html

examples/Makefile.common
vasilisp: Yes, it comes up when we try to link-in `js_of_ocaml.cma`.

hhugo:
Thanks for the fix

PR #546, opened 2016-11-10
windo##open_ returns possibly null value

### LINKED ISSUE #542: Should window##open_ return opt?
Hi, currently `window##open_` assumes that browser can always open popup window. However, if popup was blocked, `window##open_` may return `null`. Should we change signature to reflect that?

From https://developer.mozilla.org/en-US/docs/Web/API/Window/open : 

```
How can I tell when my window was blocked by a popup blocker?
With the built-in popup blockers of Mozilla/Firefox and Internet Explorer 6 SP2, you have to check the return value of window.open(): it will be null if the window wasn't allowed to open. However, for most other popup blockers, there is no reliable way.
```

little-arhat:
fixes #542

### DIFF
--- a/lib/dom_html.ml
+++ b/lib/dom_html.ml
@@ -1231,7 +1231,7 @@ class type window = object
   method parent : window t readonly_prop
   method frameElement : element t opt readonly_prop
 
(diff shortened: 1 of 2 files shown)


hhugo:
Thanks

Issue #523, opened 2016-09-25
`make -C tests` fails


bmillwood:
It could be this is actually just me running them incorrectly, in which case I'll change the issue title to "unclear how to run tests correctly" :)

Firstly, they fail early if you don't have deriving installed (even though you can compile the library without it). I installed that, and then:

```
$ make
[... a bunch of stuff ...]
$ make -C tests
make:_ Entering directory '/home/ben/code/ocaml/js_of_ocaml/tests'
TERM=dumb ocaml -noprompt -I ../lib -I ../lib/deriving_json -init ppx_init.ml -ppx ../lib/ppx/ppx_js -ppx "/home/ben/.opam/4.03.0/lib/ppx_deriving/ppx_deriving ../lib/ppx/ppx_deriving_json.cma" < ppx/write_prop.ml 2>&1 | tail -n +15 > ppx/write_prop.result
diff ppx/write_prop.result ppx/write_prop.expected
1,2d0
< /home/ben/.opam/4.03.0/lib/ppx_deriving/ppx_deriving_runtime.cma: loaded
< /home/ben/.opam/4.03.0/lib/ppx_deriving/./ppx_deriving: activated
make: *** [Makefile:49: ppx/test_write_prop] Error 1
rm ppx/write_prop.result
make: Leaving directory '/home/ben/code/ocaml/js_of_ocaml/tests'
```

That's not so interesting (seems like that `tail -n +15` just needs to be `tail -n +17`) so I also attach [a gist of what happens when you tell make not to stop on the first failure](https://gist.github.com/bmillwood/c479033c2fe50fad6720b59c6420e89f)

(also is it just me or should the diffs have `-u` passed to them, for unified context? I can never read non-unified diffs)

My OCaml version is:

```
$ ocaml -version
The OCaml toplevel, version 4.03.0
```

Let me know if there's anything else I can provide that would help.

hhugo:
one need to call `jbuilder runtest`

PR #549, opened 2016-11-26
Tests: expect test for ppx syntax

hhugo:


Drup:
You're missing the extra opam dependency.

I still feel ppx_expect has far too many dependencies for what it does.

hhugo:
Tests are currently never run (sadly) so the extra dependency doesn't really matter here.
This PR relies on Toplevel_expect_tests and makes tests more robust (see https://github.com/ocsigen/js_of_ocaml/issues/523)

Drup:
Yeah, I read it, and it looks good. I still think you should add the `{test}` dependencies (so that people can do `opam install -t --deps-only` to start doing dev on jsoo.

hhugo:
agreed.

tests/ppx/main_deriving.ml
```diff
@@ -0,0 +1,16 @@
+module Type_conv = Ppx_type_conv.Type_conv
+
+let l = Ppx_deriving.derivers ()
+let _ = ListLabels.iter l ~f:(fun (deriver : Ppx_deriving.deriver) ->
+  Type_conv.add deriver.Ppx_deriving.name
```
hhugo: @diml: `type_conv` derivers can be used with `ppx_deriving` using `ppx_type_conv.deriving`.
What about `ppx_deriving` derivers used with `type_conv` ?

diml: The feature is implemented and released in our tree, it will be on github next time we push, so hopefully next week

PR #561, opened 2017-02-12
Fixes the tanh implementation. The current implementation is coth

modlfo:
I was having a strange behavior when running a numeric program after converted to Js. I track down the problem and seems that the implementation of tanh is incorrect.

I used this mathematica code to check my implementation:

```
y = Exp[x];
z = Exp[-x];
FullSimplify[(y + z)/(y - z)] // This gives Coth[x]
FullSimplify[(y - z)/(y + z)] // This is Tanh[x]
```


hhugo:
Thanks a lot

PR #567, opened 2017-03-15
Added animation events.

sagotch:
https://developer.mozilla.org/en-US/docs/Web/API/AnimationEvent


vouillon:
There is also an `animationcancel` event.

sagotch:
Yes I know, but according to https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onanimationcancel 

> At this time, no major browser supports animationcancel.

That being said, I see no good reason not to add it to this pull request since it has the same status than other events in this PR (https://drafts.csswg.org/css-animations/#eventdef-animationevent-animationcancel)

sagotch:
Added.

PR #579, opened 2017-05-05
rework version handling

hhugo:


Drup:
I think you should keep `version: dev` in the opam files (for pins, at the very least)

hhugo:
IIRC, `version: dev` has bad interaction with https://github.com/ocaml/opam-repository/pull/8966/files

Drup:
Why would it ? dev is above everything, and you obviously remove the version field for files in the opam repository.

hhugo:
Also, I want `${version:js_of_ocaml}` to resolve to the content of the VERSION file, not to "dev" version present in opam files.
@diml, would it make sense to ignore versions in opam files if equal to "dev"

hhugo:
@drup, let me try again

Drup:
Well, "version" being equal to dev is also an important information (and you have the git tag for that purpose).

hhugo:
js_of_ocaml uses `VERSION+git-xxxxxx` for development versions.
I prefer `3.0+git-3ff19ea` over `dev+git-3ff19ea`

Drup:
@hhugo as opam version ? The problem is that it prevents distinguishing between the last stable and the dev version inside opam constraint. In particular: you cant say that `js_of_ocaml-tyxml` needs `js_of_ocaml-compiler.dev`. Similarly, eliom dev can't mandate jsoo dev version, etc.

diml:
> @diml, would it make sense to ignore versions in opam files if equal to "dev"

Actually, maybe the precedence should be changed. i.e. if you have or generate a `<package>.version`, it's likely that this is what you want to use.

PR #582, opened 2017-05-15
Change type of Dom_html.storageEvent

xvw:
When we use `aStorage##clear()`, the key value could be null.


sagotch:
Not a comment about the PR, but why is it called `keynewValue` ? It probably should be `newValue` instead.

xvw:
Agreed !

hhugo:
Thanks

PR #592, opened 2017-06-13
Toplevel: fix #585

### LINKED ISSUE #585: Embedding library in the toplevel
Hi,
I would like to build a toplevel with a library of mine available. 

I used to do it by reusing the toplevel example and just adding my_module.cmo  to the mktop line and also "-export-unit my_module".

Now I can't do "-export-unit my_module" anymore as it complains that it can't find package my_module.
using  "-export-unit my_module.cmi -export-unit my_module.cmo", It compiles but I get the error "Reference to undefined global `My_module' " when I try to use it (My_module.test () ).

I attached an example with minimal modifications in the files from js_of_ocaml 2.8.4.
I'm using  js_of_ocaml 2.8.4 with ocaml  4.04.1.
[toplevel.tar.gz](https://github.com/ocsigen/js_of_ocaml/files/1035802/toplevel.tar.gz)

And here is a version which works with  js_of_ocaml 2.6 and ocaml 4.02.3
[toplevel.tar.gz](https://github.com/ocsigen/js_of_ocaml/files/1036357/toplevel.tar.gz)

hhugo:


hhugo:
The fix is part of js_of_ocaml 3.0

PR #601, opened 2017-06-22
Compiler: fix #599

### LINKED ISSUE #599: An equivalent of emscripten's MODULARIZE
Hi there,

Following up on the mailing list thread: `--wrap-in-func` seems to be pretty close to what I need (thanks!), except for the fact that it doesn't export the function itself in the global namespace (so I guess it could be useful from `eval`, but not from e.g. `importScripts` in a web worker.

The idea of emscripten's `modularize` is that loading the script exports (in the node.js case) or adds to the global namespace (in the brwoser case) a single function, which when called returns an instance of the program.  This is especially useful when the program has global state, since it gives you an easy way to reset it without reloading (and thus reparsing) the entire script.

Thanks!

hhugo:
@cpitclaudel, this is a quick attempt at addressing https://github.com/ocsigen/js_of_ocaml/issues/599.

### DIFF
--- a/compiler/compileArg.ml
+++ b/compiler/compileArg.ml
@@ -31,7 +31,7 @@ type t = {
   input_file : string option;
   params : (string * string) list;
   static_env : (string * string) list;
-  wrap_with_fun : bool;
+  wrap_with_fun : string option;
   (* toplevel *)
   dynlink : bool;
   linkall : bool;
@@ -99,7 +99,7 @@ let options =
   in
   let wrap_with_function =
     let doc = "Wrap the generated JavaScript code inside a function that needs to be applied with the global object." in
-    Arg.(value & flag & info ["wrap-with-fun"] ~doc)
+    Arg.(value & opt (some string) None & info ["wrap-with-fun"] ~doc)
   in
   let set_param =
     let doc = "Set compiler options." in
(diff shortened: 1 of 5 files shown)


cpitclaudel:
Looks very nice! And IIUC, I can pass it an arbitrary (possibly empty) object instead of "window" to populate that object, right?

cpitclaudel:
Tested; it works very nicely, thanks!
I also noticed after reverting to 2.8.4 that this could be emulated with `--wrap-in-fun --custom-header="var JSOO_FStar="`

hhugo:
indeed :)
Can you describe a bit more your use case ?

cpitclaudel:
Yes, of course.  I'm transpiling F* to Javascript, starting with the simplest possible interface (passing F* a bunch of command line arguments and a file), and printing the verification results.

One of the issues is that F* has a lot of global state, so calling main twice doesn't work: F* just gets very confused.  Instead, I need to reload F* every time.

One way to do this is to use a new WebWorker for each iteration; this isn't free, however, because if forces us to reload and reparse the Javascript every time.

A much better alternative is to have a factory function, FStar_JSOO, which can be called to instantiate a new, pristine instance of F*.  That factory is exactly what jsoo builds; I just need to grab it by name :)

ejgallego:
@hhugo would you mind documenting how to get the old behavior back? I'm afraid I can't understand the current documentation. What is `VAL`?

ejgallego:
In particular if using `--wrap-with-fun=` no parenthesis are generated thus using eval on the resulting code will result in a JS parsing error. Should we then do the wrapping client-side?

PR #606, opened 2017-06-30
Adds readme to examples directory

bradlangel:
* Adds a readme to the examples directory which informs users how to
  build the projects using jbuilder.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,19 @@
+# Example projects
+
+## Building from source
+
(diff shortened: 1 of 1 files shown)


bradlangel:
hmm, looks like this wasn't a good idea to change it to install as it builds the example projects which have external library dependencies. I'll modify the pull request to only include new documentation.

hhugo:
Thanks

PR #617, opened 2017-08-18
Do not add "Content-Type: multipart/form-data" to HTTP POST requests

jrochel:
Instead we let the Browser figure out the header on its own. This relieves us of implementing the "boundary=..." argument which is missing in the current implementation.

### DIFF
--- a/lib/lwt/lwt_xmlHttpRequest.ml
+++ b/lib/lwt/lwt_xmlHttpRequest.ml
@@ -170,9 +170,7 @@ let perform_raw
         | `Fields _strings ->
           override_method "POST",
           override_content_type "application/x-www-form-urlencoded"
-        | `FormData _ ->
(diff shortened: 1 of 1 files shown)


lib/lwt/lwt_xmlHttpRequest.ml
```diff
@@ -170,9 +170,7 @@ let perform_raw
         | `Fields _strings ->
           override_method "POST",
           override_content_type "application/x-www-form-urlencoded"
-        | `FormData _ ->
-          override_method "POST",
-          override_content_type "multipart/form-data"
+        | `FormData _ -> override_method "POST", None
```
vasilisp: `None` -> `content_type`. If the user really wants to provide a custom `content_type`, we should respect it. Sorry if I suggested `None` earlier :).

PR #622, opened 2017-10-04
Fix caml_ml_out_channels_list

nojb:
This PR fixes a runtime bug: the function `caml_ml_out_channels_list` should return a list of `chanid`'s (file descriptors), not a list of channel *objects*.  This function is only used in the implementation of `flush_all`, and the bug has been present for a long time, but it only became visible after the commit https://github.com/ocaml/ocaml/commit/1fc6ccb0877c282c93e78619882525fc4ab9209c#diff-9e690569dd7558fbf79622f41868ade5L335 which allowed the exception to escape.

Thanks to @Nevor for invaluable help finding the bug.


nojb:
Ping.

hhugo:
Thanks

PR #623, opened 2017-10-04
Updated doc of [mount]

pveber:
Seems a previous name (`register_autoload`) was not corrected in the function's doc.


hhugo:
Thanks

PR #624, opened 2017-10-22
Addition of Jstable.delete

rixed:
Which apparently was forgotten?


hhugo:
Thanks.
I've rename `delete` into `remove` so that it matches the `Hashtbl` signature

PR #631, opened 2017-11-20
Compile ppx with OCaml 4.06

jaapb:
This updates the ppx syntax extension to compile with OCaml 4.06. Probably can't be merged as-is, as it will break compatibility with earlier versions of OCaml.


hhugo:
https://github.com/ocsigen/js_of_ocaml/commit/cc855f3376c2b230ef802867e6191e00aad00105

hhugo:
thanks

PR #638, opened 2017-12-03
Split [@@deriving json] into own pacakge

rgrinberg:
@hhugo as we've discussed. The test suite fails for me (but it also fails on master).


hhugo:
Thanks. I've fixed the tests

PR #640, opened 2017-12-04
Build: wrap js_of_ocaml lib

hhugo:
One will have to explicitly open `Js_of_ocaml`.
@balat, @vouillon, any objection ?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@
  ** Runtime: weak.js loaded by default
  ** Compiler: compile an OCaml program into a named javascript function (fix #599)
  ** Misc: move ppx_deriving_json in its own opam package (Rudi Grinberg)
+ ** Misc: always use namespace for the js_of_ocaml library
  ** Compiler: static eval of float negation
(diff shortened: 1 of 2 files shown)


hhugo:
Should we do the same thing for other libraries ?
js_of_ocaml_lwt
js_of_ocaml_deriving
js_of_ocaml_tyxml
js_of_ocaml_toplevel

rgrinberg:
Would be really great to have this

hhugo:
@rgrinberg, which part ? Just this Pr or the extension to other libs ?

rgrinberg:
Having them all ported at once would be welcome. Btw, have you considered a shorter name such as `jsoo` perhaps? `js_of_ocaml` is quite a mouthful for source code.

hhugo:
Ack.
Let's start with just the main jsoo lib.

hhugo:
This change breaks too many packages. I will revert the change and wait for https://github.com/ocaml/dune/issues/985

PR #641, opened 2018-01-09
runtime: always include weak.js

hhugo:


hhugo:
Indeed, fixed by https://github.com/ocsigen/js_of_ocaml/commit/2ef065c032e5982a137d81966aacd34b2071c190

PR #646, opened 2018-01-31
ppx: support for optdef_prop, writeonly_prop, ..

### LINKED ISSUE #645: Add support for Js.optdef_prop in PPX syntax
It would be nice to be able to create properties of type `Js.optdef_prop` from the PPX syntax.

I open this feature request as requested by @Drup here: https://stackoverflow.com/a/48471824/1160770. :)

> I don't remember implementing, and I doubt it was added later.
> 
> IIRC, One of the reason was that there was no easy syntax to use, but I could have just used an attribute. Another was that it didn't seem that useful at the time, but I guess I was wrong. :)
> 
> Please make a feature request on the bug tracker.

hhugo:
fix #645

### DIFF
--- a/doc/manual/src/ppx.wiki
+++ b/doc/manual/src/ppx.wiki
@@ -60,8 +60,7 @@ object%js (self) (* Equivalent of this *)
 
   val z = 5 [@@writeonly] (* write-only prop *)
 
-  val t1 = 6 [@@optdef_return] (* optdef prop *)
-  val t2 = Js.undefined () [@@jsoo.optdef] (* optdef prop *)
+  val t1 = Js.undefined [@@jsoo.optdef] (* optdef prop *)
 
   method foo i = self##.y1 := self##.x1 + i
 end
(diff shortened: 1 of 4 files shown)


Drup:
From a casual read, this looks good. The implementation seems complete and straightforward. You didn't documented the new attributes, though.

hhugo:
I've extended the support for `optdef_prop`.
One can do
```ocaml
object
  val p1 = Js.undefined [@@optdef]
  val p2 = Js.def 2 [@@optdef]
  val p3 = 2 [@@optdef_return]
end
```

hhugo:
I'm not super exited by `[@@optdef_return]`. Any other proposal ?

Drup:
I suppose you mean the name of the attribute? Me neither, but I don't have a better proposal. Maybe use a `.` instead of the `_`. We could also not have it at all, since it's not really essential.

hhugo:
I will probably just remove it.

hhugo:
Thanks for the review.

PR #648, opened 2018-02-10
Internationalization API: https://www.ecma-international.org/ecma-402…

slegrand45:
https://www.ecma-international.org/ecma-402/1.0/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl


hhugo:
Thanks

PR #661, opened 2018-04-16
Added media events support.

sagotch:
See https://html.spec.whatwg.org/multipage/media.html#mediaevents


sagotch:
I may have forgot something or maybe that I am not sync, but I have `Unbound value Lwt_js_events.ended` error when compiling my project. I can not be sure this PR is good for now, even if I do not see what could be wrong in it.

hhugo:
Thanks

PR #673, opened 2018-05-15
Improve node environment detection

Tyrame:
I am using the js_of_ocaml library in a typescript project, generating CommonJS modules, and executed on a browser.
The problem is that js_of_ocaml assumes an environment supporting CommonJS is always a Node.js environment, so it uses MlNodeDevice, which requires 'fs' in its constructor.
This PR uses a detection method that is true in a Node.js environment, but not necessarily true in a CommonJS environment.

Discussion on Node.js environment detection: https://stackoverflow.com/questions/4224606/how-to-check-whether-a-script-is-running-under-node-js
The code of this PR is inspired from: https://stackoverflow.com/a/35813135


hhugo:
Thanks

PR #684, opened 2018-07-16
Fix the dependencies of the Lwt logger

vouillon:
`lwt_log` depends on `lwt_unix` which we cannot compile to JavaScript...

### DIFF
--- a/js_of_ocaml-lwt.opam
+++ b/js_of_ocaml-lwt.opam
@@ -17,6 +17,6 @@ depends: [
   "js_of_ocaml-ppx"
 ]
 
-depopts: [ "graphics" ]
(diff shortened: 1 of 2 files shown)


vouillon:
Also, we need an optional dependency on `lwt_log`, so that when installing both `lwt_log` and `js_of_ocaml-lwt`,  the package `js_of_ocaml-lwt` will provide the optional logging functionality.

balat:
Hello @hhugo !
Can you merge this and release?
We have a lot of bug reports concerning 3.2.0.
Thanks by advance!
Vincent

hhugo:
https://github.com/ocaml/opam-repository/pull/12406

PR #685, opened 2018-07-30
Lib: fix DataTransfer class interface

jjurm:
Fix DataTransfer class to match the interface: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

`DataTransfer.types` should be an array of strings


hhugo:
Thanks

PR #689, opened 2018-08-26
Lib: Fix encoding in Sys.set_channel_flusher, fix #686

### LINKED ISSUE #686: Encoding issue with toplevel/eval.ml example
I built the eval example here https://github.com/ocsigen/js_of_ocaml/tree/master/toplevel/eval and input this code into toplevel 

```
print_endline "└── ";;
```

the result looks like this: 

```
âââ
```

but the same code with jsoo's offcial toplevel build print out the correct result

I wonder that did I do wrong? 

I uploaded the build here in case you want to take a quick look at it https://eval-dnifirsfop.now.sh/

hhugo:
@thangngoc89, this should fix #686, can you confirm ?


thangngoc89:
This fixed the issue. Thank you very much !

PR #696, opened 2018-09-08
Remove camlp4 support

hhugo:


hhugo:
https://opam.ocaml.org/packages/js_of_ocaml-camlp4/ mentions that js_of_ocaml-camlp4 was downloaded 124 last month.
@balat, @Drup, are we still supporting camlp4 in other ocsigen project ?

hhugo:
Ok, I don't think it make sense to keep camlp4 support it this the repo anyway. I someone complain, I would resurrect the support in another repo.

PR #700, opened 2018-10-01
Better stub generation

rgrinberg:
This PR is twofold:

* Use dune's promotion mechanism to generate the stubs file. This makes it easier to verify that the process is correct as the generated stubs are now expected. This also makes it unnecessary for users to run any shell scripts during compilation.

* The second commit replaces the shell script with a pure OCaml script. It's a little longer, but it should be portable. Also, it outputs the stubs with slightly better formatting.


hhugo:
I would have expected to see a rule for `lib/js_of_ocaml_stubs.c` with a `(mode fallback)`

hhugo:
> Do you mean `(mode promote)`? 

probably :)

hhugo:
I've pushed a commit to use the ocaml lexer instead of a custom one. I should resolve some issue I had in https://github.com/ocsigen/js_of_ocaml/pull/671 where `external` is not on the same line as primitive name

rgrinberg:
Makes sense. That's quite a bit better.

Issue #308, opened 2015-05-21
Should charCodeAt be used for binary IO?


ejgallego:
[charCodeAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt) is pervasively used in the javascript library to read a byte from a string.

However, such use is may be inconsistent when strings are used in OCAML to contain binary data, as charCodeAt may return values higher than 256. For instance:

```
> var str = "ℕlala"
> var utf8 = unescape(encodeURIComponent(str));
> var arr = [];
> for (var i = 0; i < utf8.length; i++) { arr.push(utf8.charCodeAt(i)); };
> arr
[ 226, 132, 149, 108, 97, 108, 97 ]
> utf8
'âlala'
> str.charCodeAt(1);
108
> utf8.charCodeAt(1);
132
> arr[1];
132
```

This may break some binary reading done in Ocaml, like in dynlink.ml, etc...

After a bit of thinking about Ocaml strings, it is not clear to me what should be the best choice here. One option is to implement strings always as byte arrays in JS, replacing charCodeAt by regular array access.

hhugo:
`pervasively used` when the string is correctly encoded (sequence of byte).
I don't know how you get your strings, but I suspect they are sent to the client not been utf8 encoded.

see string representation:
https://github.com/ocsigen/js_of_ocaml/blob/master/runtime/mlString.js#L25

and the utf8 encoding
https://github.com/ocsigen/js_of_ocaml/blob/master/lib/deriving_json/deriving_Json.ml#L198

hhugo:
I'll push an example of dynlink in the toplevel later today.

vouillon:
If you are getting your data through XmlHttpRequest, you can ask for the response to be stored in an ArrayBuffer, and then copy the buffer contents into an OCaml string with the following function:
```
let string_of_buffer b =
  let a = jsnew Typed_array.uint8Array_fromBuffer(b) in
  let len = a##length in
  let s = String.create len in
  for i = 0 to len - 1 do
    s.[i] <- Char.chr (Typed_array.unsafe_get a i)
  done;
  s
```
(See https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data )

In particular, the function `load_resource` in `toplevel/toplevel.cppo.ml` only works with *ASCII* data.

ejgallego:
Thank you for the explanation and help.

[...]

PR #703, opened 2018-10-08
WIP Added filesystem support via requestFileSystemSync (need help)

sagotch:
This PR enable I/O to filesystem in context supporting `requestFileSystemSync` (which is chrome browser, in a web worker). I'm using it in a cordova app.

It almost works, but data get corrupted somewhere. Probably on writing since reading should work as you can see later in this post.

**I can't find what function to use to encode/decode data, this is why I need help.**

Some files are marshaled data.

This is how I initialize files:

On javascript side (main thread):
```javascript
    input.onchange = function () {
        var files = input.files ;
        for (var i = 0; i < files.length; i++) {
            (function (i) {
                var file = files[i] ;
                console.log ("Reading " + file.name) ;
                var fr = new FileReader () ;
                fr.onload = function () { data.push({ name: file.name, data: fr.result }) ; }
                fr.readAsBinaryString (file) ;
            }) (i) ;
        }
        var loop = function () {
            if (data.length == files.length) {
                console.log ('Sending') ;
                console.log (data) ;
                w.postMessage ({type: 'loadFiles', data:data}) ;
                w.postMessage ({type: 'openBase'})
            }
            else { console.log ('loop ' + data.length + '/' + files.length) ; setTimeout (loop, 500) ; }
        } ;
        loop () ;
    }
```

On OCaml side (Web Worker `w` in the main thread)
```ocaml
  Worker.set_onmessage (fun e ->
      Firebug.console##log("Message received from main script") ;
      Firebug.console##log(e) ;
      Firebug.console##log(Js.typeof e) ;
      match Js.to_string e##.type_ with
      | "loadFiles" ->
        e##.data##forEach
          (Js.wrap_callback @@ fun x _i ->
           let name = Js.to_string @@ Js.Unsafe.get x (Js.string "name") in
           let data = Js.to_bytestring @@ Js.Unsafe.get x (Js.string "data") in
           let oc = open_out @@ bdir ^ name in
           let () = output_string oc data in
           let () = close_out oc in
           print_endline @@ "Saved " ^ bdir ^ name) ;
...
```
---

If I use the fake filesystem, I can read the files with no error, but it is not a suitable solution for my app, because I need writing rights and persistent storage.
```
let file_reader file callback =
  let reader = new%js File.fileReader in
  let () = reader##.onload := Dom.handler (fun e ->
      Js.Opt.case
        (e##.target)
        (fun () -> Js.bool false)
        (fun target ->
           Js.Opt.case
             (File.CoerceTo.string target##.result)
             (fun () -> Js.bool false)
             (fun result ->
                callback (Js.to_bytestring result) ) ) ) (* caml_new_string *)
  in
  reader##readAsBinaryString file

...
   input##.onchange := Dom.handler begin fun _e ->
       Js.Optdef.case (input##.files) (fun () -> failwith __LOC__) @@ fun files ->
       let len = files##.length in
       for i = 0 to len - 1 do
         Js.Opt.case (files##item i) (fun () -> failwith __LOC__) @@ fun file ->
         let name = Js.to_string file##.name in
         file_reader (Js.Unsafe.coerce file)
           (fun blob ->
              Sys_js.create_file ~name:fname ~content:blob ;
              (List.assoc name check)##setAttribute
                (Js.string "style")
                (Js.string "background-color:green;color:white;") ;
              Js.bool true)

...

```

### DIFF
--- a/runtime/dune
+++ b/runtime/dune
@@ -8,7 +8,7 @@
    (run %{bin:jsoo_link} -o %{targets}
     %{dep:mlString.js} %{dep:ieee_754.js} %{dep:int64.js} %{dep:md5.js} %{dep:marshal.js}
     %{dep:lexing.js} %{dep:parsing.js} %{dep:bigarray.js} %{dep:unix.js}
-    %{dep:stdlib.js} %{dep:io.js} %{dep:fs.js} %{dep:fs_fake.js} %{dep:fs_node.js} %{dep:jslib.js}
+    %{dep:stdlib.js} %{dep:io.js} %{dep:fs.js} %{dep:fs_fake.js} %{dep:fs_node.js} %{dep:fs_fss.js} %{dep:jslib.js}
     %{dep:jslib_js_of_ocaml.js}
     %{dep:internalMod.js} %{dep:gc.js} %{dep:polyfill/json2.js} %{dep:bigstring.js} %{dep:weak.js})))
 
@@ -22,6 +22,7 @@
     fs.js
     fs_fake.js
     fs_node.js
+    fs_fss.js
     gc.js
     graphics.js
     ieee_754.js
--- a/runtime/fs.js
+++ b/runtime/fs.js
@@ -20,8 +20,11 @@
 ///////////// Dummy filesystem
 
 //Provides: caml_current_dir
+//Requires: fs_fss_supported
 if(joo_global_object.process && joo_global_object.process.cwd)
   var caml_current_dir = joo_global_object.process.cwd().replace(/\\/g,'/');
+else if (fs_fss_supported())
+  var caml_current_dir = "/"    // FIXME
 else
   var caml_current_dir =  "/static";
 if(caml_current_dir.slice(-1) !== "/") caml_current_dir += "/"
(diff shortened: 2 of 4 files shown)


sagotch:
NB: I can read some part of the data. Marshaled data are described [here](https://github.com/geneweb/geneweb/blob/master/internal/database.ml) as:
```
    base - the base itself
       magic number (magic_gwb)                 : string of length 8
       number of persons                        : binary_int
       number of families                       : binary_int
       number of strings                        : binary_int
       persons array offset in file             : binary_int
       ascends array offset in file             : binary_int
       unions array offset in file              : binary_int
       families array offset in file            : binary_int
       couples array offset in file             : binary_int
       descends array offset in file            : binary_int
       strings array offset in file             : binary_int
       notes origin file                        : value
       persons array                            : value
       ascends array                            : value
       unions array                             : value
       families array                           : value
       couples array                            : value
       descends array                           : value
       strings array                            : value
```
I print `number of persons` and `number of families` and they are correct. The a problem appear when I try to read from an array. (from person array for example).

sagotch:
Could the `seek_in` function be the problem?

sagotch:
Loading text files seem to work fine...

sagotch:
Might be interesting: https://github.com/ocsigen/js_of_ocaml/issues/308#issuecomment-105914250

sagotch:
Files have the right size.

But... `input_binary_int` seems to always return 0.

PR #704, opened 2018-10-09
WIP Filesystem support via requestFileSystemSync (need help)

sagotch:
Copy of https://github.com/ocsigen/js_of_ocaml/pull/703, but from a branch instead of master

This PR enable I/O to filesystem in context supporting `requestFileSystemSync` (which is chrome browser, in a web worker). I'm using it in a cordova app.

It almost works, but data get corrupted somewhere. Probably on writing since reading should work as you can see later in this post.

**I can't find what function to use to encode/decode data, this is why I need help.**

Some files are marshaled data.

This is how I initialize files:

On javascript side (main thread):
```javascript
    input.onchange = function () {
        var files = input.files ;
        for (var i = 0; i < files.length; i++) {
            (function (i) {
                var file = files[i] ;
                console.log ("Reading " + file.name) ;
                var fr = new FileReader () ;
                fr.onload = function () { data.push({ name: file.name, data: fr.result }) ; }
                fr.readAsBinaryString (file) ;
            }) (i) ;
        }
        var loop = function () {
            if (data.length == files.length) {
                console.log ('Sending') ;
                console.log (data) ;
                w.postMessage ({type: 'loadFiles', data:data}) ;
                w.postMessage ({type: 'openBase'})
            }
            else { console.log ('loop ' + data.length + '/' + files.length) ; setTimeout (loop, 500) ; }
        } ;
        loop () ;
    }
```

On OCaml side (Web Worker `w` in the main thread)
```ocaml
  Worker.set_onmessage (fun e ->
      Firebug.console##log("Message received from main script") ;
      Firebug.console##log(e) ;
      Firebug.console##log(Js.typeof e) ;
      match Js.to_string e##.type_ with
      | "loadFiles" ->
        e##.data##forEach
          (Js.wrap_callback @@ fun x _i ->
           let name = Js.to_string @@ Js.Unsafe.get x (Js.string "name") in
           let data = Js.to_bytestring @@ Js.Unsafe.get x (Js.string "data") in
           let oc = open_out @@ bdir ^ name in
           let () = output_string oc data in
           let () = close_out oc in
           print_endline @@ "Saved " ^ bdir ^ name) ;
...
```
---

If I use the fake filesystem, I can read the files with no error, but it is not a suitable solution for my app, because I need writing rights and persistent storage.
```
let file_reader file callback =
  let reader = new%js File.fileReader in
  let () = reader##.onload := Dom.handler (fun e ->
      Js.Opt.case
        (e##.target)
        (fun () -> Js.bool false)
        (fun target ->
           Js.Opt.case
             (File.CoerceTo.string target##.result)
             (fun () -> Js.bool false)
             (fun result ->
                callback (Js.to_bytestring result) ) ) ) (* caml_new_string *)
  in
  reader##readAsBinaryString file

...
   input##.onchange := Dom.handler begin fun _e ->
       Js.Optdef.case (input##.files) (fun () -> failwith __LOC__) @@ fun files ->
       let len = files##.length in
       for i = 0 to len - 1 do
         Js.Opt.case (files##item i) (fun () -> failwith __LOC__) @@ fun file ->
         let name = Js.to_string file##.name in
         file_reader (Js.Unsafe.coerce file)
           (fun blob ->
              Sys_js.create_file ~name:fname ~content:blob ;
              (List.assoc name check)##setAttribute
                (Js.string "style")
                (Js.string "background-color:green;color:white;") ;
              Js.bool true)

...

```
---

<strike>
NB: I can read some part of the data. Marshaled data are described <a href="https://github.com/geneweb/geneweb/blob/master/internal/database.ml">here</a> as:
<code><pre>
base - the base itself
       magic number (magic_gwb)                 : string of length 8
       number of persons                        : binary_int
       number of families                       : binary_int
       number of strings                        : binary_int
       persons array offset in file             : binary_int
       ascends array offset in file             : binary_int
       unions array offset in file              : binary_int
       families array offset in file            : binary_int
       couples array offset in file             : binary_int
       descends array offset in file            : binary_int
       strings array offset in file             : binary_int
       notes origin file                        : value
       persons array                            : value
       ascends array                            : value
       unions array                             : value
       families array                           : value
       couples array                            : value
       descends array                           : value
       strings array                            : value</pre></code>
</pre></code>

I print `number of persons` and `number of families` and they are correct. The a problem appear when I try to read from an array. (from person array for example).
</strike>

Actually not true. I do not know where do the right value come from, but definitely not from unmarshalling. Each `input_binary_int` returns 0 where it shouldn't (checked with native version).

---

Could the <strike>`seek_in`</strike> `input_binary_int` or `output_binary_int` function be the problem?

---

Might be interesting: https://github.com/ocsigen/js_of_ocaml/issues/308#issuecomment-105914250

<code><pre>let string_of_buffer b =
   let a = jsnew Typed_array.uint8Array_fromBuffer(b) in
   let len = a##length in
   let s = String.create len in
   for i = 0 to len - 1 do
     s.[i] <- Char.chr (Typed_array.unsafe_get a i)
   done;
   s</pre></code>

---

Files have the right size. ASCII text files work fine. `input_binary_int` return `0` while it should not.
Loading text files seem to work fine...

### DIFF
--- a/runtime/dune
+++ b/runtime/dune
@@ -8,7 +8,7 @@
    (run %{bin:jsoo_link} -o %{targets}
     %{dep:mlString.js} %{dep:ieee_754.js} %{dep:int64.js} %{dep:md5.js} %{dep:marshal.js}
     %{dep:lexing.js} %{dep:parsing.js} %{dep:bigarray.js} %{dep:unix.js}
-    %{dep:stdlib.js} %{dep:io.js} %{dep:fs.js} %{dep:fs_fake.js} %{dep:fs_node.js} %{dep:jslib.js}
+    %{dep:stdlib.js} %{dep:io.js} %{dep:fs.js} %{dep:fs_fake.js} %{dep:fs_node.js} %{dep:fs_fss.js} %{dep:jslib.js}
     %{dep:jslib_js_of_ocaml.js}
     %{dep:internalMod.js} %{dep:gc.js} %{dep:polyfill/json2.js} %{dep:bigstring.js} %{dep:weak.js})))
 
@@ -22,6 +22,7 @@
     fs.js
     fs_fake.js
     fs_node.js
+    fs_fss.js
     gc.js
     graphics.js
     ieee_754.js
--- a/runtime/fs.js
+++ b/runtime/fs.js
@@ -20,8 +20,11 @@
 ///////////// Dummy filesystem
 
 //Provides: caml_current_dir
+//Requires: fs_fss_supported
 if(joo_global_object.process && joo_global_object.process.cwd)
   var caml_current_dir = joo_global_object.process.cwd().replace(/\\/g,'/');
+else if (fs_fss_supported())
+  var caml_current_dir = "/"    // FIXME
 else
   var caml_current_dir =  "/static";
 if(caml_current_dir.slice(-1) !== "/") caml_current_dir += "/"
@@ -55,14 +58,16 @@ function caml_make_path (name) {
 }
 
 //Provides:jsoo_mount_point
-//Requires: MlFakeDevice, MlNodeDevice, caml_root, fs_node_supported
+//Requires: MlFakeDevice, MlNodeDevice, MlFssDevice, caml_root, fs_node_supported, fs_fss_supported
 var jsoo_mount_point = []
 if (fs_node_supported()) {
     jsoo_mount_point.push({path:caml_root,device:new MlNodeDevice(caml_root)});
+} else if (fs_fss_supported()) {
+    jsoo_mount_point.push({path:caml_root,device:new MlFssDevice(caml_root)});
 } else {
     jsoo_mount_point.push({path:caml_root,device:new MlFakeDevice(caml_root)});
 }
-jsoo_mount_point.push({path:caml_root+"static/", device:new MlFakeDevice(caml_root+"static/")});
+// jsoo_mount_point.push({path:caml_root+"static/", device:new MlFakeDevice(caml_root+"static/")});
 
 //Provides:caml_list_mount_point
 //Requires: jsoo_mount_point, caml_new_string
--- a/runtime/fs_fss.js
+++ b/runtime/fs_fss.js
@@ -1,0 +1,126 @@
+// Js_of_ocaml runtime support
+// http://www.ocsigen.org/js_of_ocaml/
+// Copyright (C) 2014 Jérôme Vouillon, Hugo Heuzard
+// Laboratoire PPS - CNRS Université Paris Diderot
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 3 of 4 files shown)


vouillon:
Do you have some standalone code that I could use to reproduce the problem?

sagotch:
It depends on what you call "standalone". You can use the repo of my app (https://github.com/geneanet/gwb.js), but you need some dependencies...

```
opam pin add -ny -k git geneweb https://github.com/geneanet/geneweb#gwb.js
opam pin add -ny -k git js_of_ocaml-compiler https://github.com/geneanet/js_of_ocaml#fss
opam pin add -ny -k git js_of_ocaml-ppx https://github.com/geneanet/js_of_ocaml#fss
opam pin add -ny -k git js_of_ocaml https://github.com/geneanet/js_of_ocaml#fss
opam pin add -ny -k git jingoo https://github.com/geneanet/jingoo.git#gwb.js
opam update
opam install js_of_ocaml-ppx js_of_ocaml geneweb jingoo
```

```
git clone https://github.com/geneanet/gwb.js
cd gwb.js
git checkout fss
```

```
make cordova-init
make (clean) build (cordova-clean) cordova-build cordova-run-browser
```

Once you are on the loading page, open the console and input files from `src/grimaldi.gwb` in the input tag.

Failure raised here https://github.com/geneanet/geneweb/blob/gwb.js/internal/database.ml#L515, alled by https://github.com/geneanet/gwb.js/blob/fss/src/ml/gwb.ml#L125, and if I recall correctly, all these `input_binary_int` https://github.com/geneanet/geneweb/blob/gwb.js/internal/database.ml#L622 return `0` which is not what is expected.

sagotch:
I just fixed the opam file on https://github.com/geneanet/geneweb#gwb.js

hhugo:
@sagotch, do you still have issues with this PR ?

sagotch:
I don't know, probably.

I could not figure out where was the error. I gave up and a lot of things has changed in the code now, so I won't be able to reproduce my use cas easily.

This PR should probably be closed now, I tried this solution during a POC application development, but now time is missing for me to continue this work...

hhugo:
Ok. Sorry for not looking at that earlier.
I'm going to leave this open a bit longer and see if I can add some tests.

sagotch:
I just re-pushed old branch so my previous test case might actually work (I am not sure about it at all)

hhugo:
> Non-standard
> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.


Based on the non-standard nature of this API and the fact that this is not a blocker for you anymore, I'm gonna close this PR.

Feel free to reopen if ever needed.

sagotch:
No problem, we can close.

Just a note : The non-standard nature is obviously a problem, but as it is used in webkit it can be used in mobile apps via cordova, which was the goal here.

PR #706, opened 2018-10-23
Enable dune wrapped in transition mode

wokalski:
Fixes https://github.com/janestreet/incr_dom/issues/5
Fixes https://github.com/janestreet/virtual_dom/issues/3

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,1 +1,1 @@
-(lang dune 1.1)
+(lang dune 1.2)
--- a/lib/dune
+++ b/lib/dune
@@ -2,7 +2,7 @@
  (name js_of_ocaml)
  (public_name js_of_ocaml)
  (libraries uchar bytes)
- ;; (wrapped (transition "Will be removed past 2019-01-01"))
+ (wrapped (transition "Will be removed past 2019-01-01"))
  (c_names js_of_ocaml_stubs)
  (preprocess (pps ppx_js_internal)))
 


lib/js_of_ocaml.ml
```diff
@@ -17,7 +17,7 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
-module CSS = CSS
+module CSS = CSS__Internal
```
rgrinberg: You should not use `__` in your own module names. This is just a naming convention that's taken up by dune's wrapping mechanism. You should use some name like `css_internal` or w/e

rgrinberg:
Note that this is a breaking change. As it means that the CSS module will no longer accessible without the alias.

lib/js_of_ocaml.ml
wokalski: @hhugo is there any name for the module you’d prefer?

hhugo:
I would prefer to do the change in two steps, and maybe do a release in between.
1. enable to wrapping in dune
2. rename the css module (cSS0.ml maybe)

rauanmayemir:
@hhugo Step 1 is done. If you merge it, I'll add the rename on top of it.

wokalski:
I think it might not have been clear. The change has been performed by @rauanmayemir. Please check and let me know if everything looks good.

hhugo:
Thanks

PR #710, opened 2018-11-05
Improved detection of global object

### LINKED ISSUE #699: joo_global_object is undefined when bundled in strict mode
jsoo uses this compilation scheme to detect the global object:

```
(function(joo_global_object)
   {"use strict";
       console.log(joo_global_object);
       // ...
   }
  (function(){return this}()));
```

This fails when the generated JavaScript is bundled together with other JavaScript files that have a global `"use strict";` stanza, as `this` doesn't become the global object in strict mode.

copy:
Generate `Function("return this")()` instead of `function(){return this}()`. The former works in strict mode (for example under `node --use_strict` or when bundled using webpack), the latter doesn't.

Fixes #699.

### DIFF
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -385,14 +385,13 @@ let pack ~global js =
       | `Custom name ->
         J.ECall (f, [J.EVar (J.S {J.name;var=None})], J.N)
       | `Auto ->
-        let global =
-          J.ECall (
-            J.EFun (None, [], [
-              J.Statement (
-                J.Return_statement(
-                  Some (J.EVar (J.S {J.name="this";var=None})))),
-              J.N
-            ], J.N), [], J.N) in
+          let global =
+            J.ECall (
+              J.ECall (
+                (J.EVar (J.S {J.name="Function";var=None})),
+                [EStr ("return this", `Bytes)],
+                J.N),
+              [], J.N) in
         J.ECall (f, [global], J.N)
     in
     match global with


hhugo:
Thanks

TyOverby:
Because the function constructor is performing an `eval`, code that contains `Function("return this")` is not runnable under Content Security Policy without whitelisting "unsafe-eval"

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' blob: filesystem: chrome-extension-resource:".
```

Because having rigid Content Security Policies (especially WRT eval) is an important factor for web security, I will investigate other methods for achieving this behavior without preventing Js_of_ocaml from being run with heavier security guarantees.

hhugo:
@copy, I reverted this change. @TyOverby is working on some PR that could solve your original issue.

copy:
@hhugo @TyOverby  Good point. Glad this was reverted before breaking anyone's code.

@TyOverby  Do you have something coming up for this? From what I can tell:

- `try { Function("return this")() } catch() { /* fallback */ }` works, but one still needs to find a fallback that works in all environments with unsafe-eval
- `typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : typeof window !== 'undefined' ? window : {}` ([source](https://www.contentful.com/blog/2017/01/17/the-global-object-in-javascript/#evaluating-the-global-object-is-harder-than-expected)) is for example used by browserify, but is not as reliable as the other method, for example in the presence of a variable named `global`

TyOverby:
I just read [this](https://mathiasbynens.be/notes/globalthis) today.  Maybe we should use something similar?

PR #715, opened 2018-11-16
Move cSS to cSS0 to avoid name clashing

wokalski:
Admittedly this change is arguable. The CSS module has
been renamed to avoid clashing with virtual_dom's
css module on case insensitive filesystems. Since
jsoo will eventually be namespaced I think it
makes sense to perform this change here.

The module has been renamed to CSS0 as per @hhugo 's suggestion


wokalski:
Closing (explained here: https://github.com/janestreet/virtual_dom/issues/3)

Issue #714, opened 2018-11-13
Reproducible builds


xclerc:
While trying to make builds reproducible, I stumbled on
a problem related to source maps.

The JavaScript code generated by `js_of_ocaml` is 
reproducible, but source maps are not because they
embed the absolute paths to sources.

In the `ocamlc`/`ocamlopt` compilers, the same issue
has to be dealt with (for debug info), and they use the
`BUILD_PATH_PREFIX_MAP` [1] that basically
defines a substitution over path prefixes in order to
control reproducibility. The implementation can be
found in [2, 3, 4, 5].

However, it is not clear to me whether it would be the
way to go for `js_of_ocaml`. In particular, it makes
sense for `ocamlopt` because when using `gdb`, one
can tell `gdb` to perform the inverse substitution to
recover paths that make sense for the tool. If the
tools consuming source maps do not provide such
a feature, it will practically make the information useless.

Let me know if you need further information / I can be
of any help.

[1] https://wiki.debian.org/ReproducibleBuilds/BuildPathProposal
[2] https://github.com/ocaml/ocaml/pull/1515
[3] https://github.com/ocaml/ocaml/pull/1856
[4] https://github.com/ocaml/ocaml/pull/1869
[5] https://github.com/ocaml/ocaml/pull/1930

xclerc:
[This branch](https://github.com/xclerc/js_of_ocaml/tree/build_path_prefix_map) seems to solve the problem by simply rewriting
absolute path according to `BUILD_PATH_PREFIX_MAP` in the
`Source_map_yojson.ml.in` file.

PR #719, opened 2018-11-26
Support reproducible builds by using "BUILD_PATH_PREFIX_MAP"

xclerc:


xclerc:
Tentative fix for https://github.com/ocsigen/js_of_ocaml/issues/714.

hhugo:
I get 
```
Error: Files /home/hugo/.opam/4.07.1/lib/build_path_prefix_map/build_path_prefix_map.cmxa
       and /home/hugo/.opam/4.07.1/lib/ocaml/compiler-libs/ocamlcommon.cmxa
       both define a module named Build_path_prefix_map
```

hhugo:
@gasche, it seems that build_path_prefix_map cannot be used together with compiler-libs.common because of name conflict

xclerc:
I have pushed a new version that (i) vendors the module
(avoiding a conditional dependency), and (ii) applies the
rewrite to all path values in the source map.

(I still have to double-check there is no regression.)

xclerc:
The Travis builds seem to be broken after https://github.com/ocsigen/js_of_ocaml/commit/c0fd99bf852ca010baa2bfa2affdd1811f031e5f,
because the opam files are now 2.0 while the opam
used by the Travis jobs is still 1.2.

hhugo:
I fixed travis. Can you rebase this PR ?

xclerc:
Commit https://github.com/ocsigen/js_of_ocaml/pull/719/commits/a356e65eb716494ba045e5b1c44429e3752628ec adds a test demonstrating the effect of this PR.

It is not very satisfying, most notably because it goes through a
shell script to execute `js_of_ocaml`. It does so to access the
current working directory, and also to ensure that only this part of
the test is affected by the `BUILD_PATH_PREFIX_MAP` environment
variable.

hhugo:
merged manually. Thanks

xclerc:
Great; thanks!

Issue #519, opened 2016-09-05
Svg elements or not compatible with eventTarget


Chimrod:
According to the [w3c](https://www.w3.org/TR/SVG/shapes.html), SVG shapes should handle events like onclick, onmouseover and so on.

But the svg elements (ie [polygon](http://ocsigen.org/js_of_ocaml/2.8.1/api/Dom_svg.polygonElement-c)) does not inherit from Dom_html.​eventTarget. And this exemple does not compile :

```ocaml
Lwt_js_events.clicks dom_polygon ( fun mouseEvent _ ->
  Js.Opt.iter mouseEvent##.target (fun target ->
    target##.style##.fill := (Js.string "blue");
  )
)
```

Is there a reason why ? Or is there another way to achieve this ?

hhugo:
I didn't really look at your issue, but bindings in js_of_ocaml are surely not exhaustive, feel free to open a pull request to fix this.

PR #749, opened 2019-02-21
Let all SVG elements respond to "global event attributes."

TypesLogicsCats:
I am writing a program that requires SVG elements to handle events. I saw #519 and decided to make a contribution.

Skimming through https://www.w3.org/TR/SVG2/, it seems that every SVG element supports "global event attributes," so I figured that it was safe for me to make `Dom_svg.element` implement `Dom_html.eventTarget`. However, I may have missed some elements that do not support the interface. (Previously, a few SVG elements individually inherited `Dom_html.eventTarget`.)

However, `Dom_html.eventTarget` lists a few events (the "animation" ones) that are not "global event attributes." I factored out those events into a new class called `Dom_html.htmlEventTarget`.

I would like to comment that `Dom_html.event` inherits from `Dom.event` but sets the generic type parameter to `Dom_html.element`. Does this design make sense if `Dom_svg.element` can handle these events as well?

I assume that these changes work, but I want to add some tests just in case. Because events require interactivity, I think that I should mock up a sample web page that uses SVG events, but `lib/tests` doesn't seem to support interactivity. What is the best way for me to test this change?

### DIFF
--- a/lib/js_of_ocaml/dom_html.ml
+++ b/lib/js_of_ocaml/dom_html.ml
@@ -468,6 +468,11 @@ and eventTarget =
     method ondrag : ('self t, dragEvent t) event_listener writeonly_prop
 
     method ondrop : ('self t, dragEvent t) event_listener writeonly_prop
+  end
+
+and htmlEventTarget =
+  object
+    inherit eventTarget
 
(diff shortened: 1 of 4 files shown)


hhugo:
Closing this PR in favor of https://github.com/ocsigen/js_of_ocaml/pull/837

PR #756, opened 2019-03-13
Fix Dom_svg.createForeignObject

TheAspiringHacker:
Due to a typo ("foreignobject" instead of "foreignObject"),
Dom_svg.createForeignObject did not work. This PR fixes the
typo.


hhugo:
thanks

PR #764, opened 2019-04-08
prevent addition with a positively-signed number from being coalesced…

TyOverby:
… into a postfix addition

Also adds expect tests


TyOverby:
@hhugo: I kinda just made an educated guess about the opam file change.  Please review that file with extra scrutiny.

PR #767, opened 2019-04-12
Preserve float type

TyOverby:
This PR preserves "floatness" through the compiler by recording things that look like ints in the parser, and then having different variants for Float vs Int.  The reason for this change is that even though all numbers are floats in the Javascript spec, JS JIT compilers will specialize on whether the number is int-like or float-like.  Initially, the way that they get this information is by [looking at the source code](https://searchfox.org/mozilla-central/source/js/src/frontend/ParseNode.h#1455).

Depends on https://github.com/ocsigen/js_of_ocaml/pull/766


TyOverby:
@hhugo: looks like Int64.equal doesn't exist on 414.1

should I change the PR to use poly-equal?

hhugo:
yes

hhugo:
Should we just make ENum store the string directly.

compiler/lib/generate.ml
```diff
@@ -235,9 +235,13 @@ let one = int 1
 let zero = int 0
 
 let plus_int x y =
+  (* In general, adding Javascript numbers together in the compiler is not safe
+     as we don't have a way to emulate actual Javascript addition. *)
```
vouillon: This is 32-bit integer addition (implemented as `(x + y) | 0`), so there is no issue in having the compiler performing it.

compiler/lib/javascript.ml
```diff
@@ -116,7 +116,8 @@ and property_name_and_value_list = (property_name * expression) list
 and property_name =
   | PNI of identifier
   | PNS of string
-  | PNN of float
+  | PNInt of Int64.t
```
vouillon: Should that be `Int32.t`? JavaScript does not have 64-bit integers.

hhugo:
@TyOverby, I took the liberty to push one commit on top of your branch.

@TyOverby, @vouillon, how do you feel about that last commit ?

compiler/lib/generate.ml
```diff
@@ -235,9 +235,13 @@ let one = int 1
...
   match x, y with
-  | J.ENum 0., x | x, J.ENum 0. -> x
-  | J.ENum x, J.ENum y -> J.ENum Int32.(to_float (add (of_float x) (of_float y)))
+  | J.EInt i, x when Int64.equal i Int64.zero -> x
```
hhugo: FYI, `| J.EInt 0L, x -> x`

hhugo:
I've just rebased this PR

compiler/lib/javascript.ml
TyOverby: This was initially to represent values that look like integers, which could include numbers that are much higher than 32-bit, but which still looked like integers in the source.  (Not that it matters anymore, the changes that Hugo and I talked about and which he implemented is a much better design IMO)

PR #768, opened 2019-04-12
Bugfix: avoid redundant conversion to string.

corwin-of-amber:
When a file is created using `UInt8Array` (or `ArrayBuffer`), `res != 0` invokes `toString`, which forces conversion to bytes (and copying the entire data).

### DIFF
--- a/runtime/fs_fake.js
+++ b/runtime/fs_fake.js
@@ -32,7 +32,7 @@ MlFakeDevice.prototype.nm = function(name) {
 MlFakeDevice.prototype.lookup = function(name) {
   if(!this.content[name] && this.lookupFun) {
     var res = this.lookupFun(caml_new_string(this.root), caml_new_string(name));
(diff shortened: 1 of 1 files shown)


runtime/fs_fake.js
```diff
@@ -32,7 +32,7 @@ MlFakeDevice.prototype.nm = function(name) {
 MlFakeDevice.prototype.lookup = function(name) {
   if(!this.content[name] && this.lookupFun) {
     var res = this.lookupFun(caml_new_string(this.root), caml_new_string(name));
-    if(res != 0) this.content[name]=new MlFakeFile(res[1]);
+    if(res) this.content[name]=new MlFakeFile(res[1]);
```
hhugo: Would `typeof res != "number"` be more explicity  ?

corwin-of-amber: I would suggest `res !== 0` if you want to make it clear.

hhugo:
I've pushed the fix manually on master. Thanks a lot.

Issue #777, opened 2019-04-24
File offset is unreliable in buffered output channels


corwin-of-amber:
In `caml_ml_pos_out` and `caml_ml_seek_out` (and their int64 variants), the file offset is manipulated directly, without taking into account data that has been written to the buffer and not flushed.

This causes the offset to be slightly off, and/or some consequent buffer data to be flushed in the wrong position.

I have a branch fixing this and will create a PR.

hhugo:
I see similar behavor with the bytecode compiler.
The following program output `004`

```ocaml
    let oc = open_out "a.txt"
    let () = print_int (out_channel_length oc)
    let () = output_string oc "test"
    let () = print_int (out_channel_length oc)
    let () = flush oc
    let () = print_int (out_channel_length oc)
```

hhugo:
Correction. I have similar behavior with the ocaml toplevel only.

hhugo:
I'm not able to reproduce the problematic behavior anymore.
I've created a test in https://github.com/ocsigen/js_of_ocaml/pull/779

@corwin-of-amber, do you have any example in hand ?

corwin-of-amber:
My example is a bit big (Coq's `.vo` serialization routine), but I will try to make a small test case based on that.

hhugo:
> I'm not able to reproduce the problematic behavior anymore.

To be clear, I'm not able to observe any behavior difference between js_of_ocaml, bytecode and native code.

corwin-of-amber:
Sure, I will test to see that the buggy behavior is not produced by the bytecode and native version of the code.

corwin-of-amber:
This example is perhaps too smart for its own good, the bug can be seen at the first `pos_out` return value; I include the remainder for completeness' sake.

[chan_pos.ml.txt](https://github.com/ocsigen/js_of_ocaml/files/3117927/chan_pos.ml.txt)

In bytecode and native (32-bit), the output is
```
start=4
start=59
```

And it generates the following `out.txt`:
```
% hexdump out.txt
0000000 00 00 22 c4 00 00 00 2b 84 95 a6 be 00 00 00 0f
0000010 00 00 00 04 00 00 00 0c 00 00 00 0a a0 26 6f 75
0000020 74 70 75 74 a0 24 64 61 74 61 40 d1 58 c7 06 8e
0000030 72 93 46 91 09 16 8d 9c d1 f1 a9 00 00 00 61 84
[...]

PR #778, opened 2019-04-24
Runtime: flush output buffer before accessing offset.

corwin-of-amber:
The simplest solution to #777 would be to flush the buffer first and ask questions later.

I suspect that any performance degradation resulting from this would be unnoticeable.

### DIFF
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -394,26 +394,32 @@ function caml_output_value (chanid,v,_flags) {
 
 
 //Provides: caml_ml_seek_out
-//Requires: caml_ml_channels
+//Requires: caml_ml_channels, caml_ml_flush
 function caml_ml_seek_out(chanid,pos){
+  caml_ml_flush(chanid);
   caml_ml_channels[chanid].offset = pos;
   return 0;
 }
 
 //Provides: caml_ml_seek_out_64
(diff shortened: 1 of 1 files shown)


runtime/io.js
```diff
@@ -396,24 +396,30 @@ function caml_output_value (chanid,v,_flags) {
 //Provides: caml_ml_seek_out
 //Requires: caml_ml_channels
```
hhugo: Can you update the `//Requires:` bits ?

TyOverby:
@corwin-of-amber, do you have a repro of this bug that could be added as a test?  I've recently expanded the capabilities of unit testing in the compiler and runtime, and this seems like a great test to see if we've covered some common use-cases.

For examples, see the `/compiler/tests/` directory.

hhugo:
I've tried to come up with a test without success https://github.com/ocsigen/js_of_ocaml/pull/779

runtime/io.js
corwin-of-amber: Added them. Sorry. 🤕

corwin-of-amber:
I've added the test in #777.

hhugo:
Thanks a lot. I've merged your fixes together with some tests. I'm closing this one.

PR #780, opened 2019-04-26
Fix: don't generate source if no-source-map passed

### LINKED ISSUE #752: Source map is generated in the presence of --no-source-map
To reproduce:

```
ocamlbuild test.byte && js_of_ocaml --source-map-inline --no-source-map test.byte && tail test.js
```

With an empty test.ml.

dhcmrlchtdj:
Fix #752


hhugo:
thanks a lot

PR #795, opened 2019-05-04
Tweak locations

Drup:
This is a small patch to slightly tweak locations in js files. There are two objectives:
- Add more locations in the JS AST, notably for variables, which allow to easily look them up in srcmaps
- Try to Add more srcmap infos on linked js files.

We could probably do more on the topic, but it's sufficient for the current side-project I'm working on (which I'll show soon :p)


Drup:
A few things I noticed:
- sourcemaps seems mostly empty when `--pretty` is *not* enabled. Is that intended ?
- Inlining and lifting seems to completely forget about source locations
- There seems to be something wrong with functors. Notably the function representing `Map.Make` lost all his source information

I wouldn't mind some pointers to try to approach those, if you don't feel like tackling them yourself. The jsoo compiler codebase is a bit intimidating. :p

hhugo:
https://github.com/ocsigen/js_of_ocaml/issues/799 should help with testing

Drup:
Do we really need to wait for that in order to merge ? It seems pretty independent to me.

compiler/lib/javascript.ml
```diff
@@ -129,7 +129,10 @@ type identifier = string
 
 type ident_string =
   { name : identifier
-  ; var : Code.Var.t option }
+  ; var : Code.Var.t option
+  ; loc : location}
+
+let mk ?(loc=N) ?var name = { name ; var ; loc }
```
hhugo: The name `mk` looks a bit weird to me.
What about having a function `ident` instead, defined as
```
let ident ?(loc=N) ?var name = S { name ; var ; loc }
```

hhugo:
I agree, #799 does't have to block this PR

compiler/lib/javascript.ml
Drup: Fixed.

hhugo:
I took the liberty to add 2 commits to your PR.
I'll merge once you confirm you're happy with these changes and CI is green.

compiler/lib/js_output.ml
```diff
@@ -127,10 +127,13 @@ struct
       | Some _ -> ()
 
   let ident f = function
-    | S {name; var = None} -> PP.string f name
-    | S {name; var = Some v} ->
-        output_debug_info_ident f name v;
+    | S {name; var = Some v; _} ->
```
hhugo: Here, we ignore `loc` and only try to extract a loc from `var`. I wonder if we should have a pass that try populate all the locs

Drup: Can sourcemap accomodate multiple bindings for one identifier ?

I think a better longterm solution is to enforce (by datatype) that there is a loc in either `v` or `pi`.

Drup:
Looks good to me!

PR #807, opened 2019-06-05
Lib: Add support for 'addEventListener' with options object

facelesspanda:
The library seemed to be missing an accessible interface to give options to `addEventListener` as described in the specifications ( https://dom.spec.whatwg.org/#interface-eventtarget , https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Syntax , ... )

I added a function to do that.
Maybe the new class type will need to be moved, maybe some compatibility issues rise from the way I implemented this (some old browsers don't support the object variant. Maybe add a check for that ?)

Note: most times, users would need to pass only one option and keep the others at default. I don't know which is better between helper functions that take one argument and build the appropriate object, and additional class types with fewer methods.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,7 @@
  ** Misc: remove cppo dependency
  ** Misc: remove ppx_tools_versioned dependency in ppx_deriving_json
  ** Misc: support for ocaml 4.09
+ ** Lib: Add support for 'addEventListener' with options object
 
  * Bug fixes:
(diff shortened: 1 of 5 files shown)


hhugo:
Sorry for the delay on the review. I've made some changes to your PR (hope you don't mind)

facelesspanda:
I don't mind. This is a cleaner interface.

hhugo:
Thanks a lot for your contribution.

Issue #359, opened 2015-08-26
Marshaling loops on cyclic values


Tchou:
The following code snippet loops and eventually crashes the browser:

``` ocaml
type loop = { mutable pointer : loop option }
let l = { pointer = None }
let () = l.pointer <- Some l

let s = Marshal.to_string l []
let () = Firebug.console ## log (Js.string s)
```

compiled with:

``` shell
ocamlfind ocamlc -package js_of_ocaml,js_of_ocaml.syntax -syntax camlp4o foo.ml -linkpkg
js_of_ocaml a.out
```

and loaded by itself in a web page.
If js_of_ocaml's implementation of Marshal is not supposed to serialize cyclic values (the Json serialization does not either it seems), I think it should be prominently documented (glancing at the source code, desierialzing cyclic values seems implemented).

hhugo:
There is only a partial support for `Marshal.to_string` mostly due technical limitation (ie. cannot distinguish between float `1.` and int `1`). Cycle value is not supported (yet). We should document or take some time to implement it.

PR #814, opened 2019-06-10
Runtime: more efficient (space-wise) marshalling.

corwin-of-amber:
Modified caml_output_val to correctly reflects sharing.
As a result, it avoids duplication and saves a lot of space.

This has been thoroughly tested by compiling Coq `.vo` files, where it can reduce the file size by an order of magnitude, and produces roughly the same size as the native `coqc`.

### DIFF
--- a/compiler/tests/std_marshal.ml
+++ b/compiler/tests/std_marshal.ml
@@ -40,6 +40,11 @@ let%expect_test _ =
         Marshal.to_channel chan v [];
         pos_out chan - start
 
+    let write_out_noshare chan v =
+        let start = pos_out chan in
+        Marshal.to_channel chan v [Marshal.No_sharing];
+        pos_out chan - start
+
     let _ =
       let tmp_filename = Filename.temp_file "out" "txt" in
       let chan = open_out tmp_filename in
@@ -47,9 +52,10 @@ let%expect_test _ =
       let v2 = Op (Times, [v1; v1]) (* shared *) in
       let v1_sz = write_out chan v1 in
       let v2_sz = write_out chan v2 in
+      let v2_ns_sz = write_out_noshare chan v2 in
       flush chan;
-      Format.printf "sizes = %d %d (|v2| %s 2|v1|)\n%!" v1_sz v2_sz
-        (if v2_sz < 2 * v1_sz then "<" else ">=");
+      Format.printf "sizes = %d %d %d (|v2| %s |v2_ns|)\n%!" v1_sz v2_sz v2_ns_sz
+        (if v2_sz < v2_ns_sz then "<" else ">=");
 
       let chan = open_in tmp_filename in
       let v1' = Marshal.from_channel chan in
@@ -57,7 +63,7 @@ let%expect_test _ =
       Format.printf "readback = %B %B\n%!" (v1 = v1') (v2 = v2')
 |};
  [%expect {|
-   sizes = 33 40 (|v2| < 2|v1|)
+   sizes = 33 40 51 (|v2| < |v2_ns|)
    readback = true true |}]
 
 (* https://github.com/ocsigen/js_of_ocaml/issues/359 *)
--- a/compiler/tests/std_marshal.ml
+++ b/compiler/tests/std_marshal.ml
@@ -1,0 +1,76 @@
+(* Js_of_ocaml tests
+ * http://www.ocsigen.org/js_of_ocaml/
+ * Copyright (C) 2019 Hugo Heuzard
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, with linking exception;
+ * either version 2.1 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+(* https://github.com/ocsigen/js_of_ocaml/pull/814 *)
+
+let%expect_test _ =
+  Util.compile_and_run
+    {|
+    type sign = SPlus | SMinus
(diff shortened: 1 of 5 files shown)


hhugo:
Fixes https://github.com/ocsigen/js_of_ocaml/issues/359

vouillon:
What about using a `WeakMap`?

corwin-of-amber:
> What about using a `WeakMap`?

Or a regular `Map` for that matter... yes, that could be done and will improve the time complexity (saving the need for linear search in the internal object table). Question is if it's ok to rely on ES2015 specs in the runtime portion.

hhugo:
Thanks a lot for this.

Would you mind writing some tests ? There are plenty of examples in https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/tests/.

It would be nice to demonstrate that cyclic value can now be marshaled.

corwin-of-amber:
Alas, I failed to build the latest version from source, due to 
`Error: Unbound value Ast_mapper.make_error_of_message`

https://github.com/ocsigen/js_of_ocaml/blob/8022d1eafe263bef4c6cb4218246ef1709ee55da/ppx/ppx_js/lib_internal/ppx_js_internal.ml#L462

It seemed to work fine in Travis, so I looked at the configuration file and was not sure how to proceed, does the script (`tools/travis.sh`) even compile the source?

corwin-of-amber:
Ok I ran `opam install -y cohttp-lwt-unix menhir ppx_expect` and now it builds except for the missing `graphics` package.

(I think it caused it to upgrade `ocaml-migrate-parsetree`, which for some reason, was not done by `opam install -y --deps-only js_of_ocaml`.)

hhugo:
The `ocaml-migrate-parsetree {>= "1.3.1"}` constraint is not knonw to the opam-repository yet.
I would not expect opam to upgrade unless you pin js_of_ocaml.

About the missing graphics lib, I think building the compiler and its (new) tests should be enough (`dune build @compiler/runtest`)

corwin-of-amber:
Added one test for shared data (DAG) and one for cycles (based on #359).

I noticed that for the other tests in `channel.ml`, you added OCaml counterparts. Do you want that for the two new ones as well?

hhugo:
Tests look good as they are now. Maybe move them in their own file.

corwin-of-amber:
Ok, suggestions on how to name it? I can't use `compiler/tests/marshal.ml` because it clashes with stdlib.

hhugo:
To follow on @vouillon remark. Could we you `weakMap/Map` with some fallback if not available ?

We should probably also check the flags in `caml_output_value`
https://github.com/ocsigen/js_of_ocaml/blob/master/runtime/io.js#L389 and 
`caml_output_value_to_{buffer,string,bytes}` https://github.com/ocsigen/js_of_ocaml/blob/master/runtime/marshal.js#L445

corwin-of-amber:
> To follow on @vouillon remark. Could we you `weakMap/Map` with some fallback if not available ?

Good idea.

> We should probably also check the flags in `caml_output_value`

Can easily support `Marshal.No_sharing` and `Marshal.Compat32`, I don't know about `Marshal.Closures`.
Are the flags the same for `caml_output_value` and for `caml_output_val`?

runtime/marshal.js
```diff
@@ -337,10 +337,46 @@ function caml_marshal_data_size (s, ofs) {
   return (get32(s, ofs + 4));
 }
 
+//Provides: MlObjectTable
+var MlObjectTable;
+if (typeof joo_global_object.WeakMap === 'undefined') {
+  MlObjectTable = function() {
```
corwin-of-amber: In the current test setup this code is never tested because the condition is always `false`. 😞

corwin-of-amber:
Ok, flags are in.

runtime/marshal.js
```diff
@@ -365,13 +406,31 @@ var caml_output_val = function (){
...
+      joo_global_object.console.warn("in caml_output_val: flag Marshal.Closures is not supported.");
+
     var writer = new Writer ();
     var stack = [];
+    var intern_obj_table = new MlObjectTable();
+
+    function memo(v) {
+      var existing_offset = no_sharing ? undefined : intern_obj_table.recall(v);
+      if (existing_offset) { writer.write_shared(existing_offset); return existing_offset; }
+      else intern_obj_table.store(v);
```
hhugo: do we need to call `intern_obj_table.store` when `no_sharing` is defined ?

runtime/marshal.js
```diff
@@ -337,10 +337,46 @@ function caml_marshal_data_size (s, ofs) {
...
+if (typeof joo_global_object.WeakMap === 'undefined') {
+  MlObjectTable = function() {
+    /* polyfill */
+    function NaiveLookup(objs) { this.objs = objs; }
+    NaiveLookup.prototype.get = function(v) {
+      for (var i = 0; i < this.objs.length; i++) {
+        if (this.objs[i] === v) return i;
+      }
+    };
+    NaiveLookup.prototype.set = function() { };
```
hhugo: Why is `set` empty ?

hhugo: I understand now.
This probably deserves a small comment, saying that `objs` is shared between `MlObjectTable` and `NaiveLookup` and that `store` is responsible for updating `objs`

compiler/tests/std_marshal.ml
```diff
@@ -0,0 +1,82 @@
+(* Js_of_ocaml tests
+ * http://www.ocsigen.org/js_of_ocaml/
+ * Copyright (C) 2019 Hugo Heuzard
```
hhugo: Update the name here

corwin-of-amber: Ah oops.

runtime/marshal.js
corwin-of-amber: We still need to update the object counter.
I can do `writer.obj_counter++` instead, and remove the assignment `writer.obj_counter = intern_obj_table.length` at the end of the function.

hhugo: obj_counter seems to only be used when sharing is enable.
https://github.com/ocaml/ocaml/blob/trunk/runtime/extern.c#L180

runtime/marshal.js
hhugo: I've tested the code manually.

hhugo:
I took the liberty to push to your branch directly (hope you don't mind).
Could you review my changes ?

corwin-of-amber:
Looking great.
I noticed that `memo(v)` can in fact return a boolean. So I tweaked it further.

hhugo:
Thanks a lot !

PR #819, opened 2019-06-14
ocamlformat-next-408

hhugo:


gpetiot:
`indicate-multiline-delimiters=false` should be replaced by `indicate-multiline-delimiters=no`.

PR #831, opened 2019-07-13
Handle browserfs in fs_node detection logic

thangngoc89:
As discussed on ocamllabs slack, when a website includes Browserfs, it passes all the current `fs_node` detection logic but failed at runtime because `require` is undefined.

This PR handles this case by checking for `process.platform`


hhugo:
Thanks

PR #839, opened 2019-07-21
Tyxml: add cache for HTML entities

hhugo:
@vouillon, I've rebased an old branch of yours. Is this something we want ?


vouillon:
The issue is that `Dom_html.decode_html_entities` is slow. A better fix would be to have special cases for numeric character references `&#nnnn;` and  `&#xhhhh;` (using `Js.to_string (Js.string_constr##fromCharCode x)`), as well as for common entities (`&quot;`, `&amp;`, `&apos;`, `&lt;`, `&gt;`).

By the way, the current implementation is unsafe: it will happily execute arbitrary JavaScript code (try this string, for instance: ``).

PR #848, opened 2019-07-29
Lib: fix entity decoding

hhugo:


hhugo:
Following https://github.com/ocsigen/js_of_ocaml/pull/839

PR #841, opened 2019-07-21
Doc: document separate compilation (close #668)

hhugo:
This reverts commit 0ff280baecbf39c6567225eecd8dc76e0ab4e0b4.

fix #668

PR #852, opened 2019-08-09
Use odoc to generate the API

hhugo:


Drup:
Does this preserves the documentation on ocsigen.org ?

hhugo:
I'm not sure what you mean by `preserves the documentation on ocsigen.org`.
- It does not integrate with ocsigen.org -> #855 
- It does not push the dev doc automatically -> #853

hhugo:
see http://ocsigen.org/js_of_ocaml/dev/api/js_of_ocaml/Js_of_ocaml/index.html for a preview of #855

Drup:
Well, there is quite a bit of tooling on https://github.com/ocsigen/html_of_wiki to do that. I generated the tyxml pages with that a few time. It's quite a mess, but I wouldn't really want to reinvent it ...

hhugo:
There is tooling but nobody to support it.The ocamldoc+wikicreole+html_of_wiki approach is currently not usable. I prefer to move forward with supported tooling (odoc, dune) even if it requires tiny hacks.

Drup:
On the overall, I agree with you. I just would like the whole of ocsigen to stay mildly coherent, if possible.

dbuenzli:
I don't want to pretend to be representative of developers using ocsigen since I only use `js_of_ocaml` but as far as I'm concerned going to the ocsigen website to get documentation about it has always been a [painful experience](https://github.com/ocsigen/js_of_ocaml/pull/841#pullrequestreview-264838531).

I really like to get documentation locally from my switches directly corresponding to the version I'm currently using. So I welcome and support @hhugo's wish to move towards `odoc` and `.mld` files.

I think they are in a better position to serve both developers and users: checked cross-references between the API and manuals, low barrier entry both to documentation authoring and distribution (install `.cmti` and `.mld` files) and uniform access accross packages and according to the user's accessibility needs (via [`odig`](https://erratique.ch/software/odig) of course, sorry for the self-promotion).

If you provide and advertise (`odig doc js_of_ocaml`) a good local experience you can care a little bit less about e.g. having documentation about all previous versions of `js_of_ocaml` on the website.

Drup:
> If you provide and advertise (`odig doc js_of_ocaml`) a good local experience you can care a little bit less about e.g. having documentation about all previous versions of `js_of_ocaml` on the website.

While odig is great, most people are still looking for documentation online, on the software's website. Having online documentation for past versions has proven very useful in practice.

I do agree with your other comments, that doesn't mean we shouldn't aim for a common website for ocsigen projects. Ideally, it would be done the same way the mirage docs work, unfortunately, this is not so compatible with eliom.

dbuenzli:
> While odig is great, most people are still looking for documentation online, on the software's website.

I'm a little bit curious about that claim. Somehow I doubt you prefer to peruse documentation online once you start using packages from a distributed set of providers: it rather becomes painful to go from one website to another each with it's own way of presenting the docs. 

I think that if you properly advertise the local experience people will rather use this. It's less time consuming, corresponds exactly to what you have and how you want it. As far as I'm concerned the online experience is just useful for pointing and online discussion.

> Ideally, it would be done the same way the mirage docs work, unfortunately, this is not so compatible with eliom.

Just curious (again). What are the problems ?

PR #856, opened 2019-08-14
added `caml_round_float` function

monstasat:
Added `caml_round_float` function to achieve compatibility with OCaml 4.08


hhugo:
Thanks a lot.

Issue #680, opened 2018-07-03
Update the marshal runtime once ocaml#1683 is merged


hhugo:
https://github.com/ocaml/ocaml/pull/1683

hhugo:
https://github.com/ocsigen/js_of_ocaml/pull/861 and https://github.com/ocsigen/js_of_ocaml/pull/871

PR #861, opened 2019-08-16
OCaml 4.08 marshalling changes

vouillon:
`CODE_CUSTOM` is deprecated and replaced by `CODE_CUSTOM_FIXED` and `CODE_CUSTOM_LEN`. The custom values supported by Js_of_ocaml are now tagged using `CODE_CUSTOM_FIXED`; unmarshalling other custom values will result in a failure exception (input_value: unknown custom block identifier).


hhugo:
Closes #680

runtime/marshal.js
```diff
@@ -434,7 +438,7 @@ var caml_output_val = function (){
         if (v[0] == 255) {
           // Int64
           if (memo(v)) return;
-          writer.write (8, 0x12 /*cst.CODE_CUSTOM*/);
+          writer.write (8, 0x19 /*cst.CODE_CUSTOM_FIXED*/);
```
hhugo: We should still use CODE_CUSTOM when using older ocaml compilers. One can achieve that with an helper function: 
```javascript
//Provides: marshall_header_int64
//Version: < 4.08
function marshall_header_int64 (){ return 0x12 }
//Provides: marshall_header_int64
//Version: >= 4.08
function marshall_header_int64 (){ return 0x19 }
```

runtime/marshal.js
```diff
@@ -275,6 +277,8 @@ function caml_input_value_from_reader(reader, ofs) {
           caml_failwith ("input_value: code pointer");
           break;
         case 0x12: //cst.CODE_CUSTOM:
+        case 0x18: //cst.CODE_CUSTOM_LEN:
```
hhugo: Can we handle `CODE_CUSTOM_LEN` separately ? or add a check in the `_j`, `_i`, `_n` cases  to make sure we are not processing a `CODE_CUSTOM_LEN`

PR #867, opened 2019-09-05
added resizeObserver api

monstasat:
Added resizeObserver API bindings. 
See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) for details

### DIFF
--- a/lib/js_of_ocaml/resizeObserver.ml
+++ b/lib/js_of_ocaml/resizeObserver.ml
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
-*)
+ *)
 
 class type resizeObserverSize =
   object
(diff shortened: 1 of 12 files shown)


TyOverby:
This looks great!  I've looked at the code side-by-side with the spec, and everything seems good, but I'll play around with using the code later this week.

monstasat:
Great! Looking forward to your feedback

TyOverby:
I tried this out today and everything works as expected!  

@hhugo : there's an internal feature at 'jane/js-of-ocaml/test-resize-observer' where I built a small demo using this API.

monstasat:
Glad to hear that!

lib/js_of_ocaml/resizeObserver.ml
```diff
@@ -0,0 +1,80 @@
...
+    method contentRect : Dom_html.clientRect Js.t Js.readonly_prop
+
+    method borderBoxSize : resizeObserverSize Js.t Js.readonly_prop
+
+    method contentBoxSize : resizeObserverSize Js.t Js.readonly_prop
+  end
+
+class type resizeObserverOptions =
+  object
+    method box : Js.js_string Js.t Js.prop
```
hhugo: Should this be `Js.writeonly_prop` so that the following is disallowed ?
```ocaml
let o = empty_resize_observer_options ()
o##.box
```

monstasat: I tend to agree, especially considering the similar implementation for mutationObserver bindings. Didn't notice that before. Should I do a commit?

hhugo: Yes please.

monstasat: Done

hhugo:
thanks

PR #877, opened 2019-09-26
Added support for custom events

monstasat:
## Added support for custom events
* Added `customEvent` object type
* Added `Dom.createCustomEvent` and `Dom_html.createCustomEvent` helper functions to allow creation of custom events
* Added `eventTarget##dispatchEvent` method to allow dispatching of custom (and other) events

### DIFF
--- a/lib/js_of_ocaml/dom.ml
+++ b/lib/js_of_ocaml/dom.ml
@@ -292,6 +292,13 @@ class type ['a] event =
     method srcElement : 'a t opt readonly_prop
   end
 
+class type ['a, 'b] customEvent =
+  object
+    inherit ['a] event
+
+    method detail : 'b Js.opt Js.readonly_prop
+  end
+
 let no_handler : ('a, 'b) event_listener = Js.null
 
 let window_event () : 'a #event t = Js.Unsafe.pure_js_expr "event"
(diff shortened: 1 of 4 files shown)


hhugo:
Can you try to rebase, that should fix the CI

monstasat:
Rebased onto master branch.

TyOverby:
From the [mdn page](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) there appear to be many more properties than are present in the PR.  Why are these left out?

monstasat:
As you can see on the page you've mentioned, all properties excluding `detail` are inherited from the parent object, `Event`:

> This interface inherits properties from its parent, Event:
>
> Event.bubbles Read only
> A Boolean indicating whether the event bubbles up through the DOM or not.
> 
> Event.cancelBubble
A historical alias to Event.stopPropagation(). Setting its value to true before returning from an event handler prevents propagation of the event.

and so on...

`js_of_ocaml` bindings of `Event` object indeed lacks properties according to the docs, but if you want them to be added, I think they should be added to the `Event`, not to the `customEvent`.

Here is the [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/Event) describing the `Event` object.

lib/js_of_ocaml/dom.ml
```diff
@@ -391,6 +398,24 @@ let preventDefault ev =
...
+  let opts = Unsafe.obj [||] in
+  opt_iter (fun x -> opts##.bubbles := bool x) bubbles;
+  opt_iter (fun x -> opts##.cancelable := bool x) cancelable;
+  opt_iter (fun x -> opts##.detail := some x) detail;
+  let constr :
+      (   ('a, 'b) #customEvent Js.t Event.typ
+       -> < detail : 'b opt prop > t
+       -> ('a, 'b) customEvent t)
+      constr =
+    Unsafe.global##._CustomEvent
```
TyOverby: Where does the `_` in `_CustomEvent` come from?

monstasat: From `js_of_ocaml` documentation:

> Since ocaml doesn't allows method name to start with capitalised letter, we can add an _

Since `CustomEvent` constructor starts with the capital letter, we should prefix it with an underscore.

monstasat: Another example from the docs:

> Using a JS constructor, ex: F:
> Write in .ml:

`let f = Js.Unsafe.global##._F`

PR #883, opened 2019-10-11
Runtime: bigarray blit

hhugo:
Implement blit on bigarrays. This will be useful for project like (bin_prot, bigstringaf, ..)

- [x] add some tests

### DIFF
--- a/runtime/bigarray.js
+++ b/runtime/bigarray.js
@@ -772,6 +772,64 @@ function caml_ba_deserialize(reader, sz){
   return caml_ba_create_unsafe(kind, layout, dims, data);
 }
 
+//Provides: caml_ba_blit_ba_to_ba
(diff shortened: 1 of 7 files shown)


runtime/bigarray.js
```diff
@@ -772,6 +772,64 @@ function caml_ba_deserialize(reader, sz){
   return caml_ba_create_unsafe(kind, layout, dims, data);
 }
 
+//Provides: caml_ba_blit_ba_to_ba
+//Requires: caml_invalid_argument, caml_array_bound_error
+function caml_ba_blit_ba_to_ba(ba1, pos1, ba2, pos2, len){
+  if(12 != ba1.kind)
```
TyOverby: What is 12?

hhugo: Kind = char

PR #894, opened 2019-10-28
added bindings for pointer events

monstasat:
Added bindings for pointer events to `Dom_html` and `Lwt_js_events` modules.
See [W3C](https://www.w3.org/TR/pointerevents/#pointerevent-interface) for pointer events API.


hhugo:
Thanks

PR #896, opened 2019-10-29
Fix polymorphic compare on JavaScript values

### LINKED ISSUE #895: [BUG] JS string comparison raises an exception
**Describe the bug**
I've noticed that code like `Dom_html.document##.readyState = Js.string "loading"` throws an exception. Exception has the following message: `compare: abstract value`.
Here is the stacktrace from my production code:

```caml_raise_with_arg | @ | stdlib.js:117
  | caml_raise_with_string | @ | stdlib.js:122
  | caml_invalid_argument | @ | stdlib.js:178
  | caml_compare_val | @ | stdlib.js:424
  | caml_equal | @ | stdlib.js:455
  | domContentLoaded | @ | lwt_js_events.ml:537
```
It seems that the code fails because of the `caml_compare_val` function, in the following [line](https://github.com/ocsigen/js_of_ocaml/blob/0052b39e079a64d6886a29c4e66a41b4ea9e63c0/runtime/stdlib.js#L424).
The `Lwt_js_events.domContentLoaded` function which calls `caml_equal` just compares `Dom_html.document##.readyState` with some constant at line 537.

Here is a another simple test illustrating the issue:
```ocaml
(* This is ok and shows [true] in console *)
print_endline "comparing two equal js strings";
print_endline (string_of_bool (Js.string "foo" = Js.string "foo"));
(* This is ok and shows [false] in console *)
print_endline "comparing two different js strings";
print_endline (string_of_bool (Js.string "loading" = Js.string "complete"));
(* This one fails*)
let state = Dom_html.document##.readyState in
print_endline "comparing state";
print_endline (string_of_bool (state = Js.string "complete"))
```

**Expected behavior**
This should not fail.

**Versions**
jsoo commit 0052b39e079a64d6886a29c4e66a41b4ea9e63c0

hhugo:
- [x] Compare works on javascript `string` and `String` object
- [x] Compare works on javascript `number` and `Number` object

### DIFF
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -358,83 +358,153 @@ function caml_floatarray_create(len){
   return b
 }
 
+//Provides: caml_compare_val_tag
+//Requires: MlBytes
+function caml_compare_val_tag(a){
+  if (typeof a === "number") return 1000; // int_tag (we use it for all numbers)
+  else if (a instanceof MlBytes) return 252; // string_tag
+  else if (a instanceof Array && a[0] === (a[0]>>>0) && a[0] <= 255) {
+    // Look like an ocaml block
+    var tag = a[0] | 0;
+    // ignore double_array_tag because we cannot accurately set
+    // this tag when we create an array of float.
+    return (tag == 254)?0:tag
+  }
+  else if (a instanceof String) return 1252; // javascript string, like string_tag (252)
+  else if (typeof a == "string") return 1252; // javascript string, like string_tag (252)
+  else if (a instanceof Number) return 1000; // int_tag (we use it for all numbers)
+  else if (a.caml_custom) return 1255; // like custom_tag (255)
+  else if (a.compare) return 1256; // like custom_tag (255)
+  else if (typeof a == "function") return 1247; // like closure_tag (247)
+  else if (typeof a == "symbol") return 1251;
+  return 1001; //out_of_heap_tag
+}
+
+//Provides: caml_compare_val_get_custom
+//Requires: caml_custom_ops
+function caml_compare_val_get_custom(a){
+  return caml_custom_ops[a.caml_custom] && caml_custom_ops[a.caml_custom].compare;
+}
+
+//Provides: caml_compare_val_number_custom
+//Requires: caml_compare_val_get_custom
+function caml_compare_val_number_custom(num, custom, swap, total) {
+  var comp = caml_compare_val_get_custom(custom);
+  if(comp) {
+    var x = (swap > 0)?comp(custom,num,total):comp(num,custom,total);
+    if(total && x != x) return swap; // total && nan
+    if(+x != +x) return +x; // nan
+    if((x | 0) != 0) return (x | 0); // !nan
+  }
+  return swap
+}
+
 //Provides: caml_compare_val (const, const, const)
-//Requires: MlBytes, caml_int64_compare, caml_int_compare, caml_string_compare
-//Requires: caml_invalid_argument, caml_custom_ops
+//Requires: caml_int64_compare, caml_int_compare, caml_string_compare
+//Requires: caml_invalid_argument, caml_compare_val_get_custom, caml_compare_val_tag
+//Requires: caml_compare_val_number_custom
 function caml_compare_val (a, b, total) {
   var stack = [];
   for(;;) {
     if (!(total && a === b)) {
-      if (a instanceof MlBytes) {
-        if (b instanceof MlBytes) {
-          if (a !== b) {
-            var x = caml_string_compare(a, b);
-            if (x != 0) return x;
+      var tag_a = caml_compare_val_tag(a);
+      // forward_tag ?
+      if(tag_a == 250) { a = a[1]; continue }
+
+      var tag_b = caml_compare_val_tag(b);
+      // forward_tag ?
+      if(tag_b == 250) { b = b[1]; continue }
+
+      // tags are different
+      if(tag_a !== tag_b) {
+        if(tag_a == 1000) {
+          if(tag_b == 1255) { //immediate can compare against custom
+            return caml_compare_val_number_custom(a, b, -1, total);
           }
-        } else
-          // Should not happen
-          return 1;
-      } else if (a instanceof Array && a[0] === (a[0]|0)) {
-        var ta = a[0];
-        // ignore double_array_tag
-        if (ta === 254) ta=0;
-        // Forward object
-        if (ta === 250) {
-          a = a[1];
-          continue;
-        } else if (b instanceof Array && b[0] === (b[0]|0)) {
-          var tb = b[0];
-          // ignore double_array_tag
-          if (tb === 254) tb=0;
-          // Forward object
-          if (tb === 250) {
-            b = b[1];
-            continue;
-          } else if (ta != tb) {
-            return (ta < tb)?-1:1;
-          } else {
-            switch (ta) {
-            case 248: {
-              // Object
-              var x = caml_int_compare(a[2], b[2]);
-              if (x != 0) return x;
-              break;
-            }
-            case 251: {
-              caml_invalid_argument("equal: abstract value");
-            }
-            case 255: {
-              // Int64
-              var x = caml_int64_compare(a, b);
-              if (x != 0) return x;
-              break;
-            }
-            default:
-              if (a.length != b.length) return (a.length < b.length)?-1:1;
-              if (a.length > 1) stack.push(a, b, 1);
-            }
+          return -1
+        }
+        if(tag_b == 1000) {
+          if(tag_a == 1255) { //immediate can compare against custom
+            return caml_compare_val_number_custom(b, a, 1, total);
           }
-        } else
-          return 1;
-      } else if (b instanceof MlBytes ||
-                 (b instanceof Array && b[0] === (b[0]|0))) {
-        return -1;
-      } else if(a && a.caml_custom) {
-        if(caml_custom_ops[a.caml_custom].compare) {
-          var cmp = caml_custom_ops[a.caml_custom].compare(a,b,total);
-          if (cmp != 0) return cmp;
+          return 1
         }
-        else {
-          caml_invalid_argument("compare: abstract value");
-        }
+        return (tag_a < tag_b)?-1:1;
       }
-      else if(a && a.compare) {
-        var cmp = a.compare(b,total);
-        if (cmp != 0) return cmp;
-      }
-      else if (typeof a == "function") {
+      switch(tag_a){
+        // 246: Lazy_tag handled bellow
+      case 247: // Closure_tag
+        // Cannot happen
         caml_invalid_argument("compare: functional value");
-      } else {
+        break
+      case 248: // Object
+        var x = caml_int_compare(a[2], b[2]);
+        if (x != 0) return (x | 0);
+        break;
+      case 249: // Infix
+        // Cannot happen
+        caml_invalid_argument("compare: functional value");
+        break
+      case 250: // Forward tag
+        // Cannot happen, handled above
+        caml_invalid_argument("equal: got Forward_tag, should not happen");
+        break;
+      case 251: //Abstract
+        caml_invalid_argument("equal: abstract value");
+        break;
+      case 252: // OCaml string
+        if (a !== b) {
+          var x = caml_string_compare(a, b);
+          if (x != 0) return (x | 0);
+        };
+        break;
+      case 253: // Double_tag
+        // Cannot happen
+        caml_invalid_argument("equal: got Double_tag, should not happen");
+        break;
+      case 254: // Double_array_tag
+        // Cannot happen, handled above
+        caml_invalid_argument("equal: got Double_array_tag, should not happen");
+        break
+      case 255: // Int64
+        // Int64 is the only custom block implemented this way,
+        // TODO: We should rewrite the implementation and follow
+        // what we do for other custom blocks: zarith, bigint, bigarray, ...
+        var x = caml_int64_compare(a, b);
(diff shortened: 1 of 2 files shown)


runtime/stdlib.js
```diff
@@ -406,7 +406,18 @@ function caml_compare_val (a, b, total) {
       } else if (b instanceof MlBytes ||
                  (b instanceof Array && b[0] === (b[0]|0))) {
         return -1;
-      } else if (typeof a != "number") {
+      } else if ((typeof a === "number" || typeof a === "boolean" || typeof a === "string") &&
```
TyOverby: This would permit `a` being a `number` and `b` being a `string`.  Is this something that we have to worry about?

hhugo:
@vouillon, another pair of eyes on the diff of `stdlib.js` would be nice. https://github.com/ocsigen/js_of_ocaml/pull/896/files#diff-1c1ab2d06ff97085527fbab9c3084e42

runtime/stdlib.js
hhugo: We need to consider every possible combination of types. I rewrote the function to fix the code / make the code clearer.

hhugo:
cc @xclerc

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
+          break
+        case 250: // Forward tag
+          // Cannot happen, handled above
+          caml_invalid_argument("equal: got Forward_tag, should not happen");
+          break;
+        case 251: //Abstract
+          caml_invalid_argument("equal: abstract value");
+          break;
+        case 252: // OCaml string
           if (a !== b) {
```
hhugo: This is already checked above

lib/js_of_ocaml/dom.ml
```diff
@@ -263,7 +264,7 @@ let nodeType e =
 
 module CoerceTo = struct
   let cast (e : #node Js.t) t =
-    if e##.nodeType = t then Js.some (Js.Unsafe.coerce e) else Js.null
+    if e##.nodeType == t then Js.some (Js.Unsafe.coerce e) else Js.null
```
TyOverby:
```suggestion
    if phys_equal e##.nodeType t then Js.some (Js.Unsafe.coerce e) else Js.null
```

lib/js_of_ocaml/import.ml
```diff
@@ -0,0 +1,71 @@
+(* Js_of_ocaml
```
TyOverby: Is there a place we could put this Import module where it could be shared?

lib/lwt/lwt_xmlHttpRequest.ml
```diff
@@ -166,7 +167,7 @@ let perform_raw
     | Some (`String _ | `Blob _) -> override_method "POST", content_type
   in
   let url =
-    if get_args = []
+    if Poly.( = ) get_args []
```
TyOverby:
```suggestion
    if Poly.(get_args = [])
```

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
   var stack = [];
   for(;;) {
     if (!(total && a === b)) {
-      if (a instanceof MlBytes) {
-        if (b instanceof MlBytes) {
+      // Compute TAG for [a]
+      var ta = 1001;
```
TyOverby: What does "ta" stand for?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
   var stack = [];
   for(;;) {
     if (!(total && a === b)) {
```
TyOverby: What is "total"?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
-      if (a instanceof MlBytes) {
-        if (b instanceof MlBytes) {
+      // Compute TAG for [a]
+      var ta = 1001;
+      if (typeof a === "number") ta = 1000;
```
TyOverby: The magic numbers in this code make it really hard to follow.  Could these be pulled out into variables?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
-            if (cmp != 0) return cmp;
+          if(!comp)
+            caml_invalid_argument("compare: abstract value");
+          var x = comp(a,b,total);
+          if (x != 0) return x;
+          break;
+        case 1256: // compare function
+          var x = a.compare(b,total);
+          if(x != x | 0)
+            caml_invalid_argument("compare: abstract value");
```
TyOverby: What scenarios trigger this code path?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
+          break;
+        case 1000: // Number
+        case 1001: // The rest
+          // Here we can be in the following cases:
+          // - a and b are primitive types, comparison works as expected.
+          // - a or b aren't primitive types, comparison always return false.
+          if (a < b) return -1;
+          if (a > b) return 1;
+          if (a != b) {
+            if (!total) return NaN;
```
TyOverby: Returning NaN here is very surprising.  What does this mean? Is there a native OCaml equivalent?

TyOverby: Also, what is "total"

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
+          if (x != 0) return x;
+          break;
+        case 1256: // compare function
+          var x = a.compare(b,total);
+          if(x != x | 0)
+            caml_invalid_argument("compare: abstract value");
+          x = x | 0;
+          if (x != 0) return x;
+          break;
+        case 1000: // Number
```
TyOverby: What happens when integers are compared against floats?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
...
-        if (a != b) {
-          if (!total) return NaN;
-          if (a == a) return 1;
-          if (b == b) return -1;
+          break;
+        case 246: // Lazy_tag
+        case 254: // Double_array
+        default: // Block with other tag
+          if (a.length != b.length) return (a.length < b.length)?-1:1;
+          if (a.length > 1) stack.push(a, b, 1);
```
TyOverby: Is this not just pushing the same objects back on the stack?

runtime/stdlib.js
```diff
@@ -352,85 +352,129 @@ function caml_compare_val (a, b, total) {
   var stack = [];
```
TyOverby: It looks like this stack is only pushed onto.  Is that correct?

runtime/stdlib.js
hhugo: Tag of a

runtime/stdlib.js
hhugo: Any javascript value with a compare method returning non integer result

runtime/stdlib.js
hhugo: `total` say whether we need a total order or not.
We only need a total order for polymorphic compare, not for boolean operator  `=`, `<>`, `<`, `>`, ...

runtime/stdlib.js
hhugo: The stack looks like `[offset, objectleft, objectright, offset, objectleft, objectright]`.
Here we push the same objects with offset `1`. The next time we pick elements from the stack we'll pick `a[1]` and `b[1]` and push `stack.push(a, b, i + 1);`

```
    var i = stack.pop();
    b = stack.pop();
    a = stack.pop();
    if (i + 1 < a.length) stack.push(a, b, i + 1);
    a = a[i];
    b = b[i];
```

runtime/stdlib.js
hhugo: No, we pop from it as well.

runtime/stdlib.js
hhugo: Well, it should behave like comparison on floats

runtime/stdlib.js
hhugo: my fear was that using variable in `switch` cases could be slower.
Theses tags are the ocaml tags, with small additions
```
let lazy_tag = 246
let closure_tag = 247
let object_tag = 248
let infix_tag = 249
let forward_tag = 250

let no_scan_tag = 251

let abstract_tag = 251
let string_tag = 252
let double_tag = 253
let double_array_tag = 254
let custom_tag = 255
let final_tag = custom_tag


let int_tag = 1000
let out_of_heap_tag = 1001
let unaligned_tag = 1002
```

Additions are
```
1255 (similar to custom_tag 255)
1256 (for JavaScript objects that have a compare method)  
1247 (similar to closure_tag 1247)
```

runtime/stdlib.js
hhugo: We only return NaN when we don't need a total order.
It correspond to UNORDERED in the the ocaml runtime (see https://github.com/ocaml/ocaml/blob/trunk/runtime/compare.c#L110)

runtime/stdlib.js
hhugo: For instance, `compare nan nan = 0` but `nan <> nan`,`(nan = nan) = false`, `(nan < nan) = false`, `(nan > nan) = false`

runtime/stdlib.js
hhugo: I removed the check above

runtime/stdlib.js
hhugo: renamed

lib/js_of_ocaml/import.ml
hhugo: I believe theses import files will diverge on purpose overtime.

runtime/stdlib.js
TyOverby: If we are expected to return an int, isn't NaN breaking that requirement?

runtime/stdlib.js
```diff
@@ -358,90 +358,205 @@ function caml_floatarray_create(len){
...
+        if (a > b) return 1;
+        if (a != b) {
+          if (!total) return NaN;
+          if (a == a) return 1;
+          if (b == b) return -1;
+        }
+        break;
+      case 246: // Lazy_tag
+      case 1251: // JavaScript Symbol, not ordering.
+        if(a != b) {
```
TyOverby: should this be `===`?

TyOverby: or `!==` rather

runtime/stdlib.js
hhugo: caml_compare_val is an helper function for the functions below. We only use its return value directly when `total` is true (in `caml_comare`)
```javascript
//Provides: caml_compare (const, const)
//Requires: caml_compare_val
function caml_compare (a, b) { return caml_compare_val (a, b, true); }
//Provides: caml_equal mutable (const, const)
//Requires: caml_compare_val
function caml_equal (x, y) { return +(caml_compare_val(x,y,false) == 0); }
//Provides: caml_notequal mutable (const, const)
//Requires: caml_compare_val
function caml_notequal (x, y) { return +(caml_compare_val(x,y,false) != 0); }
//Provides: caml_greaterequal mutable (const, const)
//Requires: caml_compare_val
function caml_greaterequal (x, y) { return +(caml_compare_val(x,y,false) >= 0); }
//Provides: caml_greaterthan mutable (const, const)
//Requires: caml_compare_val
function caml_greaterthan (x, y) { return +(caml_compare_val(x,y,false) > 0); }
//Provides: caml_lessequal mutable (const, const)
//Requires: caml_compare_val
function caml_lessequal (x, y) { return +(caml_compare_val(x,y,false) <= 0); }
//Provides: caml_lessthan mutable (const, const)
//Requires: caml_compare_val
function caml_lessthan (x, y) { return +(caml_compare_val(x,y,false) < 0); }
```

runtime/stdlib.js
hhugo: `!==` and `!=` would behave the same for `Symbol`.
But you're right, `!==` convey the intent better

lib/js_of_ocaml/import.ml
hhugo: This diff has been pushed on master already.

runtime/stdlib.js
```diff
@@ -423,8 +424,9 @@ function caml_compare_val (a, b, total) {
             var comp = caml_compare_val_get_custom(a);
             if(comp) {
               var x = comp(a,b,total);
-              if(!(total && x != x) || (x === (x|0) && x != 0))
-                return x
+              if(total && x != x) return 1;
+              if(+x != +x) return +x;
+              if((x | 0) != 0) return (x | 0);
```
TyOverby: Could you factor the two blocks in this commit out into a function?  I'm terrified of this bithackery getting out of sync.

runtime/stdlib.js
```diff
@@ -358,90 +358,203 @@ function caml_floatarray_create(len){
...
+    // Look like an ocaml block
+    var tag = a[0] | 0;
+    // ignore double_array_tag because we cannot accurately set
+    // this tag when we create an array of float.
+    return (tag == 254)?0:tag
+  }
+  else if (a instanceof String) return 1252; // javascript string, like string_tag (252)
+  else if (typeof a == "string") return 1252; // javascript string, like string_tag (252)
+  else if (a instanceof Number) return 1000; // int_tag (we use it for all numbers)
+  else if (a.caml_custom) return 1255; // like custom_tag (255)
```
vouillon: @hhugo, this breaks when `a` is `null` or `undefined`

hhugo: https://github.com/ocsigen/js_of_ocaml/pull/920

Issue #898, opened 2019-10-31
[BUG] String literals containing "</script>" and embedded Javascript interact badly


sbriais:
**Describe the bug**
```
echo 'let () = print_endline "hello </script>"' > helloscript.ml
ocamlc -o helloscript helloscript.ml
js_of_ocaml helloscript
echo '<!DOCTYPE html>' > helloscript.html
echo '<script>' >> helloscript.html
cat helloscript.js >> helloscript.html
echo '</script>' >> helloscript.html
```

Visiting `helloscript.html` produces an error in the browser.
(eg with Firefox we get Syntax Error "" literal not terminated before the end of the script) 

The bug occurs because of the "</script>" substring. 

**Expected behavior**
We expect js_of_ocaml to handle this case for us.

**Versions**
ocaml 4.08
js_of_ocaml 3.4.0

alainfrisch:
Ref: https://stackoverflow.com/questions/28643272/how-to-include-an-escapedscript-script-tag-in-a-javascript-variable

 (one could also escape < or / in the JS string literals that contain `</script>`, using JS Unicode escape \u...).

dbuenzli:
> We expect js_of_ocaml to handle this case for us.

To be honest I wouldn't qualify this as a bug. When you embed a language into another you are the one who is responsible to escape it properly.

alainfrisch:
Right, but (1) the fix in js_of_ocaml would be quite simple (such as putting `/*<!--*/"..."/*-->*/` around all string literals -- or just the one containing `</script>` (modulo casing,etc); or emitting `<` in JS strings as `\u...`); (2) the workaround outside js_of_ocaml is a bit tedious since it needs to parsing JS syntax; (3) users could easily be faced with the problem, and this is likely to be reported as a deficiency of js_of_ocaml.

dbuenzli:
I don't mind if `js_of_ocaml` wants to solve this (though this should likely under a flag, people rather want from jsoo to produce minimal outputs). 

However note that in general you may also have to link with foreign, non jsoo generated JavaScript which might have the same problem.

So you likely want to pass all of your js through some kind of escaper tool if you want to build html files with inline scripts.

hhugo:
[...]

PR #899, opened 2019-10-31
Compiler: escape '</' in strings (WIP)

hhugo:


hhugo:
@dbuenzli, do you think this should be enabled under a flag ? I would expect the overhead to be relatively small.

dbuenzli:
No I think it should be fine. I thought about a flag if we had to resort to the more longer solutions @alainfrisch mentioned [here](https://github.com/ocsigen/js_of_ocaml/issues/898#issuecomment-548381397).

PR #902, opened 2019-11-02
Speedup toplevel creation 

### LINKED ISSUE #818: Creating toplevels is unbearably slow
The simple [eval](https://github.com/ocsigen/js_of_ocaml/tree/master/toplevel/examples/eval) example takes 34s on my machine. 

Also when using `--source-map-inline` memory usage on small examples is so big that it leads to SIGSEGVs.

hhugo:
Should fix #818.
On my machine, toplevel creation when from ~52s to ~17s
cc @dbuenzli


hhugo:
cc @thangngoc89

PR #906, opened 2019-11-02
Compiler: fix static evaluation of caml_equal

hhugo:


jordwalke:
I'm curious what this was fixing. I don't see any test output changing as a result.

hhugo:
The following was wrong
```
let c = if Poly.(e1 = e2) then 1l else 0l in
```

One cannot optimize the case where `e1 <> e2` in general because e1 and e2 might still end up compiled to the same representation.

hhugo:
here is a tests that would fail without that fix https://github.com/ocsigen/js_of_ocaml/commit/05a3925e0bff000190e121bacba1a6c2a8649db5

```
type pack = Pack : 'a -> pack

 let%expect_test "number comparison" =
  assert (Pack 2 = Pack 2.);
```

PR #909, opened 2019-11-03
Remove json polyfill

hhugo:


hhugo:
Does anyone know if this JSON poly-fill is safe to drop  ?

dbuenzli:
According to [caniuse](https://caniuse.com/#feat=json) I would say yes but maybe it's better to ask the person who introduced it in the first place.

hhugo:
@vouillon, @balat, would you be able to tests this PR ? Do you think it's OK to drop this poly-fill ?

balat:
Yes we can safely remove it. Even NodeJS has it and it is something very standard.

PR #910, opened 2019-11-03
Remove polyfill

hhugo:


vouillon:
While you are at it, you can probably remove the `Math.imul` polyfill in `stdlib.js`

PR #911, opened 2019-11-04
fix int64 bigarray comparison

hhugo:


hhugo:
credit to @xclerc for spotting the issue.

PR #912, opened 2019-11-07
Correct build alias from DEFAULT to default in examples/README.md

shonfeder:
Running the command as given in the readme resulted in

```
$ dune build @examples/DEFAULT
Error: Alias "DEFAULT" specified on the command line is empty.
It is not defined in examples or any of its descendants.
```

but the lower case alias worked as expected (as should be the case, given the dune files).

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,12 +2,12 @@
 
 Build all projects:
 ```
-$> dune build @examples/DEFAULT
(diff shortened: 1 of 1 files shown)


hhugo:
thanks

shonfeder:
Thank you!

PR #917, opened 2019-11-13
Deriving Json: rename module Import

vouillon:
This module name is way too common.


hhugo:
Mid-term we should probably just namespace the library / remove `(wrapped false)` from the dune file.
Short-term, we should probably apply your patch and do a minor release. I'd prefer to not do a minor release based on master, can you apply the patch on top of the 3.5 branch ?

PR #926, opened 2019-11-19
Optimize caml_call_gen

TyOverby:
A mixture of removing recursion, removing unnessary array copies,
and some case-unrolling yields marked performance improvements.


hhugo:
Waiting for some offline discussions to be reflected here

hhugo:
Merged manually.
Thanks

Issue #913, opened 2019-11-08
Add support for unix_isatty


bandersongit:
The [Pastel text highlighting library](https://www.npmjs.com/package/@reason-native/pastel) uses Unix.isatty as part of the logic to determine whether the host terminal supports color. There aren't currently stubs for this so compiling with JSOO requires consumers to provide an implementation. You can see https://github.com/facebookexperimental/reason-native/issues/207 for an an example of this.

I see a couple possible naive approaches: 
a) bind Unix.stdin, Unix.stdout, Unix.stderr to NodeJS's process.stdin, process.stdout, Process.stderr and implement unix_isatty as `stream => stream. isTTY)
b) use NodeJS's tty.isatty( ) along with numeric file descriptors

However I have noticed that runtime's fs implementation has both node based and a non non based (fake) modules that it uses, so I could see doing something like that as well to have different behavior in a browser environment (for example I've verified that ansi escape sequences work in chrome's developer console, however none of the node js modules/functions I described are available in that environment)

hhugo:
> ansi escape sequences work in chrome's developer console

Is that true ? I was not able to make this work.

I would probably go for (b), PR welcome

bandersongit:
> > ansi escape sequences work in chrome's developer console
> 
> Is that true ? I was not able to make this work.

When I was first playing with Rely and JSOO a few months ago in chrome I was able to get this colorized output
![image](https://user-images.githubusercontent.com/5252755/68522933-a700f900-0266-11ea-8ed3-610d5f74222b.png)
 (the weird unicode character thing was fixed by https://github.com/ocsigen/js_of_ocaml/issues/798, but I had the screenshot lying around).

Also for a simpler example
![image](https://user-images.githubusercontent.com/5252755/68523064-322ebe80-0268-11ea-8c82-4129353c074c.png)


> I would probably go for (b), PR welcome

Sounds good, I'll make some time for this in the next several days.

PR #928, opened 2019-11-20
add stub for unix_isatty

bandersongit:
Addresses https://github.com/ocsigen/js_of_ocaml/issues/913 and https://github.com/facebookexperimental/reason-native/issues/207

I tested this by running some JS tests compiled with Rely normally in a terminal and by piping the output to `cat` which should not be a tty. I also console logged the result of calling Unix.isatty on Unix.stdin to verify the function was returning the correct values. I thought I would need to provide values for Unix.stdin etc, but it was not necessary. I verified that the unix.isatty returns false and does not error in chrome as well.

I couldn't figure out the test structure of this repo and this change is fairly simple, so I did not add any automated tests, but could do so if desired with some help.

image

image

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 * Misc: support for ocaml 4.10
 * Misc: Cleanup Meta files (e.g. `js_of_ocaml.tyxml` is no longer valid library name)
 * Runtime: remove many old polyfill
+* Runtime: add unix_isatty
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


hhugo:
Thanks, can you add an entry to the changelog (CHANGES.md) ?

bandersongit:
> Thanks, can you add an entry to the changelog (CHANGES.md) ?

Done

hhugo:
thanks for your contribution

PR #929, opened 2019-11-22
Fix the type of blur and focus event to be focusEvent

noRubidium:
When adding focusEvent to event type, I forgot to include the update for the event  type in dom_html. This pull request updates that.


hhugo:
thanks

PR #930, opened 2019-11-22
Add closest method to element

lestertj:
Add support for the following Element method
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest


hhugo:
Thanks

PR #931, opened 2019-11-22
Runtime: fix pseudo fs initialization

### LINKED ISSUE #925: [BUG] jsoo_fs generates incorrect call to create file
**Describe the bug**

Given a simple `jsoo_fs` call like this:

```
jsoo_fs -o test.js -I . --files Demo.cmi
```

The generated `test.js` file looks like this (prettied and removed unrelated parts)

```js
// Generated by js_of_ocaml 3.5.1+git-5f956d63
(function(a) {
  "use strict";
  function b(c, b) {
    
    if (a.caml_create_file) a.caml_create_file(c, b);
    else {
      if (!a.caml_fs_tmp) a.caml_fs_tmp = [];
      a.caml_fs_tmp.push({ name: c, content: b });
    }
    return 0;
  }
  b("/static/cmis/Demo.cmi",  "some content here");
  return;
})(
  (function() {
    return this;
  })()
);
```

**Expected behavior**

`jsoo_fs` should generate this instead :

```
 if (a.caml_create_file_extern) a.caml_create_file_extern(c, b);
```

This is because `caml_create_file` isn't exposed to `this` but `caml_create_file_extern` is when building with `--toplevel` flags

**Versions**
Version of packages used to reproduce the bug
(ocamlc, js_of_ocaml, ..)

ocaml 4.06.1 
js_of_ocaml 3.5.1+git-5f956d63

hhugo:
Fix #925 

cc @thangngoc89

### DIFF
--- a/runtime/fs.js
+++ b/runtime/fs.js
@@ -229,7 +229,7 @@ function caml_fs_init (){
       caml_create_file(tmp[i].name,tmp[i].content);
     }
   }
-  joo_global_object.caml_create_file_extern = caml_create_file;
+  joo_global_object.caml_create_file = caml_create_file;
   joo_global_object.caml_fs_tmp = [];
   return 0;
 }


thangngoc89:
Thanks a lot

PR #933, opened 2019-11-28
ppx_js: behave better with merlin

trefis:


TyOverby:
@trefis please run `make fmt` 

The failure in CI is due to ocamlformat not being applied correctly to your changes

hhugo:
I'll try to run some tests later this weekend.

trefis:
Note: don't merge yet, I think this might need a bit more work :(

trefis:
Nevermind: I had some test failure, but I wasn't trying this branch... :see_no_evil:
This should be fine to merge.

hhugo:
`merlin.hide` seems to be hiding/ignoring too much information.
In particular, asking the type of the right hand side of `##.` or `##` doesn't seems to work.
By rhs, I mean `xxx` in expressions bellow 
`o##xxx`
`o##xxx a b c`
`o##.xxx`
`o##.xxx := e`

trefis:
I have just pushed a new version which should behave as well as the current ppx (I have even noticed some improved answers).

trefis:
@hhugo: CI is green.

ppx/ppx_js/tests/ppx.mlt
```diff
@@ -104,7 +104,7 @@ fun (obj : < > Js.t) -> obj##m
 
 [%%expect
 {|
-Line _, characters 24-27:
```
hhugo: This seems to be a regression. `obj`  has type `<  >` not `obj##m`

ppx/ppx_js/tests/ppx.mlt
```diff
@@ -169,7 +169,7 @@ fun (obj : < m : 'a. 'a -> unit Js.meth > Js.t) ->
 
 [%%expect
 {|
-Line _, characters 2-5:
```
hhugo: Same issue here

ppx/ppx_js/tests/ppx.mlt
trefis: Fixed.

ppx/ppx_js/tests/ppx.mlt
trefis: Fixed.

hhugo:
Thanks

PR #934, opened 2019-11-29
Hotfix runtime stdlib.js

TyOverby:
A bug in the caml_callgen code was introduced in jane/js-of-ocaml/micro-optimze-runtime.                                                                                                                           
                                                                                                                                                                                                                     
  This bug manifests itself in this call-stack:                                                                                                                                                                      
                                                                                                                                                                                                                     
  > Uncaught TypeError: f.apply is not a function                                                                                                                                                                    
  >    at caml_call_gen (stdlib.js:72)                                                                                                                                                                               
  >    at stdlib.js:100                                                                                                                                                                                              
  >    at Object.caml_call_gen (stdlib.js:72)                                                                                                                                                                        
  >    at caml_call9 (style.ml:0)                                                                                                                                                                                    
                                                                                                                                                                                                                     
  The bug is caused by aliasing a variable in the callgen code, overwriting the                                                                                                                                      
  function which will be called with one of its arguments.                                                                                                                                                           
                                                                                                                                                                                                                     
  The bug can be triggered with the following:                                                                                                                                                                       
  1. Call a function that has an arity of 7 with 1 argument (currying the rest).                                                                                                                                     
  2. Call the resulting function with all 6 remaining arguments.                                                                                                                                                     
  3. Instead of actually calling the function, it will (attempt to) call the 6th                                                                                                                                     
     argument with the args list.

### DIFF
--- a/compiler/tests/call_gen.ml
+++ b/compiler/tests/call_gen.ml
@@ -19,6 +19,7 @@
 
 open Util
 
+module M1 = struct
 let code =
   {|
   let f_prime g = g 1 2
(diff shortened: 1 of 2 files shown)


hhugo:
Could we rename the function f or the rename all arguments with a different pattern (a1,a2,..an)?

Could we also introduce a regtest?

TyOverby:
I can craft a repro on monday, but I’ve updated the variable names already (yay writing code on GitHub on my phone)

TyOverby:
Regression test added

hhugo:
Thanks

PR #609, opened 2017-07-12
Add progressive web app support

aik9508:
Add interfaces of serviceWorker API and Promise API which allow users to create progressive web apps with Ocsigen. Several necessary functions are also added in Url.ml and Form.ml.

PR #944, opened 2019-12-20
[WIP] add Promise API

adrien-n:
This PR initially extracts the first coming of #609. There is more work needed to take @aantron's comments into account.

### DIFF
--- a/lib/lwt/lwt_promise.ml
+++ b/lib/lwt/lwt_promise.ml
@@ -1,0 +1,18 @@
+open Js_of_ocaml
+open Promise
+
+let to_lwt (p : 'a promise Js.t) : 'a Lwt.t =
+  let (r, w) = Lwt.task () in
+  ignore
+    ((Js.Unsafe.coerce p)##_then
+       (Js.wrap_callback (fun v -> Lwt.wakeup w v))
+       (Js.wrap_callback (fun e -> Lwt.wakeup_exn w (try raise e with e -> e))));
+  r
+
+let of_lwt (p : unit -> 'a Lwt.t) : 'a promise Js.t =
+  new%js _Promise
+    (Js.wrap_callback
+       (fun resolve reject ->
+          Lwt.try_bind p
+            (fun v -> Js.Unsafe.fun_call resolve [|Js.Unsafe.inject v|])
+            (fun e -> Js.Unsafe.fun_call reject [|Js.Unsafe.inject e|])))
--- a/lib/promise.ml
+++ b/lib/promise.ml
@@ -1,0 +1,32 @@
+type 'a promise 
+
+let _Promise = Js.Unsafe.global##._Promise
+
+let _catch (p : 'a promise Js.t) (f : _ -> 'a promise Js.t) : 'a promise Js.t =
+  (Js.Unsafe.coerce p)##_catch (Js.wrap_callback f)
+
+let _then (p : 'a promise Js.t) ?(catch: (_ -> 'b promise Js.t) option)
+    (f: 'a -> 'b promise Js.t) : 'b promise Js.t =
+  match catch with
+  | None ->
+    (Js.Unsafe.coerce p)##_then (Js.wrap_callback f)
+  | Some catch ->
+    (Js.Unsafe.coerce p)##_then (Js.wrap_callback f) (Js.wrap_callback catch)
+
+let resolve_value (v :'a) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "resolve" [|Js.Unsafe.inject v|]
+
+let resolve_promise (v :'a promise Js.t) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "resolve" [|Js.Unsafe.inject v|]
+
+let reject_value (v :'a) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "reject" [|Js.Unsafe.inject v|]
+
+let reject_promise (v :'a promise Js.t) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "rejetct" [|Js.Unsafe.inject v|]
+
+let race (promise_list : 'a promise Js.t Js.js_array Js.t) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "race" [|Js.Unsafe.inject promise_list|]
+
+let all (promise_list : 'a promise Js.t Js.js_array Js.t) : 'a promise Js.t =
+  Js.Unsafe.meth_call _Promise "all" [|Js.Unsafe.inject promise_list|]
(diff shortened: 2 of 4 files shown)


aantron:
I'd suggest an API more along the lines of [reason-promise](https://github.com/aantron/repromise/blob/master/src/js/promise.rei) (I haven't renamed that repo yet). The docs are out of date - finishing up that repo is my next main task. See https://github.com/aantron/repromise/blob/37b4a64ab681125193ecc5fd1c08cd4d862bb2bf/src/js/promise.rei.

There is a native implementation in pure OCaml (currently in Reason syntax). The JS implementation is a direct binding to JS promises, but with a type safety fix.

adrien-n:
Hmmm, I had actually inferred that you were more in favor of pushing some (light) burden on binding authors. TBH it was a bit difficult to follow across the many bug reports, comments and projects.

Do you advise to add work inside repromise's repository or in jsoo's? Which current implementation would be the best base and do you think there is a practical way to share code with it?

aantron:
> Do you advise to add work inside repromise's repository or in jsoo's?

That depends on whether jsoo can accept a dependency for this. If so, I suggest repromise.

> Which current implementation would be the best base and do you think there is a practical way to share code with it?

The JS implementation, since we seem to be binding JS promises here. We would just have to port it from BuckleScript's FFI to jsoo's, and probably convert syntax. The result would be a promise API that is type-safe, but it would still be dependent on the underlying JS engine for avoiding certain kinds of memory leaks. Some JS engines' promise libraries avoid them, and others don't.

aantron:
Another reason not to work inside repromise is if jsoo wants to provide a fully unopinionated JS promise binding. In particular, JS promises can be rejected, while the promises of Repromise's main API can't be. They favor using `result` with promises instead, which was the result of some discussion in the Reason community.

Repromise also has a `Promise.Js` submodule, which exposes JS promises with rejection, which is what this PR currently corresponds to.

Undernearth, `Promise.t` and `Promise.Js.t` have the same definition. `Promise` just exposes less functions than `Promise.Js`.

aantron:
Going on, `Promise.Js.t` is intended for writing bindings to JS libraries, but bindings authors are encouraged to handle rejections inside the binding and expose only `Promise.t`s to their users, perhaps with nested `result`s. This should usually be pretty straightforward, because `'a Promise.t = ('a, never) Promise.Js.t`, where `never` is an uninhabited type, and there are various helpers for catching rejections.

aantron:
...and a reason to work in repromise is that by the time type soundness is fixed for this version, it will be essentially a clone of repromise, just for the jsoo FFI. I think we should stop these things from proliferating :)

lib/promise.mli
```diff
@@ -0,0 +1,13 @@
+type 'a promise
+val _Promise : 'a
+val _catch : 'a promise Js.t -> ('b -> 'a promise Js.t) -> 'a promise Js.t
+val _then :
+  'a promise Js.t ->
+  ?catch:('c -> 'b promise Js.t) ->
+  ('a -> 'b promise Js.t) -> 'b promise Js.t
+val resolve_value : 'a -> 'a promise Js.t
+val resolve_promise : 'a promise Js.t -> 'a promise Js.t
+val reject_value : 'a -> 'a promise Js.t
```
aantron: I think this should be `'a -> 'b promise Js.t`, in the style of this binding. The rejection type is not related to the fulfillment type. The same of `reject_promise`.

lib/promise.mli
```diff
@@ -0,0 +1,13 @@
+type 'a promise
+val _Promise : 'a
+val _catch : 'a promise Js.t -> ('b -> 'a promise Js.t) -> 'a promise Js.t
+val _then :
+  'a promise Js.t ->
+  ?catch:('c -> 'b promise Js.t) ->
+  ('a -> 'b promise Js.t) -> 'b promise Js.t
+val resolve_value : 'a -> 'a promise Js.t
```
aantron: As already mentioned in the PR's comment, this is where the unsoundness issue arises. If `resolve_value` is called at type `foo promise Js.t`, the result type will be `foo promise promise Js.t`. However, JS will produce a value whose representation corresponds to `foo promise Js.t`, i.e. it will silently "call" `join` to flatten the nested promise.

The user of this binding has to deliberately choose to call `resolve_promise` to encode this in OCaml. However, that is error-prone, and can't be done in polymorphic function.

lib/promise.mli
```diff
@@ -0,0 +1,13 @@
...
+val resolve_promise : 'a promise Js.t -> 'a promise Js.t
+val reject_value : 'a -> 'a promise Js.t
+val reject_promise : 'a promise Js.t -> 'a promise Js.t
+val race : 'a promise Js.t Js.js_array Js.t -> 'a promise Js.t
+val all : 'a promise Js.t Js.js_array Js.t -> 'a promise Js.t
```
aantron: IIRC `all`'s result type should include a JS array of result values.

lib/promise.mli
```diff
@@ -0,0 +1,13 @@
+type 'a promise
+val _Promise : 'a
+val _catch : 'a promise Js.t -> ('b -> 'a promise Js.t) -> 'a promise Js.t
```
aantron: `'b` is the rejection type, and can be any JS value. However, once we pass OCaml callbacks to JS `catch` and `then`, what will happen to OCaml exceptions raised in the callbacks? If they turn into JS exceptions, the JS promise functions will turn them into rejections, which will further complicate what `'b` can be, and we will need to prepare users of this promise API for that.

lib/promise.ml
```diff
@@ -0,0 +1,32 @@
+type 'a promise 
+
+let _Promise = Js.Unsafe.global##._Promise
+
+let _catch (p : 'a promise Js.t) (f : _ -> 'a promise Js.t) : 'a promise Js.t =
```
aantron: `~f` appears not to match the signature, same for `_then`.

lib/lwt/lwt_promise.ml
```diff
@@ -0,0 +1,18 @@
+open Js_of_ocaml
+open Promise
+
+let to_lwt (p : 'a promise Js.t) : 'a Lwt.t =
+  let (r, w) = Lwt.task () in
+  ignore
+    ((Js.Unsafe.coerce p)##_then
+       (Js.wrap_callback (fun v -> Lwt.wakeup w v))
+       (Js.wrap_callback (fun e -> Lwt.wakeup_exn w (try raise e with e -> e))));
```
aantron: If this is the `onReject` callback, it looks like it is capable of rejecting Lwt promises with any value, because JS promises can be rejected with any value, but Lwt promises can only be rejected with OCaml exceptions `exn`.

hhugo:
Based on @aantron comments and the inactivity on the PR, I'm going to close this.

It seems that we need something similar to https://github.com/aantron/promise.
@adrien-n, feel free to reopen a PR with a new version

hhugo:
https://github.com/mnxn/promise_jsoo seems to be the ocaml equivalent of https://github.com/aantron/promise

PR #946, opened 2019-12-23
Compiler: fix vardecl optim

### LINKED ISSUE #945: [FEATURE REQUEST] - Variable hoisting in decision trees.
`generate.ml` currently compiles decision trees like this:

```javascript
function extension_constructor(x) {
  if (is_block(x))
    if (caml_obj_tag(x) !== 248)
      if (1 <= x.length - 1) {
        var slot = x[1];
        var switch__0 = 1;
      } else {
        var switch__0 = 0;
      }
    } else {
      var switch__0 = 0;
    }
  } else {
    var switch__0 = 0;
  }
  if (!switch__0) {
    var slot = x;
  }
  if (is_block(slot))
    if (caml_obj_tag(slot) === 248) {
      var name = slot[1];
      var switch__1 = 1;
    } else {
      var switch__1 = 0;
    }
  } else {
    var switch__1 = 0;
  }
  if (!switch__1) {
    var name = call1(Pervasives[1], cst_Obj_extension_constructor__0);
  }
  return caml_obj_tag(name) === 252
    ? slot
    : call1(Pervasives[1], cst_Obj_extension_constructor);
}
```

This is limiting and problematic because the variables are declared deep in the branches, and not every branch declares the same set of variables. jsoo relies on the fact that (in this case) "if `switch__0` was not set to `1`, then we know that `slot` was not set and so we'll set it after the loop". This has been troublesome for us because various language linters and type checkers will get tripped up on the fact that it looks like `slot` may not always be declared. Languages like PHP and JS actually have function scope, so these variables are declared at the top of the function as undefined whether or not they were explicitly declared - their mere presence in the body of the function make it equivalent to `var slot = undefined`. For extending jsoo to different backends, it would make it much easier if the variables were hoisted up to the top of the function, even if they had `None` expression at the time of `Variable_declaration` at the top of the function.

Instead, I could imagine:

```javascript
function extension_constructor(x) {
  ...
  // Hoisted as high as necessary.
  var slot, switch__0, switch__1, name;
  if (is_block(x))
    if (caml_obj_tag(x) !== 248)
      if (1 <= x.length - 1) {
        slot = x[1];
        switch__0 = 1;
      } else {
        switch__0 = 0;
      }
    } else {
      switch__0 = 0;
    }
  } else {
    switch__0 = 0;
  }
  if (!switch__0) {
    slot = x;
  }
  if (is_block(slot))
    if (caml_obj_tag(slot) === 248) {
      name = slot[1];
      switch__1 = 1;
    } else {
      switch__1 = 0;
    }
  } else {
    switch__1 = 0;
  }
  if (!switch__1) {
    name = call1(Pervasives[1], cst_Obj_extension_constructor__0);
  }
  return caml_obj_tag(name) === 252
    ? slot
    : call1(Pervasives[1], cst_Obj_extension_constructor);
}
```

hhugo:
Attempt to fix the `--enable vardecl` option.
- [x] add some tests

partially fix #945

### DIFF
--- a/compiler/lib/javascript.ml
+++ b/compiler/lib/javascript.ml
@@ -284,11 +284,11 @@ and source_element =
 let compare_ident t1 t2 =
   match t1, t2 with
   | V v1, V v2 -> Code.Var.compare v1 v2
-  | S { name = s1; var = v1; loc = l1 }, S { name = s2; var = v2; loc = l2 } -> (
+  | S { name = s1; var = v1; loc = _l1 }, S { name = s2; var = v2; loc = _l2 } -> (
       match String.compare s1 s2 with
       | 0 -> (
           match Option.compare Code.Var.compare v1 v2 with
-          | 0 -> Poly.compare l1 l2
+          | 0 -> 0
           | n -> n)
       | n -> n)
   | S _, V _ -> -1
--- a/compiler/lib/javascript.ml
+++ b/compiler/lib/javascript.ml
@@ -284,12 +284,10 @@ and source_element =
 let compare_ident t1 t2 =
   match t1, t2 with
   | V v1, V v2 -> Code.Var.compare v1 v2
-  | S { name = s1; var = v1; loc = _l1 }, S { name = s2; var = v2; loc = _l2 } -> (
+  | S { name = s1; var = v1; loc = _ }, S { name = s2; var = v2; loc = _ } -> (
+      (* ignore locations *)
       match String.compare s1 s2 with
-      | 0 -> (
-          match Option.compare Code.Var.compare v1 v2 with
-          | 0 -> 0
-          | n -> n)
+      | 0 -> Option.compare Code.Var.compare v1 v2
       | n -> n)
   | S _, V _ -> -1
   | V _, S _ -> 1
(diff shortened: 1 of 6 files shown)


hhugo:
cc @jordwalke

jordwalke:
Do you believe this is not a complete fix?

hhugo:
This might be working (or not) but I'd like to spend more time on this.

compiler/lib/javascript.ml
```diff
@@ -284,11 +284,11 @@ and source_element =
 let compare_ident t1 t2 =
   match t1, t2 with
   | V v1, V v2 -> Code.Var.compare v1 v2
-  | S { name = s1; var = v1; loc = l1 }, S { name = s2; var = v2; loc = l2 } -> (
+  | S { name = s1; var = v1; loc = _l1 }, S { name = s2; var = v2; loc = _l2 } -> (
       match String.compare s1 s2 with
       | 0 -> (
           match Option.compare Code.Var.compare v1 v2 with
```
jordwalke: Could just return `Option.compare Code.Var.compare v1 v2` I suppose.

jordwalke:
It does seem to be working fine when testing! Thank you @hhugo

jordwalke:
Seems worthwhile merging since it is strictly better than what master branch has.

jordwalke:
Oh, hold on I think I'm getting that error message about escaping variables when compiling stdlib.js (CamlinternalFormatBasics.js).

PR #937, opened 2019-12-10
add spread operator support to JavaScript AST

TyOverby:
This PR adds support for the Javascript spread operator in function call position.

This is necessary for linking with any Javascript that uses this operator.

PR #957, opened 2020-01-10
Create stdlib_modern.js

TyOverby:
Depends on https://github.com/ocsigen/js_of_ocaml/pull/937

* Uses the spread operator in callgen to avoid allocating more arrays
* Uses `List.slice` to implement some primitives

### DIFF
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -21,17 +21,12 @@
 
 //Provides: raw_array_sub
 function raw_array_sub (a,i,l) {
-  var b = new Array(l);
-  for(var j = 0; j < l; j++) b[j] = a[i+j];
-  return b
+  return a.slice(i, i + l);
 }
 
 //Provides: raw_array_copy
 function raw_array_copy (a) {
-  var l = a.length;
-  var b = new Array(l);
-  for(var i = 0; i < l; i++ ) b[i] = a[i];
-  return b
+  return a.slice();
 }
 
 //Provides: raw_array_cons
@@ -44,6 +39,7 @@ function raw_array_cons (a,x) {
 }
 
 //Provides: caml_call_gen (const, shallow)
+//Weakdef
 function caml_call_gen(f, args) {
   var args_copied = false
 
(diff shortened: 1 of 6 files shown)


runtime/stdlib.js
```diff
@@ -21,17 +21,12 @@
 
 //Provides: raw_array_sub
 function raw_array_sub (a,i,l) {
-  var b = new Array(l);
-  for(var j = 0; j < l; j++) b[j] = a[i+j];
-  return b
+  return a.slice(i, i + l);
```
vouillon: Apparently, `slice` used to be slower (see 1bcded11b03aa2d335ef45073072994b2fc898d4). Is this still the case?

TyOverby: According to [this microbenchmark](http://jsben.ch/slRuP), `slice` is three times faster

TyOverby:
Ping @hhugo

hhugo:
Could you add a test that would explicitly link with this modern_stdlib.js

hhugo:
In particular, I want to make sure the file is valid

hhugo:
Ping @TyOverby

TyOverby:
@hhugo: I tried to write a test that passes the file through jsoo_minify just to make sure that the syntax is correct.  However, it's reporting that the file can't be parsed, but that the unparsable line is this one:

```javascript
    if (d === 0) {
//                  ^ this is character 15
```
```
../../install/default/bin/jsoo_minify: Error: error at file:"stdlib_modern.js" l:24 col:15
```

Do you know what's going on here?

hhugo:
There is possibly a one off error when reporting the line. Could it be failing at line 25 ?

TyOverby:
@hhugo Looks like the issue was due to me not rebasing often enough haha

Are you happy with this test that verifies syntax, or should I find a way for us to test the semantics too?

OlivierNicole:
I’m trying to figure out what `stdlib_modern.js` is for and what is its role compared with `stdlib.js`. Would it be possible to clarify this for me @TyOverby @hhugo?

hhugo:
I believe it was an attempt from @TyOverby to speedup part of the runtime by relying on newer features of JS.
It's used internally at janestreet.

OlivierNicole:
I see, thanks.

PR #975, opened 2020-04-01
Revert to dune.1

hhugo:
This will not be merged on master but will be cherry-picked for releases.

Until https://github.com/esy/esy/issues/1062 is addressed.


rgrinberg:
Note that this has been worked around on the esy side. They just publish an "override" for every new dune release. That occurs fairly quickly.

PR #977, opened 2020-04-04
recognize String concat

hhugo:
The major issue with this PR is that we're only able to recognize string concat when all instructions are part of the same block  / when debug info are disabled.


hhugo:
https://github.com/ocaml/ocaml/pull/9565

nojb:
Hi @hhugo, just to keep your use-case in mind, is it right that changing `^` to be an external (either in C or a compiler primitive) would be enough for the needs of `js_of_ocaml` ?

hhugo:
@nojb, That's right

PR #979, opened 2020-04-04
Remove console.log in `caml_sys_system_command`

smondet:


smondet:
Added a second commit because `execSync` actually seems to by default grab stdout and return it as a string:
https://stackoverflow.com/questions/30134236/use-child-process-execsync-but-keep-output-in-console

hhugo:
Thanks.

Can you add an entry to the changelog ?

Having a small reg-test for this would be nice (look at compiler/tests-jsoo/, after rebasing)

hhugo:
Looks good. I'll merge once travis is green.

PR #993, opened 2020-04-20
Runtime: fix caml_call gen

hhugo:
Fix regression introduced in #926
The main issue is that caml_call_gen can be called with `arguments` (instance of `Arguments`) which doesn't have any of the array methods.

### DIFF
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -55,10 +55,7 @@ function caml_call_gen(f, args) {
     }
     else if (d < 0) {
       if (!args_copied) {
-        if(!args.slice)
-          args = Array.prototype.slice.call(args);
-        else
-          args = args.slice();
+        args = Array.prototype.slice.call(args);
         args_copied = true;
       }
 
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -34,6 +34,7 @@ function raw_array_cons (a,x) {
 }
 
 //Provides: caml_call_gen (const, shallow)
+//Requires: caml_failwith
 //Weakdef
 function caml_call_gen(f, args) {
   var args_copied = false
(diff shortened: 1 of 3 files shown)


hhugo:
There is still a pending issue here regarding over application.
When a wrapped callback is over applied and an intermediate result can not be applied.
for example:
```ocaml
let cb = wrap_callback (fun f a b -> a + b)
fun_call bc [| inject 1; inject 2|] (* returns 3 *)
fun_call bc [| inject 1; inject 2; inject 3|] (* returns a function *)
```
 
In such case, the previous implementation would fully apply `cb` and return a function (that would accumulate any argument you give it). I strongly believe this was unintended and is a bug.

The two alternatives I see are
1. raise an exception
2. return any intermediate result that is not a function

This PR implements (2), and we have the following semantic:
```ocaml
let cb = wrap_callback (fun f a b -> a + b)
fun_call bc [| inject 1; inject 2|] (* returns 3 *)
fun_call bc [| inject 1; inject 2; inject 3|] (* returns 3, ignoring the last argument *)
```

runtime/stdlib.js
```diff
@@ -48,17 +43,22 @@ function caml_call_gen(f, args) {
...
     if (d == 0) {
       return f.apply(null, args);
     }
     else if (d < 0) {
       if (!args_copied) {
-        args = args.slice();
+        if(!args.slice)
+          args = Array.prototype.slice.call(args);
+        else
+          args = args.slice();
```
joelburget, line 61: Is there an advantage to the conditional? Otherwise I'd expect this (which works for both arrays and `Arguments`):

```suggestion
        args = Array.prototype.slice.call(args);
```

hhugo: I don't know if there are any differences in perf. @TyOverby, any idea ?

hhugo:
>The two alternatives I see are
>
> 1. raise an exception
> 2. return any intermediate result that is not a function

An slightly different approach would be do 1 or 2 depending on the context. Do 2 when wrapping callbacks, do 1 otherwise. In practice, `caml_call_gen` could take an extra (optional ?) argument to control the behavior.

hhugo:
superseded by #996

PR #994, opened 2020-04-21
Runtime: fix caml_call gen (Alternative)

hhugo:
This is an alternative to #993 

Based on https://github.com/petkaantonov/bluebird/wiki/Optimization-killers, leaking `arguments` is bad.

>What is safe arguments usage?
> Only use
> 
> - arguments.length
> - arguments[i] where i is always a valid integer index into the arguments, and can not be out of bound
> - Never use arguments directly without .length or [i]
> - STRICTLY fn.apply(y, arguments) is ok, nothing else is, e.g. .slice. Function#apply is special.

This PR changes the runtime so that caml_call_gen is never called with an `Arguments` value.

### DIFF
--- a/compiler/lib-runtime/jsoo_runtime.ml
+++ b/compiler/lib-runtime/jsoo_runtime.ml
@@ -40,6 +40,7 @@ let runtime =
     ; nat
     ; parsing
     ; stdlib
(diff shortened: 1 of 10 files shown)


hhugo:
superseded by #996

PR #995, opened 2020-04-21
Continue caml_call_gen cleanup

hhugo:
Following #994, here is an experiment to further improve caml_call_gen:

- control the behavior when too many arguments are given: raise error or ignore them
- never copy the `args` array argument in caml_call_gen and happily mutate it.


hhugo:
superseded by #996

PR #1000, opened 2020-04-24
Runtime: specialize caml_call_gen

hhugo:
Two implementations of caml_call_gen for two different use cases
- function calls generated by the compiler
- wrapping functions into javascript callbacks


hhugo:
@TyOverby, Do you mind adding the optimization back to `stdlib_modern.js` ?

hhugo:
On hold until we can prove the new implementation is faster

hhugo:
Closing, reopen if needed.

PR #1010, opened 2020-05-01
replace camlp4 syntax with ppx syntax 

bikallem:
Since the rest of the manual mentions that ppx syntax is the recommended approach, we update it in the manual too.


hhugo:
Thanks

Issue #1006, opened 2020-04-28
Runtime: update runtime for marshal


hhugo:
Following https://github.com/ocaml/ocaml/pull/8791#issuecomment-620657231

PR #1011, opened 2020-05-01
OCaml 4.11 support

kit-ty-kate:
As none of the alpha/beta of 4.11.0 have yet been release I'll put that as a draft, but as it currently stands, this should fix js_of_ocaml-compiler with OCaml 4.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ jobs:
       name: "4.10 all targets"
     - os: osx
       env: OCAML_VERSION=4.10 MODE=opam
+    - os: linux
+      env: OCAML_VERSION=4.11.0+trunk OCAML_BETA=enable MODE=opam
 
   allow_failures:
     - os: osx


hhugo:
Thanks a lot

hhugo:
We'll need to resolve #1006 as well

hhugo:
I've cherry-picked the code change in #1015.

kit-ty-kate:
The first **alpha** for OCaml 4.11.0 has just been released btw. This PR could be closed, however it still adds CI support for 4.11

hhugo:
Do you want to update this PR for the alpha ? I can merge it once the CI pass.
You'll need some more changes now that the magic number is bumped

hhugo:
I've bumped the magic number.

kit-ty-kate:
@hhugo it seems that for this PR to succeed, the CI needs its cache cleared. However I don't have enough right on this repo to do it, could you do it instead?

hhugo:
replaced by #1055

PR #1012, opened 2020-05-03
Compiler: rewrite the js parser using menhir incremental api

hhugo:
Before this PR, the js lexer was trying to adjust its production based on the previous token.

- INCR vs INCR_NB ("NB" stands for "No line Break")
- DECR vs DECR_NB
- DIV vs DIV_ASSIGN vs REGEXP

The lexer was also populating the fol (first of line) field that was later used by `Js_parser.Lexer.adjust_token` and the js parser for semicolon automatic insertion.

Using the incremental API of menhir (where the control flow is inverted), we can now remove all theses hacks and move all the logic inside `parser_aux`.

This also has the advantage of being much more friendly with the stack when running in javascript (jsoo toplevel)


hhugo:
fyi @let-def

let-def:
Cool!

TyOverby:
Is the goal with this PR to allow for better regexp parsing?

compiler/lib/js_lexer.mll
```diff
@@ -74,16 +74,7 @@ let update_loc lexbuf ?file ~line ~absolute chars =
     pos_bol = pos.pos_cnum - chars;
                               }
 
-let tokinfo prev lexbuf =
```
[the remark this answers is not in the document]
hhugo: `tokinfo` was using `prev` to populate the `fol` field which was later used elsewhere for deal with semi-colon automatic insertion. I've updated the PR description.

hhugo:
This PR allows us to parse regexp literal correctly and in a much cleaner way.

compiler/lib/parse_js.ml
```diff
@@ -19,115 +19,165 @@
...
+                    T_VIRTUAL_SEMICOLON Parse_info.zero
+                (* The practical effect of these restricted productions is as follows:
+                 * When a ++ or -- token is encountered where the parser would treat it
+                 * as a postfix operator, and at least one LineTerminator occurred between
+                 * the preceding token and the ++ or -- token, then a semicolon is automatically
+                 * inserted before the ++ or -- token. *)
+                | _, (T_DECR cpi as tok) when not (fol prev tok) -> Js_token.T_DECR_NB cpi
+                | _, (T_INCR cpi as tok) when not (fol prev tok) -> Js_token.T_INCR_NB cpi
+                | _, ((T_DIV _ | T_DIV_ASSIGN _) as tok) ->
+                    if I.acceptable checkpoint tok lexbuf.Lexing.lex_start_p
```
TyOverby, line 107: What does this condition mean?

hhugo: This is part of the incremental api. It checks if a token is valid given a parser state (checkpoint).
Div and Regexp tokens and mutually exclusive in the JavaScript parser. We first check if div is an acceptable token, if not, we reset the lexer and lex a Regexp instead

PR #1013, opened 2020-05-03
Eliminate allocation of dummy function  

hhugo:
The PR: 
- updates internalMod to not use the `.fun` property proxy for functions.
- adds an optimization to eliminate allocation of dummy function (`caml_alloc_dummy_function`). The dummy allocation is not necessary in JavaScript due to the weird non-lexical scoping.

Assuming the optimization properly eliminate all occurrences of `caml_alloc_dummy_function`,  we should be able to remove `.fun` property handling from the runtime.   

Note that `caml_alloc_dummy_function` was introduced in OCaml `4.02.3`, we would need to bump the lower bound from `4.02.0` to `4.02.3` which is fine.

### DIFF
--- a/compiler/tests-compiler/recfun.ml
+++ b/compiler/tests-compiler/recfun.ml
@@ -95,18 +95,20 @@ let myfun x =
   end
   in
   M.hash_fold_t (create ()) (List [ Atom "asd"]),
-  M.hash (List [ Atom "asd"])
+  M.hash (List [ Atom "asd"]),
+  M.hash (List [ ])
 
   |}
   in
   print_fun_decl program (Some "myfun");
   [%expect
     {|
-    function myfun(x$0)
+    function myfun(x)
      {function hash_fold_t(hsv,arg)
        {if(0 === arg[0])
          {var a0=arg[1],hsv$0=hsv | 0;return hash_fold_string(hsv$0,a0)}
         var a0$0=arg[1],hsv$1=hsv + 1 | 0;
         return hash_fold_list(hash_fold_t,hsv$1,a0$0)}
-      var _b_=hash_fold_t(42,x);
-      return [0,hash_fold_t(42,_a_),_b_]} |}]
+      function hash(x){return hash_fold_t(42,x)}
+      var _d_=hash(_a_),_e_=hash(_b_);
+      return [0,hash_fold_t(42,_c_),_e_,_d_]} |}]
--- a/compiler/tests-compiler/recfun.ml
+++ b/compiler/tests-compiler/recfun.ml
@@ -1,0 +1,117 @@
+(* Js_of_ocaml compiler
+ * http://www.ocsigen.org/js_of_ocaml/
+ * Copyright (C) 2020 Hugo Heuzard
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, with linking exception;
+ * either version 2.1 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


compiler/tests-compiler/recfun.ml
```diff
@@ -0,0 +1,110 @@
...
+      (fun hsv ->
+         fun arg ->
+           match arg with
+           | Atom _a0 ->
+             let hsv = fold_int hsv 0 in
+             let hsv = hsv in hash_fold_string hsv _a0
+           | List _a0 ->
+             let hsv = fold_int hsv 1 in
+             let hsv = hsv in hash_fold_list hash_fold_t hsv _a0 : state -> t -> state)
+    and (hash : t -> hash_value) =
```
TyOverby, line 91: I'm not seeing this function in the output.  Did it get inlined?

runtime/internalMod.js
```diff
@@ -53,22 +53,27 @@ function caml_CamlinternalMod_init_mod(loc,shape) {
...
+      switch(shape[0]){
+      case 0://module
+        for(var i=1;i<shape[1].length;i++)
+          loop(shape[1][i],real[i],x[i],real,i);
+        break;
+        //case 1://Value
+      default:
+      };
+  }
+  loop(shape,real,x,undefined,undefined);
```
TyOverby, line 77: Why is there a loop now?

compiler/tests-compiler/recfun.ml
hhugo: Yes

TyOverby: Could you add a test with functions where none of them get inlined?

runtime/internalMod.js
hhugo: Update_mod was already recursive. We introduce this inner recursive function loop to keep track of parent and i and allow to  more direct update for function

compiler/tests-compiler/recfun.ml
hhugo: Note that the inlining was enabled by the last commit. Probably better the review per commits

hhugo: I've tune the test to prevent inlining

hhugo:
@xclerc, do you want to take a look at this one ?

xclerc:
Just to be sure I understand the optimization: the idea is that every time we have
both `x = caml_alloc_dummy_function(...)` and `caml_update_dummy(x, y)`, we
can get rid of both function calls and use `y` instead of `x`, right?

It looks like [this call to filter](https://github.com/ocsigen/js_of_ocaml/pull/1013/files#diff-d36f6c73c98b086e0c53e7ce796c8703R110) is expected to be the identity. My understanding is
that it might be true only given the current state of the code generator of `ocamlc`
(`Bytegen.comp_expr`), but then if we want to be independent of the code generator,
wouldn't it be safer to detect e.g. several `caml_update_dummy(x, y_i)` calls?

hhugo:
> Just to be sure I understand the optimization: the idea is that every time we have
both x = caml_alloc_dummy_function(...) and caml_update_dummy(x, y), we
can get rid of both function calls and use y instead of x, right?

Correct

> It looks like this call to filter is expected to be the identity. My understanding is
that it might be true only given the current state of the code generator of ocamlc
(Bytegen.comp_expr), but then if we want to be independent of the code generator,
wouldn't it be safer to detect e.g. several caml_update_dummy(x, y_i) calls?

That sound correct. We should at least add some assertion.

hhugo:
For the record, caml_alloc_dummy_function was introduced in 4.02.3

hhugo:
> It looks like this call to filter is expected to be the identity.

Looking more closely at the code, it's not the case. [filter] is here to only keep info from [caml_update_dummy] that have a corresponding [caml_alloc_dummy_function]. There are other cases (e.g. `let rec l = 0 :: l`).

I've updated the code with comments.

PR #1032, opened 2020-07-06
Upgrade to ocamlformat 0.15.0

gpetiot:
Hi, this is what the project would look like after being reformated with the current version (candidate) of ocamlformat 0.15.0.
**Please do not merge until ocamlformat.0.15.0 is available with opam.**

Please let me know if you see any regressions that I may have missed.


dannywillems:
@hhugo As ocamlformat.0.15.0 is in OPAM now, I think it can be merged.

PR #1041, opened 2020-07-15
Port all PPX's to ppxlib

pitag-ha:
Before, the two extension node rewriters in the compiler and in ppx_js were based on the omp driver and on the omp Ast_mapper. The deriver in ppx_deriving_json was already partly based on ppxlib, but also on omp's Ast_408. All of those omp modules will get dropped in omp 2.0.0. For more information on that omp change, see https://discuss.ocaml.org/t/ocaml-migrate-parsetree-2-0-0/5991 by @jeremiedimino.

That's why this commit ports those 3 PPX's to ppxlib.

Closes ocaml-ppx/ppxlib#134


pitag-ha:
In order to decide what to do with the exposed mapper in Ppx_js etc, I'd like to understand what that mapper is usually used for (apart from the internal tests). What is the idea behind exposing that mapper? I've seen the example "One can add a ppx rewriter to a toplevel by registering it.
`let init () = Compiler_libs.Ast_mapper.register "js_of_ocaml" (fun _ -> Ppx_js.mapper)` " in that context but don't quite understand how that's usually used.

pitag-ha:
What's the reason why a local toplevel_expect_test is used? If I'm able to do that, is it ok to use the upstream one?

hhugo:
### local toplevel_expect_test
The idea was to be able to run tests with ppxes on multiple OCaml version. The easiest way I found was just have multiple copies and patch them as needed. If you find a easier solution, I'm happy to change it.

### ast mappers
- the mapper in ppx_js_internal.mli is exposed because it's used to implement the non internal ppx
- the mapper in ppx/ppx_js/as-lib/ppx_js.ml is used by to register the extension when building a toplevel.
- the mapper in ppx/ppx_js/lib/ppx_js_rewriter.mli is probably not necessary.

compiler/ppx/ppx_optcomp_light.ml
```diff
@@ -181,32 +179,37 @@ let rec filter_pattern = function
   | { ppat_attributes; ppat_loc; _ } as p ->
       if keep ppat_loc ppat_attributes then Some p else None
 
-let mapper =
-  { Ast_mapper.default_mapper with
-    cases =
-      (fun mapper cases ->
-        let cases =
```
hhugo, line 188: Can we maybe add support for this in ppxlib

hhugo:
What's the status of this ? Are you still working on it ?

compiler/ppx/ppx_optcomp_light.ml
pitag-ha: Yes, that's a good idea. Thanks! I've done that now.

pitag-ha:
I'm sorry for the delay on this. As far as I can see, now there are no big issues anymore. Just one thing (which is why for now I've still left the "WIP"): until ocamlformat is finished being ported to ppxlib, there is a conflict of the version of ppxlib I'm requiring (>=0.17.0) with ocamlformat.

hhugo:
@pitag-ha , I've pushed a commit to your branch. It makes jsoo compatible with both ppxlib < 0.17 and ppxlib >= 0.17

compiler/ppx/ppx_optcomp_light.ml
```diff
@@ -82,11 +80,11 @@ let get_env s =
   | None -> Sys.getenv s
   | Some p -> p
 
-let keep loc (attrs : attributes) =
+let keep loc (attrs : Ppxlib.attributes) =
```
hhugo, line 83: I would assume `attributes` to be in scope because of the `open Ppxlib.Parsetree` above

ppx/ppx_deriving_json/lib/ppx_deriving_json.ml
```diff
@@ -79,9 +78,9 @@ let int ?loc ?attrs x = Exp.constant ?loc ?attrs (Pconst_integer (string_of_int
 
 let pint ?loc ?attrs x = Pat.constant ?loc ?attrs (Pconst_integer (string_of_int x, None))
 
-let lid ?(loc = !default_loc) s = Location.mkloc (parse_lid s) loc
+let lid ?(loc = !default_loc) s = { txt = parse_lid s; loc }
```
hhugo, line 81: What about introducing two functions `mkloc` and `mknoloc` as a way to make the diff smaller ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -127,13 +136,22 @@ let merlin_hide =
 
 module Js : sig
   val type_ :
-    ?loc:Ast_helper.loc -> string -> Parsetree.core_type list -> Parsetree.core_type
+       ?loc:Ppxlib.Ast_helper.loc
+    -> string
+    -> Ppxlib.Parsetree.core_type list
```
hhugo, line 141: `Ppxlib` is opened at the top of the file, can we simplify the path to `Parsetree.core_type` and make the diff smaller ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -451,6 +471,7 @@ let new_object constr args =
     invoker
     ((app_arg (Exp.ident ~loc:constr.loc constr) :: args)
     @ [ app_arg (unit ~loc:gloc ()) ])
+  [@@ocaml.warning "-26"]
```
hhugo, line 474: what is this warning 26 about ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -563,11 +589,26 @@ let preprocess_literal_object mappper fields :
           | (Immutable | Mutable), [ `Writeonly ] -> `Writeonly
           | (Immutable | Mutable), [ `Readwrite ] -> `Readwrite
           | (Immutable | Mutable), [ `Unkown s ] ->
-              Location.raise_errorf ~loc:exp.pcf_loc "Unkown jsoo attribute ([@@@@%s])." s
+              let err =
+                Location.Error.make
+                  ~loc:exp.pcf_loc
+                  ~sub:[]
+                  (Printf.sprintf "Unkown jsoo attribute ([@@%s])." s)
```
hhugo, line 596: Is the error message different now ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -723,116 +768,116 @@ let literal_object self_id (fields : field_desc list) =
...
+            self#expression { new_expr with pexp_attributes }
+        | _ -> super#expression expr
+      in
+      default_loc := prev_default_loc;
+      new_expr
+  end
+
+let () = Driver.register_transformation "ppx_js" ~impl:transform#structure
+
+let mapper =
```
hhugo, line 877: note that the mapper will not behave correctly if one decide to update other fields. I think we should document it.

ppx/ppx_deriving_json/tests/main.ml
```diff
@@ -1,5 +1,9 @@
...
+    Ppxlib_ast.Selected_ast.of_ocaml Structure st
+    |> Ppxlib.Driver.map_structure
+    |> Ppxlib.Selected_ast.to_ocaml Structure
+  in
+  { Ast_mapper.default_mapper with structure }
 
-let _ = Ppx_deriving_json.json
-
-let () = Toplevel_expect_test.run Migrate_parsetree.Driver.run_as_ast_mapper
+let () = Toplevel_expect_test.run mapper
```
hhugo, line 9: note that the mapper will not behave correctly if one decide to update other fields. I think we should leave a small comment about it.

hhugo (changes requested):

hhugo:
Could you update the `CHANGES.md`

compiler/ppx/ppx_optcomp_light.ml
pitag-ha: Indeed, thanks.

ppx/ppx_deriving_json/lib/ppx_deriving_json.ml
pitag-ha: I've introduced the function `mkloc` now. What are you referring to with `mknoloc`?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: Good catch. I've gotten rid of all `Parsetree.X` in this file.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: I don't need to disable that warning anymore. I needed that when ppxlib was still dependent on Base. I've gotten rid of disabling it. Thanks.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: I've documented that in the .mli-file now. Is that what you were referring to? Or would you prefer a comment in the .ml-file?

ppx/ppx_deriving_json/tests/main.ml
pitag-ha: Yes. I've done that now.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: The error message is not different. But there is a problem with `Ppxlib.Location.raise_errorf`: using `Ppxlib.Location.raise_errorf` in a rewriter that is tested with toplevel_expect_test causes toplevel_expect_test to raise that error instead of returning a diff. That's why I've rewritten this using `Ppxlib.Location.make` instead of `Ppxlib.Location.raise_errorf`. I haven't found out what is happening to `Ppxlib.Location.raise_errorf`, yet.

pitag-ha:
I've added a `CHANGES.md` entry now. 

Thanks for adding the ppxlib < 0.17 support!

I've added every review in a different commit now since I didn't want to force push to my branch anymore after you've also contributed to it. Let me know if want to squash-merge it or if you want me to tidy up the commit structure a little.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -241,7 +250,7 @@ let invoker ?(extra_types = []) uplift downlift body arguments =
   let labels_and_pats =
     List.map arguments ~f:(fun d ->
         let label = Arg.label d in
-        let patt = Pat.var (Location.mknoloc (Arg.name d)) in
+        let patt = Pat.var { txt = Arg.name d; loc = Location.none } in
```
hhugo, line 253: can we introduce mknoloc and mkloc to minimize the diff ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
hhugo: Yes, I would add a comment in `ppx_js.mli` as it's a public interface.

ppx/ppx_deriving_json/lib/ppx_deriving_json.ml
pitag-ha: Ok, I've seen now that you were referring to file `ppx/ppx_js/lib_internal/ppx_js_internal.ml `.

pitag-ha:
I forgot to mention that before: the tests fail both in my branch and in master.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: Yes, I've done that now.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: Good point. I've moved the comment from `ppx_js_internal.mli` to  `ppx_js.mli` now.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
hhugo: The issue is that we're catching `Ppxlib.Location.Error` somewhere in that file, but `raise_errorf` (which comes from compiler-libs) raises `Ocaml_common.Location.Error`

pitag-ha: Thanks! That's good to understand. I'll talk with @NathanReb about the question if we should always raise the `Ppxlib.Location.Error` exception in `Ppxlib.Location`, including in `Ppxlib.Location.raise_errorf`. 

What I've done here now is to simply define `raise_errorf` at the top of the file to keep the diff smaller. Is that good enough?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -25,12 +25,15 @@ open Parsetree
 
 let nolabel = Nolabel
 
+exception Avoid_uncaught_raising of Location.Error.t
```
hhugo, line 28: Should this be called `Syntax_error` or something ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -746,6 +745,8 @@ let transform =
     inherit Ast_traverse.map as super
 
     method! expression expr =
+      input_name := expr.pexp_loc.loc_start.pos_fname;
```
hhugo, line 748: It's unclear to me that this is correct. `expr.pexp_loc` could be equal to Location.none ..

pitag-ha: In which cases is `expr.pexp_loc` equal to `Location.none`?

I've checked if in the build of js_of_ocaml itself it ever comes to the case that `!input_name` isn't equal to `!Location.input_name` from compiler-libs before using it (in `inside_Js`). And it isn't equal in only 1 case: `toplevel/examples/lwt_toplevel/colorize.fake.ml` / `toplevel/examples/lwt_toplevel/colorize.ml` . The latter gets generated by the former and has a directive pointing to the former at the beginning. So in that case `!input_name` is the former (`toplevel/examples/lwt_toplevel/colorize.fake.ml`) and `!Location.input_name` is the latter (`toplevel/examples/lwt_toplevel/colorize.ml`). I thought that wasn't a big issue. 

But `expr.pexp_loc` being `Location.none` might be a bigger issue. What I thought was that that would only be the case when `Location.input_name` is `_none_` itself or `_toplevel_` or similar. Is that not true?

hhugo: A location none could be generated by another ppx running before jsoo

pitag-ha: Good point!

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: Ok, I've done that.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: As talked about, I've defined an exception here in order to avoid `Ppxlib.Location.Error` which will probably be removed from the ppxlib API.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
pitag-ha: I've reverted this change now, since I don't think there's a way to access the source filename via ppxlib in a reliable way at the moment.

toplevel/lib/dune
```diff
@@ -4,5 +4,6 @@
   (synopsis "Js_of_ocaml toplevel library")
   (libraries
     js_of_ocaml-compiler js_of_ocaml bytes
-    compiler-libs.bytecomp compiler-libs.toplevel)
+    compiler-libs.bytecomp compiler-libs.toplevel
+    ppxlib)
```
hhugo, line 8: Why do we have an explicit dependency on ppxlib here ?

pitag-ha: That's not needed. Thanks for noticing! I've removed it.

By the way, I also explicitly increased the compiler version dependency bound to `>= "4.04.0"` (from `>= "4.02.0"`) both in `js_of_ocaml-ppx.opam` and `js_of_ocaml.opam`, which is also not needed. Do you maybe also want me to revert that?

hhugo: If it's not needed, yes please

hhugo:
- I've removed the need for `Ocaml_common.Location.input_name`
- the code is now compatible with ppxlib 0.15 and 0.18

hhugo:
One last issue: the locations check complains about the ppx_js extension.
I'd like to fix this before we merge the ppxlib support

hhugo:
The reason seems to be that metaquot behaves differently.
It use to take the location from !default_loc and now take locations from the lexical scope

hhugo:
I've pushed a fix

toplevel/lib/dune
pitag-ha: What I meant was that 4.04.0 is just an implicit bound due to using ppxlib, but I didn't need to make it explicit. I've gone back to 4.02.0.

hhugo:
thanks for your work

PR #1046, opened 2020-07-28
Compiler: fix code generation for condition

hhugo:
- add a new regtest
- remove erroneous optimization when emitting JavaScript
- mitigate the previous code change by improving if-statement simplification.


hhugo:
cc @xclerc

PR #1048, opened 2020-07-30
Avoid polymorphic comparison over float values

xclerc:
While randomly walking through the code, I noticed something similar
to an [old flambda bug](https://github.com/ocaml/ocaml/pull/1810). Indeed, with master, the following code (copied
from [this pull request](https://github.com/ocaml/ocaml/pull/1794#issuecomment-393820230)) prints `inf` instead of `-inf`:

```
let z =
  let x = -0. and y = +0. in
  if mod_float x 1. >= 0. then
    x
  else if false then x else y

let () =
  Printf.printf "%g\n" (1. /. z)
```

The culprit is the use of the polymorphic equality on float values by a join
operation in the `Flow` module. This pull request ensure that the equality
function over `Code.constant` value does not use the polymorphic equality.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
             when parsing constant in the from the bytecode
 * Compiler: make sure inline doesn't loop indefinitly (#1043)
 * Compiler: fix bug generating invalid javascript for if-then construct (#1046)
+* Compiler: do not use polymorphic comparison when joining float values (#1048)
 * Lib: Rename msg to message in Worker (#1037)
 * Lib: fix graphics_js when build with separate compilation (#1029)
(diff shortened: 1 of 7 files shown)


hhugo:
Can you add a changelog entry ?

xclerc:
I have added a changelog entry and pushed additional commits
to support OCaml 4.0{2,3,4}.

hhugo:
I've reformated the code with `make fmt` (depends on ocamlformat)
and moved foat_equal, int64_equal, array_equal in stdlib.ml

hhugo:
Thanks for the bugfix

PR #1049, opened 2020-07-31
Runtime: add missing primitives

### LINKED ISSUE #1047: [BUG] Missing `caml_sys_time_include_children` on toplevel creation.
**Describe the bug**
There's a missing primitive when you compile toplevels: 
```
There are some missing primitives
Dummy implementations (raising 'Failure' exception) will be used if they are not available at runtime.
You can prevent the generation of dummy implementations with the commandline option '--disable genprim'
Missing primitives:
  caml_sys_time_include_children
```
It's because of [this bit](https://github.com/ocaml/ocaml/blob/trunk/utils/profile.ml#L20) in the compiler. Implementing in terms of `jsoo`'s [`caml_sys_time`](https://github.com/ocsigen/js_of_ocaml/blob/db758adbd02dd3e0e85cfddff5c772153c416929/runtime/sys.js#L129) wouldn't harm.

**Expected behavior**
My clear and concise description of what I expect to happen is to have no missing primitive when I compile toplevels :-)

**Versions**
```
> ocamlc -version     
4.09.0
> js_of_ocaml --version                                      
3.6.0
```

hhugo:
fix #1047

### DIFF
--- a/runtime/sys.js
+++ b/runtime/sys.js
@@ -59,6 +59,12 @@ function caml_sys_getenv (name) {
   caml_raise_not_found ();
 }
 
+//Provides: caml_sys_unsafe_getenv
+//Requires: caml_sys_getenv
+function caml_sys_unsafe_getenv(name){
+  return caml_sys_getenv (name);
+}
+
 //Provides: caml_argv
 //Requires: caml_string_of_jsstring
 var caml_argv = ((function () {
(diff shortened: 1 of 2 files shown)


runtime/sys.js
```diff
@@ -131,6 +137,12 @@ function caml_sys_time () {
   return now * 0.001 - caml_initial_time;
 }
 
+//Provides: caml_sys_time_include_children
+//Requires: caml_sys_time
+function caml_sys_time_include_children() {
```
dbuenzli, line 142: The regular primitive takes a boolean, but maybe that's okay with Js.

PR #1063, opened 2020-09-23
adds the intersection observer API

TyOverby:
The Intersection Observer API provides a way to asynchronously observe
    changes in the intersection of a target element with an ancestor element or
    with a top-level document's viewport.
    https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

### DIFF
--- a/lib/js_of_ocaml/intersectionObserver.ml
+++ b/lib/js_of_ocaml/intersectionObserver.ml
@@ -43,19 +43,17 @@ class type intersectionObserver =
 
 let empty_intersection_observer_options () : intersectionObserverOptions Js.t =
   Js.Unsafe.obj [||]
-;;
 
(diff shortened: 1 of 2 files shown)


TyOverby:
@hhugo it looks like one of the failures is because of a flaky http request that failed for unkown reasons.

The other appears to be an ocamlformat failure, but I don't see an ocamlformat file in this repo.  How should I fix this?

hhugo:
https://github.com/ocsigen/js_of_ocaml/blob/master/.ocamlformat

You need to have ocamlformat installed with the right version and do `make fmt`

TyOverby:
@hhugo ah great; that wasn't hard.  For some reason I thought that you needed a ".ocamlformat" file at the root.

hhugo:
There is an .ocamlformat at the root

Issue #1075, opened 2020-12-13
[FEATURE REQUEST] JS-agnostic intermediate representation


Lupus:
js_of_ocaml is a really cool compiler! I'm fascinated with that it can do. Yet it targets only one language - JavaScript, while it's certainly feasible to target more languages with bytecode approach, [rehp](https://github.com/jordwalke/rehp) has Php backend working along with JS, and there is [Go backend](https://github.com/Lupus/ocaml2go) in a separate project.

The bytecode parsing machinery up to `generate.ml` is in general agnostic to target language. What Rehp does is adding intermediate representation layer for program tree, that is also agnostic to target language. Having that alone in js_of_ocaml can enable one to use it as a library to produce the IR, and take it from there to target whatever language desired.

Bytecode based approach is considered the best, and js_of_ocaml can become a solid foundation for various compilers going this route.

/cc @jordwalke

hhugo:
Contributions are welcome

hhugo:
A lot has moved since this issue was first opened. In particular, 
- New wasm backend
- Control flow compilation for javascript rewritten to use Labelled statement. (I don't think this is available in php)
- Support for Effect handlers

I no longer think that a tree like IR, such as the one implemented in #1076, is what we want. The current IR seems good enough and is JS-agnostic. Using jsoo as a lib to target other language is possible and it's how wasm_of_ocaml is implemented. Some small internal changes might be necessary though.

In practice, adding a new backend is a matter of writing a new generate.ml pass.

All that said, I'll close this issue

PR #1076, opened 2020-12-13
Generalized program tree

Lupus:
This PR tries to address #1075. Most of the work presented in this PR is due to @jordwalke (see [rehp](https://github.com/jordwalke/rehp), I only brushed it up it and prepared for upstreaming.

Test outputs are updated with the following changes
* `==` is changed to `===` in arity tests (in static languages it's expensive to implement JS comparison semantics)
* `! b` is changed to `1 - b`, the latter is used in many other places, my attempt to unify everything to `! b` failed while with `1 - b` it compiles fine, and is semantically correct:

```
> ! 0
true
> ! 1
false
> ! "aww"
false
> ! ""
true

> ! ! (1 - 0)
true
> ! ! (1 - 1)
false
> ! ! (1 - "aww")
false
> ! ! (1 - "")
true

```


hhugo:
Thanks. Would you mind restricting this PR To just introducing an additional IR. I don't want to review too many moving pieces at once.

Lupus:
Yeah, sure. Will do!

вс, 13 дек. 2020 г., 21:57 hhugo <notifications@github.com>:

Lupus:
I've trimmed this PR down to only introducing an additional IR.

compiler/lib/ir.mli
```diff
@@ -0,0 +1,173 @@
...
+and arguments = expression list
+
+and property_name_and_value_list = (Id.property_name * expression) list
+
+and raw_segment =
+  | RawText of string
+  | RawSubstitution of expression
+
+and expression =
+  | ERaw of raw_segment list
```
hhugo, line 101: Can we leave `ERaw` to a later PR

compiler/lib/id.ml
```diff
@@ -0,0 +1,39 @@
...
+  ; var : Code.Var.t option
+  ; loc : Loc.t
+  }
+
+type t =
+  | S of ident_string
+  | V of Code.Var.t
+
+let ident =
+  (fun ?(loc = N) ?var name -> (S { name; var; loc } [@explicit_arity])
```
hhugo, line 33: `[@explicit_arity]` has nothing to do here

compiler/lib/id.ml
```diff
@@ -0,0 +1,39 @@
...
+    : ?loc:Loc.t -> ?var:Code.Var.t -> identifier -> t)
+
+type property_name =
```
hhugo, line 36: `property_name` is very much specific to javascript, I don't think we need to use it in the new IR

compiler/lib/ir.ml
```diff
@@ -0,0 +1,163 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+open Stdlib
+
+type array_literal = element_list
+
+and element_list = expression option list
+
+and binop =
```
hhugo, line 26: Can you put all the float related constructor next to each other ?

hhugo:
I meant, restrict the PR to a new IR and plug it in - but exclude things that change the behavior

compiler/lib/id.ml
Lupus: Removed, sorry, leftovers after converting from Reason.

compiler/lib/id.ml
Lupus: Indeed, removed.

compiler/lib/ir.ml
Lupus: Done.

compiler/lib/ir.mli
Lupus: I've simplified `ERaw` as much as possible, but I'm not sure how to avoid it when plugging the new IR. `caml_js_expr` is currently parsing the JS code into the tree and embeds that directly. Converting JS tree to IR does not sound like a great idea.

Lupus:
> I meant, restrict the PR to a new IR and plug it in - but exclude things that change the behavior

Gotcha, I've restored the PR to original state and tweaked it a bit to produce exactly the same output. Tweaks included using `EqEq` in arity test and still generating `1 - x` in `generate.ml`.

compiler/lib/ir.ml
```diff
@@ -0,0 +1,158 @@
...
+  | Block of block
+  | Variable_statement of variable_declaration list
+  | Empty_statement
+  | Expression_statement of expression
+  | If_statement of expression * (statement * Loc.t) * (statement * Loc.t) option
+  | Loop_statement of statement * Loc.t
+  | Continue_statement of Javascript.Label.t option
+  | Break_statement of Javascript.Label.t option
+  | Return_statement of expression option
+  | Labelled_statement of Javascript.Label.t * (statement * Loc.t)
```
TyOverby, line 126: Should the IR contain primitives like this that aren't present in OCaml?

Lupus: These control flow structures are well recognized and exist in some form in all high-level languages. What would you suggest for the IR? I didn't want to make non-cosmetic changes to `generate.ml`, at least in this PR. These primities don't exist in OCaml, but control flow deconstruction produces them right now and they are not too much specific for JavaScript (in Golang backend, these got transformed to Go AST seamlessly).

Lupus:
Hey @hhugo, have you had a chance to look more at this?

mnxn:
I'm writing this as an outsider but is this new IR type sufficiently general?

I wouldn't expect constructs like `Typeof`, `ERegexp`, and `ERaw` to be present in a general IR. Similarly, I don't think JS-specific primitives/externs should be handled in `generate.ml`.

Is it feasible to handle the JS features in `javascript_from_ir.ml` instead?

Lupus:
`ERaw` is essential for interop with target language. Special primitive takes source code in target language as a string, possibly interpolates it with arguments and injects into IR, which is then processed by language backend, and target language code gets emitted during generation.

As of `Typeof` and friends, `generate.ml` has a number of primitives declared like ones below:
```
  register_bin_prim "caml_js_instanceof" `Pure (fun cx cy _ ->
      bool (J.EBin (J.InstanceOf, cx, cy)));
  register_un_prim "caml_js_typeof" `Pure (fun cx _ -> J.EUn (J.Typeof, cx))
```

I'm not completely sure if these could be somehow moved away from `generate.ml`, may be we could use `ERaw` and make Javascript backend register these primitives as macros instead.

I'm all for making the IR as general as possible, just wanted to take baby steps in the right direction :)

Lupus:
Or what you suggest @mnxn is to emit some `EPrim` in `generate.ml` and offload handling of primitives other than a set of well-known ones to target language IR processor? And consequently move all JS primitive related code from `generate.ml` into `javascript_from_ir.ml`?

mnxn:
> Or what you suggest @mnxn is to emit some `EPrim` in `generate.ml` and offload handling of primitives other than a set of well-known ones to target language IR processor? And consequently move all JS primitive related code from `generate.ml` into `javascript_from_ir.ml`?

Exactly. 
I think that the primitives that start with `caml_js` should be handled by the `javascript_from_ir` module. Then the IR can be simplified and remove many JS variants.

Lupus:
> Exactly.
> I think that the primitives that start with `caml_js` should be handled by the `javascript_from_ir` module. Then the IR can be simplified and remove many JS variants.

Sounds good to me! Question to maintainers if this belongs to this PR, or better organized as subsequent one.

Lupus:
I'm looking into moving JS specific primitives outside of `generate.ml` as this makes sense in terms of generalization of IR.

Some are quite obvious like `Prim (Extern "debugger", _)`. Some are not, like `%closure`.

Ideally primitives should be handled at two levels - first one would be `generate.ml` handling some built-in/special primitives so as not to repeat this code in each backend generator. And second one would be in specific backend, handling primitives that make sense only for one target language.

Probably passing a primitive from one level to another can be plugged in somewhere here: https://github.com/ocsigen/js_of_ocaml/blob/65c1fc1966cbedc4587da764d7871cfb965f972e/compiler/lib/generate.ml#L1194. Instead of generating generic `ECall` for primitives that we do not know - we can emit some `EPrim` IR node which can then be processed later by the target language backend.

I have not yet fully understood what that `Share` thing is doing, but it does track some primitives, in tandem with `%closure` that gets introduced in inliner (`inline.ml`). Splitting primitives in two levels would make target-language-specific primitives invisible for logic in `generate.ml`, they won't be tracked in `Share`. I'm not sure if that could hurt kittens. @hhugo could you please advise on this part? And do you think it makes sense to follow this path in general?

hhugo:
Is there still interest in such change ? Should it be rebased or closed ?

Lupus:
Not from my end unfortunately. Let's close this probably.

PR #1083, opened 2021-01-26
Fix segfault when flat-float-array mode disabled

nojb:
Fixes a segfault when `flat-float-array` mode is disabled.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@
 * Lib: expose API to attached and retrieve js errors to/from ocaml exceptions
 * Lib: intersection observer API fixes
 
+## Bug fixes
+* Compiler: fix a segmentation fault when `flat-float-array` mode is disabled.
+
 # 3.8.0 (2020-10-21) - London
 ## Features/Changes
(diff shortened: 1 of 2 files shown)


hhugo:
Can you explain a bit what's going on ?

nojb:
> Can you explain a bit what's going on ?

Yes: when `flat-float-array` mode is disabled, the runtime representation of `float array` is *boxed* (with tag 0), while a block of tag 254 is "flat" (unboxed), so the magic `Obj.magic x : float array` is casting a "flat" block of floats into a boxed block of floats. What the patch does is to an explicit cast between the two by copying the contents of the "flat" block of floats into a `float array`, taking care so that the copy works regardless of whether `flat-float-array` is enabled or disabled. Does it make sense?

nojb:
(To explain this a bit further, note that the problem arises when `js_of_ocaml` itself is compiled with `flat-float-array` mode disabled, independently of how the bytecode was compiled.)

hhugo:
Thanks for the fix

PR #1089, opened 2021-02-21
Switch to GitHub Actions

smorimoto:
TravisCI currently needs to be migrate work, and there is no reason to keep using it, so it better to move to GitHub Actions. It has more parallelism and better support for each platform.


hhugo:
Thanks for your contribution.
One should also install ocamlformat and run `make fmt` to make sure the code is formatted correctly.

hhugo:
Can will restrict the matrix to only test macos and windows on the latest version of the compiler ?

smorimoto:
I've already done those things, but since the local package versions are not consistent, opam not able to handle them well, and there are extra hacks that we don't actually need.

hhugo:
> but since the local package versions are not consistent

What does that mean ?

smorimoto:
It was completely my mistake. Disregard it!
The build works as expected on Ubuntu, but it fails on other platforms. Can you check that?

hhugo:
Why do you drop testing ocaml < 4.08 ?

smorimoto:
I just wanted to make sure it worked in several versions. I will add 4.02~4.07.

hhugo:
The current ci test to that packages are installable on old version of ocaml (not bothering about tests) and runs test ( installing additional deps ) on the latest version. Can we have the same behavior ?

hhugo:
The matrix is red for ocaml < 4.08 because of tests or tests dependencies.

hhugo:
Should we move lint-fmt to its own workflow ?

hhugo:
I've fixed the macos build and made progress with the windows one. The new failture is due to gmp

smorimoto:
Thanks for fixing it!
The reason we can't build is because each local package requires a different version of OCaml, but don't some of these really support 4.02?
By the way, lint-fmt is actually running as a separate job that has nothing to do with main build jobs.

hhugo:
> The reason we can't build is because each local package requires a different version

the failure I see are only caused by test dependencies only  (because of `--with-test`)

hhugo:
The Js_of_ocaml compiler should still work with 4.02.
I see that there are different minimal versions for different packages. We were using `opam install --best-effort` to make sure that if constraints allow it, a package will build successfully.

hhugo:
If it not easy to fix the windows part I'm happy to drop it off the PR for now, as it's not a regression to what we have now

smorimoto:
I changed what we do on the CI to exactly the same as the previous Travis CI setup. But there seems to be have another problem with 4.12 now.

hhugo:
Some tests are compiler version dependent. I think we should just promote the expect test to the 4.12 version.

smorimoto:
I made it to ignore CI test failures on Windows. However, since the tests themselves are still active, we can gradually fix them.

hhugo:
@smorimoto, next time you merge a PR, would you mind squashing commits to keep the history cleaner ?

smorimoto:
@hhugo Apologies, I should have confirmed that to you!

hhugo:
np

PR #1099, opened 2021-04-10
add innerText to `Dom_html`

anuragsoni:
This seems to be supported in all common browsers now [1], and also seems part of the current html living spec [2].

[1] https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText
[2] https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 # ??? (??) - ??
++## Features/Changes
++* Lib: add innerText property for Dom_html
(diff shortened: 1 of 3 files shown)


smorimoto (approved):

TyOverby (approved):

TyOverby:
Thanks!

PR #1103, opened 2021-05-30
Avoid spurious `ocamlfind` warning when not generating Graphics lib

CraigFe:
The `dune` file for the `js_of_ocaml` Graphics library is conditional on the availability of the `graphics` library, as determined by querying `ocamlfind`. This commit passes `-qe` to the OCamlfind query to avoid a redundant warning when `graphics` is not available:

```
$ dune build
       ocaml (internal)
ocamlfind: Package `graphics' not found
```

(This warning is visible to all packages that vendor `js_of_ocaml`.)

### DIFF
--- a/lib/lwt/graphics/dune
+++ b/lib/lwt/graphics/dune
@@ -3,7 +3,7 @@ open Jbuild_plugin.V1
 
 let graphics_exists =
   try
-    match run_and_read_lines "ocamlfind query -a-format -predicates byte graphics" with
+    match run_and_read_lines "ocamlfind query -a-format -predicates byte graphics -qe" with
     | path :: _ -> Sys.file_exists path
     | _ -> false
   with _ -> false


CraigFe:
At a higher level, it seems odd to require `ocamlfind` here at all, since `ocamlfind` is only a dep-opt of `js_of_ocaml-compiler`. There's a different error for not having an available `ocamlfind`:

```
$ dune build
       ocaml (internal)
sh: line 1: ocamlfind: command not found
```

What's the desired behaviour here when `ocamlfind` is not available?

hhugo:
Ideally, we should not call ocamlfind our-self. It was added some time ago to work around an issue with optional libraries https://github.com/ocsigen/js_of_ocaml/commit/6e801507259d54b45cb1310bfcf207dd07885010.

Do you want to try remove it and see if the CI is happy ?

hhugo:
alternative fix in #1118

hhugo:
I've removed the explicit call to ocamfind

Issue #1104, opened 2021-05-31
[BUG] Stack overflow when linking against many JS files


CraigFe:
**Describe the bug**
When running `js_of_ocaml link` on a large number of source files, it's possible to trigger a stack overflow exception:

```
$ js_of_ocaml link --source-map-inline -o test.js <350 files>
/root/bin/js_of_ocaml: Error: Stack overflow
Raised by primitive operation at file "stdlib.ml", line 296, characters 22-31
Called from file "stdlib.ml", line 296, characters 22-31
Called from file "stdlib.ml", line 296, characters 22-31
...
Called from file "stdlib.ml", line 296, characters 22-31
Called from file "compiler/lib/source_map.ml", line 221, characters 18-172
Called from file "compiler/lib/source_map.ml", line 242, characters 8-143
Called from file "compiler/lib/link_js.ml", line 82, characters 12-63
Called from file "compiler/lib/stdlib.ml", line 643, characters 11-15
Re-raised at file "compiler/lib/stdlib.ml", line 646, characters 9-16
Re-raised at file "compiler/lib/stdlib.ml", line 652, characters 6-15
Called from file "src/cmdliner_term.ml", line 25, characters 19-24
Called from file "src/cmdliner.ml", line 117, characters 32-39
Called from file "src/cmdliner.ml", line 147, characters 18-36
Called from file "src/cmdliner.ml", line 265, characters 22-48
Called from file "compiler/bin-js_of_ocaml/js_of_ocaml.ml", line 48, characters 4-225
```

This happened when building a certain JS file inside Tezos which has a rather large dependency tree (`src/proto_010_PtGRANAD/lib_benchmark/lib_benchmark_type_inference/test/test_inference.bc.js`).

**Expected behavior**
The link process should succeed w/o raising an exception.

**Versions**
True of both 3.9.1 and today's `master` (`34ae55d`).

hhugo:
fixed by #1112

PR #1105, opened 2021-05-31
Compiler: fix stack overflow when linking against many files

CraigFe:
Fix https://github.com/ocsigen/js_of_ocaml/issues/1104.

This is the simplest possible fix for #1104. There may be other uses of `( @ )` worth replacing in the code-base, but I'll leave this for someone more experienced with Jsoo to determine.


hhugo:
I would change compiler/lib/stdlib.ml to make `@` tail recursive. Same approach is already used with List.map

hhugo:
List.append (and @) is now tail rec on master

Issue #1106, opened 2021-06-26
4.13 support


copy:
Currently, on 4.13.0~alpha1, any invocation of `js_of_ocaml` fails as follows:

```
% ocamlc test.ml
% js_of_ocaml a.out
js_of_ocaml: Error: Bytecode version mismatch.
js_of_ocaml: Error: Your ocaml bytecode and the js_of_ocaml compiler have to be compiled with the same version of ocaml.
js_of_ocaml: Error: The Js_of_ocaml compiler has been compiled with ocaml version 4.13.0~alpha1.
```

hhugo:
support has been merged. There are still some missing primitives though

Issue #897, opened 2019-10-30
Runtime: implement new runtime for floats operation .. 


hhugo:
if/when https://github.com/ocaml/ocaml/pull/944 is merged

hhugo:
# natively supported 
`cbrt`
`log2`
`acosh`
`asinh`
`atanh`

# other
`exp2(x)` => `pow(2,x)`

`erf` => https://hewgill.com/picomath/javascript/erf.js.html
```
function erf(x) {
    // constants
    var a1 =  0.254829592;
    var a2 = -0.284496736;
    var a3 =  1.421413741;
    var a4 = -1.453152027;
    var a5 =  1.061405429;
    var p  =  0.3275911;

    // Save the sign of x
    var sign = 1;
    if (x < 0) {
        sign = -1;
    }
    x = Math.abs(x);

    // A&S formula 7.1.26
    var t = 1.0/(1.0 + p*x);
    var y = 1.0 - (((((a5*t + a4)*t) + a3)*t + a2)*t + a1)*t*Math.exp(-x*x);

    return sign*y;
}
```
`erfc` => `erfc ⁡ ( x ) = 1 − erf ⁡ ( x )`

hhugo:
needed for #1106

hhugo:
Here is the list of primitives to implement
```
caml_acosh_float
caml_asinh_float
caml_atanh_float
caml_cbrt_float
caml_erf_float
caml_erfc_float
caml_exp2_float
caml_fma_float
caml_log2_float
```

pmwhite:
I'm going to work on this.

hhugo:
@pmwhite, no rush but what's your timeframe for working on this ?

pmwhite:
@hhugo I expect to have a PR open by end-of-week, if not today.

PR #1107, opened 2021-07-02
Some changes for 4.13

hhugo:
- [x] handle the new magic number
- [x] fixing runtime for recursive module is no longer needed
- [x] deal with deprecation warning
- [ ] add missing primitives for float #897
- [ ] add missing primitives from sys (rmdir/mkdir) #1109, not blocking as theses were introduced in 4.12
- [x] implement other missing primitives (some small tooling would be welcome #1110 )
```
caml_eventlog_pause
caml_eventlog_resume
caml_gc_huge_fallback_count
caml_gc_major_slice
caml_gc_minor_words
caml_get_major_bucket
caml_get_major_credit
caml_get_minor_free
caml_obj_add_offset
caml_obj_reachable_words
```


hhugo:
fix #1106

kit-ty-kate:
It might be worth adding `< "4.14"` to the constraint on `ocaml` for `js_of_ocaml-compiler` to avoid having to add a constraint later for next release, since it does require a compiler specific change everytime (at the very least to add the `V4_N` case)

hhugo:
> It might be worth adding `< "4.14"` to the constraint on `ocaml` for `js_of_ocaml-compiler` to avoid having to add a constraint later for next release, since it does require a compiler specific change everytime (at the very least to add the `V4_N` case)

done

PR #1113, opened 2021-07-07
Implement float primitives for 4.13

### LINKED ISSUE #897: Runtime: implement new runtime for floats operation ..

pmwhite:
fix #897 

This isn't finished by any means. What's missing is

1. An implementation for `fma`
2. Tests

My plan for 1 is to port the native ocaml runtime's C implementation to javascript. However, I'm not sure how to do 2, since the stdlib for ocaml 4.12 does not provide functions like `Float.atanh` or `Float.acosh`, so I can't even access the primitives in a straightforward way.

If I were to write tests and run them locally using an opam switch with the 4.13 compiler, it would still break the current CI. This seems like blocker for merging this PR, since it presumably will be impossible to test until 4.13 is released.

Am I understanding things correctly?


hhugo:
You can declare externals in your test directly. Just copy paste signature from ocaml master

hhugo:
I'd probably put new functions in the ieee_754.js file  next to other related functions

pmwhite:
Moved the functions to `ieee_754.js`. Looks like some of the existing functions in that file re-implement functions that are builtin. If this is to act as a polyfill, maybe we should just use the polyfill as a fallback. For example:
```javascript
var caml_sinh_float = Math.sinh || function (x) { return (Math.exp(x) - Math.exp(-x)) / 2; }
```

pmwhite:
Looks like `compiler/test-ocaml/lib-float/test.ml` is taken directly from the ocaml compiler. Oddly enough, it seems to be lacking tests for several of the builtin functions. It would be nice if the ocaml's stdlib tests were available in opam so that we could just depend on that instead of manually keeping these files up-to-date.

Anyway, I take it that tests should go in `compiler/tests-jsoo/test_floats.ml`

runtime/ieee_754.js
```diff
@@ -284,20 +293,82 @@ function caml_hypot_float (x, y) {
...
+
+//Provides: caml_round_float const
 function caml_round_float (x) { return Math.round(x); }
 
+//Provides: caml_cbrt_float const
+var caml_cbrt_float = Math.cbrt || (function (pow) { 
+  return function cbrt(x) {
+    return x < 0 ? -pow(x, 1/3) : pow(x, 1/3);
+  };
+})(Math.pow);
```
TyOverby, line 339: Why pass pow in like this?

pmwhite: All these polyfill implementations are taken from the MDN docs. This one, in particular was taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/cbrt

pmwhite:
Add some tests. The situation is not ideal because the ocaml stdlib doesn't have these functions defined yet, so these tests can't be run in native mode (I disabled native mode for all the tests in this directory, but if there's no workaround, I should probably put the float tests in their own library so we can disable native code for _just_ the float tests.

hhugo:
> Moved the functions to `ieee_754.js`. Looks like some of the existing functions in that file re-implement functions that are builtin. If this is to act as a polyfill, maybe we should just use the polyfill as a fallback. For example:
> 
> ```js
> var caml_sinh_float = Math.sinh || function (x) { return (Math.exp(x) - Math.exp(-x)) / 2; }
> ```

We don't need polyfill for things that are well supported. It's possible that polyfill were added in the past if support was not great at the time.

hhugo:
> Added some tests. The situation is not ideal because the ocaml stdlib doesn't have these functions defined yet, so these tests can't be run in native mode (I disabled native mode for all the tests in this directory, but if there's no workaround, I should probably put the float tests in their own library so we can disable native code for _just_ the float tests.

Maybe move the tests in theirs on lib for now.

pmwhite:
@hhugo I'm finished implementing and testing these functions, so this is ready to be reviewed. The fma implementation I took from this [gist](https://gist.github.com/Yaffle/fb47de4c18b63147699e0b621f1031f7), and the fma tests I took from the ocaml compiler. The aforementioned gist passed all the tests without modification.

hhugo:
I've tested your PR with 4.13 and made a small fix regarding nan  printing in the tests. I've also rebased your work on top of the latest master.

hhugo:
Would you mind adding a changelog entry ?

pmwhite:
Ahh, sorry for being slow to correct the formatting and changelog. Thanks!

hhugo:
Thanks for working on this

Issue #1109, opened 2021-07-07
Runtime: implement caml_sys_rmdir/mkdir


hhugo:
Theses primitives were introduced in 4.12.

PR #923, opened 2019-11-17
Compiler: compile ocaml string as javascript ones

hhugo:
This feature requires that all deps use the right primitives to manipulate bytes and string.
- https://github.com/OCamlPro/ocplib-endian/issues/21
- https://github.com/ocaml/num/pull/20
- https://github.com/mirage/ocaml-base64/pull/37
- ...

Here is a PoC for typing some of the primitives known to the OCaml compiler 
https://github.com/hhugo/ocaml/commits/prim-typ

Issue #483, opened 2016-06-02
Better runtime support for nodejs


hhugo:
io, unix, ..

hhugo:
Closing this META issue as there is no clear goal. nodejs support has improved a lot anyway.

PR #1115, opened 2021-07-21
Mapping ocaml unix APIs to nodejs calls

phated:
This is just a draft right now because I need a lot of advice on TODOs and testing (sorry, first time contributing here).

@hhugo saw my work on these utilities in the Grain repository and said I should look into upstreaming them. So here is my first pass.

I tried changing some of our assumptions to allow for non-nodejs usage, but I don't know what the functions should do if not inside nodejs.

Please help guide me into taking this over the finish line! 🍻

### DIFF
--- a/compiler/tests-compiler/sys_fs.ml
+++ b/compiler/tests-compiler/sys_fs.ml
@@ -55,18 +55,18 @@ let%expect_test _ =
   compile_and_run
     {|
   (match Sys.mkdir "/not/exists" 0o777 with
-  | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
+  | exception Sys_error path -> print_endline "EXPECTED ERROR"
   | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "BUG");
   |};
-  [%expect {|Sys_error: Error: ENOENT: no such file or directory, mkdir '/not/exists'|}]
+  [%expect {|EXPECTED ERROR|}]
 
 let%expect_test _ =
   compile_and_run
     {|
   (match Sys.rmdir "/not/exists" with
-  | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
+  | exception Sys_error path -> print_endline "EXPECTED ERROR"
   | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "BUG");
   |};
-  [%expect {|Sys_error: /not/exists: No such file or directory|}]
+  [%expect {|EXPECTED ERROR|}]
--- a/compiler/tests-compiler/sys_fs.ml
+++ b/compiler/tests-compiler/sys_fs.ml
@@ -59,7 +59,7 @@ let%expect_test _ =
   | exception _ -> print_endline "INCORRECT ERROR"
   | _ -> print_endline "BUG");
   |};
-  [%expect {|Sys_error: /not/exists|}]
+  [%expect {|Sys_error: Error: ENOENT: no such file or directory, mkdir '/not/exists'|}]
 
 let%expect_test _ =
   compile_and_run
--- a/compiler/tests-compiler/sys_fs.ml
+++ b/compiler/tests-compiler/sys_fs.ml
@@ -56,7 +56,7 @@ let%expect_test _ =
     {|
   (match Sys.mkdir "/not/exists" 0o777 with
   | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
-  | exception _ -> print_endline "INCORRECT ERROR"
+  | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "BUG");
   |};
   [%expect {|Sys_error: Error: ENOENT: no such file or directory, mkdir '/not/exists'|}]
@@ -66,7 +66,7 @@ let%expect_test _ =
     {|
   (match Sys.rmdir "/not/exists" with
   | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
-  | exception _ -> print_endline "INCORRECT ERROR"
+  | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "BUG");
   |};
   [%expect {|Sys_error: /not/exists: No such file or directory|}]
--- a/compiler/tests-compiler/sys_fs.ml
+++ b/compiler/tests-compiler/sys_fs.ml
@@ -50,3 +50,23 @@ let%expect_test _ =
   |};
   [%expect {|
     bbb|}]
+
+let%expect_test _ =
+  compile_and_run
+    {|
+  (match Sys.mkdir "/not/exists" 0o777 with
+  | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
+  | exception _ -> print_endline "INCORRECT ERROR"
+  | _ -> print_endline "BUG");
+  |};
+  [%expect {|Sys_error: /not/exists|}]
+
+let%expect_test _ =
+  compile_and_run
+    {|
+  (match Sys.rmdir "/not/exists" with
+  | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
+  | exception _ -> print_endline "INCORRECT ERROR"
+  | _ -> print_endline "BUG");
+  |};
+  [%expect {|Sys_error: /not/exists: No such file or directory|}]
--- a/compiler/tests-compiler/unix_fs.ml
+++ b/compiler/tests-compiler/unix_fs.ml
@@ -22,47 +22,41 @@ open Util
 let%expect_test "Unix.mkdir_Unix.rmdir" =
   compile_and_run
     {|
+let f () =
   Unix.mkdir "aaa" 0o777;
   Unix.mkdir "aaa/bbb" 0o777;
   let l = Sys.readdir "aaa" |> Array.to_list in
   List.iter print_endline l;
   (match Unix.rmdir "aaa" with
-  | exception _ -> ()
+  | exception (Unix.Unix_error (Unix.ENOTEMPTY, _, _)) -> ()
+  | exception e -> print_endline (Printexc.to_string e)
   | _ -> print_endline "UNEXPECTED SUCCESS");
   Unix.rmdir "aaa/bbb";
   Unix.rmdir "aaa"
-  |};
-  [%expect {|bbb|}]
-
-let%expect_test "Unix.mkdir_Unix.rmdir_static" =
-  compile_and_run
-    {|
-  Sys.chdir "/static/";
-  Unix.mkdir "aaa" 0o777;
-  Unix.mkdir "aaa/bbb" 0o777;
-  let l = Sys.readdir "aaa" |> Array.to_list in
-  List.iter print_endline l;
-  (match Unix.rmdir "aaa" with
-  | exception _ -> ()
-  | _ -> print_endline "UNEXPECTED SUCCESS");
-  Unix.rmdir "aaa/bbb";
-  Unix.rmdir "aaa";
-  |};
-  [%expect {|bbb|}]
+in
+f ();Sys.chdir "/static"; f ()|};
+  [%expect {|
+    bbb
+    bbb|}]
 
 let%expect_test "Unix.mkdir_ENOENT" =
   compile_and_run
     {|
+let f () =
   (match Unix.mkdir "aaa/bbb/ccc" 0o777 with
   | exception Unix.Unix_error (Unix.ENOENT, syscall, path) -> print_endline ("ENOENT: " ^ syscall)
   | exception err -> print_endline (Printexc.to_string err)
-  | _ -> print_endline "UNEXPECTED SUCCESS");
-  |};
-  [%expect {|ENOENT: mkdir|}]
+  | _ -> print_endline "UNEXPECTED SUCCESS")
+in
+f (); Sys.chdir "/static"; f ()|};
+  [%expect {|
+    ENOENT: mkdir
+    ENOENT: mkdir|}]
 
 let%expect_test "Unix.mkdir_ENOTDIR" =
   compile_and_run
     {|
+let f () =
   let oc = open_out "aaa" in
   output_string oc "bbb";
   close_out oc;
@@ -73,23 +67,31 @@ let%expect_test "Unix.mkdir_ENOTDIR" =
   | exception Unix.Unix_error (Unix.ENOENT, syscall, path) -> print_endline ("EXPECTED ERROR: " ^ syscall)
   | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "UNEXPECTED SUCCESS");
-  Sys.remove "aaa";
-  |};
-  [%expect {|EXPECTED ERROR: mkdir|}]
+  Sys.remove "aaa"
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    EXPECTED ERROR: mkdir
+    EXPECTED ERROR: mkdir|}]
 
 let%expect_test "Unix.rmdir_ENOENT" =
   compile_and_run
     {|
+let f () =
   (match Unix.rmdir "aaa/bbb/ccc" with
   | exception Unix.Unix_error (Unix.ENOENT, syscall, path) -> print_endline ("ENOENT: " ^ syscall)
   | exception err -> print_endline (Printexc.to_string err)
-  | _ -> print_endline "UNEXPECTED SUCCESS");
-  |};
-  [%expect {|ENOENT: rmdir|}]
+  | _ -> print_endline "UNEXPECTED SUCCESS")
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    ENOENT: rmdir
+    ENOENT: rmdir|}]
 
 let%expect_test "Unix.rmdir_ENOTDIR" =
   compile_and_run
     {|
+let f () =
   Unix.mkdir "aaa" 0o777;
   let oc = open_out "aaa/bbb" in
   output_string oc "ccc";
@@ -102,9 +104,12 @@ let%expect_test "Unix.rmdir_ENOTDIR" =
   | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "UNEXPECTED SUCCESS");
   Sys.remove "aaa/bbb";
-  Unix.rmdir "aaa";
-  |};
-  [%expect {|EXPECTED ERROR: rmdir|}]
+  Unix.rmdir "aaa"
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    EXPECTED ERROR: rmdir
+    EXPECTED ERROR: rmdir|}]
 
 let%expect_test "Unix.stat_file" =
   compile_and_run
--- a/compiler/tests-compiler/unix_fs.ml
+++ b/compiler/tests-compiler/unix_fs.ml
@@ -114,6 +114,7 @@ f (); Sys.chdir "/static"; f () |};
 let%expect_test "Unix.stat_file" =
   compile_and_run
     {|
+let f () =
   let oc = open_out "aaa" in
   output_string oc "bbb";
   close_out oc;
@@ -122,48 +123,66 @@ let%expect_test "Unix.stat_file" =
   | { st_kind = Unix.S_REG; st_size } ->
       print_string "File size: ";
       print_int st_size;
+      print_newline ();
   | _ -> print_endline "UNEXPECTED SUCCESS");
   Sys.remove "aaa"
-  |};
-  [%expect {|File size: 3|}]
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    File size: 3
+    Failure("unix_stat: not implemented")|}]
 
 let%expect_test "Unix.stat_dir" =
   compile_and_run
     {|
+let f () =
   Unix.mkdir "aaa" 0o777;
   (match Unix.stat "aaa" with
   | exception err -> print_endline (Printexc.to_string err)
-  | { st_kind = Unix.S_DIR } -> print_string "Found dir"
+  | { st_kind = Unix.S_DIR } -> print_endline "Found dir"
   | _ -> print_endline "UNEXPECTED SUCCESS");
   Unix.rmdir "aaa"
-  |};
-  [%expect {|Found dir|}]
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    Found dir
+    Failure("unix_stat: not implemented")|}]
 
 let%expect_test "Unix.stat_symlink" =
   compile_and_run
     {|
+let f () =
   let oc = open_out "aaa" in
   output_string oc "bbb";
   close_out oc;
-  Unix.symlink "aaa" "ccc";
+  (try Unix.symlink "aaa" "ccc" with
+  | err -> print_endline (Printexc.to_string err));
   (match Unix.stat "ccc" with
   | exception err -> print_endline (Printexc.to_string err)
   | { st_kind = Unix.S_REG; st_size } ->
     print_string "File size: ";
     print_int st_size;
+    print_newline ();
   | _ -> print_endline "UNEXPECTED SUCCESS");
-  Sys.remove "ccc";
+  (try Sys.remove "ccc" with | _ -> ());
   Sys.remove "aaa";
-  |};
-  [%expect {|File size: 3|}]
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect
+    {|
+    File size: 3
+    Failure("unix_symlink: not implemented")
+    Failure("unix_stat: not implemented")|}]
 
 let%expect_test "Unix.symlink_Unix.readlink" =
   compile_and_run
     {|
+let f () =
   let oc = open_out "aaa" in
   output_string oc "bbb";
   close_out oc;
-  Unix.symlink "aaa" "ccc";
+  (try Unix.symlink "aaa" "ccc" with
+  | err -> print_endline (Printexc.to_string err));
   (match Unix.readlink "ccc" with
   | exception err -> print_endline (Printexc.to_string err)
   | path -> (
@@ -171,14 +190,20 @@ let%expect_test "Unix.symlink_Unix.readlink" =
     let contents = input_line ic in
     print_endline contents;
     close_in ic));
-  Sys.remove "ccc";
+  (try Sys.remove "ccc" with | _ -> ());
   Sys.remove "aaa";
-  |};
-  [%expect {|bbb|}]
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect
+    {|
+    bbb
+    Failure("unix_symlink: not implemented")
+    Failure("unix_readlink: not implemented")|}]
 
 let%expect_test "Unix.readlink_EINVAL" =
   compile_and_run
     {|
+let f () =
   (match Unix.readlink "." with
   (* Linux & Mac raise Unix.EINVAL *)
   | exception Unix.Unix_error (Unix.EINVAL, syscall, path) -> print_endline "EXPECTED ERROR"
@@ -186,12 +211,16 @@ let%expect_test "Unix.readlink_EINVAL" =
   | exception Unix.Unix_error (Unix.EUNKNOWNERR(code), syscall, path) -> print_endline "EXPECTED ERROR"
   | exception err -> print_endline (Printexc.to_string err)
   | _ -> print_endline "UNEXPECTED SUCCESS");
-  |};
-  [%expect {|EXPECTED ERROR|}]
+in
+f (); Sys.chdir "/static"; f () |};
+  [%expect {|
+    EXPECTED ERROR
+    Failure("unix_readlink: not implemented")|}]
 
 let%expect_test "Unix.lstat_file" =
   compile_and_run
     {|
+let f () =
   let oc = open_out "aaa" in
   output_string oc "bbb";
   close_out oc;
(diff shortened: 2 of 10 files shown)


phated:
If I'm looking at this correctly, it seems the only way to write tests that run in nodejs is to include them inside the `compiler/tests-compiler/` directory and use `Util.compile_and_run` - is this correct?

phated:
@hhugo if you have some time to look over this, I'm hoping to get this over the finish line within the next week-or-so!

TyOverby:
> If I'm looking at this correctly, it seems the only way to write tests that run in nodejs is to include them inside the `compiler/tests-compiler/` directory and use `Util.compile_and_run` - is this correct?

Yes, this is correct.  I’m not sure if any of the runtime js files are linked by default though, you may need to make changes in Util

phated:
> Yes, this is correct. I’m not sure if any of the runtime js files are linked by default though, you may need to make changes in Util

Hmm, I'm not sure what you mean

runtime/unix.js
```diff
@@ -64,3 +64,180 @@ function unix_isatty(fileDescriptor) {
...
+    'EINPROGRESS', 'EALREADY', 'ENOTSOCK', 'EDESTADDRREQ', 'EMSGSIZE',
+    'EPROTOTYPE', 'ENOPROTOOPT', 'EPROTONOSUPPORT', 'ESOCKTNOSUPPORT',
+    'EOPNOTSUPP', 'EPFNOSUPPORT', 'EAFNOSUPPORT', 'EADDRINUSE', 'EADDRNOTAVAIL',
+    'ENETDOWN', 'ENETUNREACH', 'ENETRESET', 'ECONNABORTED', 'ECONNRESET', 'ENOBUFS',
+    'EISCONN', 'ENOTCONN', 'ESHUTDOWN', 'ETOOMANYREFS', 'ETIMEDOUT', 'ECONNREFUSED',
+    'EHOSTDOWN', 'EHOSTUNREACH', 'ELOOP', 'EOVERFLOW'
+  ];
+  var variant = errors.indexOf(err.code);
+  if (variant < 0) {
+    // TODO: Make sure this works `EUNKNOWNERR(int)`
```
hhugo, line 91: This comment needs to be addressed

runtime/unix.js
```diff
@@ -64,3 +64,180 @@ function unix_isatty(fileDescriptor) {
...
+   *   | S_BLK                       (** Block device *)
+   *   | S_LNK                       (** Symbolic link *)
+   *   | S_FIFO                      (** Named pipe *)
+   *   | S_SOCK                      (** Socket *)
+   */
+  var file_kind;
+  if (js_stats.isFile()) {
+    file_kind = 0;
+  }
+  if (js_stats.isDirectory()) {
```
hhugo, line 118: Are we sure all those ifs are independent ? I would probably use if-else-if ..

runtime/unix.js
```diff
@@ -64,3 +64,180 @@ function unix_isatty(fileDescriptor) {
...
+    js_stats.mtimeMs,
+    js_stats.ctimeMs
+  );
+}
+
+//Provides: unix_stat
+//Requires: fs_node_supported, caml_stats_from_js
+//Requires: caml_jsbytes_of_string
+//Requires: caml_raise_unix_exn_of_nodejs_error, caml_failwith
+function unix_stat(filename) {
```
hhugo, line 173: Theses functions bypass the filesystem proxy of jsoo. I wonder if this is the right. Let me think about this a bit more

runtime/unix.js
```diff
@@ -64,3 +64,180 @@ function unix_isatty(fileDescriptor) {
...
+    }
+  } else {
+    caml_failwith('unix_lstat unsupported outside nodejs');
+  }
+}
+
+//Provides: unix_mkdir
+//Requires: fs_node_supported, caml_jsbytes_of_string
+//Requires: caml_raise_unix_exn_of_nodejs_error, caml_failwith
+function unix_mkdir(dirname, perm) {
```
hhugo, line 210: Related to #1109

hhugo:
> > Yes, this is correct. I’m not sure if any of the runtime js files are linked by default though, you may need to make changes in Util
> 
> Hmm, I'm not sure what you mean

That mean `Util.compile_and_run` might not give enough flags when calling the  js_of_ocam compiler,  one might need to tune `util.ml` to do so.

I don't think you need to worry about this as unix.js is already part of the default runtime.

hhugo:
> If I'm looking at this correctly, it seems the only way to write tests that run in nodejs is to include them inside the `compiler/tests-compiler/` directory and use `Util.compile_and_run` - is this correct?

This is not the only way. Look at tests in lib/test (e.g. https://github.com/ocsigen/js_of_ocaml/blob/master/lib/tests/test_nodejs_filesystem_errors.ml). Theses tests run in node only, the dune file says ` (inline_tests
    (modes js))`

runtime/unix.js
hhugo: I've implemented Sys.mkdir in #1134, `unix_mkdir` could directly call into caml_sys_mkdir

hhugo:
I though more about this. I would prefer to integrate with the "js_of_ocaml filesystem proxy" with something similar to what we have with `caml_sys_rename`
```js
function unix_stat(n){
  var root = resolve_fs_device(n);
  if(!root.device.stat)
    caml_failwith("unix_stat: no implemented");
  root.device.stat(root.rest);
}
```
You would then have update `fs_node.js` with new methods

hhugo:
Improves https://github.com/ocsigen/js_of_ocaml/issues/483

phated:
> I though more about this. I would prefer to integrate with the "js_of_ocaml filesystem proxy"

Excellent! Happy to do this. You are okay with JS built for the browser failing with `"unix_stat: not implemented"`?

hhugo:
> > I though more about this. I would prefer to integrate with the "js_of_ocaml filesystem proxy"
> 
> Excellent! Happy to do this. You are okay with JS built for the browser failing with `"unix_stat: not implemented"`?

That or we implement some dummy stat.

phated:
> That or we implement some dummy stat.

One thing I considered for Grain was embedding the file stats of the built JS file for anything in memory, since any embedded files would be a snapshot at that timestamp. Is this a good idea? If so, any thoughts on how I'd implement that?

hhugo:
Can you describe the grain use case a bit more ?

phated:
> Can you describe the grain use case a bit more ?

Currently, Grain compiles the compiler (written in Reason) to JS using JSOO and then includes it into a nodejs runtime called [`pkg`](https://www.npmjs.com/package/pkg). `pkg` provides the filesystem abstraction for us as long as the resulting JS calls nodejs APIs that we expect (the current implementation of this logic). Eventually the primary compiler will move to native executables.

However, I also did some research into a browser bundle to add a playground on our website—this should be a natural workflow, as we can compile to WebAssembly in the browser, and then directly run it in a WebAssembly context. Unfortunately, we'd need the unix methods supported in the browser because our compiler compares timestamps of files for caching and recompilation. I was planning to add this feature in the future when work began on our in-browser editor.

hhugo:
The current implementation of fs_fake.js is a bit naive currently. With some work you should be able to add  additional information like stat.

hhugo:
I don't have more to say right now, but I'm happy to answer any question if you spend some time working on it.

phated:
Thanks! For this first draft, I'll not handle the browser APIs, but I will plan to revisit in the future.

runtime/unix.js
phated: I investigated this some more and I found that `caml_sys_mkdir` can rethrow a `caml_raise_sys_error`, but we can't reconstruct a `Unix.Unix_error` if done the other way around. So I've refactored a bit to raise the Unix_error and then catch and re-raise in the `caml_sys_*` functions.

phated: I've separated that refactor into it's own commits

runtime/fs_node.js
```diff
@@ -40,30 +41,31 @@ MlNodeDevice.prototype.exists = function(name) {
   try {
     return this.fs.existsSync(this.nm(name))?1:0;
   } catch (err) {
-    caml_raise_sys_error(err.toString());
+    // TODO: Is it correct to always return a boolean, even on error?
+    return 0;
```
phated, line 45: @hhugo should this **ever** raise or would it just always return a boolean?

runtime/unix.js
phated: Done in my refactor. I believe they are all independent, but we can be careful.

runtime/unix.js
phated: This was an old comment, I'm 99% sure this is how the variant would be constructed. I don't know how to actually force this type of error in nodejs, and I don't think we can mock it. I removed the comment in my refactor.

runtime/unix.js
phated: Refactored!

phated:
@hhugo this should be ready for review! I added some tests that I could think of for the Unix calls that were implemented. Let me know if you think there should be more tests.

runtime/fs.js
```diff
@@ -189,24 +189,32 @@ function caml_sys_file_exists (name) {
...
   var root = resolve_fs_device(name);
-  var ok = root.device.unlink(root.rest);
-  if(ok == 0) caml_raise_no_such_file(name);
-  return 0;
+  try {
+    var ok = root.device.unlink(root.rest);
+    if(ok == 0) caml_raise_no_such_file(name);
+    return 0;
+  } catch (err) {
+    caml_raise_sys_error(name.toString());
```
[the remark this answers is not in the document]
hhugo: actually, fs_node already catch errors from unlink and convert it to sys_error

runtime/fs.js
```diff
@@ -189,24 +189,32 @@ function caml_sys_file_exists (name) {
...
 function caml_sys_read_directory(name){
   var root = resolve_fs_device(name);
-  var a = root.device.readdir(root.rest);
-  var l = new Array(a.length + 1);
-  l[0] = 0;
-  for(var i=0;i<a.length;i++)
-    l[i+1] = caml_string_of_jsbytes(a[i]);
-  return l;
+  try {
+    var a = root.device.readdir(root.rest);
```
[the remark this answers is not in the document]
hhugo: In fact, the try catch is already done in fs_node.js

runtime/fs.js
```diff
@@ -218,15 +226,19 @@ function caml_sys_is_directory(name){
...
+//Requires: caml_failwith, resolve_fs_device, caml_raise_sys_error
 function caml_sys_rename(o,n){
   var o_root = resolve_fs_device(o);
   var n_root = resolve_fs_device(n);
   if(o_root.device != n_root.device)
     caml_failwith("caml_sys_rename: cannot move file between two filesystem");
   if(!o_root.device.rename)
     caml_failwith("caml_sys_rename: no implemented");
-  o_root.device.rename(o_root.rest, n_root.rest);
+  try {
```
hhugo, line 237: try-catch already done in fs_node.js

runtime/fs_node.js
hhugo: Always returning a boolean seems fine. I don't think the native ocaml runtime would ever raise.

runtime/fs_node.js
```diff
@@ -40,30 +41,31 @@ MlNodeDevice.prototype.exists = function(name) {
...
   }
 }
 MlNodeDevice.prototype.mkdir = function(name, mode) {
   try {
     this.fs.mkdirSync(this.nm(name),{mode:mode});
     return 0
   } catch (err) {
-    caml_raise_sys_error(err.toString());
+    this.raise_unix_exn_of_nodejs_error(err);
```
hhugo, line 53: I've another proposal, have mkdir take a third argument and decide what error to raise based on that. What do you think ?

runtime/fs_node.js
```diff
@@ -40,30 +41,31 @@ MlNodeDevice.prototype.exists = function(name) {
...
   }
 }
 MlNodeDevice.prototype.rmdir = function(name) {
   try {
     this.fs.rmdirSync(this.nm(name));
     return 0
   } catch (err) {
-    caml_raise_sys_error(err.toString());
+    this.raise_unix_exn_of_nodejs_error(err);
```
hhugo, line 61: same here

runtime/fail.js
```diff
@@ -26,6 +26,9 @@ function caml_return_exn_constant (tag) { return tag; }
 //Provides: caml_raise_with_arg (const, const)
 function caml_raise_with_arg (tag, arg) { throw [0, tag, arg]; }
 
+//Provides: caml_raise_with_args (const, const, const)
+function caml_raise_with_args (tag, nargs, args) { throw [0, tag].concat(args); }
```
hhugo, line 30: Can we validate, maybe truncate `args` based on `nargs` ? Alternatively, we could completely remove `nargs`, this function is only a helper for writing stubs

hhugo (commented):
It took me a while get the full picture of changes. Some comments probably reflect that.
The one thing I don't like about it is the raise-catch-raise business. see my comment about adding an extra argument.

runtime/fs_node.js
```diff
@@ -104,16 +106,136 @@ MlNodeDevice.prototype.open = function(name, f) {
...
+    "EINPROGRESS", "EALREADY", "ENOTSOCK", "EDESTADDRREQ", "EMSGSIZE",
+    "EPROTOTYPE", "ENOPROTOOPT", "EPROTONOSUPPORT", "ESOCKTNOSUPPORT",
+    "EOPNOTSUPP", "EPFNOSUPPORT", "EAFNOSUPPORT", "EADDRINUSE", "EADDRNOTAVAIL",
+    "ENETDOWN", "ENETUNREACH", "ENETRESET", "ECONNABORTED", "ECONNRESET", "ENOBUFS",
+    "EISCONN", "ENOTCONN", "ESHUTDOWN", "ETOOMANYREFS", "ETIMEDOUT", "ECONNREFUSED",
+    "EHOSTDOWN", "EHOSTUNREACH", "ELOOP", "EOVERFLOW"
+  ];
+  var variant = errors.indexOf(err.code);
+  if (variant < 0) {
+    variant = BLOCK(0, err.errno);
```
hhugo, line 172: A comment explaining this is EUNKNOWNERR would be nice. What if errno is not defined ?

runtime/fail.js
phated: Oh, so the function that exists in ocaml core isn't used anywhere except for external stubs? Would JSOO want to provide the same interface that it provides? I'm happy to add a validation or truncate.

runtime/fs_node.js
phated: I think that will work. My issue was trying to raise the appropriate exception type depending on where this was called from. A flag would work better than the extra try-catch logic I added.

runtime/fs.js
```diff
@@ -189,24 +189,32 @@ function caml_sys_file_exists (name) {
...
-  return l;
+  try {
+    var a = root.device.readdir(root.rest);
+    var l = new Array(a.length + 1);
+    l[0] = 0;
+    for(var i=0;i<a.length;i++)
+      l[i+1] = caml_string_of_jsbytes(a[i]);
+    return l;
+  } catch (err) {
+    caml_raise_sys_error(name.toString());
```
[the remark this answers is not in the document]
phated: I'm fine with switching back to `err.toString()` for Sys_error exceptions (in combination with the flag mentioned below to raise Unix_errors); however, I reviewed the ocaml stdlib and it often raises a Sys_error with only the filepath. I'm guessing JSOO doesn't use only the filepath for better runtime debugging?

runtime/fs_node.js
phated: I believe errno is always supposed to be attached to a node "fs" error. Do you want me to add a fallback int?

phated: Comment added, let me know if there's a fallback in that makes sense 👍

phated:
@hhugo Thanks for the feedback! I've updated the MlNodeDevice methods to take the `raise_unix` flag as the last argument and use that to determine if Sys_error or Unix_error exceptions are raised.

There are a couple decisions to be made still (unresolved comments) but I'm happy to get those fixed up in the morning.

runtime/fail.js
hhugo: I don't think we need to keep the same interface in the case. I would vote for removing the unused argument

runtime/fs.js
```diff
@@ -189,7 +189,7 @@ function caml_sys_file_exists (name) {
 
 //Provides: caml_sys_read_directory
 //Requires: caml_string_of_jsbytes
-//Requires: caml_raise_not_a_dir, resolve_fs_device
+//Requires: caml_raise_not_a_dir, resolve_fs_device, caml_raise_sys_error
```
hhugo, line 192: try to keep the `//Requires:` accurate

runtime/fs_node.js
```diff
@@ -104,16 +121,161 @@ MlNodeDevice.prototype.open = function(name, f) {
...
+  if (variant < 0) {
+    // If none of the above variants, fallback to EUNKNOWNERR(int)
+    variant = BLOCK(0, err.errno);
+  }
+  var args = [
+    variant,
+    caml_string_of_jsbytes(err.syscall || ""),
+    caml_string_of_jsbytes(err.path || "")
+  ];
+  caml_raise_with_args(caml_named_value("Unix.Unix_error"), args.length, args);
```
hhugo, line 219: We might want to have a fallback in case `Unix.Unix_error` was not registered.

runtime/fs.js
phated: Thanks! Updated.

runtime/fs_node.js
phated: Is there a way to determine if `Unix.Unix_error` was registered inside the JS?

runtime/fs_node.js
```diff
@@ -104,16 +121,161 @@ MlNodeDevice.prototype.open = function(name, f) {
...
   }
 }
 
-MlNodeDevice.prototype.rename = function(o,n) {
+MlNodeDevice.prototype.rename = function(o, n, raise_unix) {
   try {
     this.fs.renameSync(this.nm(o), this.nm(n));
   } catch (err) {
-    caml_raise_sys_error(err.toString());
+    if (raise_unix) {
```
hhugo, line 136: I would refactor some of the error handling logic and lift things to a helper function.
```js
function raise_nodejs_error(err, raise_unix) {
    if (raise_unix) {
      this.raise_unix_exn_of_nodejs_error(err);
    } else {
      caml_raise_sys_error(err.toString());
    }
}
```

runtime/fs_node.js
hhugo: `if(caml_named_value("Unix.Unix_error"))`

runtime/fs_node.js
phated: Updated! I'm going to refactor this a bit more so the nodejs error transformation just happens in this helper.

runtime/fs_node.js
phated: Is it fine to fallback to the Sys_error if `Unix.Unix_error` is not registered? I changed the `if` statement to `if (raise_unix && caml_named_value("Unix.Unix_error")) {`

runtime/fs.js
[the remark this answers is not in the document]
hhugo: It seems the ocaml runtime often calls `caml_sys_error(filename)` that in turn tried to construct an informative message before raising `Sys_error`

runtime/fs_node.js
hhugo: looks good. Maybe make a single call to caml_named_value.

runtime/fs_node.js
```diff
@@ -104,17 +126,125 @@ MlNodeDevice.prototype.open = function(name, f) {
...
+    this.fs.symlinkSync(this.nm(src), this.nm(dst), to_dir ? 'dir' : 'file');
+    return 0;
+  } catch (err) {
+    this.raise_nodejs_error(err, raise_unix);
+  }
+}
+MlNodeDevice.prototype.readlink = function(name, raise_unix) {
+  try {
+    var js_stats = this.fs.readlinkSync(this.nm(name), 'utf8');
+    return caml_string_of_jsbytes(js_stats);
```
hhugo, line 167: - please rename js_stats.
- do we want to call caml_string_of_jsbytes or caml_string_of_jsstring ?

phated: I've renamed the variable. I'm actually still unclear about the difference between `caml_string_of_jsbytes` and `caml_string_of_jsstring`? If I change this to `caml_string_of_jsstring`, do I also need to change the 2 references inside `raise_nodejs_error`?

hhugo:
It almost feels you're sitting next to me :)

runtime/fs_node.js
phated: I'm noticing that `caml_string_of_jsstring` has the `//If: js-string` requirement. What are the implications of that if this uses `caml_string_of_jsstring`?

We aren't using the `js-string` option in Grain so I'm unfamiliar.

hhugo: ```//If: js-string``` is to have conditional stubs based on the js-string option see https://github.com/ocsigen/js_of_ocaml/pull/923. In short, one can choose the memory representation of ocaml strings.

`caml_string_of_jsbyte` should be used when the javascript string represent sequence of bytes, where all chars fall into the 0..255 range. `caml_string_of_jsstring` should be used for normal javascript strings, utf16 encoded. The utf8 encoding  argument is confusing here but I think you still get back a regular javascript string, utf16 encoded. I would use caml_string_of_jsstring and update the two other references.

compiler/tests-compiler/unix_fs.ml
```diff
@@ -19,8 +19,14 @@
 
 open Util
 
+(* Unix.LargeFile requires unix_stat_64 which causes tests
```
hhugo, line 22: I don't understand why this is only an issue on windows.
What about we just alias the primitive to unix_stat ?

phated: I was also confused by that. The type signatures look the same, so it would probably work. I should probably provide the `unix_lstat_64` too: https://github.com/ocaml/ocaml/blob/f68acd1a618ac54790a8347fad466084f15a9a9e/otherlibs/win32unix/unix.ml#L407-L408

compiler/tests-compiler/sys_fs.ml
```diff
@@ -50,3 +50,23 @@ let%expect_test _ =
   |};
   [%expect {|
     bbb|}]
+
+let%expect_test _ =
+  compile_and_run
+    {|
+  (match Sys.mkdir "/not/exists" 0o777 with
+  | exception Sys_error path -> print_endline ("Sys_error: " ^ path)
+  | exception _ -> print_endline "INCORRECT ERROR"
```
hhugo, line 59: Printing the exception will be more informative in the error case.

phated: I was just trying to debug some Windows test errors and thought the same thing 😅 Updated now.

phated:
The MacOS runs have been really flakey on my fork. It seems to be something with the opam env?

I believe there's still 1 windows test failure I need to handle, but my run should finish soon.

hhugo:
> The MacOS runs have been really flakey on my fork. It seems to be something with the opam env?
> 
> I believe there's still 1 windows test failure I need to handle, but my run should finish soon.

@smorimoto, the macos CI look rather unstable, I don't know if there is anything we can do about it ?

phated:
I believe with that last push that the tests should run on Windows. There were a couple issues:
1. The JSOO program on Windows raises a Sys_error when `Unix.mkdir` or `Unix.rmdir` use an absolute path, like `"/not/exists"`
2. The Unix_errors raised by Windows are often different than their POSIX counterparts. I couldn't find a good way to handle that with ppx_expect so I just made the tests handle both cases where necessary.
3. Tests don't log filepaths anymore since those aren't portable across operating systems.

compiler/tests-compiler/unix_fs.ml
```diff
@@ -172,11 +175,14 @@ let%expect_test "Unix.readlink_EINVAL" =
   compile_and_run
     {|
   (match Unix.readlink "." with
-  | exception Unix.Unix_error (Unix.EINVAL, syscall, path) -> print_endline ("EINVAL: " ^ syscall)
+  (* Linux & Mac raise Unix.EINVAL *)
+  | exception Unix.Unix_error (Unix.EINVAL, syscall, path) -> print_endline "EXPECTED ERROR"
+  (* Windows raises Unix.EUNKNOWNERR *)
+  | exception Unix.Unix_error (Unix.EUNKNOWNERR(code), syscall, path) -> print_endline "EXPECTED ERROR"
```
phated, line 181: We also have a `Unix.EUNKNOWNERR` now because Windows doesn't know how to handle readlink on a directory.

hhugo:
> I believe with that last push that the tests should run on Windows. There were a couple issues:
> 
> 1. The JSOO program on Windows raises a Sys_error when `Unix.mkdir` or `Unix.rmdir` use an absolute path, like `"/not/exists"`
> 2. The Unix_errors raised by Windows are often different than their POSIX counterparts. I couldn't find a good way to handle that with ppx_expect so I just made the tests handle both cases where necessary.
> 3. Tests don't log filepaths anymore since those aren't portable across operating systems.

For 1., Maybe we should fix the tests to use a valid root based on the system.

hhugo:
Were you able to test grain with this PR ? Do you still this "hack"/jsoo_headers/runtime changes after that ? (Except maybe for always mounting c:/)

smorimoto:
The dune cache does make things pretty bad in some cases. It's fixed in dune 3, but the problem can still happen in 2. So I was thinking about backporting cache functionality from 3 to 2, but I hadn't even started discussing that yet.

phated:
> For 1., Maybe we should fix the tests to use a valid root based on the system.

I don't think it's entirely necessary to test the absolute paths, but if you think it makes sense, I can look into how that would be done. I think it would be hard because JSOO always tells the program it is in a unix env.

> Were you able to test grain with this PR ? Do you still this "hack"/jsoo_headers/runtime changes after that ? (Except maybe for always mounting c:/)

I'll be testing this today. I believe the only thing Grain will need is to always mount the C: drive due to our packaging mechanisms.

hhugo:
> I don't think it's entirely necessary to test the absolute paths, but if you think it makes sense, I can look into how that would be done.

Using unix path on windows make us test `resolve_fs_device` only.
If relative path allow us to test correctly, I'm fine with it.
 
> I think it would be hard because JSOO always tells the program it is in a unix env.

This is no longer true, as of #1132

compiler/tests-compiler/unix_fs.ml
```diff
@@ -0,0 +1,217 @@
...
+
+let%expect_test "Unix.mkdir_Unix.rmdir" =
+  compile_and_run
+    {|
+  Unix.mkdir "aaa" 0o777;
+  Unix.mkdir "aaa/bbb" 0o777;
+  let l = Sys.readdir "aaa" |> Array.to_list in
+  List.iter print_endline l;
+  (match Unix.rmdir "aaa" with
+  | exception _ -> ()
```
hhugo, line 30: Could we restrict to exception we expect

compiler/tests-compiler/unix_fs.ml
```diff
@@ -0,0 +1,217 @@
...
+let%expect_test "Unix.mkdir_Unix.rmdir_static" =
+  compile_and_run
+    {|
+  Sys.chdir "/static/";
+  Unix.mkdir "aaa" 0o777;
+  Unix.mkdir "aaa/bbb" 0o777;
+  let l = Sys.readdir "aaa" |> Array.to_list in
+  List.iter print_endline l;
+  (match Unix.rmdir "aaa" with
+  | exception _ -> ()
```
hhugo, line 46: Same here

phated:
> This is no longer true, as of #1132

In that PR, I believe you fixed the Windows pathing issue, but I think `Sys.win32` still always returns false. How should I build the absolute path independent of the filesystem?

hhugo:
One more thing, could you add tests for the new functions but under the `/static/`directory. The tests should show that some features are not implemented yet. It will be useful if you/one decide to add the support.

hhugo:
> > This is no longer true, as of #1132
> 
> In that PR, I believe you fixed the Windows pathing issue, but I think `Sys.win32` still always returns false. How should I build the absolute path independent of the filesystem?

I think Sys.os_type should return Cygwin on windows. But indeed, I haven't checked other values.

compiler/tests-compiler/unix_fs.ml
hhugo: I've done that

compiler/tests-compiler/unix_fs.ml
hhugo: I've done that

hhugo:
I've changed the tests related to mkdir and rmdir to also run on the '/static' root and fixed fs_fake accordingly.

runtime/fs_fake.js
```diff
@@ -62,15 +63,64 @@ MlFakeDevice.prototype.exists = function(name) {
   this.lookup(name);
   return this.content[name]?1:0;
 }
-MlFakeDevice.prototype.mkdir = function(name,mode) {
+MlFakeDevice.prototype.mkdir = function(name,mode, raise_unix) {
+  var unix_error = raise_unix && caml_named_value('Unix.Unix_error');
+  if(this.exists(name)) {
+    if (unix_error) {
+      caml_raise_with_args(unix_error, [8, caml_string_of_jsstring("mkdir"), caml_string_of_jsstring(this.nm(name))]);
```
hhugo, line 70: We need a readable way to reference the error constructor.

runtime/fs_fake.js
```diff
@@ -62,15 +63,64 @@ MlFakeDevice.prototype.exists = function(name) {
...
+  }
   this.create_dir_if_needed(this.slash(name));
 }
-MlFakeDevice.prototype.rmdir = function(name) {
+MlFakeDevice.prototype.rmdir = function(name, raise_unix) {
+  var unix_error = raise_unix && caml_named_value('Unix.Unix_error');
   var name_slash = (name == "")?"":(this.slash(name));
   var r = new RegExp("^" + name_slash + "([^/]+)");
+  if(!this.exists(name)) {
+    if (unix_error) {
```
hhugo, line 101: This error handling here is very repetitive, can we do something about it ?

phated:
Sorry for disappearing for a little bit, I was putting together a patch for Grain to ensure this all worked when combined. I drafted a PR on Grain with this code applied at https://github.com/grain-lang/grain/pull/922

runtime/fs_fake.js
phated: @hhugo I added another function in https://github.com/ocsigen/js_of_ocaml/pull/1115/commits/fad310827a9b0829ae60572f5281b1f11276c3ff called `make_unix_err_args` that allows the string named to be used. Do you think that is a good solution?

hhugo:
I've fixed the CI on master. Can you rebase your PR.

runtime/fs_fake.js
phated: At first glance, I don't see a good way to reduce the repetition. Any helper methods we add for this will require many arguments to handle the variation between different syscalls ("mkdir" and "rmdir" currently). We could implement them to take JS objects like `{ message, code, syscall, path }` and make a helper similar to the `raise_nodejs_error` helper in fs_node.

I'll keep thinking about this.

phated:
> Can you rebase your PR.

Done!

.github/workflows/build.yml
```diff
@@ -69,6 +69,9 @@ jobs:
       - run: opam exec -- git diff --exit-code
         if: ${{ matrix.ocaml-compiler == '4.12.0' }}
 
+      - run: opam exec -- git diff --exit-code
```
hhugo, line 72: This is duplicated. You should remove this diff

phated: Good catch. Updated!

hhugo:
> One more thing, could you add tests for the new functions but under the `/static/`directory. The tests should show that some features are not implemented yet. It will be useful if you/one decide to add the support.

I few more tests and we should be ready to merge

phated:
> few more tests and we should be ready to merge

I'll try to get to that this afternoon. Have a couple of job interviews, but hope to have time outside of them.

phated:
I added the tests in the same way you added Unix.mkdir and Unix.rmdir tests, but I added the Failure messages to the expected output. If there is a better way to handle these, please let me know 🙏

hhugo:
Thanks a lot for the hard work, I will merge once the CI is happy.

PR #1116, opened 2021-07-26
Enable dune cache on the macOS runners as well

smorimoto:


hhugo:
@smorimoto, do you know what's going on ?

smorimoto:
I just noticed a weird error. I will take a closer look, as that should definitely not happen.

PR #1125, opened 2021-08-29
fix: Handle Windows root drive letters in caml_make_path

phated:
This upstreams a patch we use in Grain to handle Windows roots within `caml_make_path`

### DIFF
--- a/runtime/fs.js
+++ b/runtime/fs.js
@@ -39,7 +39,8 @@ function MlFile(){  }
 //Requires: caml_jsstring_of_string
 function caml_make_path (name) {
   name=caml_jsstring_of_string(name);
-  if(name.charCodeAt(0) != 47)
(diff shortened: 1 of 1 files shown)


hhugo:
An alternative fix landed in #1132

phated:
Thanks @hhugo! Just looked through your PR and it looks great. Any thoughts on #1115? I'd love to get the Unix calls mapped to nodejs calls

hhugo:
I'll try to have a look later

hhugo:
@phated , does #1132 cover all aspect of https://github.com/grain-lang/grain/blob/main/compiler/grainc/jsoo_hacks.js ?

phated:
Yep! Those are the only externals that grain needs; however, I believe we can use the utilities I added to cover more of unix.ml

PR #1127, opened 2021-09-07
DRAFT: Dom_html: adding connection and share into navigator (3.10.2)

Thibaut-Gudin:


lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
 
 class type undoManager = object end
 
+class type optionPromise =
```
hhugo, line 2168: I think this should be named differently

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+  object
+    method url : js_string t option prop
```
hhugo, line 2170: You don't want `option` here. `optdef` maybe ?

lib/js_of_ocaml/dom_html.ml
Thibaut-Gudin: So what name should I use instead? (I based this name from the object name "optionEvent")

lib/js_of_ocaml/dom_html.ml
Thibaut-Gudin: I had a lot of trouble to create the constructor "optionPromise", I "deduce" the 'option' type from the only successful experimentation  tu I indeed hesitate with "optdef" and "opt".

hhugo: an OCaml `option` is represented either by `0` (for `None`) or `[0; x]` (for `Some x`)

lib/js_of_ocaml/dom_html.ml
hhugo: `optionEvent` ? where ?
If theses options are about `share`, should it be `shareOption` maybe ?

lib/js_of_ocaml/dom_html.ml
Thibaut-Gudin: I don't know how to create a constructor of "optionPromise" (that I'm renaming "shareOption" if the parameters are optdef, it will return:

<<
2416 |   option##.text := text;
         ^^^^^^^^^^^^^
Error: This expression has type
         js_string t optdef prop =
           < get : js_string t optdef; set : js_string t optdef -> unit >
           gen_prop
       but an expression was expected of type
         < set : 'a option -> unit; .. > gen_prop
       Types for method set are incompatible
>>

I haven't found any other way to create an instance of "shareOption" other than a typed "Js.Unsafe.obj" and I don't understood who to use the optdef values here...

hhugo: try `option##.text := Optdef.option text;`

vouillon:
These features are not widely supported, so we need a way to check whether they are available
https://caniuse.com/mdn-api_workernavigator_connection
https://caniuse.com/web-share

Thibaut-Gudin:
I just added boolean functions that indicates if modules are available.
I tried to add them inside the "navigator" object directly for better implementation but I wasn't able to directly implement the functions inside the object definition.

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+    method url : js_string t optdef prop
+
+    method text : js_string t optdef prop
+
+    method title : js_string t optdef prop
+  end
+
+class type promise =
+  object
+    method _then : ('self t, event t -> unit) Js.meth_callback opt -> promise t meth
```
hhugo, line 2179: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share says the promise is resolved with undefined. Are we really getting and `event t` ?

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+
+    method catch : ('self t, js_string -> unit) Js.meth_callback -> unit meth
```
hhugo, line 2181: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share says  the promise can be rejected with a DomException not a string.

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+  end
+
+class type connection =
+  object
+    method downlink : int t readonly_prop
```
hhugo, line 2186: should be `int readonly_prop`.

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+
+    method downlinkMax : int t readonly_prop
```
hhugo, line 2188: same here

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2165,6 +2165,37 @@ class type history =
...
+
+    method effectiveType : js_string t readonly_prop
+
+    method rtt : int t readonly_prop
```
hhugo, line 2192: same here

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2188,8 +2219,19 @@ class type navigator =
...
+
+    method connection : connection t readonly_prop
+
+    method share : #shareOption t -> promise t meth
   end
 
+let navigator_supported () = Js.Optdef.test Js.Unsafe.global##.navigator
+
+let navigator_connection_supported () =
+  Js.Optdef.test Js.Unsafe.global##.navigator##.connection
```
hhugo, line 2231: This will raise if navigator is not supported.

lib/js_of_ocaml/dom_html.ml
```diff
@@ -2188,8 +2219,19 @@ class type navigator =
...
+
+let navigator_share_supported () = Js.Optdef.test Js.Unsafe.global##.navigator##.share
```
hhugo, line 2233: This will raise if navigator is not supported.

hhugo:
@vouillon, @jrochel, @Thibaut-Gudin is this still being worked on ?

jrochel:
No, not at all.

hhugo:
@jrochel, It's not ready to be merged. Should I close it ?

PR #1128, opened 2021-09-07
revert accidental version bump

jrochel:


CHANGES.md
```diff
@@ -1,8 +1,5 @@
 # ?? (??) - ??
 
-# 3.10.1 (2021-09-06) - Paris
-Lib: add hidden, onfullscreenchange and onwebkitfullscreenchange to to Dom_html.document
```
hhugo, line 4: Can we keep this changelog entry

Thibaut-Gudin:
We should probably also cancel the change of the "VERSION" file too? (it is wrote 3.10.1 here too).

jrochel:
> We should probably also cancel the change of the "VERSION" file too? (it is wrote 3.10.1 here too).

Indeed. Thanks.

PR #1131, opened 2021-09-17
Fix toplevel generation

### LINKED ISSUE #1129: [BUG] Error with lwt_example
**Describe the bug**
After compiling the toplevel example and opening the index.html in the browser, this error pops up in the console :

> Fatal error: Toplevel bytecode executable is corrupted
js_print_stderr @ toplevel-4.12.0.js:1413

**Expected behavior**
The toplevel should load as expected

**Versions**
I used the latest release, js_of_ocaml 3.10.0
Note that this issue didn't occur on 3.9.1

hhugo:
extend #1130 
fix #1129 

- String.split_char had a surprising/wrong behavior of not round-tripping with String.concat. We can use the upstream one now that the minimal ocaml version is 4.04
- Add (minimal) testing for the toplevel

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,10 @@
 # ?? (??) - ??
 
 ## Features/Changes
-* Lib: add hidden, onfullscreenchange and onwebkitfullscreenchange to to Dom_html.document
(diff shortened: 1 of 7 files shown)


AltGr:
I often say that there are as many `string_split` variants as projects :)

Thanks!

PR #1132, opened 2021-09-17
Make all tests pass on windows

hhugo:
will include/extends #1125 
- [x] properly detect absolute path on windows
- [x] make windows failure hard errors on windows 
- [x] fix toplevel on windows
- [x] fix expect test on window
- [x] fix all other tests on windows


smorimoto (approved):

hhugo:
I'm going to fix the windows build before merging

hhugo:
Expect test are still broken

smorimoto:
By doing the same thing, I think you can fix it without making patches.
https://github.com/ocaml/ocaml-lsp/blob/b841d46f6a4bbb0719a3a726c020959a5e4868e2/.github/workflows/build-and-test.yml#L27

hhugo:
> By doing the same thing, I think you can fix it without making patches.
> https://github.com/ocaml/ocaml-lsp/blob/b841d46f6a4bbb0719a3a726c020959a5e4868e2/.github/workflows/build-and-test.yml#L27

If this is enough, I feel it should be documented on the ppx_expect repo.

hhugo:
Will merge once CI is green

PR #1135, opened 2021-09-22
Switch tests to 4.13

hhugo:


hhugo:
@smorimoto, can I test OCaml 4.13-rc with setup-ocaml ?

smorimoto:
@hhugo I just pushed a commit to enable `4.13-rc`!

hhugo:
Thanks

smorimoto:
Can we disable the test for 4.12? If so, I can push an additional commit.

hhugo:
Tests are supposed to pass on 4.12 in this PR

smorimoto:
That's good!

hhugo:
@smorimoto, have I broken something during rebase ? do you expect the missing compiler for windows ?

smorimoto:
@hhugo Maybe -- Yes, 4.13 is not yet available on Windows, so you should set it to `4.12.x`.

Issue #397, opened 2016-01-07
Stdio not handled well


Leonidas-from-XIV:
Consider this source code, which works just fine, bytecode-compiled:

``` ocaml
let () =
  input_line stdin
  |> output_string stdout
```

Compiling it to JS works without any options, which would imply that it should more or less work. Trying it out with Node.js fails, though:

``` foo.js:185
i=ak(r);if(0===i){if(!m)throw D;var
                              ^
248,End_of_file,-5
```

I know that stdio is not defined in JS, but I would expect either a compiler error or (even better), a message that I have to pick a JS runtime backend which is used for stdio, like Node.js.

dbuenzli:
You can define `stdio` even in the browser via the pseudo file system, see [`Sys_js`](http://ocsigen.org/js_of_ocaml/2.6/api/Sys_js). So maybe one should define a library to link against when compiled for node.js that just setups `node.js` notion of channel with `Sys_js` and maybe other things in `Sys` and `Unix` that may become available in such a setting.

dbuenzli:
Sorry in fact `Sys_js` is only for defining what happens when you write on output channels and when you open input channels, it is true that `stdin` is formally undefined.

hhugo:
See how this is done in the toplevel.  https://github.com/ocsigen/js_of_ocaml/blob/master/toplevel/toplevel.cppo.ml#L353

Leonidas-from-XIV:
I think this would make a nice PR to get one's feet wet (might even try to implement a solution myself), but what would you think would be a good approach to solve it? From the linked issue, it could be useful to have a switch which would select a "backend" for stdio, like browser or Node.js, but I'd love to hear your input on that.

hhugo:
@phated, I think you've fixed this in grain, would you want to contribute ?

PR #1136, opened 2021-09-22
Runtime: make stdin work on nodejs

hhugo:
Fix https://github.com/ocsigen/js_of_ocaml/issues/397
Makes https://github.com/grain-lang/grain/blob/main/compiler/grainc/jsoo_header.re obsolete (cc @phated)


phated:
I tested this on Windows through the Grain compiler and it seems to be working great! Thanks!

PR #1140, opened 2021-09-24
Allow creating nodejs mount points

phated:
@hhugo related to our discussion in https://github.com/grain-lang/grain/pull/922 - I was wondering if we could add an optional parameter to `Sys_fs.mount` in order to create `MlNodeDevice` mount points


hhugo:
The lookup function is useless for the MlNodeDevice.

What do you think about "mounting" windows drive automatically on demand instead ?

phated:
> What do you think about "mounting" windows drive automatically on demand instead ?

Your solution looks way better than this workaround. I'm going to test your solution right now.

phated:
Closing in favor of #1143

PR #1159, opened 2021-10-19
[Preview] upgrade to ocamlformat.0.20.0

gpetiot:
This is a preview of the not-yet-released `ocamlformat.0.20.0`, please wait until the package is published in opam to merge this PR. The output is still likely to slightly change before the package is released.
The changes are due to:
- `module-item-spacing` is now set to `compact` for the default profile
- `module-item-spacing` is now correctly applied to mutually recursive type definitions
- formatting of list elements has been improved

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -12,4 +12,5 @@ break-infix-before-func
 break-separators=before
 dock-collection-brackets=false
 margin=90
-version=0.19.0
+module-item-spacing=sparse
+version=0.20.0
--- a/benchmarks/sources/ml/nucleic.ml
+++ b/benchmarks/sources/ml/nucleic.ml
@@ -6919,13 +6919,11 @@ let anticodon_domains =
   ; stacked3' rG 37 38
   ; stacked3' rA 36 37
   ; stacked3' rA 35 36
-  ; stacked3' rG 34 35
-  ; (* <-. Distance      *)
-    p_o3' rCs 32 31
-  ; (*   | Constraint    *)
-    p_o3' rUs 33 32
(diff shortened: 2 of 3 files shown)


hhugo:
Could you update the PR and set `module-item-spacing` so that the diff is not that big ?

gpetiot:
I updated the PR, but please note that our plans on ocamlformat's side is to deprecate the mix-and-match of options in version 0.21.0, in the long run we plan to only make ocamlformat customizable by profiles (default, janestreet, ocamlformat).

hhugo:
Thanks, the current diff is MUCH easier to digest/review.
Do you plan to fix the remaining diffs ?

gpetiot:
The remaining diff won't be fixed, it's due to the `[_; ...]` lists not being built as `_ :: _` anymore, the comments attached to each expression element are now formatted like any expression (there was some ad-hoc tweaking to handle comments in `::` expressions, as comments attached to the last element were attached to `[]`), which causes the trailing comment to wrap after the expression when it fits.

hhugo:
Ok, I restored the comments to what they were before we switched to ocamlformat.

smorimoto:
How about disabling only a specific option with `[@ocamlformat "option=VAL"]`? That is more my thing.

hhugo:
> How about disabling only a specific option with `[@ocamlformat "option=VAL"]`? That is more my thing.

I'm not sure what you mean. Where do you want to do that ?
Anyway, I'd prefer to keep the upgrade PR as small as possible. Other changes can be made later on.

Issue #1162, opened 2021-11-01
[FEATURE REQUEST] Improved development env repeatability


cdaringe:
**Is your feature request related to a problem? Please describe.**

On my development machine, I receive unexpected side-effects when building and running the project. Such effects are surely unrelated to my changeset(s). Examples include:

1. Missing cmi

```sh
❯ make all
dune build @all
jsoo_listunits toplevel/examples/lwt_toplevel/export.txt
Could not find cmi Dynlink_compilerlibs.cmi or dynlink_compilerlibs.cmi in /Users/myuser/.opam/4.12.0/lib/ocaml
```

2. Formatting changes

```diff
diff --git a/compiler/tests-compiler/mutable_closure.ml b/compiler/tests-compiler/mutable_closure.ml.corrected
index 44fdad6658..d79b020dda 100644
--- a/compiler/tests-compiler/mutable_closure.ml
+++ b/compiler/tests-compiler/mutable_closure.ml.corrected
@@ -158,7 +158,7 @@ let%expect_test _ =
         var
          _c_=indirect[1],
          _d_=function(f){return caml_call1(f,0)},
-         indirect$0=caml_call2(Stdlib_List[19],_d_,_c_),
+         indirect$0=caml_call2(Stdlib_list[19],_d_,_c_),
```

3. Undefined symbols

```txt
Undefined symbols for architecture x86_64:
  "_caml_acosh_float", referenced from:
      _camlJsoo_testsuite__Test_floats__p_1443 in jsoo_testsuite.a(jsoo_testsuite__Test_floats.o)
      _camlJsoo_testsuite__Test_floats__1 in jsoo_testsuite.a(jsoo_testsuite__Test_floats.o)
  "_caml_asinh_float", referenced from:
      _camlJsoo_testsuite__Test_floats__p_1453 in jsoo_testsuite.a(jsoo_testsuite__Test_floats.o)
      _camlJsoo_testsuite__Test_floats__1 in jsoo_testsuite.a(jsoo_testsuite__Test_floats.o)
  "_caml_atanh_float", referenced from:
```

...and a few others which have occurred over the passed few days.

**Describe the solution you'd like**

I desire a recipe for reproducible & successful development, where libraries & binaries are known stable/working.

Ideas:

[...]

PR #1163, opened 2021-11-01
feat: support docker development

cdaringe:
# Problem

See https://github.com/ocsigen/js_of_ocaml/issues/1162

# Solution

Demonstrate a casual pass at slapping together a docker environment.

To try it out:

- pull this branch
- `make dev_container`
- `make all test`

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,12 @@
+# Develop using a known working environment.
+# @usage
+# run `make dev_container`
+FROM ocaml/opam:ubuntu-18.04-ocaml-4.12
+USER root
+RUN apt-get update && apt-get install -y libgmp-dev pkg-config
+USER opam
+RUN opam install --deps-only js_of_ocaml js_of_ocaml-lwt js_of_ocaml-compiler js_of_ocaml-ocamlbuild js_of_ocaml-toplevel js_of_ocaml-ppx js_of_ocaml-ppx_deriving_json js_of_ocaml-tyxml
+RUN opam install lwt reactiveData tyxml ppxlib cohttp-lwt-unix menhir ppx_expect yojson sexplib ocp-indent graphics
+RUN curl -fsSL https://fnm.vercel.app/install | bash
+ENV PATH=/home/opam/.fnm:$PATH
+RUN eval "`fnm env`" && fnm install 16
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -3,10 +3,29 @@
 # run `make dev_container`
 FROM ocaml/opam:ubuntu-18.04-ocaml-4.13
 USER root
-RUN apt-get update && apt-get install -y libgmp-dev pkg-config
+# https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash \
+  && apt-get update \
+  && apt-get install -y libgmp-dev pkg-config nodejs
 USER opam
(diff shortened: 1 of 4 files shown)


cdaringe:
I am not invested into the effort of this PR. The container does work though, and is really nice single-command execution into a functional dev environment :) If you are not interested, please close :)

Dockerfile.dev
```diff
@@ -0,0 +1,12 @@
+# Develop using a known working environment.
+# @usage
+# run `make dev_container`
+FROM ocaml/opam:ubuntu-18.04-ocaml-4.13
+USER root
+RUN apt-get update && apt-get install -y libgmp-dev pkg-config
+USER opam
+RUN opam install --deps-only js_of_ocaml js_of_ocaml-lwt js_of_ocaml-compiler js_of_ocaml-ocamlbuild js_of_ocaml-toplevel js_of_ocaml-ppx js_of_ocaml-ppx_deriving_json js_of_ocaml-tyxml
+RUN opam install lwt reactiveData tyxml ppxlib cohttp-lwt-unix menhir ppx_expect yojson sexplib ocp-indent graphics
+RUN curl -fsSL https://fnm.vercel.app/install | bash
```
hhugo, line 10: Can we apt install node  instead ?

cdaringe: done. i usually avoided apt before for node, because debian is still pegged to node 8 .x (8 major versions behind), but the node site now recommends it with a deb.nodesource.com as a deb-src :)

smorimoto:
I'm a little confused about whether Docker is really needed here. A simple development workflow using local switch should be just enough.

smorimoto:
In fact, I don't think we need to add any code to the repository to support any kind of extra development workflow.

cdaringe:
Nothing here is needed. It about reproducibility, convenience. I'm not attached to it, but i do think there is room to improve the project bootstrap UX

smorimoto:
In terms of reproductivity, Docker doesn't really help much, it just makes the development experience worse, and I think you should use Nix if you really care about reproducibility.

The best way to bootstrap the development environment is as follows, which I use on a daily basis:

```
opam switch create . ocaml-base-compiler.4.12.1 --no-install --yes
opam pin scan . | awk 'NR>1{print $1}' | xargs -I PKG opam pin add PKG.dev . --no-action --yes
opam install . --deps-only --with-test
```

smorimoto:
In fact, I can agree this is a pretty crazy solution, but this problem should essentially be solved on the opam side.

smorimoto:
I can agree this is a pretty crazy solution, but this problem should essentially be solved on the opam side.

cdaringe:
The motivation for this PR is that when I began making my first patches to the project, I hit bumps in the developer journey. Not big problems at all, but just a few little ones that were sufficient to motivate me to send this 😄 . Simply improving CONTRIBUTING.md with more rigorous steps could suffice as the solution. Offering a single command to fully hydrate a developer environment I think is a fantastic development UX that I would have certainly used to get started. Happy to ditch docker for that if the maintainers aren't keen on it. Adding a `bootstrap` target to the Makefile is also an option.

> I don't think we need to add ... any kind of extra development workflow.

> The best way to bootstrap the development environment is \<user's personal workflow\>

I respect that workflow. If that's how reproducibility is achieved, let's check that in :). However, I'll kindly suggest that it's already disjoint with the project's requirements. Wrong compiler, missing some packages that are not listed in opam files, etc. It's things like this that motivated the PR--an unambiguous "run this <cmd>" and you'll be good to go with high confidence.

hhugo:
I'll fix the contributing section in https://github.com/ocsigen/js_of_ocaml/pull/1210 but I don't feel like maintaining this docker workflow, so I'll close this PR.

PR #1123, opened 2021-08-23
Lib: add message event types and PerformanceObserver

pmwhite:

PR #1164, opened 2021-11-06
Lib: add message event types and PerformanceObserver

pmwhite:
This is re-opened version of https://github.com/ocsigen/js_of_ocaml/pull/1123 to fix the bad state that I got the repo into. Sorry for the extra noise.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 # dev (2021-??-??) - ??
 ## Features/Changes
 * Compiler: static evaluation of backend_type
+* Lib: add messageEvent to Dom_html
+* Lib: add PerformanceObserver API
 
 ## Bug fixes
 * Compiler: fix sourcemap warning for empty cma
--- a/lib/js_of_ocaml/performanceObserver.mli
+++ b/lib/js_of_ocaml/performanceObserver.mli
@@ -1,0 +1,59 @@
+(** PerformanceObserver API
+
+    A code example:
+    {[
+      if (PerformanceObserver.is_supported()) then
+        let entry_types = [ "measure" ] in
+        let f entries observer =
+          let entries = entries##getEntries in
+          Firebug.console##debug entries ;
+          Firebug.console##debug observer
+        in
+        PerformanceObserver.observe ~entry_types ~f
+        ()
+   ]}
+
+   @see <https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver> for API documentation.
+*)
+
+class type performanceObserverInit =
+  object
+    method entryTypes : Js.js_string Js.t Js.js_array Js.t Js.writeonly_prop
+  end
+
+class type performanceEntry =
+  object
+    method name : Js.js_string Js.t Js.readonly_prop
+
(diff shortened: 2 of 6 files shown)


pmwhite:
The checks failed on MacOS, but I highly doubt that has anything to do with this PR, so my guess is that the build is just flaky. In response to the suggestion to bind supportedEntryTypes, I'd rather not, because I haven't actually used that API in anything, so writing the binding would probably be untested. This current patch is just lifted from an app, so I'm pretty confident that the bindings are done correctly.

lib/js_of_ocaml/performanceObserver.mli
```diff
@@ -0,0 +1,59 @@
...
+
+    method takeRecords : performanceEntry Js.t Js.js_array Js.t Js.meth
+  end
+
+val performanceObserver :
+  (   (performanceObserverEntryList Js.t -> performanceObserver Js.t -> unit) Js.callback
+   -> performanceObserver Js.t)
+  Js.constr
+
+val is_supported : unit -> bool
```
cdaringe, line 54: i wonder if `is_supported` should be baked in, or left to user space? genuine ask, not a passive challenge :)

CHANGES.md
```diff
@@ -25,6 +25,8 @@
 * Compiler: drop support for OCaml 4.03 and bellow
 * Lib: add offsetX and offsetY to Dom_html.mouseEvent
 * Lib: add innerText property for Dom_html
+* Lib: add messageEvent to Dom_html
+* Lib: add PerformanceObserver API
```
hhugo, line 29: The changelog entry is misplaced

lib/js_of_ocaml/performanceObserver.mli
hhugo: What do you mean by backed it ?

cdaringe: sorry, it's a quirky phrase :)

> baked in: incorporate something as an integral part of a product, service, or system.

one may observe that https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/isSupported doesn't exist. thus, i am asking "should jsoo decorate web apis with extra functions, methods, or properties?"

pmwhite: I've included `is_supported` for consistency with the existing `intersectionObserver` and `resizeObserver` APIs. One could quite reasonably argue that jsoo should just expose web APIs and nothing more, and I might even agree with it. However, I think this PR does the correct thing by following the existing convention.

CHANGES.md
pmwhite: I moved the changelog entries to the section at the top. I thought I had corrected that mistake when I re-created the PR, but I guess not. I'm a little unsure that I've put them in the correct place still, because they're under the section dated 2021-10-06, which makes me think the release already happened.

hhugo: If you rebase your PR, you'll see that there is a new section for unrelated changes

pmwhite: Ah, I see. I've rebased (without introducing a merge commit this time) and moved the changelog entries to the right place.

PR #1166, opened 2021-11-08
Static eval backend type

hhugo:


msprotz:
Thanks Hugo. I gave this a go and it looks like the partial evaluation isn't working quite as intended:

```
jonathan@absinthe:/tmp $ cat test.ml 

let main () = if Sys.backend_type = Other "js_of_ocaml" then "yes" else "no"
jonathan@absinthe:/tmp $ cat test.js
(function(g){"use strict";var
a=g.jsoo_runtime,b=a.caml_string_of_jsbytes,c=[0,b("js_of_ocaml")],e=b("yes"),f=b("no"),d=a.caml_get_global_data().Stdlib__sys;a.caml_register_global(4,[0,function(b){return a.caml_equal(d[5],c)?e:f}],"Test");return}(function(){return this}()));
```

This is with an OPAM pin of this pull request. I would expect the equality test to be skipped and the `no` case to disappear entirely from the generated JS.

Thanks!

hhugo:
It will not work with separate compilation. Try compiling with `dune build --profile release`. When using separate compilation, we don't leak any cross module information

msprotz:
Here's what I have in my source:

```ocaml
  open Hacl_star

  (* Stub for JS *)
  external whacl_ed25519_secret_to_public: bytes -> bytes = "whacl_ed25519_secret_to_public"

  let ed25519_secret_to_public ~sk =
    if Sys.backend_type = Other "js_of_ocaml" then
      whacl_ed25519_secret_to_public sk
    else
      Hacl.Ed25519.secret_to_public ~sk
```

a `git grep` confirms that there are no other places that call into the `Hacl_star` package.

Here's what I'm seeing with `--pretty`:

```
193662               _cck_=
193663                caml_equal(backend_type,_bIk_)
193664                 ?whacl_sha2_256_hash(b)
193665                 :caml_call1(_bty_,b);
```

with:

```
 29127      _bIk_=[0,caml_string_of_jsbytes("js_of_ocaml")],
```

and

```
 30308      backend_type=caml_sys_const_backend_type(0),
```

am I doing something wrong? this is with `dune build --profile=release`

msprotz:
just to confirm I have pinned the right branch:

```
jonathan@absinthe:~/Code/mls-proposals (protz_perf) $ opam pin list | grep js_of_ocaml
js_of_ocaml.3.11.0    git  git+ssh://git@github.com/ocsigen/js_of_ocaml#static-eval-backend-type
```

msprotz:
and here is the relevant section from my `dune` file:

```
 (js_of_ocaml
   (flags "--disable" "genprim" "--opt" "3" "--set" "tc_depth=0" "--pretty")
   (javascript_files "BigInteger.js" "zarith.js" "overrides.js"))
```

hhugo:
I don't think we optimize polymorphic comparison function. Try pattern matching instead.

msprotz:
Source:
```
  let sha2_256_hash b =
    match Sys.backend_type with
    | Other "js_of_ocaml" ->
        whacl_sha2_256_hash b
    | _ ->
        Hacl.SHA2_256.hash b
```

JS:
```
193661              if
193662               (typeof backend_type
193663                !==
193664                "number"
193665                &&
193666                !
193667                caml_string_notequal(backend_type[1],_bIk_))
193668               {var _cck_=whacl_sha2_256_hash(b);switch$0 = 1}
193669              if(! switch$0)var _cck_=caml_call1(_bty_,b);
```

Note that the `caml_call1` is still here, meaning that the codepath that calls into `ctypes` was not eliminated...

Thanks,

Jonathan

hhugo:
I've change the tests to reflect more accurately reality.
I've just noticed you're not using the right version of the jsoo compiler. You've pinned js_of_ocaml but you should have pinned js_of_ocaml-compiler.

On a related note, it seems we have similar interest of using hacl_star in javascript. Do you know about https://gitlab.com/tezos/tezos/-/blob/master/src/lib_hacl_glue/js/src/hacl.ml ? I'd be interested to talk about your use case.

msprotz:
Ah that would explain why I had trouble, then. I'll try again with the right pin.

My use-case is essentially "factor out" the crypto in my application and allow multiple implementations to be used in a JS context, such as:
- HACL/WASM via the formalized compilation toolchain, cf https://hacl-star.github.io/Obtaining.html#javascript
- native WebCrypto if running on the web *and* ciphersuites are available
- native system crypto if running in e.g. electron

So, I'm trying to "abstract out" the crypto at a high level.

I took at look at your bindings and it looks like the purpose is to mimic the `Hacl_star` namespace, but redirecting the calls to JS implementations. Is this using the "virtual library" feature that François Bobot mentioned on the other github issue? If so, this could be a pretty good way to "solve" my problem, and I would actually be quite supportive of integrating this into the official hacl-star repository alongside the regular OCaml bindings, rather than keeping it in the Tezos repository. That way, both could be kept in sync and the hacl-star package could even declare two "virtual libraries" with the "normal" implementation as default and your hacl-js stubs as an alternative.

What do you think?

msprotz:
Thanks, I can confirm that with the new pin on js_of_ocaml-compiler, the calls to ctypes bindings are successfully eliminated from the generated JS in release mode. Unfortunately,
- I still get a lot of warnings about missing primitives, it looks like merely linking in ctypes generates a lot of static initializers that keep these definitions reachable
- perhaps as a consequence, I get an exception thrown at JS load-time:
Screen Shot 2021-11-10 at 11 41 46 AM

So, maybe virtual libraries is my only option here. Any thoughts?

hhugo:
I'd like to try compile ctypes to jsoo and see where that lead. see https://github.com/project-everest/hacl-star/pull/494

PR #1170, opened 2021-11-10
Add Style Declaration Property Accessors

TyOverby:
These APIs allow direct access to the declarations programatically.  They are especially useful for setting [css custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).

- [CSSStyleDeclaration.setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty)
- [CSSStyleDeclaration.getPropertyValue](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue)
- [CSSStyleDeclaration.getPropertyPriority](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyPriority)
- [CSSStyleDeclaration.removeProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # dev (2021-??-??) - ??
 
+## Features/Changes
+* Lib: add CSSStyleDeclaration.{setProperty, getPropertyValue, getPropertyPriority, removeProperty}
+
 ## Bug fixes
 * Compiler: fix sourcemap warning for empty cma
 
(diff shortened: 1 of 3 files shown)


smorimoto:
Just needs a CHANGELOG entry

smorimoto:
By the way, every time I wonder how good it is to force us to just copy and paste the spec.
I think it's a better idea to auto-generate these based on something like WenIDL.

TyOverby:
@smorimoto Changelog updated!

As for generating the API from webidl, a coworker of mine @pmwhite is doing exactly that!

smorimoto:
That's great! Generating it from WebIDL or TypeScript resources is always a good thing!
Perhaps another easy way is to reuse the ts2ocaml library and generate it. In fact, we created a standard library based on it.
So either way, it should work well!

PR #1172, opened 2021-11-16
Strengthen bound checks for switch indexes

xclerc:
This pull request tweaks the bounds checks for
switch indexes in `compiler/lib/eval.ml` in order
to handle negative indexes.

While the change seems sound and in line with
the comment about dead code, it is not entirely
clear to me whether this constitutes a proper fix
or is rather masking a deeper problem.

(I apologize for mistakenly pushing to `master`.)

### DIFF
--- a/compiler/lib/eval.ml
+++ b/compiler/lib/eval.ml
@@ -311,8 +311,8 @@ let eval_branch info = function
          Meanwhile, add guards to prevent Invalid_argument("index out of bounds")
          see https://github.com/ocsigen/js_of_ocaml/issues/485 *)
       match the_case_of info (Pv x) with
-      | CConst j when j < Array.length const -> Branch const.(j)
-      | CTag j when j < Array.length tags -> Branch tags.(j)
+      | CConst j when j >= 0 && j < Array.length const -> Branch const.(j)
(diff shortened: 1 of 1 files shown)


hhugo:
LGTM.
I suppose you saw such 
Do you have any idea where this code

hhugo:
I've removed your commits from master.
Can you rebase and keep your last commit only ?
Can you add a test (probably in compiler/tests-compiler/static_eval.ml)

smorimoto:
I removed unnecessary commits. You just need to add a test.

xclerc:
(Sorry for the delayed reply; I will try to reduce the build
failure we were seeing into a reproduction case.)

hhugo:
gentle ping

xclerc:
Sorry, I forgot to report back: I can reproduce the problem,
but failed to minimize it. I will git it another try.

hhugo:
I would have preferred to have a small repro but it's unclear when / if we'll get it. The change is obviously good so I decided to merge now

xclerc:
(Sorry, I did not have the time to work on it.
I will post a comment here if I can produce
a reproduction case.)

PR #1178, opened 2021-11-21
Compiler: optimize cond more

hhugo:


smorimoto:
It's probably better to merge #1177 and then merge this, right?

smorimoto:
Just needs a CHANGELOG entry!

PR #1179, opened 2021-11-21
Use tarides/changelog-check-action in changelog.yml

smorimoto:


hhugo:
What's the difference between the old and new implementation?

smorimoto:
There's no difference, but the only advantage is that it's better to just keep up in the repository than to keep up in each repository with changes from GitHub.

PR #1180, opened 2021-11-21
Regular cleanup

smorimoto:


smorimoto:
The build field updates in the opam file are from dune 2.9, which is the default.

dune-workspace.dev
```diff
@@ -1,16 +0,0 @@
-(lang dune 1.11)
```
hhugo, line 1: I would like to keep that file

dune
```diff
@@ -1,10 +1,3 @@
-(vendored_dirs vendor)
-
-(env
```
hhugo, line 3: I would like to keep this field

dune-workspace.dev
smorimoto: I kept it, and updated it.

dune
smorimoto: I restored just it.

dune-project
```diff
@@ -1,14 +1,13 @@
-(lang dune 2.8)
+(lang dune 2.9)
 (name js_of_ocaml)
 
 (generate_opam_files true)
 
 (authors "Ocsigen team <dev@ocsigen.org>")
 (maintainers "Ocsigen team <dev@ocsigen.org>")
 (source (github ocsigen/js_of_ocaml))
-(bug_reports "https://github.com/ocsigen/js_of_ocaml/issues")
```
hhugo, line 9: Why is bug_reports gone ?

smorimoto: If github is specified in the source stanza, it's generated automatically.

.github/workflows/build.yml
```diff
@@ -61,8 +61,6 @@ jobs:
           opam-depext: ${{ !matrix.skip-test }}
           opam-depext-flags: --with-test
 
-      - run: opam pin add ppx_expect https://github.com/hhugo/ppx_expect.git#v0.14.1+winfix --no-action
```
smorimoto, line 64: Now that 0.14.2 has been released, we can safely remove this line!

smorimoto:
I think this is now ready for merging.

hhugo (approved):
LGTM

PR #1186, opened 2021-11-27
Simplify and improve the build instructions for Menhir parsers.

fpottier:
These changes have the effect of activating `--infer`.
This is strongly recommended (in general).
Furthermore, in the specific case of `annot_parser`, activating `--infer` works around a bug in OCaml 4.07-4.10, which causes a failure during the compilation of `js_of_caml-compiler`.
Thus, I would recommend merging at least the first two commits in this PR.

Best regards,
François.


compiler/lib/dune
```diff
@@ -19,43 +19,32 @@
...
-   TUnknown)))
+   TUnknown
+  )
+)
 
 (rule
  (alias runtest)
  (deps js_parser.conflicts.reference js_parser.conflicts)
  (action
   (diff js_parser.conflicts.reference js_parser.conflicts)))
```
smorimoto, line 42: Perhaps we can just remove this one as well?

smorimoto (approved):

smorimoto:
Could you please confirm this PR?

smorimoto:
@hhugo Could you please confirm this PR?

hhugo:
Thank you for your contribution

Issue #1184, opened 2021-11-26
`json cannot be derived for mutable records`


rluong89:
I got this message when I tried to do a [@@deriving json] on a type that contained a mutable field. I wonder why I have this error.

PR #1187, opened 2021-11-29
PPX: removing the error concerning mutable fields

rluong89:
I discussed with @vouillon and he doesn't see why there should be this error. 
fix https://github.com/ocsigen/js_of_ocaml/issues/1184


hhugo:
LGTM.
Can you update the Changelog and add a few tests in ppx/ppx_deriving_json/tests/ppx.mlt ?

rluong89:
Done.

hhugo:
Thanks, will merge after CI runs

PR #1193, opened 2021-12-18
Compiler: use globalThis, drop joo_global_object

hhugo:
fix https://github.com/ocsigen/js_of_ocaml/issues/699

- [x] implement compatibility for the old joo_global_object

### DIFF
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -325,7 +325,7 @@ let output formatter ~standalone ~custom_header ?source_map () js =
   Js_output.program formatter ?source_map js;
   if times () then Format.eprintf "  write: %a@." Timer.print t
 
-let pack ~global { Linker.runtime_code = js; always_required_codes } =
+let pack ~global ~standalone { Linker.runtime_code = js; always_required_codes } =
   let module J = Javascript in
   let t = Timer.make () in
   if times () then Format.eprintf "Start Optimizing js...@.";
@@ -363,20 +363,7 @@ let pack ~global { Linker.runtime_code = js; always_required_codes } =
       | `Function -> f
       | `Bind_to _ -> f
       | `Custom name -> J.ECall (f, [ J.EVar (J.ident name), `Not_spread ], J.N)
-      | `Auto ->
-          let global =
-            J.ECall
-              ( J.EFun
-                  ( None
-                  , []
-                  , [ ( J.Statement (J.Return_statement (Some (J.EVar (J.ident "this"))))
-                      , J.N )
-                    ]
-                  , J.N )
-              , []
-              , J.N )
-          in
-          J.ECall (f, [ global, `Not_spread ], J.N)
+      | `globalThis -> J.ECall (f, [ J.EVar (J.ident "globalThis"), `Not_spread ], J.N)
     in
     match global with
     | `Bind_to name ->
@@ -397,6 +384,34 @@ let pack ~global { Linker.runtime_code = js; always_required_codes } =
   in
   let runtime_js = wrap_in_iifa ~can_use_strict:true js in
   let js = List.flatten always_required_js @ runtime_js in
+  let js = match global, standalone with
+    | (`Function | `Bind_to _ | `Custom _), _  -> js
+    | `globalThis, false -> js
+    | `globalThis, true ->
+      let s =
+        {|
+(function (Object) {
+  typeof globalThis !== 'object' && (
+    this ?
+      get() :
+      (Object.defineProperty(Object.prototype, '_T_', {
+        configurable: true,
+        get: get
+      }), _T_)
+  );
+  function get() {
+    var global = this || self;
+    global.globalThis = global;
+    delete Object.prototype._T_;
+  }
+}(Object));
+|}
+      in
+      let lex = Lexing.from_string s in
+      let lex = Parse_js.Lexer.of_lexbuf lex in
+      let e = Parse_js.parse lex in
+      e @ js
+  in
   (* post pack optim *)
   let t3 = Timer.make () in
   let js = (new Js_traverse.simpl)#program js in
@@ -428,7 +443,7 @@ type profile = Code.program -> Code.program
 
 let f
     ?(standalone = true)
-    ?(global = `Auto)
+    ?(global = `globalThis)
     ?(profile = o1)
     ?(dynlink = false)
     ?(linkall = false)
@@ -450,7 +465,7 @@ let f
   let emit =
     generate d ~exported_runtime
     +> link ~standalone ~linkall ~export_runtime:dynlink
-    +> pack ~global
+    +> pack ~global ~standalone
     +> coloring
     +> check_js
     +> output formatter ~standalone ~custom_header ?source_map ()
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -363,7 +363,7 @@ let pack ~global ~standalone { Linker.runtime_code = js; always_required_codes }
       | `Function -> f
       | `Bind_to _ -> f
       | `Custom name -> J.ECall (f, [ J.EVar (J.ident name), `Not_spread ], J.N)
-      | `globalThis -> J.ECall (f, [ J.EVar (J.ident "globalThis"), `Not_spread ], J.N)
+      | `globalThis -> J.ECall (f, [ J.EVar (J.ident global_object), `Not_spread ], J.N)
     in
     match global with
     | `Bind_to name ->
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -355,6 +355,20 @@ let pack ~global ~standalone { Linker.runtime_code = js; always_required_codes }
     else js
   in
   let wrap_in_iifa ~can_use_strict js =
+    let js =
+      let o = new Js_traverse.free in
+      let js = o#program js in
+      if StringSet.mem Constant.old_global_object o#get_free_name
+      then
+        ( J.Statement
+            (J.Variable_statement
+               [ ( J.ident Constant.old_global_object
+                 , Some (J.EVar (J.ident global_object), J.N) )
+               ])
+        , J.N )
+        :: js
+      else js
+    in
     let f =
       J.EFun (None, [ J.ident global_object ], use_strict js ~can_use_strict, J.U)
     in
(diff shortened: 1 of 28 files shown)


hhugo:
cc @copy, @jchavarri, @TyOverby 
Anyone interested reviewing and testing this ?

TyOverby:
I can try to test this on the Jane Street codebase next week

hhugo:
Tests pass on node 10 (which doesn't have support for globalThis) and node 14 and 16.
I was able to run test referencing the old joo_global_object.
 
@TyOverby, I'd like to make a new release soon including this change. Please let me know once you've run your testing

hhugo:
It also fixed the issue reported by @jchavarri in https://github.com/ocsigen/js_of_ocaml/issues/699#issuecomment-991800569

compiler/lib/linker.ml
```diff
@@ -296,6 +296,16 @@ let check_primitive ~name pi ~code ~requires =
   let freename = StringSet.diff freename Reserved.keyword in
   let freename = StringSet.diff freename Reserved.provided in
   let freename = StringSet.remove Constant.global_object freename in
+  if StringSet.mem Constant.old_global_object freename && false
+     (* Don't warn yet, we want to give a transition period where both
+        "globalThis" and "joo_global_object" are allowed without extra
+        noise *)
+  then
+    warn
+      "warning: %s: 'joo_global_object' is being deprecated, please sue `globalThis` \
```
jchavarri, line 305:
```suggestion
      "warning: %s: 'joo_global_object' is being deprecated, please use `globalThis` \
```

compiler/lib/linker.ml
```diff
@@ -296,6 +296,16 @@ let check_primitive ~name pi ~code ~requires =
   let freename = StringSet.diff freename Reserved.keyword in
   let freename = StringSet.diff freename Reserved.provided in
   let freename = StringSet.remove Constant.global_object freename in
+  if StringSet.mem Constant.old_global_object freename && false
+     (* Don't warn yet, we want to give a transition period where both
```
jchavarri, line 300: This is awesome.

compiler/lib/driver.ml
```diff
@@ -325,11 +325,25 @@ let output formatter ~standalone ~custom_header ?source_map () js =
...
   if times () then Format.eprintf "Start Optimizing js...@.";
   (* pre pack optim *)
+  let js =
+    let o = new Js_traverse.free in
+    let js = o#program js in
+    if StringSet.mem Constant.old_global_object o#get_free_name
+    then
+      ( J.Statement
+          (J.Variable_statement
+             [ ( J.ident Constant.old_global_object
```
jchavarri, line 340: Not sure i'm reading this right, does this mean that existing usages of `joo_global_object` will compile down to `globalThis`?

I am thinking about cases where code would break after updating the version, due to the environment not supporting `globalThis` for some reason. If these assumptions are correct, would this patch be released under major version then?

compiler/lib/driver.ml
```diff
@@ -397,6 +398,35 @@ let pack ~global { Linker.runtime_code = js; always_required_codes } =
...
+    this ?
+      get() :
+      (Object.defineProperty(Object.prototype, '_T_', {
+        configurable: true,
+        get: get
+      }), _T_)
+  );
+  function get() {
+    var global = this || self;
+    global.globalThis = global;
```
jchavarri, line 419: Ah, I get this answers my question above :) no need to release major version as backwards compat is guaranteed.

compiler/lib/driver.ml
```diff
@@ -397,6 +398,35 @@ let pack ~global { Linker.runtime_code = js; always_required_codes } =
   in
   let runtime_js = wrap_in_iifa ~can_use_strict:true js in
   let js = List.flatten always_required_js @ runtime_js in
+  let js =
+    match global, standalone with
+    | (`Function | `Bind_to _ | `Custom _), _ -> js
+    | `globalThis, false -> js
```
jchavarri, line 404: What is `standalone` and why there's no need to stub `globalThis` when it's `false`?

jchavarri (approved):
Thank you!

I have more questions than anything else, but otherwise looks great to me.

compiler/lib/driver.ml
hhugo: when emitting the js, if there are any references to "joo_global_object", we add a `var joo_global_object = globalThis`.
The code is then wrapped with
```js
(function(globalThis){
  var joo_global_object = globalThis;
  rest_of_the_code
})
(globalThis)`

File: compiler/lib/driver.ml (hunk 4 above)
Comment (hhugo) [reply compiler/lib/driver.ml:419]: indeed

File: compiler/lib/driver.ml (hunk 5 above)
Comment (hhugo) [reply compiler/lib/driver.ml:404]: Standalone is false when building individual cmo/cma to javascript during separate compilaton. The globalThis compatibility layer will already be provided by the jsoo runtime loaded first.

**Comment by hhugo** [comment]
> Thank you!
> 
> I have more questions than anything else, but otherwise looks great to me.

Thanks for the review, can you close threads once you're happy with the answer

**Comment by jchavarri** [comment]
@hhugo I don't think I can resolve conversations as I didn't open the pull request not have write access to the repo.

**Comment by TyOverby** [comment]
@hhugo I got this branch building internally, but we have so much javascript that uses "joo_global_object" that it'll take a long time for me to make the necessary changes in order to tell if there are any other issues.

Also, I must admit that this feature confuses me a bit; I would have expected that you'd use globalThis to generate wrapping-functions that look like this:

```js
(function (joo_global_object) {
   // ... code goes here ...
})(globalThis)
```
Where we still keep `joo_global_object`, but the way that the object is acquired is by looking at `globalThis`.

Instead, it appears that the function-parameter name is also changed, which requires all of the rest of the changes in this feature.

TyOverby:
Also, what does "implement compatibility for the old joo_global_object" refer to?

TyOverby:
Here's how I assumed that the feature would be implemented: #1195

hhugo:
@TyOverby , it is better to read this PR commit by commit.

The first commit just changes how we compute the global object (using globalThis), similar to your #1195. 

The second commit cleans up the old joo_global_object name, renaming it to globalThis. We don't need a Jsoo-specific name, just use what the JS ecosystem uses.

The third commit make the stubs using the old joo_global_object work again (by introducing a`var joo_global_object = globalThis` statement if needed).

hhugo:
> but we have so much javascript that uses "joo_global_object" that it'll take a long time for me to make the necessary changes in order to tell if there are any other issues.

I don't understand. You don't need to change anything, it should just work out of the box

TyOverby:
> I don't understand. You don't need to change anything, it should just work out of the box

Sadly something appears to be broken as every test in the tree fails with the message "undefined variable joo_global_object"

TyOverby:
> The second commit cleans up the old joo_global_object name, renaming it to globalThis. We don't need a Jsoo-specific name, just use what the JS ecosystem uses.

What is the benefit of this approach?

TyOverby:
>> I don't understand. You don't need to change anything, it should just work out of the box
>
>Sadly something appears to be broken as every test in the tree fails with the message "undefined variable joo_global_object"

I won't be able to investigate this further for a few days due to the holidays, but I'll dig deeper on Monday

TyOverby:
the failure happens here:

https://github.com/janestreet/core/blob/master/core/src/strftime.js#L43

```
if(joo_global_object)joo_global_object.strftime = adaptedStrftime}
   ^
ReferenceError: joo_global_object is not defined
```

TyOverby:
Fixing that instance just causes the same error to happen in other places that we use `joo_global_object`.  Is there a flag or something that we need to turn on in order to get the "it just works" shim?

TyOverby:
> introducing avar joo_global_object = globalThis statement if needed

I can confirm that _some_ sections get this added to the top of them, but not all.  Of note, at least two sections use `joo_global_object` but don't get the shim.

hhugo:
> > introducing avar joo_global_object = globalThis statement if needed
> 
> I can confirm that _some_ sections get this added to the top of them, but not all. Of note, at least two sections use `joo_global_object` but don't get the shim.

Thanks for testing, I have a theory that I should be able to test/fix early next year

hhugo:
@TyOverby, I think I've fixed your issues in my last commit.
`core_kernel/src/runtime.js` and `virtual_dom/src/hooks.js` are using joo_global_object but were not getting the shim.

(Also note that core_kernel declare `strftime.js` but doesn't seem to use it)

TyOverby:
> @TyOverby, I think I've fixed your issues in my last commit.

Yep!  I can confirm that it works now!

I don't know if it's directly caused by this pull-request, but now I'm seeing some new, strange behavior.  One of our files is being split between IIFEs like so:

**runtime.js** 
```js
var x = (function () { ...})();
if (typeof window !== undefined) window.x = x;
if (typeof globalThis !== undefined) globalThis.x = x;
```

When run through Js_of_ocaml, we get 

**output.bc.js** 
```js
(function(globalThis) {
    var x = (function () { ...})();
}(globalThis));

(function(globalThis) {
    if (typeof window !== undefined) window.x = x;
    if (typeof globalThis !== undefined) globalThis.x = x;
}(globalThis));
```

Obviously, by moving the definition of `x` out of scope from the place that uses it, the 2nd IIFE will crash.

TyOverby:
I tried building a test to demonstrate this, but the unit tests no longer build due to an assumption that they're being run using dune

TyOverby:
I spent a few hours trying to get a minimal repro but could not.
I'll be on vacation for the next two weeks, but would be happy to continue my investigation when I get back.

TyOverby:
It looks like I'm able to work around that specific issue by wrapping the whole file in another IIFE.

hhugo:
@TyOverby, I've managed to reproduce your issue with the following
```javascript
var x = (function () { ...})();
//Always .. a comment starting by //Always
if (typeof window !== 'undefined') window.x = x;
if (typeof globalThis !== 'undefined') globalThis.x = x;
```
The behavior was introduced by https://github.com/ocsigen/js_of_ocaml/commit/c6d86086adf5a33e7b59f21ea65d6aeae0a89f22
and should now be fixed on master by https://github.com/ocsigen/js_of_ocaml/commit/3e7f728811ae603883baaaba39a22a92dc551102

I'll rebase this PR to make it easier to test for you

hhugo:
@TyOverby, I'm going to merge now, please make sure master works for you when you return from vacation.

TyOverby:
@hhugo all of our automated tests are passing; we're going to do some more manual testing next week.

Thanks for all the help!

PR #1195, opened 2021-12-24
global_this alternative implementation

### LINKED ISSUE #699: joo_global_object is undefined when bundled in strict mode

TyOverby:
fix #699 by passing globalThis to the IIFE


hhugo:
We need to ship a polyfill for globalThis. Support in various engine is relatively recent

hhugo:
Closing in favor of #1193

PR #1198, opened 2021-12-31
 Accumulator Passing Style specialize_instrs

philzook58:
I have an example in bap that ran into a stack overflow during dynamic linking. 

https://gist.github.com/philzook58/2bbec1b4cb58923b88325df37ccc3499

By finding what I think is an appropriate javascript stack trace as described in the gist, there were hundreds of `specialize_instrs` calls on top of the stack. Upon examining this function, it is indeed not tail recursive. Switching to accumulator passing style fixes the problem and now the example stack overflows at another point unrelated to dynamic linking, instead involving a CPS monad inside bap itself. This other problem can be fixed by reverting to bap 2.3.0, which did things differently, or perhaps by patching bap's monad to use a trampoline.

I also removed the `rem` term from the call to `specialize_instr` which perhaps was unwise to do in the same pull request. It is used homogenously in the function and this function is called from only one location, so it seemed like confusing bloat. I can remove this change if you like.

### DIFF
--- a/compiler/lib/specialize_js.ml
+++ b/compiler/lib/specialize_js.ml
@@ -22,55 +22,49 @@ open! Stdlib
 open Code
 open Flow
 
-let specialize_instr info i rem =
+let specialize_instr info i =
   match i with
-  | Let (x, Prim (Extern "caml_format_int", [ y; z ])) ->
-      (match the_string_of info y with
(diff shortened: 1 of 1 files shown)


hhugo:
Thanks
I've extended your PR to also make `f_once` tail recursive.

PR #1205, opened 2022-01-05
Runtime: clean jsoo_static_env

hhugo:


hhugo:
@TyOverby, could you check in what context `jsoo_static_env` is used in the Janestreet codebase ?

hhugo:
ping @TyOverby

TyOverby:
I'll check into this on Monday

TyOverby:
The only usage of jsoo_static_env (outside of our vendored copy jsoo) is located here: https://github.com/janestreet/timezone/blob/master/src/runtime.js

PR #1219, opened 2022-01-11
Introduce js_of_ocaml.runtime-lib to improve compatibility with gen_js_api and brr

### LINKED ISSUE #1146: [FEATURE REQUEST] Small runtime library to improve binding libraries interoperability
Over the time two other FFI alternatives emerged out of the `js_of_ocaml` project itself to provide alternative binding strategies to JavaScript (based on abstract data types rather than phantom typing of JavaScript objects). 

These projects do not make use of the `js_of_ocaml.cma` library. All the projects are perfectly able to interoperate together except for one thing, the exception registered to represent the JavaScript exception. For reference here is where this registration happens: 

* [`Brr`](https://github.com/dbuenzli/brr/blob/41580885f40bfd184c3d8e5be2ddd56b0712b411/src/jv.ml#L227-L231)
* [`js_gen_api`](https://github.com/LexiFi/gen_js_api/blob/377d58632d4b34aaf54a3e2f3c80741cc1ab2044/lib/ojs_exn.ml)
* [`js_of_ocaml`](https://github.com/ocsigen/js_of_ocaml/blob/cde86b2cba215344a6f81b1db28b922b80c7b55a/lib/js_of_ocaml/js.ml#L672)

As discussed in https://github.com/dbuenzli/brr/issues/2 a simple answer to this problem would be to provide a simple runtime OCaml library seperate from `js_of_ocaml.cma` against which `js_of_ocaml` compiled programs must link; this would also be a good place to add the C primitive stubs so that the byte code can be linked without `-no-check-prims` in both `js_gen_api` and `brr` (these primitives are currently in the `js_of_ocaml.cma` library). 

This OCaml runtime library should only have this module:
```ocaml
module Js_error : sig
  type t 
  exception Exn of t
end = struct
   type t (* Voluntarily left abstract, this is a javascript object. *)
   exception Exn of t 
   let () = Callback.register_exception "jsError" (Exn (obj [||]))
end
```
Each of the API can then simply expose its own exception implemented as being equal to `Js_error.Exn`. For example the [current Brr API](https://erratique.ch/software/brr/doc/Jv/index.html#exns) could just change its signature to: 

```ocaml
module Error : sig
   type t = Js_error.t (* <--- this is the only visible external change *)
   val name : t -> Jstr.t
   ...
end
exception Error of Error.t
```
implemented as: 
```ocaml
module Error = struct
   type t = Js_error.t
   let name e = to_jstr (get (Obj.magic e) "name")
   ...
end
exception Error = Js_error.Exn
```

/cc @mlasson

hhugo:
Fix #1146 

The first commit implement the solution suggested by @dbuenzli  in #1146
The next two commits restore the previous api but keep the extra library `js_of_ocaml.runtime-lib`.

I believe that `js_of_ocaml.runtime-lib` could be used by both brr and gen_js_api.
```ocaml
module Error : sig
  type t

  exception Exn of t
end = struct
  type nonrec t = Js_of_ocaml_runtime.Js.error Js_of_ocaml_runtime.Js.t

  exception Exn = Js_of_ocaml_runtime.Js.Exn
end
```

@dbuenzli, what do you think ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 * Lib: add PerformanceObserver API (#1164)
 * Lib: add CSSStyleDeclaration.{setProperty, getPropertyValue, getPropertyPriority, removeProperty} (#1170)
 * Lib: make window.{inner,outer}{Width,Height} non-optional
+* Lib: Introduce js_of_ocaml.runtime-lib to improve compatibility with gen_js_api and brr
 * PPX: json can now be derived for mutable records (#1184)
 * Runtime: use crypto.getRandomValues when available (#1209)
 
--- a/compiler/bin-js_of_ocaml/dune
+++ b/compiler/bin-js_of_ocaml/dune
@@ -7,7 +7,7 @@
   js_of_ocaml-compiler
   cmdliner
   compiler-libs.common
-  js_of_ocaml-compiler.runtime
+  js_of_ocaml-compiler.builtin-runtime
   (select
    findlib_support.ml
    from
--- a/compiler/lib-runtime/dune
+++ b/compiler/lib-runtime/dune
@@ -1,7 +1,7 @@
 (library
  (name jsoo_runtime)
  (libraries js_of_ocaml-compiler)
- (public_name js_of_ocaml-compiler.runtime))
+ (public_name js_of_ocaml-compiler.builtin-runtime))
 
 (rule
  (target files.ml)
--- a/lib/gen_stubs/gen_stubs.ml
+++ b/lib/gen_stubs/gen_stubs.ml
@@ -29,8 +29,9 @@ let print_stub s =
 
 let () =
   let mls = ref [] in
+  let ignores = ref [] in
   Arg.parse
-    []
+    [ "--ignore", String (fun x -> ignores := x :: !ignores), "" ]
     (fun ml -> if not (Filename.check_suffix ml ".pp.ml") then mls := ml :: !mls)
     "generate dummy js stubs";
   let externals = ref String_set.empty in
@@ -39,6 +40,7 @@ let () =
     externals := List.fold_right String_set.add l !externals;
     desc
   in
+  let ignores = List.fold_right String_set.add !ignores String_set.empty in
   let mapper = { Ast_mapper.default_mapper with value_description } in
   List.iter
     (fun ml ->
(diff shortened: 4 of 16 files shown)


hhugo:
Most of the stubs are now part of the new library.

hhugo:
@dbuenzli, gentle ping to comment on this. Are you happy with this new library inside the js_of_ocaml opam package or would you prefer a new opam package ?

dbuenzli:
Sorry it seems I missed the ping. Will try to have a look later this evening.

dbuenzli:
Personally I would have preferred if only the minimal had been added to the library and in particular,  `Js_of_ocaml`'s encoding of JavaScript values had been left outside. 

I don't think a separate opam package is needed however why call that library `js_of_ocaml.runtime-lib` ? It's a bit redundant, `js_of_ocaml.runtime` would do.

dbuenzli:
The more I think about it, the more I would prefer if we could stick to the first commit and simply add the C stubs to the library – so that we can drop the `-no-check-prims` from our compilation instructions.

This makes it more clear what we effectively share and is less confusing for users following type definitions from `brr` which would then be subject to all the definitions of `js_of_ocaml`'s module `Js` whose existence `brr` actively tries to hide :-)
It also makes sure nothing that `gen_js_api` or `brr` would not like to have accidentally creeps in this library in the future.

smorimoto:
I think it would be better to include these 3 people in the discussion @alainfrisch @mlasson @cannorin

mlasson:
That's neat; both handling the "Js_error.Exn" situation and exposing the whole runtime library !
Thanks.

hhugo:
Making the runtime lib part of the js_of_ocaml-compiler package is a good idea.

I don't want to simply break the API for `Js.Error`, I'm planning to merge #1224 first and revisit this PR after the next release of jsoo.

hhugo:
I was able to trick the typesystem with the following piece of code
```ocaml
module Magic = struct
  module type T = sig
    exception Error of error t
  end

  type ('a, 'b) eq = Eq : ('a, 'a) eq

  let (eq : (error t, Js_error.t) eq) = Obj.magic Eq

  let m =
    match eq with
    | Eq ->
        (module struct
          exception Error = Js_error.Exn
        end : T)

  module Error = (val m : T)
end

include Magic.Error
```
allowing me to preserve the current api  see #1224

Issue #1185, opened 2021-11-27
Release build performs considerably worse compared to dev build due to inlining


linoscope:
I have been using `js_of_ocaml` to run [my game boy emulator](https://github.com/linoscope/CAMLBOY) in the browser, and one thing I noticed is the release build (built with `dune build --profile=release` where js_of_ocaml compiles without separate compilation) performs considerably worse, in terms of speed, compared to the dev build (built with `dune build` where js_of_ocaml compiles with separate compilation). With some investigation ([thread in discuss.ocaml.org](https://discuss.ocaml.org/t/js-of-ocaml-output-performs-considerably-worse-when-built-with-profile-release-flag/8862/14)), we found that the cause is inlining and that adding `--no-inline` to the release build improves the performance by a lot.

I have measured the FPS of my emulator (without UI) for each of the cases in my local environment (`i7-8665U CPU @ 1.90GHz`, `Ubuntu 20.04.2`, `Google Chrome Version 95.0.4638.69`), and summarized the results in the following chart:
![jsoo_compare](https://user-images.githubusercontent.com/24925588/143675245-0627152d-9596-4b1f-88cb-b9f19e686287.png)

Also, I have hosted the emulator, both with UI and without UI, in the following URLs to demonstrate the FPS difference:

- Dev build: [with UI](https://linoscope.github.io/CAMLBOY/dev?throttled=false), [without UI](https://linoscope.github.io/CAMLBOY/dev/bench.html)
- Release build: [with UI](https://linoscope.github.io/CAMLBOY/release?throttled=false), [without UI](https://linoscope.github.io/CAMLBOY/release/bench.html)
- Release build with `--no-inline`: [with UI](https://linoscope.github.io/CAMLBOY/release-no-inline?throttled=false), [without UI](https://linoscope.github.io/CAMLBOY/release-no-inline/bench.html)

The version of my `js_of_ocaml` and `ocamlc` is as follows:

```sh
$ js_of_ocaml --version
3.11.0
$ ocamlc --version
4.13.1
```

**Questions:**

Is this expected behavior? It is pretty surprising that the release build performs so much worse than the dev build (although this can be due to `dune`'s default config) and that inlining causes such a performance hit.

And are there any improvements we can make to avoid the negative impact of inlining without having to resort to completely disabling it?

linoscope:
[...]

PR #1220, opened 2022-01-11
Compiler: less agressive inlining

hhugo:
fix https://github.com/ocsigen/js_of_ocaml/issues/1185

### DIFF
--- a/compiler/lib/config.ml
+++ b/compiler/lib/config.ml
@@ -118,7 +118,8 @@ module Param = struct
   let switch_max_case =
     p ~name:"switch_size" ~desc:"set the maximum number of case in a switch" (int 60)
 
-  let inline_max_size = p ~name:"inline_limit" ~desc:"<TODO>" (int 500)
+  let inlining_limit =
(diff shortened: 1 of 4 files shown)


compiler/lib/config.ml
```diff
@@ -118,6 +118,9 @@ module Param = struct
   let switch_max_case =
     p ~name:"switch_size" ~desc:"set the maximum number of case in a switch" (int 60)
 
+  let inlining_limit =
+    p ~name:"inlining-limit" ~desc:"set the size limit for inlining" (int 500)
+
```
glennsl, line 123: What unit is this in? Bytes?

hhugo: The unit is not really meaningful. What this feature compute is "number of instructions + number of branching"

smorimoto:
FYI: @linoscope

linoscope:
Nice! I will reintroduce inlining and rerun the benchmark once this change is released.  Thanks, @hhugo for working on this!

PR #1224, opened 2022-01-17
Introduce a new runtime library to improve compatibility with gen_js_api and brr (replace #1219)

### LINKED ISSUE #1146: [FEATURE REQUEST] Small runtime library to improve binding libraries interoperability

hhugo:
replace #1219, fix #1146
- Introduce a new module `Js_of_ocaml.Js_error`
- Deprecate `Js_of_ocaml.Js.Error`


dbuenzli:
I had a look. It looks good to me. 

There's still more than I expected in the runtime library, but if the idea is to have a place that declares the primitives that "we" are allowed to use, that looks fine by me.

PR #1225, opened 2022-01-18
Typo in typed_array.mli

craff:


hhugo:
I don't think this is typo, see https://github.com/ocsigen/js_of_ocaml/commit/82b7440df077d44238042334c140c7ff8b119393

As I understand it, it was written this way so that it's easier to deal with uint32 with the highest bit set and be able to use `4294967294.` (as a float) instead of `-2` (as a signed int).

My hunch is that it would be more intuitive to use int here, or even define a new type `uint32` and provide conversion function to/from int and float

@vouillon, do you want to comment ? you made the change in Feb-2014  :)

hhugo:
I've made a PR https://github.com/ocsigen/js_of_ocaml/pull/1227 to make the api cleaner.

hhugo:
Closing this as it's clearly not a typo. One my want to open new issue or PR to change how we deal with uint32

craff:
On 22-01-18 05:58:44, hhugo wrote:
> Closing this as it's clearly not a typo. One my want to open new issue or PR to
> change how we deal with uint32

Ok, forgat that js_of_ocaml only had 32 bits int ... It makes sense to
use float even if a uint would probably be a good alternative too.

> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.
> You are receiving this because you authored the thread.*Message ID: <ocsigen/
> ***@***.***>
>

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

Issue #1045, opened 2020-07-28
[FEATURE REQUEST] enable with-js-error by default?


Lupus:
Are there any caveats for enabling `with-js-error` optimization by default? According to [this SO answer](https://stackoverflow.com/questions/35306496/performance-hit-of-newing-up-error-object-in-modern-node-js) perf hit for capturing the stack is quite affordable. It's quite hard to debug exceptions without backtraces and dune currently gives no option to enable this flag globally.

Lupus:
Or maybe the behavior could be changed to match that of `OCAMLRUNPARAM=b`, i.e. always call some hook during exception throwing, and have it determine if `new Error()` needs to be created based on some property of environment (if we're running on node.js, we could parse OCAMLRUNPARAM environment variable, not sure what would be the best way to toggle this flag in browser, read some key from local storage?).

hhugo:
xref #1071

PR #1229, opened 2022-01-18
Runtime: use monomorphic representation for MlBytes

hhugo:
Using `deoptigate` to check performances/optimizations

probably blocked by #1045


hhugo:
If we make `Bytes.unsafe_of_string` `O(n)`, we would need to fix `String.sub`,  `String.get_intN_*`, `String.get_uintN_*`, `String.get_utf_*`, `String.is_valid_utf_*`. They convert the whole string argument to bytes but only look at a small subset.

PR #1232, opened 2022-01-21
Return from caml_sys_exit when quit/exit called

phated:
This is a pretty obscure issue, but I'm trying to mock out the node process object in a WebWorker (so we can run the Grain compiler in it).

I can provide `process.exit()`, but I can't close the Worker synchronously. This would be fine if exit was only called on an error because I can throw; however, it is also called with the `0` code upon a successful execution.

By returning after calling `quit` or `exit`, you are leaving it up to the implementer to handle the closing (and it can do async cleanup, etc).

### DIFF
--- a/runtime/sys.js
+++ b/runtime/sys.js
@@ -27,10 +27,10 @@ function caml_raise_sys_error (msg) {
 //Requires: caml_invalid_argument
 function caml_sys_exit (code) {
   var g = globalThis;
-  if(g.quit) g.quit(code);
+  if(g.quit) return g.quit(code);
   //nodejs
   if(g.process && g.process.exit)
-    g.process.exit(code);
+    return g.process.exit(code);
   caml_invalid_argument("Function 'exit' not implemented");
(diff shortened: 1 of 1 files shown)


hhugo:
Have you considered raising an exception that you would catch at toplevel ? It might be a bit fragile if you have catch-all patterns in your codebase though.
```ocaml
exception Exit_with_code of int
```

phated:
Wouldn't we still need to call sys_exit to exit with the correct codes in the native build?

phated:
@hhugo to be clear, this isn't an ocaml issue and I don't think it needs an ocaml fix. This is a JS issue as the `caml_sys_exit` function is calling an exit call for the platform but not trusting that the exit will actually be completed.

phated:
Nevermind, this was definitely a Grain problem: https://github.com/grain-lang/grain/pull/1117 (Someone put `exit(0)` upon successful cmdliner runs).

PR #1239, opened 2022-02-12
DRAFT:  flush output_channel if length over max length

### LINKED ISSUE #1238: [BUG] Memory keeps growing with `Buffer` and js_of_ocaml
Running the following piece of code with `js_of_ocaml` crashes the node environment (it generates a too large string, and memory consumption grows unbounded). Running the same code as a binary directly doesn't have this behavior.

```ocaml
let () =
  let ob = Buffer.create 2096 in
  let oc = open_out "test-file" in
  let seq =
    Seq.unfold
      (fun b ->
        if b < 100 * 100 * 100 * 100 then Some (b, b + 1) else None)
      0
  in
  let () =
    Seq.iter
      (fun _v ->
        Buffer.add_string ob ("output");
        Buffer.output_buffer oc ob;
        Buffer.clear ob)
      seq
  in
  close_out oc
```

It's possible to fix this, by flushing the output channel regularly (for example `if i mod n = 0 then flush oc`).

**Expected behavior**
I expect the js_of_ocaml behavior to be the same as the executable, namely that the memory consumption doesn't grow, even if the output channel is not flushed manually.

Some more information can be found here:
* https://github.com/ocaml-community/yojson/pull/133#issuecomment-1035980421
* https://discuss.ocaml.org/t/memory-keeps-growing-with-buffer-and-js-of-ocaml/9299

**Versions**
* ocaml 4.13.1
* js_of_ocaml 4.0.0

tcoopman:
The code is not ready yet, but I wanted to open the PR already because I have some questions. This PR fixes #1238 

### What should the actual logic be?

* The current implementation is a lot slower if you add small strings with line breaks vs small string without line breaks (4-5 times slower in some quick tests).
* What if we have a string that is `n * max_length`? Do we...
  * add the whole string and flush after that
  * flush `n` times, every time after `max_length`
* If we want to keep the current behavior and flush after line breaks, and we have a string that looks like: `foo\n very long .... string after line break"`, what do we want to do?
  * only flush after `\n` and add the long string without flushing
  * flush both after `\n` and do the same as a string with `n * max_length`
  * ignore the `\n` and do the same as a string with `n * max_length`

My proposal would be to do the simplest thing:

```
  if total_length > max_length : add full string and flush (* even if there are linebreaks in the string *)
  else if we have a line break: flush after the line break
  else just add the string
```

What do you think?

### Should I write a test for this? And how and where should I do it?

I could write some js unit tests, but I don't see any in the code so far, so I'm not sure what to do, just let me know

### DIFF
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -366,6 +366,7 @@ function caml_ml_flush (chanid) {
 //Requires: caml_ml_flush,caml_ml_bytes_length
 //Requires: caml_create_bytes, caml_blit_bytes, caml_raise_sys_error, caml_ml_channels, caml_string_of_bytes
 //Requires: caml_jsbytes_of_string
+var MAX_LENGTH = 65536;
 function caml_ml_output_bytes(chanid,buffer,offset,len) {
   var chan = caml_ml_channels[chanid];
   if(! chan.opened) caml_raise_sys_error("Cannot output to a closed channel");
@@ -380,16 +381,15 @@ function caml_ml_output_bytes(chanid,buffer,offset,len) {
   var jsstring = caml_jsbytes_of_string(string);
   var id = jsstring.lastIndexOf("\n");
   var total_length = jsstring.length + chan.buffer.length;
-  if(id < 0) {
+  if (total_length > MAX_LENGTH) {
     chan.buffer+=jsstring;
-    if (total_length > 65536) {
-      caml_ml_flush (chanid);
-    }
-  }
-  else {
+    caml_ml_flush (chanid);
+  } else if(id >= 0) {
     chan.buffer+=jsstring.substr(0,id+1);
     caml_ml_flush (chanid);
     chan.buffer += jsstring.substr(id+1);
+  } else {
+    chan.buffer+=jsstring;
   }
   return 0;
 }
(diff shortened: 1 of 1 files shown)


runtime/io.js
```diff
@@ -366,6 +366,7 @@ function caml_ml_flush (chanid) {
 //Requires: caml_ml_flush,caml_ml_bytes_length
 //Requires: caml_create_bytes, caml_blit_bytes, caml_raise_sys_error, caml_ml_channels, caml_string_of_bytes
 //Requires: caml_jsbytes_of_string
+var MAX_LENGTH = 65536;
```
tcoopman, line 369: where should I place this? At the top of the file and rename it to `MAX_BUFFER_LENGTH`?

hhugo: It's only going to be used here. Just put it inside the funciton.

hhugo:
> The current implementation is a lot slower if you add small strings with line breaks vs small string without line breaks (4-5 times slower in some quick tests).

The current implementation is only beneficial for stdout and stderr when printing on the console. We could restrict that logic for theses cases only (https://github.com/ocsigen/js_of_ocaml/pull/1241 allows to identify theses cases)


> What if we have a string that is n * max_length? Do we...
add the whole string and flush after that
flush n times, every time after max_length

I'dont have any preference here

hhugo:
Could you please rebase your work on top of https://github.com/ocsigen/js_of_ocaml/pull/1241 ?

hhugo:
Merged as part of #1270

PR #1248, opened 2022-03-06
feat: Add more provided global values

phated:
This adds a bunch of global values that are provided by the web platform. I annotated a couple with the platforms where they are available.

In the future, it'd be cool for these lists to be adjusted based on the `--target-env`

### DIFF
--- a/compiler/lib/reserved.ml
+++ b/compiler/lib/reserved.ml
@@ -140,7 +140,24 @@ let provided =
     ; "isNaN"
     ; "parseFloat"
     ; "parseInt"
-    ; "module"
-    ; "require"
+    ; "module" (* only available in node *)
+    ; "require" (* only available in node *)
     ; "Symbol"
+    ; "ArrayBuffer"
+    ; "Float32Array"
+    ; "Float64Array"
+    ; "Int16Array"
+    ; "Int32Array"
+    ; "Int8Array"
+    ; "TextDecoder"
+    ; "Uint16Array"
+    ; "Uint32Array"
+    ; "Uint8Array"
+    ; "atob"
+    ; "clearInterval"
+    ; "console"
+    ; "global" (* only available in node *)
+    ; "importScripts" (* only available in WebWorker *)
+    ; "performance"
+    ; "setTimeout"
     ]
--- a/compiler/lib/reserved.ml
+++ b/compiler/lib/reserved.ml
@@ -154,6 +154,7 @@ let provided =
     ; "Uint32Array"
     ; "Uint8Array"
     ; "atob"
+    ; "btoa"
     ; "clearInterval"
     ; "console"
     ; "global" (* only available in node *)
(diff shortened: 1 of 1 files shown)


compiler/lib/reserved.ml
```diff
@@ -140,7 +140,24 @@ let provided =
...
+    ; "Int32Array"
+    ; "Int8Array"
+    ; "TextDecoder"
+    ; "Uint16Array"
+    ; "Uint32Array"
+    ; "Uint8Array"
+    ; "atob"
+    ; "clearInterval"
+    ; "console"
+    ; "global" (* only available in node *)
```
hhugo, line 159: Do we really need this ? One can now use the portable `globalthis` to refer to the global object.

hhugo:
> In the future, it'd be cool for these lists to be adjusted based on the --target-env

I don't understand the benefit of such a thing. It would basically mean that some Jsoo stubs would fail/warn if used with the wrong target_env. Is that the goal ?

compiler/lib/reserved.ml
phated: In node, `global === globalThis`, so either can be used. Since `globalThis` wasn't added to node until v10, a lot of older JS that exists would still have `global` being used.

I added this because Binaryen compiled to JS has both `globalThis` and `global` for its feature detection logic.

phated:
> Is that the goal ?

My thinking is that if I write some JSOO bindings that require provided global values from nodejs or the browser, but a consumer sets their `--target-env` to a different platform, they should be warned that the script will fail. Do you think that's a good idea?

compiler/lib/reserved.ml
```diff
@@ -140,7 +140,24 @@ let provided =
...
+    ; "Float32Array"
+    ; "Float64Array"
+    ; "Int16Array"
+    ; "Int32Array"
+    ; "Int8Array"
+    ; "TextDecoder"
+    ; "Uint16Array"
+    ; "Uint32Array"
+    ; "Uint8Array"
+    ; "atob"
```
hhugo, line 156: If we include `atob`, we should also include `btoa`

phated, line 156:
```suggestion
    ; "atob"
    ; "btoa"
```

phated: Good idea! Added!

phated:
@hhugo anything else you'd like on this (other than the changelog entry)?

compiler/lib/reserved.ml
```diff
@@ -140,7 +140,25 @@ let provided =
...
+    ; "btoa"
+    ; "clearInterval"
+    ; "console"
+    ; "global" (* only available in node *)
+    ; "importScripts" (* only available in WebWorker *)
+    ; "performance"
```
hhugo, line 162: This is not available on node. Add a comment ?

hhugo:
LGTM modulo a small comment and a changelog entry.

compiler/lib/reserved.ml
phated:
```suggestion
    ; "performance" (* Not available in Node until node 16+ *)
```
This was actually added to node's global scope in node 16.

Issue #1247, opened 2022-03-06
Optimizer breaks some code


phated:
I'm working on some Binaryen stuff again and I found that the JSOO optimizer breaks some of the code.

In particular, it changes:
```js
"object" == typeof window && window
```
into
```js
typeof"object"==window&&window
```
Which isn't correct and actually fails when the script tries to run in an env without Window (in this specific case).

I had to hunt for while to figure out why `--pretty` didn't fail, but a normal build kept crashing.

phated:
It looks like this is related to `share_constant`. When I disable that optimization, the reorganization doesn't happen.

hhugo:
I think there is a typo in your issue.
You say it translates to 
`typeof"object"==window&&window`. Is this really what you meant to write ?

phated:
Yes, that's the mistake the optimizer produces.

hhugo:
The issue comes from this line https://github.com/ocsigen/js_of_ocaml/blob/49e1aa56de10f82f59f2dae98606d113f233958f/compiler/lib/js_traverse.ml#L204.

phated:
Thanks ❤️

PR #1249, opened 2022-03-06
Compiler: fix rewriter bug

hhugo:
fix https://github.com/ocsigen/js_of_ocaml/issues/1247


smorimoto (approved):
LGTM! Could you please add a change log entry?

PR #1252, opened 2022-03-09
Lib: add missing Intl options

### LINKED ISSUE #1251: Missing Intl.DateTimeFormat options
There's several options missing in `Intl.DateTimeFormat.options`, such as `dateStyle`, `timeStyle` and `calendar`. I also see from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat) that some of these are not supported by IE, and wonder if that's why they're not included. Then again, `timeZoneName` is included despite also not being supported on IE.

My question then is, what is the criteria for which APIs to include and which not to include? Then once I know that I can make a PR to fix this.

glennsl:
Fixes #1251

There may be (probably is) more missing options in the other `Intl` APIs. These two were the only one's I checked. If you want to be thorough I can go through the rest of them later though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 * Misc: remove uchar dep
 * Misc: use 4.14 in the CI
 * Lib: add missing options for Intl.DateTimeFormat
+* Lib: add missing options for Intl.NumberFormat
 
 ## Bug fixes
 * Compiler: fix rewriter bug in share_constant (fix #1247)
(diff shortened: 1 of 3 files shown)


hhugo:
LGTM. Thanks
Can you rebase and fix the confict.

glennsl:
Done.

PR #1257, opened 2022-04-06
Dialog element support

sagotch:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
and
https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element

I think that I remember that `cancel` and `close` events are not well supported, but they are part of the spec.

~THIS PR DEPENDS ON CURRENT TYXML MASTER, which is unreleased right now. Would a new tyxml release be good enough for this to be merged?~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Features/Changes
 * Compiler: only flush the necessary env for closures (#1568)
+* Library: dialog element support
 
 ## Bug fixes
 * Compiler: fix --enable=vardecl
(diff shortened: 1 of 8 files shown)


hhugo:
@sagotch, I've rebased this PR. Is it still blocked on tyxml ?

sagotch:
Yes it is.

Tyxml has not been released since the last available release v4.5.0 (Apr 22, 2021).

I think @Drup used to manage the releases of Tyxml, but I am not sure it is the case anymore.

hhugo:
@drup, when should we expect the next release of tyxml ?

PR #1259, opened 2022-04-11
switch to dune3

hhugo:
We should probably wait for dune.3 to be compatible with esy / non-opam workflow, see  https://github.com/ocaml/dune/issues/5488


phated:
Looking good! We have successful dune 3.2 runs across the board:
https://github.com/grain-lang/libbinaryen/actions/runs/2373688424
https://github.com/grain-lang/libbinaryen/actions/runs/2373688425

PR #1261, opened 2022-04-25
Runtime: fix 1260

### LINKED ISSUE #1260: [BUG] Empty path components in file paths not handled correctly on Linux
I recently was attempting to upgrade some code to use v4.0.0, but ran into this bug. The code was attempting to read a file at `a/b//c/d`, but the file was not found. Changing the code to read `a/b/c/d` instead fixed the program, and I will probably persist this change, but I think this is also a bug in the js_of_ocaml filesystem layer, since Linux ignores empty path components.

I believe the bug was introduced in [PR #1132 in runtime/fs.js](https://github.com/ocsigen/js_of_ocaml/pull/1132/files#diff-a41a3ccec464659b1841ed7e0057ce638942fc5032d77da2ae68bcf906af835eL50). Since I only need to target Linux, I'm temporarily patching fs.js to put that line back. However, I imagine the proper way to fix this bug is to handle both the Linux and Windows cases separately. Is this correct? If so, we should also probably have a tests that only run on Windows and tests that only run on Linux?

hhugo:
fix #1260

### DIFF
--- a/runtime/fs.js
+++ b/runtime/fs.js
@@ -91,6 +91,7 @@ function caml_make_path (name) {
     switch(comp[i]){
     case "..": if(ncomp.length>1) ncomp.pop(); break;
     case ".": break;
+    case "": break;
     default: ncomp.push(comp[i]);break
     }
   }


hhugo:
@phated, Can you confirm this PR doesn't break your code/workflow ?

phated:
@hhugo I *think* that it should be fine for Grain. We are still waiting for a dune fix so we can't upgrade to jsoo4 yet, otherwise I'd test it directly.

hhugo:
You should be able to use it with dune 2.9.

Or just patch the opam file removing `--promote-install-files=false` and the `dune install ...` line.

phated:
I'll give it a shot!

PR #1273, opened 2022-05-15
Lib: wheel event binding

### LINKED ISSUE #1272: [FEATURE REQUEST] `wheel` event bindings
**Is your feature request related to a problem? Please describe.**
`Dom_html` currently has bindings only for deprecated [mousewheel](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event) and [_DOMMouseScroll](https://developer.mozilla.org/en-US/docs/Web/API/Element/DOMMouseScroll_event) events, but now [wheel](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) event is supported by all modern browsers.

**Describe the solution you'd like**
~~I suggest to define `wheelEvent` object type inherited from already defined `mousewheelEvent`. It is possible because legacy props from `mousewheelEvent` is supported in [WheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent).~~ I suggest to extend `mousewheelEvent` with other props from [WheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent). In a word, I made PR #1273

**Describe alternatives you've considered**
Make new object type `wheelEvent` inherited from `mousewheelEvent`

prekel:
fixes #1272

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 * Lib: add missing options for Intl.NumberFormat
 * Runtime: Implement weak semantic for weak and ephemeron
 * Runtime: Implement Gc.finalise_last
+* Lib: wheel event binding
 
 ## Bug fixes
 * Compiler: fix rewriter bug in share_constant (fix #1247)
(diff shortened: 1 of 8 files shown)


hhugo:
Thanks for the contribution

PR #1284, opened 2022-06-22
fix: Only write small chunks to stdout/stderr in node

phated:
We encountered a problem in the Grain LSP when compiled to JS. We were attempting to write a string that was >8096 bytes but  only 8096 bytes were being written. We dug into the nodejs source code and found that node only keeps a buffer with size 8096 around when you try to write on stdout or stderr.

The `process.stdout.write()` and `process.stderr.write()` functions that return a flag on whether the data was flushed, and any large content will result it in returning `false`. We dropped the chunk size down to 8 characters and it seems to flush instantly in a loop.

I left the `// TODO` comments in the code in case @hhugo thinks we should handle the `flushed` return value in some way.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 * Runtime: fix Out_channel.is_buffered, set_buffered
 * Runtime: fix format wrt alternative
 * Misc: fix installation with dune 3 without opam
+* Node: Only write small chunks to stdout/stderr so they flush
 
 # 4.0.0 (2021-01-24) - Lille
 ## Features/Changes
--- a/runtime/jslib.js
+++ b/runtime/jslib.js
@@ -80,7 +80,7 @@ function js_print_stdout(s) {
 function js_print_stderr(s) {
   var s = caml_utf16_of_utf8(s);
   var process = globalThis.process;
-  if (process && process.stdout && process.stdout.write) {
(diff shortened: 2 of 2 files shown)


hhugo:
Is there any documentation about this nodejs behavior ? I cannot reproduce the issue.

Why use chunks of size 8 instead of 8096 ?

I would be nice to avoid too much duplication between js_print_stderr and js_print_stdout.

phated:
`process.stdout` and `process.stderr` are streams, so using `.write` follows this API ("If false is returned, further attempts to write data to the stream should stop until the 'drain' event is emitted."): https://nodejs.org/docs/latest-v12.x/api/stream.html#stream_writable_write_chunk_encoding_callback

If you write after `false` is returned, it seems the output is dropped. I tested a few values and it seems that 8 bytes is the safest to always flush immediately (I tested with a bunch of different values). 

> I cannot reproduce the issue.

I believe @marcusroberts discovered the 8096 limit is specific to macos, but these changes work cross-platform.

phated:
@hhugo I don't see a great way to deduplicate this code since they each write to a different stream. Any suggestions?

hhugo:
> @hhugo I don't see a great way to deduplicate this code since they each write to a different stream. Any suggestions?

An helper function taking a stream as argument

PR #1285, opened 2022-06-22
fix: Allow reading nodejs stdin that does not close

phated:
When using `fs.readFileSync(0, 'utf8')` in nodejs, the stdin has to be closed before the function will return. However, in the Grain LSP, stdin is always open, and separate messages are sent.

This implementation supports reading synchronously from the stdin file descriptor, while refilling only what is available in the stream at the time.

### DIFF
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -138,9 +138,8 @@ function caml_ml_open_descriptor_in (fd)  {
   var refill = null;
   if(fd == 0 && fs_node_supported()){
     var fs = require('fs');
-    var Buffer = require('buffer').Buffer;
     refill = function () {
-      var buf = Buffer.alloc(256);
+      var buf = globalThis.Buffer.alloc(256);
       // Ref https://gist.github.com/espadrine/172658142820a356e1e0
       var stdinFd;
       var needsClose = false;
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -138,8 +138,34 @@ function caml_ml_open_descriptor_in (fd)  {
   var refill = null;
   if(fd == 0 && fs_node_supported()){
     var fs = require('fs');
+    var Buffer = require('buffer').Buffer;
     refill = function () {
-      return caml_string_of_jsstring(fs.readFileSync(0, 'utf8'))};
+      var buf = Buffer.alloc(256);
+      // Ref https://gist.github.com/espadrine/172658142820a356e1e0
+      var stdinFd;
+      var needsClose = false;
+      try {
+        stdinFd = fs.openSync('/dev/stdin', 'rs');
+        needsClose = true;
+      } catch (e) {
+        // Opening /dev/stdin can fail (on Windows or in pkg)
+        if (globalThis.process.platform === 'win32') {
+          // On Windows, we need to use the stdin fd
+          stdinFd = process.stdin.fd;
+        } else {
+          // On Linux, we just want to use fd 0
+          // due to this nodejs bug: https://github.com/nodejs/node/issues/42826
+          stdinFd = 0;
+        }
+      }
+      var bytesRead = fs.readSync(fd, buf);
+      if (needsClose) {
+        fs.closeSync(stdinFd);
+      }
+      // Ensures we don't have any trailing, empty data
+      var str = buf.slice(0, bytesRead).toString("utf8");
+      return caml_string_of_jsstring(str);
+    };
   }
   var channel = {
     file:data.file,
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -150,7 +150,7 @@ function caml_ml_open_descriptor_in (fd)  {
         // Opening /dev/stdin can fail (on Windows or in pkg)
         if (globalThis.process.platform === 'win32') {
           // On Windows, we need to use the stdin fd
-          stdinFd = process.stdin.fd;
+          stdinFd = globalThis.process.stdin.fd;
         } else {
           // On Linux, we just want to use fd 0
           // due to this nodejs bug: https://github.com/nodejs/node/issues/42826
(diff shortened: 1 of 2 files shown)


runtime/io.js
```diff
@@ -138,8 +138,34 @@ function caml_ml_open_descriptor_in (fd)  {
   var refill = null;
   if(fd == 0 && fs_node_supported()){
     var fs = require('fs');
+    var Buffer = require('buffer').Buffer;
```
hhugo, line 141: `fs_node.js` uses  `globalThis.Buffer` directly

runtime/io.js
```diff
@@ -138,8 +138,34 @@ function caml_ml_open_descriptor_in (fd)  {
...
     refill = function () {
-      return caml_string_of_jsstring(fs.readFileSync(0, 'utf8'))};
+      var buf = Buffer.alloc(256);
```
hhugo, line 143: Should we allocate a new Buffer everytime we refill ?

runtime/io.js
```diff
@@ -138,8 +138,34 @@ function caml_ml_open_descriptor_in (fd)  {
...
+      // Ref https://gist.github.com/espadrine/172658142820a356e1e0
+      var stdinFd;
+      var needsClose = false;
+      try {
+        stdinFd = fs.openSync('/dev/stdin', 'rs');
```
hhugo, line 148: We we need to try reading `/dev/stdin` first ? can't we process.stdin.fd / 0 directly ?

runtime/io.js
phated: 👍

runtime/io.js
phated: Yes, the fd0 on linux and mac might be async—which happens any time someone tries to read from stdin using `process.stdin` (or even access the fd property, as noted in the linked issue).

So it is much safer to open a new file descriptor `/dev/stdin` in sync mode. The fallback to fd0 is used in the try/catch because there might be weird environments where /dev/stdin doesn't exist (such as the `pkg` tool that Grain uses).

runtime/io.js
phated, line 143:
```suggestion
      var buf = globalThis.Buffer.alloc(256);
```

runtime/io.js
phated, line 141:
```suggestion
```

runtime/io.js
```diff
@@ -139,7 +139,32 @@ function caml_ml_open_descriptor_in (fd)  {
...
+        needsClose = true;
+      } catch (e) {
+        // Opening /dev/stdin can fail (on Windows or in pkg)
+        if (globalThis.process.platform === 'win32') {
+          // On Windows, we need to use the stdin fd
+          stdinFd = process.stdin.fd;
```
phated, line 153:
```suggestion
          stdinFd = globalThis.process.stdin.fd;
```

runtime/io.js
```diff
@@ -139,7 +139,32 @@ function caml_ml_open_descriptor_in (fd)  {
...
+          stdinFd = globalThis.process.stdin.fd;
+        } else {
+          // On Linux, we just want to use fd 0
+          // due to this nodejs bug: https://github.com/nodejs/node/issues/42826
+          stdinFd = 0;
+        }
+      }
+      var bytesRead = fs.readSync(fd, buf);
+      if (needsClose) {
+        fs.closeSync(stdinFd);
```
hhugo, line 162: Can't we leave the fd open for the next refill ?

phated: You don't want to leave nodejs file descriptors open for a long-lived period of time because they might get garbage collected. Currently, nodejs just logs a warning but they plan to throw in an upcoming version if a file descriptor would be GC'd

hhugo: How can they decide to GC a file descriptor ? Isn't it just an integer ? I find it weird to keep opening and closing stdin everytime you want to read from stdin. Can you point to some doc ?

phated: I assume that node maintains some resource pool underneath the fd integer. I don't have any docs to link, but recently encountered the deprecation message in a different project I work on.

phated: On further investigation, the error seems to be https://nodejs.org/api/deprecations.html#DEP0137
> Allowing a [fs.FileHandle](https://nodejs.org/api/fs.html#class-filehandle) object to be closed on garbage collection is deprecated. In the future, doing so might result in a thrown error that will terminate the process.
>
> Please ensure that all fs.FileHandle objects are explicitly closed using FileHandle.prototype.close() when the fs.FileHandle is no longer needed

Using the `fs.promises.open` API seems to return a different thing than `open` and `openSync` from the non-promise API (so confusing!)

So it is probably okay to open the fd once. However, where do you expect to close it? If it is open when the process tries to exit, won't the nodejs process be held open?

hhugo: > However, where do you expect to close it? If it is open when the process tries to exit, won't the nodejs process be held open?

I would close it whenever the use does `close_in stdin`

runtime/io.js
phated: @hhugo this is a security risk because the buffer will contain extra information from the previous read. Node core has long recommended fresh or clear buffers for inputs because they will zero out all data in the buffer. There was a concerted effort in the JS ecosystem to do this work.

runtime/io.js
phated: @hhugo I'm confused then, do you want to move this "open file descriptor" logic to the place where you bind stdin for jsoo?

hhugo: I'll take a deeper look at this.

hhugo: https://github.com/ocsigen/js_of_ocaml/pull/1289 is refactoring the io logic in the runtime.
It adds a buffer to in_channel and makes reading from stdin similar to reading a regular file. Can you take a look at it, maybe test the PR and report back ?

hhugo:
Replaced by #1285

hhugo:
@phated, thanks for your contributions and testing

PR #1288, opened 2022-06-30
bump magic number to match 5.0.0~alpha0

jmid:
This PR bumps the magic numbers to suffix 32 to match the upcoming 5.0.0 ones: ocaml/ocaml@293014b. Without it jsoo will fail on 5.0.0~alpha0-compiled bytecode:
```
$ cat hello.ml 
let _ = print_endline "hello, world!"
$ ocamlc -o hello.byte hello.ml
$ js_of_ocaml hello.byte 
js_of_ocaml: Error: Bytecode version mismatch.
js_of_ocaml: Error: Your ocaml bytecode and the js_of_ocaml compiler have to be compiled with the same version of ocaml.
js_of_ocaml: Error: The Js_of_ocaml compiler has been compiled with ocaml version 5.0.0~alpha0.
js_of_ocaml: Error: Its seems that your ocaml bytecode has been compiled with a newer version of ocaml.
```


smorimoto:
@jmid The change log entry is needed. Could you add that?

jmid:
Sure! Done in da1bf3c :smiley:

PR #1289, opened 2022-07-04
Runtime: update io

hhugo:
- Implement  in_channel buffers
- Fix the implementation of md5 to work with stdin
- Always use Uint8Array in mlBytes when memory representation is an array.

TODO
- [x] more test for digest
- [x] more test for set_channel_filler
- [x] create aliases `caml_array/bytes_of_string`
- [x] check if https://github.com/ocsigen/js_of_ocaml/pull/1285 is still needed.


runtime/md5.js
```diff
@@ -20,15 +20,15 @@
...
-//Requires: caml_raise_end_of_file, caml_create_bytes, caml_string_of_bytes
+//Requires: caml_raise_end_of_file, caml_bytes_of_array, caml_md5_bytes
 function caml_md5_chan(chanid,len){
   var chan = caml_ml_channels[chanid];
   var chan_len = chan.file.length();
   if(len<0) len = chan_len - chan.offset;
   if(chan.offset + len > chan_len) caml_raise_end_of_file();
-  var buf = caml_create_bytes(len);
+  var buf = new Uint8Array(len);
   chan.file.read(chan.offset,buf,0,len);
```
hhugo, line 30: This is not correct. there might be stuff in the buffer.

runtime/jslib.js
```diff
@@ -50,55 +50,6 @@ function caml_trampoline_return(f,args) {
   return {joo_tramp:f,joo_args:args};
 }
 
-//Provides: nodejs_print (const)
-function nodejs_print(stream, s) {
-  // Writing too much data at once causes node's stdout socket to choke
```
hhugo, line 55: This logic was dropped. Check it again

runtime/fs_fake.js
```diff
@@ -221,18 +221,44 @@ MlFakeFile.prototype.write = function(offset,buf,pos,len){
...
+    this.log = console.log;
+  if(fd == 2 && typeof console.error == "function")
+    this.log = console.log;
+}
+
+MlFakeFile_out.prototype = Object.create(MlFakeFile.prototype);
+MlFakeFile_out.prototype.constructor = MlFakeFile_out;
+
+MlFakeFile_out.prototype.write = function (offset,buf,pos,len) {
+  if(len > 0 && buf[pos+len-1] == 10)
```
hhugo, line 257: check pos and len are within bounds.

runtime/fs_fake.js
```diff
@@ -221,18 +221,44 @@ MlFakeFile.prototype.write = function(offset,buf,pos,len){
...
   return 0
 }
 MlFakeFile.prototype.read = function(offset,buf,pos,len){
   var clen = this.length();
-  caml_blit_bytes(this.data, offset, buf, pos, len);
-  return 0
-}
-MlFakeFile.prototype.read_one = function(offset){
-  return caml_bytes_get(this.data, offset);
+  if(offset + len >= clen) {
```
hhugo, line 229: bugfix ??

runtime/io.js
```diff
@@ -80,11 +71,27 @@ function caml_sys_open (name, flags, _perms) {
...
+  function file(fd, name) {
+    if(fs_node_supported()) {
+      try {
+        var fs = require("fs");
+        var fd2 = fs.openSync(name, "rs");
+        new MlNodeFile(fd2, true);
+      } catch(e) {
+        new MlNodeFile(fd, true)
+      }
+      return new MlNodeFile(fd, true)
```
phated, line 86: This won't work on Windows. Windows stdin, stdout, and sterr are on different file descriptors

runtime/io.js
```diff
@@ -122,31 +130,31 @@ function caml_ml_open_descriptor_out (fd) {
     fd:fd,
     opened:true,
     out:true,
-    buffer:"",
+    buffer_curr:0,
+    buffer:new Uint8Array(65536),
```
phated, line 134: why 65536?

runtime/jslib.js
```diff
@@ -188,13 +139,12 @@ function caml_list_to_js_array(l){
...
 
 //Provides: caml_js_var mutable (const)
-//Requires: js_print_stderr
 //Requires: caml_jsstring_of_string
 function caml_js_var(x) {
   var x = caml_jsstring_of_string(x);
   //Checks that x has the form ident[.ident]*
   if(!x.match(/^[a-zA-Z_$][a-zA-Z_$0-9]*(\.[a-zA-Z_$][a-zA-Z_$0-9]*)*$/)){
-    js_print_stderr("caml_js_var: \"" + x + "\" is not a valid JavaScript variable. continuing ..");
+    console.log("caml_js_var: \"" + x + "\" is not a valid JavaScript variable. continuing ..");
```
phated, line 147:
```suggestion
    console.error("caml_js_var: \"" + x + "\" is not a valid JavaScript variable. continuing ..");
```

runtime/io.js
hhugo: I think took that number from the native runtime I think

runtime/io.js
hhugo: Here is what I read from the nodejs documentation, is this incorrect ?
```
process.stdin.fd <number>
This property refers to the value of underlying file descriptor of process.stdin. The value is fixed at 0.

process.stdout.fd <number>
This property refers to the value of underlying file descriptor of process.stdout. The value is fixed at  1.

process.stderr.fd <number>
This property refers to the value of underlying file descriptor of process.stderr. The value is fixed at 2. 
```

hhugo: Also note that the CI run with windows as well. It seems to work on windows with the setup used in the CI.

runtime/fs_fake.js
hhugo: fixed

hhugo:
@phated, I've made further changes that bring the js runtime closer to the native one.

runtime/io.js
phated: When I was investigating Windows support for Grain, Windows didn't use `0`, `1`, and `2`, so I think the nodejs documentation is wrong.

phated:
> Did you run your test against the last version of this PR?

Yes, I used `e1827f336d420dc92d76068da0d0c34b6311a1c6` in https://github.com/grain-lang/grain/compare/phated/jsoo-updates

The reproduction is a pain because you need to build the grain LSP as a JS file and then open files in vscode so it sends the messages to the LSP.

phated:
The issue only occurs on very large amounts of text, which I don't see a test for. The grain file it fails on is over 4000 lines long and 110k in size

hhugo:
> The issue only occurs on very large amounts of text, which I don't see a test for. The grain file it fails on is over 4000 lines long and 110k in size

The failure occurs while reading from an in_channel ?

phated:
> The failure occurs while reading from an in_channel ?

I'm trying to determine that right now. I'm uncertain if it is stdin or stdout.

hhugo:
This PR has a test reading from large channel/file https://github.com/ocsigen/js_of_ocaml/pull/1289/files#diff-78f20956f584b957d9ed7393c7d8f927fc8f3d47f03cf92fd5a112bf8f23a84dR24

phated:
This is also only happening on Windows. It doesn't seem to happen on MacOS.

phated:
@hhugo please ignore me. My Windows environment was broken. I've fixed it and these changes seem to be working great.

(Note that the hacks.js was removed from the dune stanza and isn't used in the branch. I just didn't delete it).

phated:
Yep! I think this should be merged and supersede my PR 🎉

PR #1290, opened 2022-07-08
feat: Implement unix_opendir and windows variants

phated:
We needed these in Grain to provided externals needed by [reason-native `fs`](https://github.com/reasonml/reason-native/tree/master/src/fs) so I implemented in nodejs.

We're using these in our published binaries, but let me know if you'd like any changes.


phated:
Weird, I have the "allow edits" checkbox selected. I'll take a look at #1291

PR #1291, opened 2022-07-11
Pr1290

hhugo:


runtime/unix.js
```diff
@@ -203,3 +203,78 @@ function unix_getpwuid(unit) {
...
+  if (entry === null) {
+      caml_raise_end_of_file();
+  } else {
+      return caml_string_of_jsstring(entry.name);
+  }
+}
+
+//Provides: unix_closedir
+//Requires: make_unix_err_args, caml_raise_with_args, caml_named_value
+//Requires: fs_node_supported, caml_failwith
```
phated, line 239:
```suggestion
```
Not needed anymore?

phated:
One comment but this looks great! 👍

runtime/unix.js
```diff
@@ -203,3 +203,78 @@ function unix_getpwuid(unit) {
...
+  dir_handle.pointer = new_dir_handle.pointer;
+  return 0;
+}
+
+//Provides: win_findfirst
+//Requires: caml_jsstring_of_string, caml_string_of_jsstring
+//Requires: unix_opendir, unix_readdir
+function win_findfirst(path) {
+  // The Windows code adds this glob to the path, so we need to remove it
+  var path = caml_string_of_jsstring(caml_jsstring_of_string(path).replace(/\\\*\.\*$/, ""));
```
phated, line 263: Just looked at the CI failure and it seems that this glob maybe isn't being removed correctly?

runtime/unix.js
```diff
@@ -203,3 +203,78 @@ function unix_getpwuid(unit) {
...
+      var unix_error = caml_named_value('Unix.Unix_error');
+      caml_raise_with_args(unix_error, make_unix_err_args("EBADF", "closedir", dir_handle.path));
+  }
+}
+
+//Provides: unix_rewinddir
+//Requires: unix_closedir, unix_opendir
+function unix_rewinddir(dir_handle) {
+  unix_closedir(dir_handle);
+  var new_dir_handle = unix_opendir(dir_handle.path);
```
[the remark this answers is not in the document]
hhugo: Rewind is implemented on the ocaml side on windows.

runtime/unix.js
hhugo: removed

runtime/unix.js
hhugo: dir_sep is `/` in practice, I've updated the regex.
Note that the new version also handle the case where the original path is empty and we endup with `win_findfirst("*/*")`

phated: Ah yes, our implementation replaces the "cygwin" env with an actual "win32" environment so that would explain the difference

hhugo: What's the reason for using Win32 instead of Cygwin ? Do you think it's a better default ? Should it be selected based on some node variable ?

hhugo:
Closing in favor of #1290

runtime/unix.js
phated: Grain needs the system to look like Windows in order for [reason-native `fp`](https://github.com/reasonml/reason-native/blob/master/src/fp/Fp.re#L223-L227) to parse Windows separators in filepaths. I'm not sure if it should be the default.

PR #1296, opened 2022-07-28
Removes bytes from dune files.

novemberkilo:
Hello, I came across [this build](http://check.ocamllabs.io/log/1658671506-f1eef1a5b413443a841a24a1c7f655ef64772692/5.0+alpha-repo/bad/js_of_ocaml-compiler.4.0.0.1~alpha-repo) of your branch on `check.ocamllabs.io` and thought I would contribute what I hope to be a fix.

I'm quoting a colleague here at Tarides:

> 5.0.0~alpha1 corrects a mistake since the very beginning of jbuilder where packages insisted on adding the bytes package to libraries stanzas, even though there was no version of OCaml you could be using which didn't have the Bytes module

Thus my suggested fix is to remove `bytes` from any dune file that currently has it.

### DIFF
--- a/compiler/lib/dune
+++ b/compiler/lib/dune
@@ -5,7 +5,6 @@
  (libraries
   compiler-libs.common
   compiler-libs.bytecomp
-  bytes
   menhirLib
   (select
    source_map_io.ml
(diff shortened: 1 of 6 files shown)


novemberkilo:
cc/ @hhugo

PR #1298, opened 2022-08-25
WIP: target JScript

ghuysmans:
I'm trying to run OCaml programs on Windows through [JScript](https://docs.microsoft.com/en-us/previous-versions//hbxc2t98(v=vs.85)?redirectedfrom=MSDN) (an old implementation of ECMAScript 3). It's an easy and portable way to access COM (actually ActiveX) objects. A generator for wrappers using [type libraries](https://docs.microsoft.com/en-us/windows/win32/midl/com-dcom-and-type-libraries) is underway.

It mostly works but I had to debug strange issues related to typed arrays (separate compilation seems to break weak symbols resolution) and a curious interpretation of try/catch in JScript, as reported in [this document](https://regmedia.co.uk/2007/10/31/jscriptdeviationsfromes3.pdf) (§2.11). The latter didn't always trigger since it's related to the way scoping works: as the generated variable names change even slightly, the bug might disappear. I wonder if I solved these issues (especially the latter) correctly. Any thoughts?


smorimoto:
How about "@babel/plugin-transform-jscript" for your use case?

ghuysmans:
Hi, thanks for your answer!

> How about "@babel/plugin-transform-jscript" for your use case?

It doesn't look like they handle anything else than _named function expressions_ (which I don't support yet, thank you for reminding me!), so that won't work. For example, exception variables (after the `catch` keyword) are accessible to the whole enclosing function. Moreover, the real ES3 doesn't even support redefining array accessors.

smorimoto:
Oops, I thought it was a little more decent...
https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-jscript/src/index.ts

ghuysmans:
> Oops, I thought it was a little more decent... https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-jscript/src/index.ts

Yeah, right... Besides, it would be safer (and more pleasant) to write those transformations using the existing mapper infrastructure in jsoo.

ghuysmans:
I ended up hitting a recursion limit, and there's no way to configure it. So I guess that wouldn't ever run reliably.

hhugo:
> I ended up hitting a recursion limit, and there's no way to configure it. So I guess that wouldn't ever run reliably.

Can you show the exact error ? Do you have a stack trace ? Do you know what code triggers it ?

ghuysmans:
Just to be clear, `cscript` (the launcher for scripts running in the console) crashes with a stack overflow, not jsoo! I tried compiling code generated by ppx_rapper, which uses a monadic operator; I still have the code in a separate (currently private) repository.

hhugo:
@ghuysmans, do you plan to do anything with this PR ? Using `--enable effects` might solve your stackoverflow issue.

hhugo:
> It mostly works but I had to debug strange issues related to typed arrays (separate compilation seems to break weak symbols resolution) and a curious interpretation of try/catch in JScript, as reported in [this document](https://regmedia.co.uk/2007/10/31/jscriptdeviationsfromes3.pdf) (§2.11). The latter didn't always trigger since it's related to the way scoping works: as the generated variable names change even slightly, the bug might disappear. I wonder if I solved these issues (especially the latter) correctly. Any thoughts?

Maybe https://github.com/ocsigen/js_of_ocaml/commit/731146444ea41ad9a187d0c4be9a72a8789f15d0 fixes the scoping issue...

ghuysmans:
> @ghuysmans, do you plan to do anything with this PR ? Using `--enable effects` might solve your stackoverflow issue.

I'll have a try, this looks pretty exciting but I'm too busy right now...

hhugo:
js_of_ocaml is currently going the other direction of modernizing its js output.
We can certainly make some small changes if it helps your use case but we won't be able maintain compatibility with es3 or JScript.

Maybe looking in the direction of https://babeljs.io/docs/babel-plugin-transform-jscript could help.

I'll close this PR as I don't see what to make of it.

PR #1307, opened 2022-09-19
Bugfix in ocamlyacc runtime code 

mlasson:
# Description 

The parsing module runtime has a bug in `parsing.js`. 

In happened in the "ERROR_DETECTED" case. 
There's a break instruction that is breaking a for loop instead of breaking a switch. 
Which makes the "flow" fall through the "SHIFT" case instead of the "SHIFT_RECOVER" case. 

This PR adds a label to fix this. 

Here is a reproduction case, https://github.com/mlasson/js-yacc-bug

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 
 ## Bug fixes
 * Compiler: fix rewriter bug in share_constant (fix #1247)
+* Runtime: fix ocamlyacc parse engine (#1307)
 * Runtime: fix Out_channel.is_buffered, set_buffered
 * Runtime: fix format wrt alternative
 * Runtime: fix Digest.channel
--- a/runtime/parsing.js
+++ b/runtime/parsing.js
@@ -96,7 +96,7 @@ function caml_parse_engine(tables, env, cmd, arg)
   var errflag = env[env_errflag];
 
   exit:for (;;) {
-    switch(cmd) {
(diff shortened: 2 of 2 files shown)


hhugo:
Can we include the repro case as well ?

mlasson:
Where do you want to include it ?
In a PR description ? In a test ? In a comment ?
But, sure, yes, feel free.

mlasson:
> Where do you want to include it ? In a PR description ? In a test ? In a comment ? But, sure, yes, feel free.

I can add a test myself if you want. In that case, should I add a directory `tests-ocamlyacc` in the `compiler` directory ?

hhugo:
I 

> > Where do you want to include it ? In a PR description ? In a test ? In a comment ? But, sure, yes, feel free.
> 
> I can add a test myself if you want. In that case, should I add a directory `tests-ocamlyacc` in the `compiler` directory ?

Adding it in compiler/tests-jsoo would make it run in both native and js. I would also be a good place to make sure https://github.com/ocsigen/js_of_ocaml/pull/1308 follows the native version.

hhugo:
@mlasson, Gentle ping

hhugo:
Adding test in https://github.com/ocsigen/js_of_ocaml/pull/1315

mlasson:
Thanks

PR #1308, opened 2022-09-19
Implemement set_parser_trace

mlasson:
# Description 

While investigating #1307, I made this to try to understand what was going on. 
It implements the tracer of the `Parsing` module using `console.log` with minor differences with the native counterpart (`console.log` is adding spaces around values as well as some colors when executed with node). 

Feel free to close if not interested.

### DIFF
--- a/runtime/parsing.js
+++ b/runtime/parsing.js
@@ -17,8 +17,13 @@
 
 /* The pushdown automata */
 
+//Provides: caml_parser_trace
+var caml_parser_trace = 0;
+
 //Provides: caml_parse_engine
-//Requires: caml_lex_array
+//Requires: caml_lex_array, caml_parser_trace,caml_jsstring_of_string
+//Requires: caml_ml_output, caml_ml_string_length, caml_string_of_jsbytes
+//Requires: caml_jsbytes_of_string, MlBytes
 function caml_parse_engine(tables, env, cmd, arg)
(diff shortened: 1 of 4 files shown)


runtime/parsing.js
```diff
@@ -73,8 +76,27 @@ function caml_parse_engine(tables, env, cmd, arg)
...
+
+  function print_token(tok)
+  {
+    var token;
+    if (tok instanceof Array) {
+      token = token_name(tables[tbl_names_block], tok[0]);
+    } else {
+      token = token_name(tables[tbl_names_const], tok);
+    }
+    console.log("State", state, ": read token", token);
```
hhugo, line 98: The native runtime prints some additional information. Could we try to replicate the behavior.

Can we add an expect test for the trace feature ?

hhugo: I've pushed a commit to do this.

mlasson:
Thanks for taking care of that.

hhugo:
Files were missing. I've pushed a commit to fix this.

hhugo:
One need to investigate the failure on windows

mlasson:
It's seems that the "runner" is not capturing stderr on windows.

hhugo:
I've disabled the trace in the testsuite until the issue is fixed in ppx_expect.

PR #1309, opened 2022-10-01
Add support for transition events.

bogdan2412:


hhugo:
Thanks.

It would be nice to perform the corresponding changes in `lib/lwt/lwt_js_events.ml` and cleanup the outdated implementation that exists for `transitionend`.

Can you also add a changelog entry ?

bogdan2412:
I made an attempt, let me know if it's reasonable or not.

One behaviour difference here that seems maybe hard to avoid - previously we tried to identify a kind of transitionend event that is supported by the browser by trying out a bunch of different options and if we failed to find any, then we'd end up returning from [transitionend] immediately, whereas now we hang forever. That both seems to only potentially be a problem in fairly ancient browsers, but also seems more correct anyway?

lib/lwt/lwt_js_events.ml
```diff
@@ -387,6 +387,36 @@ let pointerover ?use_capture ?passive target =
 let pointerup ?use_capture ?passive target =
   make_event Dom_html.Event.pointerup ?use_capture ?passive target
 
+let transition_evn =
```
hhugo, line 390: I should have been more explicit. I think we can get rid of all old browser specific support.

PR #1311, opened 2022-10-12
caml_js_get and caml_js_delete (can) have side-effects

### LINKED ISSUE #1310: [BUG] caml_js_get should not be assumed to be pure
**Describe the bug**
The dead code elimination will happily remove the call to a property getter (using caml_js_get) if the result is unused.  But accessing the property can trigger arbitrary code, so the optimization is not sound.  We have observed a bug in our application (only in on-dev mode!), due to a DOM property being accessed only for its side-effect (namely, accessing `offsetHeight` to force a reflow).

````ocaml
type t

external get: t -> string -> t = "caml_js_get"
external get_glb: unit -> t = "get_glb"
external set_glb: t -> unit = "set_glb"

let () =
  set_glb (get (get_glb()) "foo");
  ignore (get (get_glb()) "bar")
````

generates:

````javascript
    set_glb(get_glb(0)[caml_string_of_jsbytes("foo")]);
    get_glb(0);
````

Any code triggered in the getter for the "bar" property will not be executed as it should.

**Versions**
js_of_ocaml 4.00; ocaml 4.14.1

### PR DESCRIPTION
**Pull request by alainfrisch**
This is obvious for caml_js_delete, but also for caml_js_get which can trigger arbitrary code in the property getter.

Fix #1310

### DIFF
--- a/runtime/jslib.js
+++ b/runtime/jslib.js
@@ -24,7 +24,7 @@ function caml_js_pure_expr (f) { return f(); }
 
 //Provides: caml_js_set (mutable, const, const)
 function caml_js_set(o,f,v) { o[f]=v;return 0}
-//Provides: caml_js_get mutable (const, const)
+//Provides: caml_js_get mutator (const, const)
 function caml_js_get(o,f) { return o[f]; }
 //Provides: caml_js_delete (mutable, const)
 function caml_js_delete(o,f) { delete o[f]; return 0}
--- a/runtime/jslib.js
+++ b/runtime/jslib.js
@@ -24,7 +24,7 @@ function caml_js_pure_expr (f) { return f(); }
 
 //Provides: caml_js_set (mutable, const, const)
 function caml_js_set(o,f,v) { o[f]=v;return 0}
-//Provides: caml_js_get mutator (const, const)
+//Provides: caml_js_get (const, const)
 function caml_js_get(o,f) { return o[f]; }
 //Provides: caml_js_delete (mutable, const)
 function caml_js_delete(o,f) { delete o[f]; return 0}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,7 @@
 * Runtime: sync channel seek / pos with the native runtime
 * Misc: fix installation with dune 3 without opam
 * Node: Only write small chunks to stdout/stderr so they flush
+* Compiler: no dead code elimination for caml_js_get
 
 # 4.0.0 (2021-01-24) - Lille
 ## Features/Changes
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -861,7 +861,7 @@ let _ =
 
   register_tern_prim "caml_js_set" (fun cx cy cz _ ->
       J.EBin (J.Eq, J.EAccess (cx, cy), cz));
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by hhugo** [comment]
I'm surprised this actually fixes your issue. Can you confirm it does ?
Reading the code, I would expect the information from the runtime (runtime/jslib.js) to take precedence / override.

The current logic is confusing, I'll try to clean this up.

**Comment by hhugo** [comment]
You should probably rebase your PR on top of #1312

**Comment by nojb** [comment]
> I'm surprised this actually fixes your issue. Can you confirm it does ?

Indeed, it seems this is *not* enough to fix the problem.

**Comment by alainfrisch** [comment]
I tested manually with the examples from #1310, but perhaps I did something wrong...

**Comment by hhugo** [comment]
#1310 has been merged. Can you rebase this one on top ?

**Comment by nojb** [comment]
> #1312 has been merged. Can you rebase this one on top ?

Done. And I confirm this fixes our issue as well.

File: runtime/jslib.js [hunk 1]
```diff
@@ -24,7 +24,7 @@ function caml_js_pure_expr (f) { return f(); }
 
 //Provides: caml_js_set (mutable, const, const)
 function caml_js_set(o,f,v) { o[f]=v;return 0}
-//Provides: caml_js_get mutable (const, const)
+//Provides: caml_js_get mutator (const, const)
```
alainfrisch, line 27: Shouldn't we declare caml_js_delete (just below) also mutator?

nojb: The absence of an annotation is actually interpreted as "mutator":
https://github.com/ocsigen/js_of_ocaml/blob/8c993985feb70314635d83d31ea7c0daa10efe37/compiler/lib/annot_parser.mly#L36 so I actually removed the annotation from `caml_js_get` for uniformity with the rest of the file.

hhugo:
I wonder if we should provide the two alternatives.
Here is the diff I see on the minesweeper example.
```diff
------ master
++++++ pr
@-1,4 +1,5 ============================================================
!// Generated by js_of_ocaml 4.0.0+git-4.0.0-89-geee04cead4-dirty
 (function
    (Object)
    {typeof globalThis
     !==
@-5573,10 +5574,13 ============================================================
     caml_register_named_value(caml_string_of_jsbytes("jsError"),slot);
     (function(exn){throw exn});
-    var
-     t19=globalThis,
-     no_handler=null,
-     undefined$0=undefined,
-     false$0=false,
-     array_length=t19.Array;
+    var t19=globalThis,no_handler=null,undefined$0=undefined,false$0=false;
+    t19.String;
+    t19.RegExp;
+    t19.Object;
+    var array_length=t19.Array;
+    t19.Date;
+    t19.Math;
+    t19.Error;
+    t19.JSON;
     register_printer
      (function(param)
```

hhugo:
I wasn't able to push to your branch. I've merged this in https://github.com/ocsigen/js_of_ocaml/commit/0239708904e61ed454bf77e74b1ead0f53a27118

nojb:
> I wasn't able to push to your branch. I've merged this in [0239708](https://github.com/ocsigen/js_of_ocaml/commit/0239708904e61ed454bf77e74b1ead0f53a27118)

Thanks! And apologies for the lag (I was on holidays and this PR got lost in the shuffle).

PR #1317, opened 2022-10-24
Js_traverse.freevar: performance improvement

vouillon:
With nested functions, we are merging repeatedly the same set of variable usage information. It is more efficient to maintain a global table.


vouillon:
In order to implement effect handlers, we are performing a CPS conversion of the code, which generates deeply nested functions. Js_of_ocaml is really slow on some test programs as a result. But we may have to change this since some JavaScript engines do not support overly nested functions.

I have not observed this slow-down otherwise. But since this change does not make the code significantly more complex, I think it's worthwhile anyway.

Issue #331, opened 2015-06-26
The minifyer does not deal correctly with toplevel variables


vouillon:
This piece of code:
```javascript
var a = "toto";
function f (e){ return a; }
```
is turned into the incorrect code:
```javascript
var
a="toto";function
f(a){return a}
```
So, the minifyer does not rename toplevel variable, which is correct. But it does not take them into account when renaming other variables.

PR #1318, opened 2022-10-24
More efficient implementation of Js_traverse.rename_variable + fixes

vouillon:
It was inefficient for deeply nested functions since we were performing variable substitution repeatedly.

Also,
```javascript
var xx = 1;

function f () {
    xx = 2; 
    try {
        throw 1
    } catch (xx) { 
        var xx = 3 
    }; 
    return xx;
}

function g () {
    var xx = 2; 
    return xx; 
}
```
was minified into
```javascript
var
xx=1;function
f(){xx=2;try{throw 1}catch(a){var
a=3}return xx}function
g(){var
xx=2;return xx}
```
which is wrong for `f` since `xx` is no longer declared locally.

And, we rename variables with the same name as a topelevel variable as in function `g`. I think the fix for #331 was overly conservative.

### DIFF
--- a/compiler/bin-jsoo_minify/jsoo_minify.ml
+++ b/compiler/bin-jsoo_minify/jsoo_minify.ml
@@ -74,9 +74,7 @@ let f { Cmd_arg.common; output_file; use_stdin; files } =
     let true_ () = true in
     let open Config in
     let passes : ((unit -> bool) * (unit -> Js_traverse.mapper)) list =
-      [ ( Flag.shortvar
-        , fun () -> (new Js_traverse.rename_variable toplevel_def :> Js_traverse.mapper)
-        )
+      [ (Flag.shortvar, fun () -> (new Js_traverse.rename_variable :> Js_traverse.mapper))
       ; (true_, fun () -> new Js_traverse.simpl)
       ; (true_, fun () -> new Js_traverse.clean)
       ]
--- a/compiler/bin-jsoo_minify/jsoo_minify.ml
+++ b/compiler/bin-jsoo_minify/jsoo_minify.ml
@@ -74,7 +74,7 @@ let f { Cmd_arg.common; output_file; use_stdin; files } =
     let true_ () = true in
     let open Config in
     let passes : ((unit -> bool) * (unit -> Js_traverse.mapper)) list =
-      [ (Flag.shortvar, fun () -> (new Js_traverse.rename_variable :> Js_traverse.mapper))
+      [ (Flag.shortvar, fun () -> new Js_traverse.rename_variable)
       ; (true_, fun () -> new Js_traverse.simpl)
       ; (true_, fun () -> new Js_traverse.clean)
       ]
--- a/compiler/lib/js_traverse.ml
+++ b/compiler/lib/js_traverse.ml
@@ -625,55 +625,82 @@ class free =
   end
 
 class rename_variable =
-  object
-    inherit free as super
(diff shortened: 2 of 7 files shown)


compiler/lib/js_traverse.ml
```diff
@@ -507,63 +672,93 @@ class free =
...
-      let x = super#source x in
-      match x with
+    method expression e =
+      match e with
+      | EFun (ident, params, body, _) ->
+          let subst0 = subst in
+          let decl0 = decl in
+          let subst', decl' = update_subst ident params body subst in
+          subst <- subst';
+          decl <- decl';
```
hhugo, line 725: Why not use the override construct like in `Function_declaration` ?

compiler/lib/js_traverse.ml
```diff
@@ -507,63 +672,93 @@ class free =
...
+          subst <- subst0;
+          decl <- decl0;
+          e'
+      | _ -> super#expression e
+
+    method statement s =
+      match s with
+      | Try_statement (b, Some ((S { name; _ } as id), block), final)
+        when not (StringSet.mem name decl) ->
+          (* If [name] is declared in [body] but not outside, then
```
hhugo, line 736:
```suggestion
          (* If [name] is declared in [block] but not outside, then
```

compiler/bin-jsoo_minify/jsoo_minify.ml
```diff
@@ -74,9 +74,7 @@ let f { Cmd_arg.common; output_file; use_stdin; files } =
     let true_ () = true in
     let open Config in
     let passes : ((unit -> bool) * (unit -> Js_traverse.mapper)) list =
-      [ ( Flag.shortvar
-        , fun () -> (new Js_traverse.rename_variable toplevel_def :> Js_traverse.mapper)
-        )
+      [ (Flag.shortvar, fun () -> (new Js_traverse.rename_variable :> Js_traverse.mapper))
```
hhugo, line 77:
```suggestion
      [ (Flag.shortvar, fun () -> new Js_traverse.rename_variable)
```

compiler/lib/js_traverse.ml
```diff
@@ -507,63 +672,93 @@ class free =
...
+             we cannot replace [id] by a fresh variable. As a fast
+             approximation, we only use a fresh variable when [name]
+             is not declared. *)
+          Try_statement
+            ( m#statements b
+            , (let subst0 = subst in
+               subst <- StringMap.add name (Code.Var.fresh_n name) subst;
```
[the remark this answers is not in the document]
vouillon: Indeed, this is better.

compiler/lib/js_traverse.ml
```diff
@@ -507,63 +672,83 @@ class free =
...
 
-class rename_variable keeps =
-  object
-    inherit free as super
+class rename_variable =
+  object (m)
+    inherit map as super
 
-    val mutable sub_ = new subst (fun x -> x)
+    val mutable subst = StringMap.empty
```
hhugo, line 679: mutable isn't needed anymore

hhugo: I've removed it.

vouillon:
> I took the liberty to push on your branch.
> 
>     * I've added tests
> 
>     * split-out the bug fixes from the performance improvement
> 
> 
> Let me know if the result looks good to you still.

Thanks! This looks good to me.

PR #1319, opened 2022-10-27
Compiler: attempt to optimize computation of dominance frontier 

### LINKED ISSUE #1300: [PERF] Dominance frontier calculation appears to be O(n^2)
A large jsoo application I've encountered at work spends about 9 minutes compiling with js_of_ocaml using direct-compilation (separate compilation is _much_ faster).  Profiling the compiler shows that almost all of this time is spent in [generate.ml, specifically computing dominance frontiers](https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/lib/generate.ml#L637).

I believe that the issue is that although dominance frontier calculation can be done for the whole program in one pass, the implementation in generate.ml will compute the frontier for each node individually, resulting in no opportunity to share results.

hhugo:
Cache computation of dominance frontier to avoid the quadratic behavior when dealing with long chains of simple branches.

- Add debug for dominance frontier timings.
- keeps the old implementation next to the new one, allowing to check that the two implementations give the same result.  
- the cache is invalidated every time we update the graph (when collapsing frontier / processing interm nodes).

Fixes #1300.
With the following exemple, dominance frontier timing goes from 14.43 s to 0.01 s.
 
 ```
ocaml <(echo "for i = 0 to 10000  do print_endline \"print_char '.';\"; done") > a.ml && ocamlc -g a.ml -o a.byte && _build/default/compiler/bin-js_of_ocaml/js_of_ocaml.exe --debuginfo --debug times  --pretty a.byte
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 * Test: track external used in the stdlib and unix
 
 ## Bug fixes
+* Compiler: fix quadratic behavior of dominance frontier (fix #1300)
 * Compiler: fix rewriter bug in share_constant (fix #1247)
 * Compiler: fix miscompilation of mutually recursive functions in loop (#1321)
 * Runtime: fix ocamlyacc parse engine (#1307)
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -39,6 +39,8 @@ let debug = Debug.find "gen"
 
 let times = Debug.find "times"
 
+let dominance_frontier_time = ref 0.
+
 open Code
 module J = Javascript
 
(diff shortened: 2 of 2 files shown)


hhugo:
@TyOverby, can you confirm this fixes your issue and doesn't introduce regressions ?
@vouillon, would you be able to review this change ?

pmwhite:
@TyOverby handed this off to me. I'll work on confirming that this doesn't cause regressions.

pmwhite:
Just tested this patch internally; all the tests pass. I also traced the build before and after the patch to see if it did indeed go faster. Before, the final job of the build (which I believe is the relevant one for this patch) took 47.42 seconds, but afterwards only took 27.50 seconds (42% speedup). Ship it!

hhugo:
@pmwhite, what flags / setup are you testing against? #1300 was mentioning stack-overflow with source-map and whole-program compilation after running for more than minutes.

pmwhite:
@hhugo Sorry, I should have included that information. The app I tested against was a lot smaller than the app Ty tested against, hence the much smaller timings. I just built the app that Ty had built before, and the final job went from +5 minutes (I quit waiting) down to 1 minute, 4 seconds. I was building with source-maps and whole-program compilation for both these builds and the first ones.

PR #1322, opened 2022-10-28
Compiler: get rid of block.handler

hhugo:
The `handler` of a block doesn't seem to be useful. There was some renaming logic in `generate.ml` based on `handler` but I don't it ever triggers. Remove it allows to simplify the code a bit.

The PR also removes the second argument of the Poptrap constructor which was only used in parse_bytecode.

### DIFF
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -333,7 +333,6 @@ type last =
 
 type block =
   { params : Var.t list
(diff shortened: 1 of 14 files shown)


hhugo:
@vouillon, do you have time to look at this ?

vouillon:
This looks good to me. I must admit I don't remember what this renaming logic was for...

PR #1323, opened 2022-11-02
Doc: document findlib interaction

hhugo:
cc @dbuenzli, in case you want to review

### DIFF
--- a/manual/linker.wiki
+++ b/manual/linker.wiki
@@ -2,10 +2,10 @@
 The Js_of_ocaml compiler accepts JavaScript files provided on the command-line.
 The main purpose is to provide (external) primitives needed by the bytecode program.
 
-Most of the primitives from the standard library are already implemented and loaded by default (located in "runtime.js").
+Most of the primitives from the standard library are already implemented and loaded by default.
 
-Additionally, some other primitives are installed but not loaded by default:
- * "+toplevel.js" : when compiling toplevel
+Additionally, some other primitives are available but not loaded by default:
+ * "+toplevel.js" and "+dynlink.js" when compiling toplevel and/or using dynlink.
 
 == Command-line
 Pass the JavaScript file (must have a ".js" extension)
@@ -13,12 +13,8 @@ Pass the JavaScript file (must have a ".js" extension)
 js_of_ocaml jsfile.js a.byte
 >>
 
-Or load it from a findlib package
-<<code |
-js_of_ocaml +mypackage/jsfile.js a.byte
->>
-The file **jsfile.js** will be looked up inside **mypackage** lib directory.
-When no package is provided, the compiler will look inside js_of_ocaml-compiler lib directory.
+See the <<a_manual chapter="runtime-files" |runtime files>> chapter for how to
+discover runtime files.
 
 == Provide your own JavaScript
 You may need to provide extra JavaScript files to provide missing primitives or to override existing ones.
--- a/manual/runtime-files.wiki
+++ b/manual/runtime-files.wiki
@@ -1,0 +1,112 @@
+= Runtime files =
+
+For compiling a bytecode executable to JavaScript, one often need to pass
+JavaScript runtime files for some OCaml libraries used by the
+executable. Tracking the required runtime files is done using the de-facto
+standard for library management in OCaml, Findlib.
+
+== The old deprecated way
+
+Before dune and its native support for js_of_ocaml, the (hacky) way
+to rely on Findlib was to use a **linkopts** variable with a **javascript**
+predicate {{{linkopts(javascript)}}} and have js_of_ocaml be responsible for
+looking up the files in the right directory.
+
+This approach is not very satisfying:
+* It makes it harder for a build system to know the set of dependencies.
+* The js_of_ocaml compiler needs to perform findlib resolution itself.
+
+=== Listing linkopts
+
+To list all {{{linkopts(javascript)}}} for a set of libraries:
+
+{{{
+  ocamlfind query -o-format -r -predicates javascript LIB1 LIB2 ...
+}}}
+
+For example, for **base** and **ppx_expect**
+
+{{{
+  $ ocamlfind query -o-format -r -predicates javascript base ppx_expect
+  +base/base_internalhash_types/runtime.js
+  +base/runtime.js
+  +time_now/runtime.js
+  +ppx_expect/collector/runtime.js
+}}}
+
+=== Declaring runtime JavaScript files inside META files
+
+The content of {{{linkopts(javascript)}}} is given directly to the
+js_of_ocaml compiler.  To be able reference files independently of the
+installation path, we rely on js_of_ocaml's findlib support (see Command-Line findlib support below).
+
+For example, to declare that **mypackage.sublib** requires the file **mypackage/subdir/runtime.js**
+
+{{{
+  cat mypackage/META
+  package "sublib" (
+    directory = "subdir"
+    linkopts(javascript) = "+mypackage/sublib/runtime.js"
+  )
+}}}
+
+
+== The new way (introduced by/for dune)
+
+Using findlib more efficiently, with a usage closer to what's
+done for OCaml archives.
+
+=== Listing js runtime files
+
+To list all JavaScript runtime files for a set of libraries:
+
+{{{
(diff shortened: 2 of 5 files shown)


manual/runtime-files.wiki
```diff
@@ -0,0 +1,112 @@
...
+    jsoo_runtime = "runtime.js"
+  )
+}}}
+
+== Command line findlib support
+
+If js_of_ocaml is compiled with findlib support, one can use the
+following syntax to reference JavaScript files:
+* {{{ +{findlib-package}/{javascript-file} }}}
+* {{{ +{javascript-file} }}}
```
dbuenzli, line 96: I'm a bit uneasy about that convention, is it here for backward compat or is that new ? If it's new I would rather say let's not add it or use a different symbol for packages. If it's already there, I would suggest to deprecate it.

IIRC in the `ocaml` compilers the `+` simply means substitute `ocamlc -where`. So `+dir/file` will be `$(ocamlc -where)/dir/file` not file `file` in package `dir`.

manual/linker.wiki
```diff
@@ -2,23 +2,19 @@
 The Js_of_ocaml compiler accepts JavaScript files provided on the command-line.
 The main purpose is to provide (external) primitives needed by the bytecode program.
 
-Most of the primitives from the standard library are already implemented and loaded by default (located in "runtime.js").
+Most of the primitives from the standard library are already implemented and loaded by default.
 
-Additionally, some other primitives are installed but not loaded by default:
- * "+toplevel.js" : when compiling toplevel
+Additionally, some other primitives are available but not loaded by default:
+ * "+toplevel.js" and "+dynlink.js" when compiling toplevel and/or using dynlink.
```
dbuenzli, line 8: I think we should expose (ocaml empty) libraries to bring these in rather than use this special syntax. See my comment below.

dbuenzli (commented):
Thanks. The convention and mechanics looks good to me. I left a few comments.

dbuenzli:
One more thing. In case you have a single `.js` file I think it would be maybe a good idea to suggest calling it `runtime.js` by convention (or something else but that's what you did in `ptime`).

manual/runtime-files.wiki
hhugo: This is not new, this has been present for many years. It was introduced before there was any build system support for js_of_ocaml.

Findlib has `@package/file` which I should probably have used back then.

> The variable "archive" specifies the list of archive files. These files should be given either as (1) plain names without any directory information; they are only searched in the package directory. (2) Or they have the form "+path" in which case the files are looked up relative to the standard library. (3) Or they have the form "@name/file" in which case the files are looked up in the package directory of another package. (4) Or they are given as absolute paths.
```

Comment (hhugo) [reply manual/runtime-files.wiki:96]: Dropping special findlib support seems to be the way to go.

File: manual/runtime-files.wiki [hunk 3]
```diff
@@ -0,0 +1,112 @@
+= Runtime files =
+
+For compiling a bytecode executable to JavaScript, one often need to pass
+JavaScript runtime files for some OCaml libraries used by the
+executable. Tracking the required runtime files is done using the de-facto
+standard for library management in OCaml, Findlib.
+
+== The old deprecated way
```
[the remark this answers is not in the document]
hhugo: Moved at the end

manual/runtime-files.wiki
```diff
@@ -0,0 +1,112 @@
...
+{{{
+  cat mypackage/META
+  package "sublib" (
+    directory = "subdir"
+    linkopts(javascript) = "+mypackage/sublib/runtime.js"
+  )
+}}}
+
+
+== The new way (introduced by/for dune)
```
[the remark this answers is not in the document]
hhugo: addressed

manual/runtime-files.wiki
```diff
@@ -0,0 +1,112 @@
...
+
+For example, for **base** and **ppx_expect**
+
+{{{
+  $ ocamlfind query -format "%+(jsoo_runtime)" -r base ppx_expect | grep -v "^$"
+  ~/.opam/4.14.0/lib/base/base_internalhash_types/runtime.js
+  ~/.opam/4.14.0/lib/base/runtime.js
+  ~/.opam/4.14.0/lib/time_now/runtime.js
+  ~/.opam/4.14.0/lib/ppx_expect/collector/runtime.js
+}}}
```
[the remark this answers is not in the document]
hhugo: I've added an example. The toplevel/dynlink situation will be dealt with elsewhere.

manual/runtime-files.wiki
```diff
@@ -0,0 +1,112 @@
...
+  cat mypackage/META
+  package "sublib" (
+    directory = "subdir"
+    jsoo_runtime = "runtime.js"
+  )
+}}}
+
+== Command line findlib support
+
+If js_of_ocaml is compiled with findlib support, one can use the
```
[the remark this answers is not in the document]
hhugo: Support is automatically enabled if the findlib library is available when building the jsoo compiler.

manual/linker.wiki
hhugo: #1324 adds `+dynlink.js` and `+toplevel.js` whenever  `--toplevel` or `--dynlink` is used. More cleanup is required, the distinction between `--toplevel` and `--dynlink` should be documented or removed.

manual/runtime-files.wiki
hhugo: The documentation now mentions that findlib support is deprecated.

hhugo:
> One more thing. In case you have a single `.js` file I think it would be maybe a good idea to suggest calling it `runtime.js` by convention (or something else but that's what you did in `ptime`).

added

manual/linker.wiki
dbuenzli: If there's no distinction then indeed it may be worth deprecating `--toplevel` and instruct people to use `--dynlink` when they want to compile a toplevel or use the `Dynlink` library (it's not entirely unnatural IIRC ocamlnat uses `Dynlink`).

dbuenzli (commented):
I had a quick look at the changes it looks mostly ok to me. Thank you. 

If you really want to polish I would rather call the `runtime-files` page, "How to compile a program" and make it centered on that but I guess most people are not interested in the base invocations and the current organisation is enough for build system fiddlers.

Btw. maybe I dreamt but at some point I thought I had bought you into converting the docs to `.mld` files :-) I still think it would be worthwhile, the docs should show up on ocaml.org (and in `odig doc js-of-ocaml-compiler` if you are a local documentation addict).

hhugo:
> If there's no distinction then indeed it may be worth deprecating --toplevel and instruct people to use --dynlink when they want to compile a toplevel or use the Dynlink library (it's not entirely unnatural IIRC ocamlnat uses Dynlink).

I was more tempted to deprecated `--dynlink` and ask people to use `--toplevel`. I cannot come up with a good argument either way.

dbuenzli:
> I cannot come up with a good argument either way.

Well you can use dynlink without using a toplevel no ?

dbuenzli:
> And you can use the toplevel without dynlink.

Somehow a `#load` directive is equivalent to dynlinking a module (I remember thinking when I was working on the library linking proposal that a lot of the bytecode toplevel loading business could likely be made to reuse `Dynlink`, and IIRC `ocamlnat` does actually use it). 

> Dynlink can only be used if Js_of_ocaml_toplevel is initialized.

That looks like a jsoo implementation detail. For me toplevel is really the interactive behaviour so it feels a bit strange to me to use the `--toplevel` flag if I just want to be able to dynamically load bytecode.

But in any case what do these flags do ? What do they enable ? And do they actually do exactly the same thing ? 

Based on these answers maybe we can find a new entirely different technical name and deprecate the other twos if you find neither suitable.

hhugo:
> If you really want to polish I would rather call the runtime-files page, "How to compile a program" and make it centered on that but I guess most people are not interested in the base invocations and the current organisation is enough for build system fiddlers.

I'd like to restructure the entire doc eventually but will leave the doc as is for now

>  Btw. maybe I dreamt but at some point I thought I had bought you into converting the docs to .mld files :-) I still think it would be worthwhile, the docs should show up on ocaml.org (and in odig doc js-of-ocaml-compiler if you are a local documentation addict).

I'd like some kind of agreement  with other ocsigen devs before. Mostly concerned about the ocsigen.org website.

Issue #1236, opened 2022-01-30
[FEATURE REQUEST] Provide `Jsoo_runtime.Sys.js_of_ocaml_version`


dbuenzli:
I was trying to get rid of my use of `Js_of_ocaml.Sys_js` in [this source](https://github.com/dbuenzli/brr/blob/0b48f2871eb5ee6a0103146b4adc5b4c51dddbf2/src/poke/brr_poke.ml). It failed short because I'm using `Js_of_ocaml.Sys_js.js_of_ocaml_version`. 

Would it maybe make sense to provide it in `Jsoo_runtime` as `Sys.js_of_ocaml_version` in the future ? 

Thanks.

jonahbeckford:
I'd like this feature as well. I have unreleased code that warns if `jsoo_runtime` is already in `globalThis` during a `require()` of the separately compiled JSOO runtime. It would be even better to explicitly get a console error if there was a version mismatch, which I can implement only if the version is accessible from the JSOO runtime.

PR #1326, opened 2022-11-10
Lib: add Jsoo_runtime.Sys.version

hhugo:
fix https://github.com/ocsigen/js_of_ocaml/issues/1236


dbuenzli:
There's no .mli ?

hhugo:
> There's no .mli ?

Apparently not, I don't have time for this right now. We could create an issue to track it.

hhugo:
I'm not sure we need an mli for this very use case. Everything in this file is exported and most of the code is just declaration of externals. We could add doc-comment in the ml file though

PR #1333, opened 2022-11-21
Compiler: refactor Generate.compile_block 

hhugo:


vouillon:
Maybe we should consider rewriting this code based on this nice paper from last ICFP: [Beyond Relooper: recursive translation of unstructured control flow to structured control flow, Norman Ramsey](https://dl.acm.org/doi/abs/10.1145/3547621)

PR #1338, opened 2022-11-24
Compiler: some small size-optimisation 

hhugo:
- rewrite `! (a > b)` into `(a<=b)` when applied to integers  (for other comparison ops)
- remove trailing `continue` statement in for-loops
- improve `Js_simpl.if_statement` to decrease nesting.


hhugo:
With this PR, the size of the stdlib compiled to javascript decreases by ~1%. (from 267334 to  264797 bytes)

Issue #943, opened 2019-12-14
[FEATURE REQUEST] Separate compilation: only link modules that are needed (like ocamlc)  


hhugo:

hhugo:
WIP https://github.com/ocsigen/js_of_ocaml/tree/link-what-we-need

PR #1340, opened 2022-11-30
Effect handlers

vouillon:
This pull request adds support for effect handlers. This is a joint work with Olivier Nicole (@OlivierNicole) based on previous work from Armaël Guéneau (@Armael) and Patrick Ferris (@patricoferris).

The implementation is based on a CPS transform of the whole code. This a thorough implementation. In particular, you can perform effects from within the OCaml toplevel in a browser.

The graph below shows the performance impact on some short programs from the Js_of_ocaml benchmark. Short recursive functions that did not perform any allocation, such as the Fibonacci function, become much slower (up to 9 time slower). For other benchmarks, we are more between 1.5 and 2 times slower. Interestingly, code that makes heavy use of exceptions can become faster (`boyer`).
time

The code size is larger, within a factor 2.5. For a much larger example, the Ocsigen Start demo, the generated code is 80% larger. But only 26% larger when compressed with `bzip2` (we are adding  a lot of `function` and `return` that compress well).
size

We hope to be able to reduce the performance gap by performing a partial CPS transform based on a static analysis to detect code that does not perform any effects.

The compatibility with existing code should be rather high. However, the calling convention mismatch between JavaScript and OCaml is now larger since OCaml functions take a continuation as additional parameter. So, it is no longer possible to call a JavaScript function as if it was an OCaml function (and conversely). When calling a JavaScript from OCaml, you have to use `Js.Unsafe.call`. For using an OCaml function from JavaScript, you have to properly wrap it using `Js.wrap_callback` or one of the `Js.Unsafe.callback` variants.


TyOverby:
@pmwhite is looking at running Jane Streets codebase under this patch

TyOverby:
In the past, large applications have demonstrated surprising performance charactaristics when compiled with JSOO, like this issue where testing a [gameboy emulator](https://github.com/ocsigen/js_of_ocaml/issues/1185) showed some scenarios where inlining hurt runtime performance.  I wonder if it might be worth it to build and benchmark that gameboy emulator with these changes in order to see how it performs on larger programs.

vouillon:
> In the past, large applications have demonstrated surprising performance charactaristics when compiled with JSOO, like this issue where testing a [gameboy emulator](https://github.com/ocsigen/js_of_ocaml/issues/1185) showed some scenarios where inlining hurt runtime performance. I wonder if it might be worth it to build and benchmark that gameboy emulator with these changes in order to see how it performs on larger programs.

That's a good idea, indeed. It is about 7 times slower in Chrome. I guess that should be expected for this kind of code.

With support for effect handlers:
```
ROM path: ./tobu.gb
  Frames: 1500
Duration: 13.999800
     FPS: 107.144388
```
Without:
```
ROM path: ./tobu.gb
  Frames: 1500
Duration: 2.048000
     FPS: 732.421875
```

Interestingly, the generated code is only 30% larger (5% larger when compressed). So, it is possible that we can do way better for the Ocsigen Start demo I mentioned above (the lambda lifting phase currently generates a few functions with thousands of parameters, which might well have a large impact on code size...)

compiler/lib/config.ml
```diff
@@ -58,6 +58,8 @@ module Flag = struct
 
   let inline = o ~name:"inline" ~default:true
 
+  let effects = o ~name:"effects" ~default:true
```
hhugo, line 61: We should probably change the default for now.

compiler/lib/parse_bytecode.ml
```diff
@@ -913,7 +913,7 @@ and compile infos pc state instrs =
           infos
           (pc + 2)
           (State.pop 3 state)
-          (Let (x, Apply (f, args, false)) :: instrs)
+          (Let (x, Apply { f; args; exact = false; tailcall = false }) :: instrs)
```
hhugo, line 940: maybe add a helper function
```ocaml
let apply f args = Apply { f; args; exact = false; tailcall = false }
```

compiler/lib/specialize.ml
```diff
@@ -48,19 +48,26 @@ let rec function_cardinality info x acc =
 
 let specialize_instr info (acc, free_pc, extra) i =
   match i with
-  | Let (x, Apply (f, l, _)) when Config.Flag.optcall () -> (
-      let n' = List.length l in
+  | Let (x, Apply { f; args; tailcall; _ }) when Config.Flag.optcall () -> (
+      (* Field [tailcall] is set at a latter stage *)
```
hhugo, line 52: Maybe use the following helper ?
```ocaml
let apply f args ~exact = Apply { f; args; exact; tailcall = false }
```

compiler/lib/generate_closure.ml
```diff
@@ -97,15 +97,19 @@ module Trampoline = struct
     match counter with
     | None ->
         { params = []
-        ; body = [ Let (return, Apply (f, args, true)) ]
+        ; body = [ Let (return, Apply { f; args; exact = true; tailcall = false }) ]
```
hhugo, line 100: Is tailcall not yet populated ? should we clarify this  ?

runtime/jslib.js
```diff
@@ -51,6 +51,15 @@ function caml_trampoline_return(f,args) {
   return {joo_tramp:f,joo_args:args};
 }
 
+//Provides:caml_stack_depth
+var caml_stack_depth = 0;
+
+//Provides:caml_stack_check_depth
+//Requires:caml_stack_depth
+function caml_stack_check_depth() {
```
hhugo, line 59: Is this for effects only ? if so, mark it with `//If: effect`

compiler/lib/lambda_lifting.ml
```diff
@@ -0,0 +1,227 @@
...
+        let body, (program, functions) =
+          rewrite_body true (program, functions) block.body
+        in
+        ( { program with blocks = Addr.Map.add pc { block with body } program.blocks }
+        , functions ))
+    pc
+    program.blocks
+    (program, functions)
+
+let f program =
```
hhugo, line 224: Why is this only applied with effect ? 

Maybe make `threshold` configurable (using `Config.Param`) and add some expect tests in  compiler/test-compiler (with a small threshold)

compiler/lib/effects.ml
```diff
@@ -0,0 +1,528 @@
...
+  , Return ret )
+
+let cps_jump_cont ~st cont =
+  let call_block =
+    let body, branch = cps_branch ~st cont in
+    add_block st { params = []; body; branch }
+  in
+  call_block, []
+
+let memoize f =
```
hhugo, line 252: Should we move that in stdlib.ml ?

compiler/lib/driver.ml
```diff
@@ -80,6 +80,13 @@ let phi p =
   if debug () then Format.eprintf "Variable passing simplification...@.";
   Phisimpl.f p
 
+let effects p =
+  if Config.Flag.effects ()
+  then (
+    if debug () then Format.eprintf "Effects...@.";
+    Effects.f p |> inline |> deadcode |> phi |> flow |> fst |> Lambda_lifting.f)
```
hhugo, line 87: Can applying effect allow for more inlining  ? Can you provide an example ?

compiler/lib/parse_bytecode.ml
```diff
@@ -2088,64 +2112,82 @@ and compile infos pc state instrs =
...
     | STOP -> instrs, Stop, state
-    | RESUME ->
-        let _stack = State.accu state in
-        let _func = State.peek 0 state in
-        let _arg = State.peek 1 state in
-        let state = State.pop 2 state in
+    | (RESUME | RESUMETERM | PERFORM | REPERFORMTERM) when not (Config.Flag.effects ()) ->
+        if not !effect_warning
+        then (
+          warn
```
hhugo, line 2118: Should we move this warning in generate, after we give DCE a chance to remove usage of effects ?

hhugo:
> The implementation is based on a CPS transform of the whole code.

What's the story for separate compilation and dynlink ? I wonder if we should have the jsoo linker check that all js files have been compiled with compatible options. (use-js-string and effects)

hhugo:
Iv'e done a first pass of review without really looking at the commit implementing effects  yet.

compiler/lib/generate.ml
```diff
@@ -77,14 +77,26 @@ let rec map_last f l =
 (****)
 
 module Share = struct
+  module AppMap = Map.Make (struct
+    type t = int * bool * bool
+
+    let compare (n, e, t) (n', e', t') =
```
hhugo, line 83: What about `let compare = Poly.compare` ?

compiler/lib/generate.ml
```diff
@@ -77,14 +77,26 @@ let rec map_last f l =
 (****)
 
 module Share = struct
+  module AppMap = Map.Make (struct
+    type t = int * bool * bool
```
hhugo, line 81: record instead of tuple ?

compiler/lib/generate.ml
```diff
@@ -208,15 +221,26 @@ module Share = struct
...
-      try J.EVar (IntMap.find n t.vars.applies)
+      try J.EVar (AppMap.find (n, exact, tailcall) t.vars.applies)
       with Not_found ->
-        let x = Var.fresh_n (Printf.sprintf "caml_call%d" n) in
+        let x =
+          Var.fresh_n
+            (Printf.sprintf
+               "caml_%scall%d"
+               (match exact, tailcall with
+               | true, false -> assert false
```
hhugo, line 235: Why assert false ?

compiler/lib/generate.ml
```diff
@@ -817,17 +841,39 @@ let parallel_renaming params args continuation queue =
...
+    else
+      J.ECond
+        ( J.EBin (J.EqEq, J.EDot (f, "length"), int n)
+        , apply_directly
+        , ecall
+            (runtime_fun ctx "caml_call_gen")
+            [ f; J.EArr (List.map params ~f:(fun x -> Some x)) ]
+            J.N )
+  in
+  if tailcall
```
hhugo, line 861: `if tailcall && effect ()` ?

compiler/lib/code.ml
```diff
@@ -308,7 +308,12 @@ type prim_arg =
   | Pc of constant
 
 type expr =
-  | Apply of Var.t * Var.t list * bool
+  | Apply of
+      { f : Var.t
+      ; args : Var.t list
+      ; exact : bool
+      ; tailcall : bool
```
hhugo, line 315: It seems that tailcall is only used in  `Generate.translate_expr`. I wonder if it would be better to reconstruct the information on the fly by passing additional information in `Generate.translate_instrs`.
I like the use of inline record but maybe we don't need to have tailcall in it.

hhugo:
Can you include the benchmark changes used to generate the graphs.

runtime/jslib.js
```diff
@@ -50,6 +51,10 @@ function caml_trampoline_return(f,args) {
   return {joo_tramp:f,joo_args:args};
 }
 
+//Provides: caml_callback
+//Requires:caml_call_gen
```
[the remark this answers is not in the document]
vouillon: I don't think this works with conditionals, though. But maybe we should extend the alias mechanism?
```javascript
//Provides: caml_callback
//If: !effects
//Requires:caml_call_gen
```

compiler/lib/config.ml
vouillon: Indeed. I think this is convenient for testing, for now.

compiler/lib/generate.ml
vouillon: This case is handled directly in `translate_expr`.
https://github.com/ocsigen/js_of_ocaml/blob/a78b688979a5f9325e3be73f7c336be1bc3b849e/compiler/lib/generate.ml#L1075-L1077

compiler/lib/parse_bytecode.ml
vouillon: We can do that, indeed.

compiler/lib/code.ml
vouillon: The set of tail calls is now stored separately. Does this look better to you? I should probably revert the type of `Code.Apply` as well to minimize the changes in the PR.

hhugo: It looks better to me but I still wonder if it would be better to compute tailcalls in generate.ml directly. That way, we don't have to worry about other passes messing up with tailcalls after applying `Effects.t`.
 `Generate.Share.get` already performs a full traversal and could additionally track tailcalls. Is there something I'm missing that forces us to compute tailcalls in Effects ?

hhugo: > I should probably revert the type of Code.Apply as well to minimize the changes in the PR.

I find the code clearer with such change. I'm happy to keep it.

bluddy:
Will this become the default implementation or will it remain optional? Seems like it would be useful to keep the old behavior for code that makes no use of effects.

dbuenzli:
@bluddy: 

> We hope to be able to reduce the performance gap by performing a partial CPS transform based on a static analysis to detect code that does not perform any effects.

hhugo:
@vouillon, I've experimented with various changes in https://github.com/ocsigen/js_of_ocaml/commit/8dfcafb7dd5b3411abbf175b6b0464b03b297e7a

pmwhite:
I've been attempting to test this PR against Jane Street's code. However, I'm getting stuck at the very beginning...all the standard library tests fail with one of the following errors:

1. The first error I determined was caused by the continuation being passed to `caml_tail_call1` being a number instead of a function.
```
TypeError: f is not a function
  at caml_tail_call1 ... rest of stacktrace
```

2. The second error I determined was caused by trying to read `Parsexp[5]` when `Parsexp` was undefined. The issue is at least partially on my end, since `parsexp` shouldn't have been a dependency of the `lib-bool` tests, but there's also another bug, since the error should have occurred much sooner.
```
TypeError: Cannot read properties of undefined (reading '5')
  at <stacktrace>
```

I recognize that I haven't given nearly enough context to allow in-depth debugging, but maybe it is enough to pinpoint a bug? I'll keep debugging this to try to find the issue, and I'll report back if I find anything else out. The weird thing to me is that these tests pass when I build this repo using dune, but not with our internal build tool, so at least some of the problem is because of a difference in how the build is set up.

vouillon:
@pmwhite
> I've been attempting to test this PR against Jane Street's code. However, I'm getting stuck at the very beginning...all the standard library tests fail with one of the following errors:
> 
>     1. The first error I determined was caused by the continuation being passed to `caml_tail_call1` being a number instead of a function.
> 
> 
> ```
> TypeError: f is not a function
>   at caml_tail_call1 ... rest of stacktrace
> ```
> 
>     2. The second error I determined was caused by trying to read `Parsexp[5]` when `Parsexp` was undefined. The issue is at least partially on my end, since `parsexp` shouldn't have been a dependency of the `lib-bool` tests, but there's also another bug, since the error should have occurred much sooner.
> 
> 
> ```
> TypeError: Cannot read properties of undefined (reading '5')
>   at <stacktrace>
> ```
> 
> I recognize that I haven't given nearly enough context to allow in-depth debugging, but maybe it is enough to pinpoint a bug? I'll keep debugging this to try to find the issue, and I'll report back if I find anything else out. The weird thing to me is that these tests pass when I build this repo using dune, but not with our internal build tool, so at least some of the problem is because of a difference in how the build is set up.

This really looks like calling convention mismatches. Is it possible that two versions of Js_of_ocaml are involved?

Or are you using any Js_of_ocaml command line option that I may have failed to consider? 

For issue 1, this looks like a function like `Format.printf` CPS-transformed but called as if it were in direct style.

For issue 2, it seems that the initialization of module Parsexp finishes early, before registering the module, maybe because it calls a function from another library that returns a value instead of calling the continuation. I had errors like that at some point... I failed to reproduce this issue: I can compile and run without error the following piece of code:
```
(ref None) := Some Parsexp.Eager.State.create
```

I was a bit suspicious of commit 93a2f52a6304575aef7d7f8bd1a520a4f3996404 that I pushed without much test, but I did some more tests and it does not seem to break anything.

hhugo:
@pmwhite, it's possible that you use `stdlib_modern.js` which has not been updated for effects.

vouillon:
> @pmwhite, it's possible that you use `stdlib_modern.js` which has not been updated for effects.

That must be it! I have pushed an updated version of `stdlib_modern.js`.

pmwhite:
> it's possible that you use stdlib_modern.js which has not been updated for effects.

Indeed, we do; good catch! I just tried running it with that disabled, and this fixed the first error, but not the `Parsexp` error.

> That must be it! I have pushed an updated version of stdlib_modern.js.

I also just tried it with the commits you just pushed and got the same result (first error is fixed, but not the second).

compiler/lib/code.ml
vouillon: For now, we can compute tailcalls in `generate.ml` directly, indeed.

But if we CPS-transform the program only partially, we will need to keep track somehow which parts have been transformed. But hopefully the inlining phase after `Effects.f` will then no longer be necessary.

hhugo:
> > it's possible that you use stdlib_modern.js which has not been updated for effects.
> 
> Indeed, we do; good catch! I just tried running it with that disabled, and this fixed the first error, but not the `Parsexp` error.
> 
> > That must be it! I have pushed an updated version of stdlib_modern.js.
> 
> I also just tried it with the commits you just pushed and got the same result (first error is fixed, but not the second).

Can you tell us if you use separate compilation or not ?

compiler/lib/generate.ml
hhugo: If we expect tailcall to only be true with effects, maybe use an assert instead.

pmwhite:
Something that seems worth mentioning is that my build setup applies `ppx_jane` by default; it seems that all the instances of this error were caused by this setup pulling in dependencies that the Github repository does not. However, this can't be the only difference in the build, since I've been unable to get the Github repository to fail in the same way.

> Can you tell us if you use separate compilation or not ?

Yes, I have been. I just now tried building without separate compilation and got no more instances of the `Parsexp` error. I'm not sure what to make of it, since it could just mean that the unnecessary dependencies got dead-code-eliminated.

In non-separate-compilation mode, I'm now getting another error - this unexpected test diff in `lib/tests/test_fun_call.ml`:
```diff
 let%expect_test "wrap_callback_strict" =                                                  
   call_and_log                                                                            
     (Js.Unsafe.callback_with_arity 2 cb3)                                                 
     {| (function(f){ return f(1,2,3) }) |};                                               
   [%expect {|                                                                             
     Result: function#0 |}];                                                               
   call_and_log                                                                            
     (Js.Unsafe.callback_with_arity 2 cb3)                                                 
     {| (function(f){ return caml_callback(f(1,2,3),[4]) }) |};                            
-  [%expect {|                                                                             
-    got 1, 2, 4, done                                                                     
-    Result: 0 |}];                                                                        
+  [%expect.unreachable];                                                                  
   call_and_log                                                                            
     (Js.Unsafe.callback_with_arity 2 cb3)                                                 
     {| (function(f){ return caml_callback(f(1,2),[3]) }) |};                              
-  [%expect {|                                                                             
-    got 1, 2, 3, done                                                                     
-    Result: 0 |}];                                                                        
+  [%expect.unreachable];                                                                  
-  call_and_log (Js.Unsafe.callback_with_arity 2 cb3) {| (function(f){ return f(1,2) }) |};
-  [%expect {|                                                                             
-    Result: function#0 |}]                                                                
+  call_and_log (Js.Unsafe.callback_with_arity 2 cb3) {| (function(f){ return f(1,2) }) |};
+  [%expect.unreachable]                                                                   
+[@@expect.uncaught_exn {| ("ReferenceError: caml_callback is not defined") |}]            
```

compiler/lib/code.ml
hhugo: maybe the name `tailcall` is not precise enough / too generic..

hhugo:
(dune is currently broken when using jsoo inline_tests with whole program compilation, fix available in https://github.com/ocaml/dune/pull/6645)

hhugo:
> Something that seems worth mentioning is that my build setup applies `ppx_jane` by default; it seems that all the instances of this error were caused by this setup pulling in dependencies that the Github repository does not. However, this can't be the only difference in the build, since I've been unable to get the Github repository to fail in the same way.
> 
> > Can you tell us if you use separate compilation or not ?
> 
> Yes, I have been. I just now tried building without separate compilation and got no more instances of the `Parsexp` error. I'm not sure what to make of it, since it could just mean that the unnecessary dependencies got dead-code-eliminated.
> 
> In non-separate-compilation mode, I'm now getting another error - this unexpected test diff in `lib/tests/test_fun_call.ml`:
> 
> ```diff
>  let%expect_test "wrap_callback_strict" =                                                  
>    call_and_log                                                                            
>      (Js.Unsafe.callback_with_arity 2 cb3)                                                 
>      {| (function(f){ return f(1,2,3) }) |};                                               
>    [%expect {|                                                                             
>      Result: function#0 |}];                                                               
>    call_and_log                                                                            
>      (Js.Unsafe.callback_with_arity 2 cb3)                                                 
>      {| (function(f){ return caml_callback(f(1,2,3),[4]) }) |};                            
> -  [%expect {|                                                                             
> -    got 1, 2, 4, done                                                                     
> -    Result: 0 |}];                                                                        
> +  [%expect.unreachable];                                                                  
>    call_and_log                                                                            
>      (Js.Unsafe.callback_with_arity 2 cb3)                                                 
>      {| (function(f){ return caml_callback(f(1,2),[3]) }) |};                              
> -  [%expect {|                                                                             
> -    got 1, 2, 3, done                                                                     
> -    Result: 0 |}];                                                                        
> +  [%expect.unreachable];                                                                  
> -  call_and_log (Js.Unsafe.callback_with_arity 2 cb3) {| (function(f){ return f(1,2) }) |};
> -  [%expect {|                                                                             
> -    Result: function#0 |}]                                                                
> +  call_and_log (Js.Unsafe.callback_with_arity 2 cb3) {| (function(f){ return f(1,2) }) |};
> +  [%expect.unreachable]                                                                   
> +[@@expect.uncaught_exn {| ("ReferenceError: caml_callback is not defined") |}]            
> ```

I would expect the test in `lib/tests/test_fun_call.ml` to fail when using whole program compilation. They'll need to be rewritten.

pmwhite:
Yes, that was the one that I tested whole program compilation with (and also separate compilation, which is the mode in which the error occurs). To be especially clear, I'm referring to f41b9b530e3d1bb62a58ce4cfada1f369a3ed301

vouillon:
@pmwhite Would it be possible to have a look at the generated code? The beginning, up to the code generated for `parsexp.cma` (included), should be enough.

compiler/lib/lambda_lifting.ml
vouillon: We have never got any report of nested functions being an issue. While it is very much an issue after the CPS transform.

I will add a test.

pmwhite:
@vouillon I'm a bit reluctant to do that, since it would require sifting through and modifying ~60,000 lines of generated javascript to avoid leaking local file paths and stuff.

However, I was able to track down a more precise bug in the generated code that I think should help. It looks like `caml_cps_callN` is called like `caml_cps_callN(a1, a2, ..., aN, continuation)`, but it is defined like `caml_cps_callN(continuation, a1, a2, ..., aN)`. I manually edited the generated code so that `caml_cps_callN` was defined with the continuation as the last argument, so that calls would succeed, and the particular test I was editing succeeded.

You can probably guess at this, but in case it isn't obvious, the reason this causes the behavior we're seeing is that the call to `caml_register_global` for `Parsexp` happens within one of the continuations passed to `caml_cps_callN`, and thus never gets executed.

vouillon:
> However, I was able to track down the place in the generated code that causes the bug. It looks like `caml_cps_callN` is called like `caml_cps_callN(a1, a2, ..., aN, continuation)`, but it is defined like `caml_cps_callN(continuation, a1, a2, ..., aN)`. I manually edited the generated code so that `caml_cps_callN` was defined with the continuation as the last argument, so that calls would succeed, and the particular test I was editing succeeded.

@pmwhite I'm confused. What make you think that? I have the following definition, and in all cases the continuation is assumed to be last.
```JavaScript
    function caml_cps_call3(f,a0,a1,a2)
     {return runtime.caml_stack_check_depth()
              ?f.length == 3?f(a0,a1,a2):runtime.caml_call_gen(f,[a0,a1,a2])
              :runtime.caml_trampoline_return(f,[a0,a1,a2])}
```

compiler/lib/generate.ml
```diff
@@ -1016,31 +1090,38 @@ let throw_statement ctx cx k loc =
         , loc )
       ]
 
-let rec translate_expr ctx queue loc _x e level : _ * J.statement_list =
+let rec translate_expr ctx queue loc in_tail_position e level : _ * J.statement_list =
+  let cps = in_tail_position && Config.Flag.effects () in
   match e with
-  | Apply (x, l, true) ->
-      let (px, cx), queue = access_queue queue x in
+  | Apply { f; args; exact = true } when not cps ->
```
hhugo, line 1096: Is this special case still needed now that  `apply_fun` can perform direct call ?

runtime/jslib.js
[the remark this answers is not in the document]
hhugo: The issue is not with the conditional, It's that other runtime functions cannot depend on an alias.
Let's keep it this way.

runtime/jslib.js
```diff
@@ -50,6 +51,62 @@ function caml_trampoline_return(f,args) {
...
+function caml_callback(f,args) {
+  function uncaught_effect_handler(eff,k,ms) {
+    // Resumes the continuation k by raising exception Unhandled.
+    caml_resume_stack(k[1],ms);
+    var exn = caml_named_value("Effect.Unhandled")
+              || [248,caml_string_of_jsbytes("Stdlib.Effect.Unhandled"),
+                  caml_fresh_oo_id(0)];
+    caml_raise_with_arg(exn, eff);
+  }
+  var depth = caml_stack_depth;
```
hhugo, line 80:
```suggestion
  var saved_stack_depth = caml_stack_depth;
```

runtime/jslib.js
```diff
@@ -208,84 +265,107 @@ function caml_ojs_new_arr(c, a) {
...
+//Provides: caml_js_function_arity
+//If: !effects
+function caml_js_function_arity(f) {
+  return f.length
+}
+
+//Provides: caml_js_function_arity
+//If: effects
+function caml_js_function_arity(f) {
+  return f.length - 1
```
hhugo, line 366: This would probably deserve a small comment.

compiler/lib/generate.ml
hhugo: We could move the check in `apply_fun`
```diff
 let apply_fun ctx f params exact cps loc =
-  if Config.Flag.inline_callgen ()
+  (* We always go through an intermediate function when doing CPS
+     calls. This function first checks the stack depth to prevent
+     a stack overflow. This makes the code smaller than inlining
+     the test, and we expect the performance impact to be low
+     since the function should get inlined by the JavaScript
+     engines. *)
+  if Config.Flag.inline_callgen () || (exact && not cps)
   then apply_fun_raw ctx f params exact cps
   else
     let y =
```

hhugo: I've pushed this change

compiler/lib/parse_bytecode.ml
hhugo: I've pushed this change

pmwhite:
> @pmwhite I'm confused. What make you think that? I have the following definition, and in all cases the continuation is assumed to be last.
> 
> ```js
>     function caml_cps_call3(f,a0,a1,a2)
>      {return runtime.caml_stack_check_depth()
>               ?f.length == 3?f(a0,a1,a2):runtime.caml_call_gen(f,[a0,a1,a2])
>               :runtime.caml_trampoline_return(f,[a0,a1,a2])}
> ```

I assumed `f` was the continuation, but on second thought, that actually doesn't make a ton of sense. I guess `f` is the function being called, and `a2` is the continuation, which means my diagnosis is wrong. I guess there must be another reason the continuation wasn't called...stay tuned.

pmwhite:
Revised hypothesis: `caml_CamlinternalMod_init_mod` defined in `runtime/internalMod.js` is not CPS-transformed, and therefore doesn't call its continuation. This would also explain why I haven't been able to reproduce this bug using dune, since my opam switch uses 4.14, and the internal build is still on 4.12. I think the function should still have a CPS-transformed version, though, even though 4.12 won't ever support effects (is that right)?

vouillon:
@pmwhite You are right. I forgot about these functions. Thanks a lot! That should work now.

compiler/lib/config.ml
hhugo: I've pushed changes to explicitly enable effects for tests that need it.

compiler/tests-jsoo/lib-effects/dune
```diff
@@ -0,0 +1,19 @@
+(env
+ (_
+  (js_of_ocaml
+   (flags
+    (:standard --enable effects))
+    ;; separate compilation doesn't work when using
+    ;; features such as 'effects' or 'use-js-string'
+    ;; because dune doesn't know that it should compile
+    ;; multiple versions of the dependencies
+   (compilation_mode whole_program))))
```
vouillon, line 10: This seems to work, though `unhandled_unlinked` does not pass since, with separate compilation, the whole standard library is included.
```suggestion
  (js_of_ocaml
   (flags
    (:standard --enable effects))
   (build_runtime_flags (:standard --enable effects)))))
```

hhugo: related to my other question https://github.com/ocsigen/js_of_ocaml/pull/1340#issuecomment-1335261147. Is it ok to mix `--enable effects` and `--disable effects` when doing separate compilation ?

hhugo: There is #943 to track the issue mentioned above

toplevel/examples/lwt_toplevel/dune
```diff
@@ -90,6 +90,8 @@
    %{bin:js_of_ocaml}
    compile
    --pretty
+   --enable
+   effects
```
vouillon, line 94: We need to enable effects in the embedded compiler as well...

compiler/tests-jsoo/lib-effects/dune
hhugo: Once we disable effect by default, stdlib.cma.js (and other deps) will be compiled without effects support. Testing this, I'm getting lots of errors

vouillon:
> What's the story for separate compilation and dynlink ? I wonder if we should have the jsoo linker check that all js files have been compiled with compatible options. (use-js-string and effects)

Yes, we should do that. The users get really puzzling errors otherwise.
Maybe in the toplevel as well? Though I'm not sure how.

runtime/jslib.js
```diff
@@ -50,6 +51,67 @@ function caml_trampoline_return(f,args) {
...
+      caml_stack_depth = 40;
+      try {
+        res = caml_call_gen(res.joo_tramp, res.joo_args);
+      } catch (e) {
+        /* Handle exception coming from JavaScript or from the runtime. */
+        if (!caml_exn_stack.length) throw e;
+        var handler = caml_exn_stack[1];
+        caml_exn_stack = caml_exn_stack[2];
+        res = {joo_tramp: handler,
+               joo_args: [caml_wrap_exception(e)]};
```
vouillon, line 104: We should take into account flag `excwrap`, but I'm not sure how.

pmwhite:
> @pmwhite You are right. I forgot about these functions. Thanks a lot! That should work now.

Fantastic. I got a lot further with testing our code this time. Most things I've built and run the tests for seem to be working great. Everything that didn't work immediately was due to treating the OCaml and Javascript calling conventions as being the same, and were fixed easily enough by using `Js.Unsafe.fun_call` (haven't had to use `wrap_callback` yet but I'm sure I will).

toplevel/examples/lwt_toplevel/dune
hhugo: I've pushed a fix for this

runtime/jslib.js
```diff
@@ -50,6 +51,67 @@ function caml_trampoline_return(f,args) {
...
+  try {
+    caml_exn_stack = 0;
+    caml_fiber_stack =
+      {h:[0, 0, 0, uncaught_effect_handler], r:{k:0, x:0, e:0}};
+    var res = {joo_tramp: f,
+               joo_args: args.concat(function (x){return x;})};
+    do {
+      caml_stack_depth = 40;
+      try {
+        res = caml_call_gen(res.joo_tramp, res.joo_args);
```
hhugo, line 97: Isn't it too expensive to setup an exception handler every time we bounce on the trampoline ? We could have a while-loop inside the `try` for fast rebound

runtime/jslib.js
hhugo: 1. have a global in the runtime that control the behavior
2. or have conditional js runtime for caml_wrap_exception

Either way, the logic using excwrap could disappear from generate.ml

runtime/jslib.js
```diff
@@ -20,7 +20,8 @@
 ///////////// Jslib
 
 //Provides: caml_js_pure_expr const
-function caml_js_pure_expr (f) { return f(); }
+//Requires: caml_callback
+function caml_js_pure_expr (f) { return caml_callback(f, [0]); }
 
 //Provides: caml_js_set (mutable, const, const)
```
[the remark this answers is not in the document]
pmwhite: Gotcha; that makes sense. Upon further investigation, I think the issue is in gen_js_api's `caml_ojs_iterate_properties` function, which receives and ocaml function and calls it with the old calling convention. The fixes to use `wrap_callback` on the ocaml function. In short, this is just another instance of the same calling convention pitfall.

runtime/jslib.js
hhugo: see https://github.com/ocsigen/js_of_ocaml/pull/1346

runtime/jslib.js
vouillon: I think setting up an exception handler is cheap now. For a long time, V8 was not able to generate native code for function containing exception handlers, but that was fixed when they switch from Crankshaft to TurboFan about five years ago.

hhugo:
> > What's the story for separate compilation and dynlink ? I wonder if we should have the jsoo linker check that all js files have been compiled with compatible options. (use-js-string and effects)
> 
> Yes, we should do that. The users get really puzzling errors otherwise. Maybe in the toplevel as well? Though I'm not sure how.

Done in #1346

runtime/jslib.js
hhugo: Isn't the boyer benchmark still showing significant performance impact when using exception ? Is it only throwing exception that is expensive ?

vouillon: I think so, yes

runtime/jslib.js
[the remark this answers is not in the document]
vouillon: I made a PR: https://github.com/LexiFi/gen_js_api/pull/170

compiler/lib/lambda_lifting.ml
```diff
@@ -0,0 +1,224 @@
+(*
```
hhugo, line 1: Missing header

compiler/lib/lambda_lifting.mli
```diff
@@ -0,0 +1 @@
+val f : Code.program -> Code.program
```
hhugo, line 1: Missing header

compiler/lib/parse_bytecode.ml
```diff
@@ -334,19 +334,41 @@ end = struct
...
       | KUnary -> scan debug blocks code (pc + 2) len
       | KBinary -> scan debug blocks code (pc + 3) len
       | KNullaryCall ->
           let blocks =
             if Debug.mem debug (pc + 1) then Addr.Set.add pc blocks else blocks
           in
+          let blocks =
+            (* The code transformation used to deal with effect
+               handlers expects that function applications and effect
+               primitives are at the very end of a block. *)
```
hhugo, line 349: We could have a pass right before the effect  one to put the program in the required form, That way we have less invariant to preserve and we don't have to disable optimisations breaking this invariants.
Also, less blocks would probably increase compilation speed (see timing in https://github.com/ocsigen/js_of_ocaml/issues/1300#issuecomment-1276796143)

runtime/jslib.js
hhugo: #1346 has been merged

hhugo:
I've rebased the branch on master

compiler/lib/parse_bytecode.ml
vouillon: I've done that. There are some speed improvements, indeed (about 10% faster on the Ocsigen Start demo).

compiler/lib/effects.ml
```diff
@@ -0,0 +1,552 @@
...
+    let stack = Var.fresh () in
+    let k = Var.fresh () and exn_stack = Var.fresh () and b = Var.fresh () in
+    let handlers = Var.fresh () in
+    [ Let (stack, Field (continuation, 0))
+    ; Let (k, Field (fiber, 0))
+    ; Let (exn_stack, Field (fiber, 1))
+    ; Let (handlers, Field (fiber, 2))
+    ; Let (b, Block (0, [| k; exn_stack; handlers; stack |], NotArray))
+    ; Set_field (continuation, 0, b)
+    ]
```
hhugo, line 235: Why do we inline instructions instead of solely relying on primitive calls ?

compiler/lib/effects.ml
```diff
@@ -0,0 +1,552 @@
...
+      let ret = Var.fresh () in
+      ( pop_instrs @ [ Let (ret, Apply { f = exn_handler; args = [ x ]; exact = true }) ]
+      , Return ret )
+  | Stop -> [], Stop
+  | Branch cont -> cps_branch ~st cont
+  | Cond (x, cont1, cont2) ->
+      [], Cond (x, cps_jump_cont ~st cont1, cps_jump_cont ~st cont2)
+  | Switch (x, c1, c2) ->
+      (* To avoid code duplication during JavaScript generation, we need
+         to create a single block per continuation *)
```
hhugo, line 282: Shouldn't this apply to `Cond` as well ?

compiler/lib/specialize.ml
```diff
@@ -48,15 +48,22 @@ let rec function_cardinality info x acc =
...
-      let n' = List.length l in
+  | Let (x, Apply { f; args; _ }) when Config.Flag.optcall () -> (
+      let n' = List.length args in
       match function_cardinality info f [] with
       | None -> i :: acc, free_pc, extra
-      | Some n when n = n' -> Let (x, Apply (f, l, true)) :: acc, free_pc, extra
+      | Some n when n = n' ->
+          Let (x, Apply { f; args; exact = true }) :: acc, free_pc, extra
       | Some n when n < n' ->
+          (* We skip this optimization for now since it generates a
```
hhugo, line 58: Stale comment

compiler/lib/parse_bytecode.ml
hhugo: > about 10% faster on the Ocsigen Start demo.

Is this will or without debug events ? (e.g. --pretty)

compiler/lib/effects.ml
hhugo: `cont` are probably always different with `Cond`. Maybe add a comment or an assert ?

hhugo:
Tests can now run with both ocaml.4.14 and ocaml.5.

compiler/lib/effects.ml
vouillon: They are always different. I have added an `assert`.

compiler/lib/effects.ml
vouillon: There are only three occurrences of`%reperform`, so the code size was not critical.

For `%perform`, you need to call the effect handler from the primitive if you want a significant size gain. Otherwise, the primitive returns many results that you have to extract using inline instructions.

lib/tests/test_fun_call.ml
```diff
@@ -150,13 +150,13 @@ let%expect_test "wrap_callback_strict" =
     Result: function#0 |}];
   call_and_log
     (Js.Unsafe.callback_with_arity 2 cb3)
-    {| (function(f){ return f(1,2,3)(4) }) |};
+    {| (function(f){ return caml_callback(f(1,2,3),[4]) }) |};
```
hhugo, line 153: These tests don't work with `dune runtest --profile release` because short-var is enabled and `caml_callback` is renamed.

compiler/lib/effects.ml
```diff
@@ -0,0 +1,436 @@
...
+          ||
+          match branch with
+          | Branch _ -> false
+          | Return x' -> (
+              (not (Var.equal x x'))
+              ||
+              match i with
+              | Let (_, Prim (Extern "%perform", _)) -> true
+              | _ -> false)
+          | _ -> true)
```
hhugo, line 336: I don't follow lines 334 and 336. Could you add a comment ?

compiler/lib/effects.ml
```diff
@@ -0,0 +1,436 @@
...
+      | _ -> false
+    in
+    let rec split (p : Code.program) pc block accu l branch =
+      match l with
+      | [] ->
+          let block = { block with body = List.rev (l @ accu) } in
```
hhugo, line 342:
```suggestion
          let block = { block with body = List.rev accu } in
```

compiler/lib/effects.ml
```diff
@@ -0,0 +1,436 @@
...
+            @@ fun ret k' ->
+            [ Let
+                ( ret
+                , Prim (Extern "caml_perform_effect", [ Pv effect; Pc (Int 0l); Pv k' ])
+                )
+            ]
+        | _ -> None)
+    | None, _
+    | Some (_, (Set_field _ | Offset_ref _ | Array_set _ | Assign _)), _
+    | _, (Raise _ | Stop | Cond _ | Switch _ | Pushtrap _ | Poptrap _) -> None
```
hhugo, line 300:
```suggestion
    | Some _, (Raise _ | Stop | Cond _ | Switch _ | Pushtrap _ | Poptrap _) -> None
```

hhugo:
Thanks for all the work on this.

The manual probably deserves some changes as well
- overview,wiki and README,md could mention the support for effects
- The graphs from the description of the PR could be added to the performance page
-  tailcall.wiki should mention that there is general TCO when using effects
- ..

compiler/lib/effects.ml
vouillon: `cps_block` was not rewriting `%perform` when followed by a return. I have added a case in `cps_block` to handle this case and removed this code.

compiler/lib/effects.ml
```diff
@@ -0,0 +1,416 @@
...
+  let split_block pc block p =
+    let is_split_point i r branch =
+      match i with
+      | Let (x, (Apply _ | Prim (Extern ("%resume" | "%perform" | "%reperform"), _))) -> (
+          (not (List.is_empty r))
+          ||
+          match branch with
+          | Branch _ -> false
+          | Return x' -> not (Var.equal x x')
+          | _ -> true)
```
hhugo, line 316: Do we need to split for `Raise` and `Stop` ?

compiler/lib/effects.ml
```diff
@@ -0,0 +1,415 @@
...
+      Let (x, Closure (params @ [ st.closure_continuation pc ], (pc, args)))
+  | Let (x, Prim (Extern "caml_alloc_dummy_function", [ size; arity ])) -> (
+      match arity with
+      | Pc (Int a) ->
+          Let
+            ( x
+            , Prim (Extern "caml_alloc_dummy_function", [ size; Pc (Int (Int32.succ a)) ])
+            )
+      | _ -> assert false)
+  | Let (_, (Apply _ | Prim (Extern ("%resume" | "%perform"), _))) -> assert false
```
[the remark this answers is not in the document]
vouillon: We don't really need to since it can only occur in tail position in the standard library, in functions that cannot be inlined.
But it does not cost much and can only make the code more robust by including it as well.

compiler/lib/effects.ml
vouillon: Function `cps_block` only handles function calls and primitives followed by `Branch` and `Return`. For `Raise`, we have to generate code that is basically what we obtain by splitting anyway. It does not seem to be very useful to have a special case for `Stop`. I'm not even sure we can have a function call right before a `Stop`.

hhugo:
I've looked at timings compiling the toplevel with and without effects.
I think theses results are kind of expected.  There is significant increase for any step looking at js free variables and js scopes (coloring, checks, ..) and steps looking at number of closures (generate_closure).
  
```diff
 |Start Optimizing...
-|  tail calls: 0.04
 |    flow analysis 1: 0.15
-|    flow analysis 2: 0.24
+|    flow analysis 2: 0.21
-|    flow analysis 3: 0.12
+|    flow analysis 3: 0.08
-|    flow analysis 4: 0.48
+|    flow analysis 4: 0.47
-|    flow analysis 5: 0.67
+|    flow analysis 5: 0.73
-|  flow analysis: 1.65
+|  flow analysis: 1.64
-|  dead code elim.: 0.35
-|  inlining: 0.14
-|  dead code elim.: 0.55
-|  tail calls: 0.04
+|  dead code elim.: 0.35
+|  inlining: 0.14
+|  dead code elim.: 0.54
-|    phi-simpl. 1: 0.33
+|    phi-simpl. 1: 0.32
 |    phi-simpl. 2: 0.41
 |  phi-simpl.: 0.86
-|    flow analysis 1: 0.12
+|    flow analysis 1: 0.11
 |    flow analysis 2: 0.09
 |    flow analysis 3: 0.03
 |    flow analysis 4: 0.08
-|    flow analysis 5: 0.25
+|    flow analysis 5: 0.26
-|  flow analysis: 0.59
+|  flow analysis: 0.58
-|  dead code elim.: 0.30
+|  dead code elim.: 0.29
-|  inlining: 0.13
+|  inlining: 0.12
-|  dead code elim.: 0.51
+|  dead code elim.: 0.59
-|    flow analysis 1: 0.25
+|    flow analysis 1: 0.20
-|    flow analysis 2: 0.19
+|    flow analysis 2: 0.21
-|    flow analysis 3: 0.12
+|    flow analysis 3: 0.13
-|    flow analysis 4: 0.15
+|    flow analysis 4: 0.11
-|    flow analysis 5: 0.22
+|    flow analysis 5: 0.20
-|  flow analysis: 0.94
+|  flow analysis: 0.85
-|  dead code elim.: 0.26
-|  inlining: 0.13
+|  dead code elim.: 0.26
+|  inlining: 0.12
-|  dead code elim.: 0.27
+|  dead code elim.: 0.26
 |    phi-simpl. 1: 0.05
 |    phi-simpl. 2: 0.05
 |  phi-simpl.: 0.15
 |    flow analysis 1: 0.06
-|    flow analysis 2: 0.06
+|    flow analysis 2: 0.05
 |    flow analysis 3: 0.03
 |    flow analysis 4: 0.04
-|    flow analysis 5: 0.83
+|    flow analysis 5: 0.89
-|  flow analysis: 1.02
+|  flow analysis: 1.07
+|  effects: 0.56
+|  dead code elim.: 0.66
+|  inlining: 0.68
+|  dead code elim.: 1.43
+|    phi-simpl. 1: 0.29
+|    phi-simpl. 2: 0.37
+|  phi-simpl.: 0.79
+|    flow analysis 1: 0.14
+|    flow analysis 2: 0.11
+|    flow analysis 3: 0.06
+|    flow analysis 4: 0.09
+|    flow analysis 5: 0.35
+|  flow analysis: 0.74
+|  lambda lifting: 1.38
-|  free vars: 0.39
+|  free vars: 0.85
-|  generate closures: 0.50
+|  generate closures: 1.18
-|  dead code elim.: 0.31
+|  dead code elim.: 0.88
-| optimizations : 9.31
+| optimizations : 16.75
-|Start Generation...
-|Warning: your program contains effect handlers; you should probably run js_of_ocaml with option '--enable=effects'
-|  code gen.: 1.01
+|Start Generation...
+|  code gen.: 1.75
 |Start Linking...
-|  linking: 0.35
+|  linking: 1.50
 |Start Optimizing js...
-|    simpl: 0.63
+|    simpl: 0.16
-|    clean: 0.31
+|    clean: 0.15
-|  optimizing: 1.30
+|  optimizing: 1.11
 |Start Coloring...
-|  coloring: 1.22
+|  coloring: 5.41
 |Start Checks...
-|  checks: 0.61
+|  checks: 2.30
 |Start Writing file...
-|  write: 0.47
+|  write: 0.99
-|compilation: 17.90
+|compilation: 33.45
```

hhugo:
> Thanks for all the work on this.
> 
> The manual probably deserves some changes as well
> 
> * overview,wiki and README,md could mention the support for effects
> * The graphs from the description of the PR could be added to the performance page
> * tailcall.wiki should mention that there is general TCO when using effects
> * ..

Last item before we merge ?

hhugo:
@pmwhite, do you have time to test the latest version of this PR ?

compiler/lib/driver.ml
```diff
@@ -29,7 +29,7 @@ let should_export = function
 
 let tailcall p =
   if debug () then Format.eprintf "Tail-call optimization...@.";
-  Tailcall.f p
+  if Config.Flag.effects () then p else Tailcall.f p
```
hhugo, line 32: `if debug () ...` should probably go under the `else` branch

hhugo:
@vouillon, I've made some changes to the runtime. can you take a look ?

pmwhite:
@hhugo yeah, I've got time to test the latest commits, though it seems like there are still some trailing commits happening, so maybe I should wait until those are all done before doing a final test (it's not _that_ easy for me to test out a particular commit).

I haven't been paying too close of attention to all the changes since the last test, but am I right in guessing that I will need to use `--enable effects` to turn on effects explicitly, since the default will remain the non-effect version of js_of_ocaml? I should also probably the non-effect version as well, just to sanity check that nothing broke there.

hhugo:
> @hhugo yeah, I've got time to test the latest commits, though it seems like there are still some trailing commits happening, so maybe I should wait until those are all done before doing a final test (it's not _that_ easy for me to test out a particular commit).

The trailing commits are mostly polishing ones. I think the PR is in a merge-able state.  

> I haven't been paying too close of attention to all the changes since the last test, but am I right in guessing that I will need to use `--enable effects` to turn on effects explicitly, since the default will remain the non-effect version of js_of_ocaml? I should also probably the non-effect version as well, just to sanity check that nothing broke there.

Correct, `--enable effects` is no longer the default and you'll need to specify it explicitly.

pmwhite:
Testing without `--enable effects` went fine. _with_ `--enable effects`, I got a crash with the following error:

```
compiler/bin-js_of_ocaml/js_of_ocaml.exe: You found a bug. Please report it at https://github.com/ocsigen/js_of_ocaml/issues :
      Error: File "effects.ml", line 177, characters 6-12: Assertion failed
      Raised at Js_of_ocaml_compiler__Effects.cps_last in file "effects.ml", line 177, characters 6-43
      Called from Js_of_ocaml_compiler__Effects.cps_block in file "effects.ml", line 293, characters 32-60
      Called from Js_of_ocaml_compiler__Effects.f.(fun) in file "effects.ml", line 381, characters 30-77
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 589, characters 14-22
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.fold_children in file "code.ml", line 564, characters 17-27
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.fold_children in file "code.ml", line 564, characters 17-27
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.fold_children in file "code.ml", line 564, characters 17-27
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse'.(fun) in file "code.ml", line 585, characters 29-71
      Called from Stdlib__Array.fold_right in file "array.ml", line 174, characters 9-30
      Called from Js_of_ocaml_compiler__Code.traverse' in file "code.ml", line 581, characters 6-192
      Called from Js_of_ocaml_compiler__Code.traverse in file "code.ml" (inlined), line 593, characters 40-87
      Called from Js_of_ocaml_compiler__Effects.f.(fun) in file "effects.ml", line 378, characters 10-243
      Called from Stdlib__List.fold_left in file "list.ml", line 125, characters 24-34
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 19-42
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 26-41
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 26-41
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 26-41
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 26-41
      Called from Stdlib__Map.Make.fold in file "map.ml", line 325, characters 26-41
      Called from Js_of_ocaml_compiler__Effects.f in file "effects.ml", line 363, characters 4-882
      Called from Js_of_ocaml_compiler__Effects.f in file "effects.ml" (inlined), line 414, characters 10-13
      Called from Js_of_ocaml_compiler__Driver.effects in file "driver.ml", line 87, characters 4-15
      Called from Js_of_ocaml_compiler__Driver.(+>) in file "driver.ml" (inlined), line 94, characters 21-26
      Called from Js_of_ocaml_compiler__Driver.full in file "driver.ml", line 525, characters 4-105
      Called from Js_of_ocaml_compiler__Driver.(+>) in file "driver.ml" (inlined), line 94, characters 21-26
      Called from Js_of_ocaml_compiler__Driver.full in file "driver.ml" (inlined), line 525, characters 4-122
      Called from Js_of_ocaml_compiler__Driver.full in file "driver.ml", line 542, characters 10-15
      Called from Js_of_ocaml_compiler__Stdlib.Filename.gen_file in file "stdlib.ml", line 737, characters 11-15
      Re-raised at Js_of_ocaml_compiler__Stdlib.Filename.gen_file in file "stdlib.ml", line 740, characters 9-16
      Re-raised at Js_of_ocaml_compiler__Stdlib.Filename.gen_file in file "stdlib.ml", line 746, characters 6-15
      Called from Bin_prefix_js_of_ocaml__Compile.run.output in file "compile.ml", line 170, characters 8-609
      Called from Bin_prefix_js_of_ocaml__Compile.run.output_partial in file "compile.ml" (inlined), line 206, characters 4-60
      Called from Bin_prefix_js_of_ocaml__Compile.run in file "compile.ml", line 308, characters 8-45
      Called from Cmdliner__Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
      Called from Cmdliner__Cmdliner_eval.run_parser in file "cmdliner_eval.ml" (inlined), line 34, characters 37-44
      Called from Cmdliner__Cmdliner_eval.eval_value in file "cmdliner_eval.ml", line 195, characters 14-39
      Re-raised at Cmdliner__Cmdliner_eval.eval_value in file "cmdliner_eval.ml", line 195, characters 14-39
      Called from Bin_prefix_js_of_ocaml__Js_of_ocaml.(fun) in file "js_of_ocaml.ml", line 48, characters 8-395
      Called from Bin_prefix_js_of_ocaml__Js_of_ocaml in file "js_of_ocaml.ml", line 46, characters 4-821
      Called from Bin_prefix_js_of_ocaml__Js_of_ocaml in file "js_of_ocaml.ml", line 46, characters 4-821
```

TyOverby:
@vouillon I believe that Safari implements TCO for JavaScript, so it might be interesting to benchmark the effects feature in a browser that might not need trampolining

vouillon:
@pmwhite I pushed a fix

hhugo:
> @pmwhite I pushed a fix

@vouillon aren't we duplicateting blocks then ?

runtime/compare.js
```diff
@@ -235,25 +235,25 @@ function caml_compare_val (a, b, total) {
 //Provides: caml_compare (const, const)
 //Requires: caml_compare_val
 function caml_compare (a, b) { return caml_compare_val (a, b, true); }
-//Provides: caml_int_compare mutable (const, const)
+//Provides: caml_int_compare (const, const)
```
vouillon, line 240: I'm not sure we should make this change. We are preventing some optimizations just to get an error in some bogus code that does not even use the result of the comparison.

pmwhite:
I tested after @vouillon's commit and found now issues, although I wasn't able to build and run as many tests as usual because it's difficult to run a CI build with different compiler flags.

pmwhite:
Is there a way I can figure out the location source code that a continuation came from? I was hoping to scan over the code that triggered the assertion to see why it was triggered. I know the library that the error is occurring in, but unfortunately it is a bit too large to scan over the whole thing for suspicious control flow.

runtime/compare.js
hhugo: ok, I'll revert

vouillon:
@pmwhite
> Is there a way I can figure out the location source code that a continuation came from? I was hoping to scan over the code that triggered the assertion to see why it was triggered. I know the library that the error is occurring in, but unfortunately it is a bit too large to scan over the whole thing for suspicious control flow.

I think that the generated bytecode contains an unnecessary branch, that jumps to the very next instruction.  I guess you could find it using `ocamldumpobj`, but I don't think it is worth investigating. I thought this was not possible, so I put an assert failure if it happened, that I was not able to trigger. We are generating a somewhat suboptimal code in this case, but it is so rare that it does not really matter...

runtime/marshal.js
```diff
@@ -231,8 +230,6 @@ function caml_input_value_from_reader(reader, ofs) {
     if (code >= 0x40 /*cst.PREFIX_SMALL_INT*/) {
       if (code >= 0x80 /*cst.PREFIX_SMALL_BLOCK*/) {
         var tag = code & 0xF;
-        if (caml_is_continuation_tag(tag))
```
hhugo: We need to make sure your ocaml PR makes it into 5.0

runtime/compare.js
hhugo: done

hhugo:
> This is so rare that I don't think we really care. Actually, I'm wondering how they managed to to that. Anyway, if we want to handle this case, we should rather turn `Cond (x, cont, cont)` into `Branch` in an earlier phase.

I've done that in `eval.ml`.

pmwhite:
> > This is so rare that I don't think we really care. Actually, I'm wondering how they managed to to that. Anyway, if we want to handle this case, we should rather turn `Cond (x, cont, cont)` into `Branch` in an earlier phase.
> 
> I've done that in `eval.ml`.

Should we put back the assertion, then? Or are there other cases where it could happen?

hhugo:
> > > This is so rare that I don't think we really care. Actually, I'm wondering how they managed to to that. Anyway, if we want to handle this case, we should rather turn `Cond (x, cont, cont)` into `Branch` in an earlier phase.
> > 
> > 
> > I've done that in `eval.ml`.
> 
> Should we put back the assertion, then? Or are there other cases where it could happen?

One can still do `--disable eval` to skip that pass.

runtime/marshal.js
hhugo: Closing to not block the merge

vouillon:
> @vouillon I believe that Safari implements TCO for JavaScript, so it might be interesting to benchmark the effects feature in a browser that might not need trampolining

@TyOverby Indeed, I need to have a look at that. Thanks!

vouillon:
> Ok, I'm pretty happy with the state of this PR. I think ti's ready to be merged. I'm not sure what to do with the history. It doesn't have to be perfect but a little bit of cleanup could be good.

@hhugo I have cleaned up the history. Thanks a lot for your review and all the work you put into it. That was tremendously useful!

hhugo:
Thanks everyone for the work!!!

hhugo:
@vouillon, how is your effects-optim branch progressing. Do you think we should wait for it before the next release ?

vouillon:
@hhugo It probably still needs several weeks of work. So I don't think we should wait for it.

hhugo:
@vouillon, changing the baseline of lambda_lifting from 1 to 2 seem to reduce the amount of free variable quiet a bit.
Compiling ocamlc, the size goes from 3M to 2.7M 

Using es6 lambda could reduce the size (uncompressed) a bit
`(function (x) { return ... })`  => `(x => ...)`

vouillon:
> @vouillon, changing the baseline of lambda_lifting from 1 to 2 seem to reduce the amount of free variable quiet a bit. Compiling ocamlc, the size goes from 3M to 2.7M

Well spotted! See #1360 for an alternative fix.

> Using es6 lambda could reduce the size (uncompressed) a bit `(function (x) { return ... })` => `(x => ...)`

Yes, indeed.

hhugo:
> Using es6 lambda could reduce the size (uncompressed) a bit
> `(function (x) { return ... })` => `(x => ...)`

I did a quick experiment, `ocamlc` compiled with `effects` goes from 2.7M to ~2.4M.

compiler/tests-jsoo/lib-effects/dune
hhugo: This restriction will be gone with the next dune 3.7 (see https://github.com/ocaml/dune/pull/6714)

PR #1346, opened 2022-12-07
Prepare for #1340

hhugo:


vouillon:
What do you think of statically simplifying `if (FLAG(...)) { ... } else {...}`?

hhugo:
> What do you think of statically simplifying `if (FLAG(...)) { ... } else {...}`?

I wrote the initial version assuming we might want tune the behavior at runtime in the future. That said, I'm happy with your version.

compiler/lib/js_traverse.ml
```diff
@@ -1017,6 +1022,8 @@ class simpl =
       let s = super#statements s in
       List.fold_right s ~init:[] ~f:(fun (st, loc) rem ->
           match st with
+          | If_statement (ENum n, iftrue, iffalse) ->
+              if Num.is_zero n then opt_cons iffalse rem else iftrue :: rem
```
hhugo, line 1026: This is not quiet correct because is_zero only check for `"0"` we need to consider `-0`, `nan`, `0.0`, ...
We could either use `Num.to_int32` or implement and use Num.to_float

vouillon: Indeed! I'm just going to test for `0` and `1` for now.

PR #1348, opened 2022-12-08
Compiler: more blocks, more debug

hhugo:


hhugo:
I'm not sure this should be merged, I don't remember why I disabled this code previously.

hhugo:
It's not clear to me why the lambda_lifting test is affected this way.

vouillon:
> It's not clear to me why the lambda_lifting test is affected this way.

If you have more blocks, the CPS transform will generate more functions. Since inlining is disabled for this test, you see them in the output.

vouillon:
We need to have debugging information only for the locations where the JavaScript debugger might stop, which does not necessarily corresponds to where the OCaml debugger would stop. So it is not clear to me that we need more.

hhugo:
> > It's not clear to me why the lambda_lifting test is affected this way.
> 
> If you have more blocks, the CPS transform will generate more functions. Since inlining is disabled for this test, you see them in the output.

That doesn't explain why `g` is no longer lifted.

vouillon:
> > > It's not clear to me why the lambda_lifting test is affected this way.
> > 
> > 
> > If you have more blocks, the CPS transform will generate more functions. Since inlining is disabled for this test, you see them in the output.
> 
> That doesn't explain why `g` is no longer lifted.

This is probably because there are now two consecutive function definitions. They are not lifted, in case they are mutually recursive.

Issue #1354, opened 2022-12-10
[BUG] Miscompilation of code involving references and exceptions


vouillon:
I have found out what the block handler was for... This following piece of code is miscompiled. The issue is that the OCaml bytecode program does not allocate a reference but update the stack in place. But we no longer keep track of this update.
```ocaml
let x = ref 0 in
try
  incr x;
  raise Exit
with Exit ->
  Format.eprintf "%d@." !x
```
You have to compile it without the `-g` option. For instance:
```shell
ocamlc bug.ml -o bug.byte && js_of_ocaml bug.byte && ./bug.byte && node bug.js
```
The above command outputs `1` (correct) then `0` (wrong):

vouillon:
I think that this piece of code, with nested exception handlers, has always been miscompiled.
```ocaml
let x = ref 0 in
try
  (try incr x ; raise Exit with _ -> ());
  raise Exit
with Exit ->
  Format.eprintf "%d@." !x
```

hhugo:
I wonder if we could make that logic more explicit by keeping track of `ASSIGN`ed variable that have been previously captured in an exception handler. Maybe using a new instruction 
```ocaml
type instr =
  | Let of Var.t * expr
  | Set_field of Var.t * int * Var.t
  | Offset_ref of Var.t * int
  | Array_set of Var.t * Var.t * Var.t
  | Update_pushtrap_cont  of Var.t * Var.t
```

hhugo:
The following program was never compiled correctly.
```ocaml
let x = ref 0 in
try
  incr x ;
  (try incr x ; raise Exit with _ -> Format.eprintf "%d@." !x);
  raise Exit
with Exit ->
  Format.eprintf "%d@." !x
```

I have wip fix in #1355

PR #1355, opened 2022-12-10
Compiler: demonstrate bug from 1354

hhugo:
https://github.com/ocsigen/js_of_ocaml/issues/1354

### DIFF
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -317,6 +317,7 @@ type expr =
 
 type instr =
   | Let of Var.t * expr
+  | Assign of Var.t * Var.t
   | Set_field of Var.t * int * Var.t
   | Offset_ref of Var.t * int
   | Array_set of Var.t * Var.t * Var.t
@@ -452,6 +453,7 @@ module Print = struct
   let instr f i =
     match i with
     | Let (x, e) -> Format.fprintf f "%a = %a" Var.print x expr e
+    | Assign (x, y) -> Format.fprintf f "(assign) %a = %a" Var.print x Var.print y
     | Set_field (x, i, y) -> Format.fprintf f "%a[%d] = %a" Var.print x i Var.print y
     | Offset_ref (x, i) -> Format.fprintf f "%a[0] += %d" Var.print x i
     | Array_set (x, y, z) ->
@@ -635,6 +637,7 @@ let invariant { blocks; start; _ } =
       | Let (x, e) ->
           define x;
           check_expr e
+      | Assign _ -> ()
       | Set_field (_, _i, _) -> ()
       | Offset_ref (_x, _i) -> ()
       | Array_set (_x, _y, _z) -> ()
--- a/compiler/lib/flow.ml
+++ b/compiler/lib/flow.ml
@@ -109,6 +109,7 @@ let program_deps { blocks; _ } =
               add_var vars x;
               add_expr_def defs x e;
               expr_deps blocks vars deps defs x e
+          | Assign (x, y) -> add_dep deps x y
           | Set_field _ | Array_set _ | Offset_ref _ -> ());
       match block.branch with
       | Return _ | Raise _ | Stop -> ()
@@ -237,6 +238,7 @@ let program_escape defs known_origins { blocks; _ } =
       List.iter block.body ~f:(fun i ->
           match i with
           | Let (x, e) -> expr_escape st x e
+          | Assign (x, _) -> possibly_mutable.(Var.idx x) <- true
           | Set_field (x, _, y) | Array_set (x, _, y) ->
               Var.Set.iter
                 (fun y -> possibly_mutable.(Var.idx y) <- true)
--- a/compiler/lib/flow.ml
+++ b/compiler/lib/flow.ml
@@ -109,7 +109,9 @@ let program_deps { blocks; _ } =
               add_var vars x;
               add_expr_def defs x e;
               expr_deps blocks vars deps defs x e
-          | Assign (x, y) -> add_dep deps x y
+          | Assign (x, y) ->
+              add_dep deps x y;
+              add_assign_def vars defs x y
           | Set_field _ | Array_set _ | Offset_ref _ -> ());
       match block.branch with
       | Return _ | Raise _ | Stop -> ()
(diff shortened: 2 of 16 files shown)


compiler/tests-compiler/util/util.ml
```diff
@@ -444,12 +444,12 @@ let compile_and_run_bytecode ?unix s =
...
 
-let compile_and_run ?flags ?unix s =
+let compile_and_run ?debug ?flags ?unix s =
   with_temp_dir ~f:(fun () ->
       s
       |> Filetype.ocaml_text_of_string
       |> Filetype.write_ocaml ~name:"test.ml"
-      |> compile_ocaml_to_bc ?unix
+      |> compile_ocaml_to_bc ?debug ?unix
       |> compile_bc_to_javascript ?flags
```
vouillon, line 453:
```suggestion
      |> compile_bc_to_javascript ?flags ~sourcemap:debug
```

compiler/tests-compiler/gh1354.ml
```diff
@@ -0,0 +1,196 @@
...
+        var
+         global_data=runtime.caml_get_global_data(),
+         Stdlib=global_data.Stdlib,
+         Stdlib_Format=global_data.Stdlib__Format,
+         _c_=[0,[4,0,0,0,[17,4,0]],caml_string_of_jsbytes("%d@.")],
+         _b_=[0,[4,0,0,0,[17,4,0]],caml_string_of_jsbytes("%d@.")],
+         _a_=0;
+        try
+         {try
+           {var _d_=_a_ + 1 | 0;0;_a_ = _d_;_a_ = _d_;throw Stdlib[3]}
```
hhugo, line 181: duplicated `assign` need to be fixed  `_a_ = _d_;_a_ = _d_`

compiler/tests-compiler/gh1354.ml
```diff
@@ -0,0 +1,196 @@
...
+let%expect_test _ =
+  let prog =
+    {|
+let x = ref 0 in
+let p x = Format.eprintf "%d@." x in
+try
+  incr x;
+  raise Exit
+with Exit ->
+  p !x|}
```
vouillon, line 31:
```suggestion
  p (!x + 0)|}
```

compiler/lib/flow.ml
```diff
@@ -237,6 +238,7 @@ let program_escape defs known_origins { blocks; _ } =
       List.iter block.body ~f:(fun i ->
           match i with
           | Let (x, e) -> expr_escape st x e
+          | Assign (x, _) -> possibly_mutable.(Var.idx x) <- true
```
vouillon, line 241: We want `maybe_unknown` to be true (that is, we do not know all possible values of `x`), not `possibly_mutable` to be true (the fields of the values of `x` may be modified).

compiler/lib/freevars.ml
```diff
@@ -53,6 +53,7 @@ let iter_instr_free_vars f i =
       f x;
       f y;
       f z
+  | Assign (_, _) -> ()
```
vouillon, line 56:
```suggestion
  | Assign (x, y) -> f x; f y
```

compiler/lib/pure_fun.ml
```diff
@@ -34,6 +34,7 @@ let pure_expr pure_funs e =
 let pure_instr pure_funs i =
   match i with
   | Let (_, e) -> pure_expr pure_funs e
+  | Assign _ -> false
```
vouillon, line 37: The side-effect cannot be observed from outside the function
```suggestion
  | Assign _ -> true
```

vouillon:
This outputs 1...
```ocaml
let x = ref 0 in
let y = !x in
try
  incr x;
  raise Exit
with Exit ->
  Format.eprintf "%d@." y
```

compiler/lib/flow.ml
vouillon: Or if we have `Assign (x, y)`, `y` should be a known origin of `x`

vouillon:
The intuition is that `Assign (x, y)` should be like passing an argument `y` to a parameter `x`.

compiler/lib/freevars.ml
vouillon: Actually only `y` is free. `x` is a bound variable (of the exception handler block).

compiler/lib/code.ml
```diff
@@ -317,6 +317,7 @@ type expr =
 
 type instr =
   | Let of Var.t * expr
+  | Assign of Var.t * Var.t
```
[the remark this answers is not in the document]
vouillon: Probably not worth it, though...

compiler/tests-compiler/gh1354.ml
vouillon: This is so rare that it is probably not worth it...

compiler/lib/generate.ml
```diff
@@ -1508,6 +1514,13 @@ and compile_block_no_loop st queue (pc : Addr.t) loop_stack frontier interm =
       let try_catch_frontier = Addr.Set.union new_frontier handler_frontier_cont in
       if debug () then Format.eprintf "@[<hv 2>try {@;";
       if Interm.mem pc1 handler_interm then decr_preds st pc1;
+      (* Parameters of the exception handler are assigned here, not in
+         handler body. Indeed, they correspond to variables that are
+         going to be modified by an instruction Assign. *)
+      let _, assign_handler_variables =
```
vouillon, line 1520: Alternatively, we could rewrite the code, adding an intermediate branch before the `Pushtrap` than binds the exception handler's variable. Maybe as soon as in `Parse_bytecode`? Then, we do not have to do anything special here.

I'll have to do this transformation to handle effect handlers anyway.

compiler/lib/js_assign.ml
```diff
@@ -316,6 +316,11 @@ let program' (module Strategy : Strategy) p =
   let color = function
     | V v ->
         let name = names.(Var.idx v) in
+        let name =
+          match name with
+          | "" -> "escape"
+          | n -> n
+        in
```
hhugo, line 323: This diff was only for debugging

hhugo:
I recreated the branch by mistake, will reopen an PR.

PR #1356, opened 2022-12-11
Compiler: demonstrate and fix bug from 1354

### LINKED ISSUE #1354: [BUG] Miscompilation of code involving references and exceptions

hhugo:
continues/replaces #1355 
fix #1354

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 * Runime: Gc.finalise_last should not be eliminated
 * Toplevel: Make sure the toplevel uses the correct memory representaion for strings
 * Compiler: fix minifier, missing constraint on try-catch blocks.
+* Compiler: Miscompilation of code involving references and exceptions (#1354, #1356)
 
 # 4.1.0 (2022-11-15) - Lille
 ## Features/Changes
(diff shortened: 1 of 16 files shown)


vouillon:
This works much better than I would have thought. With the right intuition (this is just argument passing), the changes to the optimization phases are straightforward. The complexity is localized in `Parse_bytecode`.

hhugo:
Thanks for the fix. I've stashed some commit. Do you have more on this PR ? Can we merge ?

vouillon:
> Thanks for the fix. I've stashed some commit. Do you have more on this PR ? Can we merge ?

I think we can merge

Issue #1246, opened 2022-03-04
[FEATURE REQUEST] Unsafe apply annotation (avoiding caml_call)


wlitwin:
**Is your feature request related to a problem? Please describe.**
Js_of_ocaml's implementation of `caml_callN` is very costly because it calls `f.length` to check for partial application (I believe?). It seems Js_of_ocaml can properly eliminate these caml_calls when inside the same module/library, but across library boundaries it has issues. For some of my projects I have tight loops and this `f.length` is causing a lot of performance problems.

**Describe the solution you'd like**
I'd like some annotation I can add to an application to inform Js_of_ocaml that it should generate a normal application, and not a call to caml_call. E.g.:

```ocaml
let lib_fun1 t =
    (t.intf.fun1 t.impl)[@js_of_ocaml.unsafe_apply]
```

Right now it would generate the following code:
```javascript
    function lib_fun1(t){return caml_call1(t[1][2],t[2])}
```

when I would like:
```javascript
    function lib_fun1(t){return t[1][2](t[2])}
```

**Describe alternatives you've considered**
From what I gather it's difficult to get some of this information due to how Js_of_ocaml works, so this seems like a decent middle ground. I'm open to other solutions.

----

I'm also willing to put in the work to make this happen, but a small description/introduction to what places I would need to modify would be helpful.

wlitwin:
I see, I apologize for my naivety. It seems my issue is not library/module boundaries, but indirection. The case I'm trying to optimize is abstracting over two implementations. For example the architecture looks something like this:

```mermaid
flowchart TD
A[Userland Lib] --> B[Backend 1 Common]
A --> C[Backend 2 Common]
B --> D[Backend 1 Web]
B --> E[Backend 1 Desktop]
C --> F[Backend 2 Web]
C --> G[Backend 2 Desktop]
```

Originally I was using functors, but that proved to be too heavy, and then I tried first class modules and records of functions, but I can't seem to get rid of the caml_calls in the intermediate libraries. Flambda (for desktop) removed them in almost all the cases, but I'm trying to coax Js_of_ocaml to do the same.

Here's a simplified example of what I was trying to do: https://github.com/wlitwin/caml_call_example In this example the functions in `Lib` have to go through `caml_call` when calling the implementations. Maybe there's a better way to go about this?

hhugo:
#1358 should have fixed the perf issue.

PR #1358, opened 2022-12-15
Apply functions: optimizations

vouillon:
Getting the arity of a function is slow with `v8`. On Chrome, function `p` is about ten times slower than function `q`:
```javascript
function f (x) {}
f.l = 1;
var s = 0;
function p() {for (i = 0; i < 100000000; i++) { ((f.length)==1?s+=1:0)} }
function q(){for (i = 0; i < 100000000; i++) { ((f.l)==1?s+=1:0)}}
p(); q()
```
image

`node` takes 3.1s to execute this piece of code without this optimization and 1.3s with it (and 0,8s when manually removing the call to `caml_call1`).
```ocaml
let l = List.of_seq (Seq.init 10000 (fun i -> i))
let s = ref 0
let f x = s := !s + x
let iter f  = for i = 1 to 10000 do List.iter f l; s := 0; List.iter f l done
let () = iter f
```

I have known about this issue for a long time. I was considering wrapping functions inside an object: `{arity:2,fun:function(...){..}}`, with some optimization for when the function is statically known. But that would had been a large change with some impact on the generated code size and of lot of added complexity in the `Js_of_ocaml` compiler. This change seems to address the issue at a very low cost.

Related issue:  #1246

### DIFF
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -25,7 +25,7 @@ function caml_call_gen(f, args) {
     return caml_call_gen(f.fun, args);
   //FIXME, can happen with too many arguments
   if(typeof f !== "function") return f;
-  var n = f.length | 0;
+  var n = (f.l >= 0)?f.l:(f.l = f.length);
   if(n === 0) return f.apply(null,args);
   var argsLen = args.length | 0;
   var d = n - argsLen | 0;
@@ -76,7 +76,7 @@ function caml_call_gen(f, args) {
   if (f.fun)
     return caml_call_gen(f.fun, args);
   if (typeof f !== "function") return args[args.length-1](f);
-  var n = f.length | 0;
+  var n = (f.l >= 0)?f.l:(f.l = f.length);
   if (n === 0) return f.apply(null, args);
   var argsLen = args.length | 0;
   var d = n - argsLen | 0;
--- a/runtime/stdlib.js
+++ b/runtime/stdlib.js
@@ -35,13 +35,35 @@ function caml_call_gen(f, args) {
     return caml_call_gen(f.apply(null,args.slice(0,n)),args.slice(n));
   }
   else {
-    var g = function (){
-      var extra_args = (arguments.length == 0)?1:arguments.length;
-      var nargs = new Array(args.length+extra_args);
-      for(var i = 0; i < args.length; i++ ) nargs[i] = args[i];
-      for(var i = 0; i < arguments.length; i++ ) nargs[args.length+i] = arguments[i];
-      return caml_call_gen(f, nargs)
-    };
+    switch (d) {
+    case 1: {
+      var g = function (x){
+        var nargs = new Array(argsLen + 1);
+        for(var i = 0; i < argsLen; i++ ) nargs[i] = args[i];
+        nargs[argsLen] = x;
+        return f.apply(null, nargs)
+      };
+      break;
+    }
+    case 2: {
+      var g = function (x, y){
+        var nargs = new Array(argsLen + 2);
+        for(var i = 0; i < argsLen; i++ ) nargs[i] = args[i];
+        nargs[argsLen] = x;
+        nargs[argsLen + 1] = y;
+        return f.apply(null, nargs)
+      };
+      break;
+    }
+    default: {
+      var g = function (){
+        var extra_args = (arguments.length == 0)?1:arguments.length;
+        var nargs = new Array(args.length+extra_args);
+        for(var i = 0; i < args.length; i++ ) nargs[i] = args[i];
+        for(var i = 0; i < arguments.length; i++ ) nargs[args.length+i] = arguments[i];
+        return caml_call_gen(f, nargs)
+      };
+    }}
     g.l = d;
     return g;
   }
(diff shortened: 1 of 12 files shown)


hhugo:
Is this draft on purpose ? what's the plan ?

pmwhite:
> f.length is slow with v8, so we cache the function arity in another property

Also, I'd be very interested in seeing some documentation of this aspect of v8. How do you know `f.length` is slow?

vouillon:
> Is this draft on purpose ? what's the plan ?

I would like to do more benchmarking.

vouillon:
> Also, I'd be very interested in seeing some documentation of this aspect of v8. How do you know `f.length` is slow?

I have not seen any documentation on this. I just noticed that somehow. This was also reported in #1246.

TyOverby:
I pulled this change into our repo and saw a huge improvement for microbenchmarks (50% runtime reduction in some cases) but much smaller (2% - 3%) improvement in our benchmarks that attempt to replicate real-world applications and workloads.   I suspect that this is because the fallback case in `caml_call_gen` returns a function that has a length of 0, which means that partially-applied functions don't get length caching.

I suspect that if `caml_callgen` checked to see how many arguments are left unapplied and returned functions specialized to that size, then the the arity-cache proposed here would be hit far more often.

TyOverby:
> partially-applied functions don't benefit from length caching

This is because if `f.l == 0` then the first argument to the `||` operator in `f.l || f.l = f.length` will evaluate to `false`, causing the right-hand-side of the operator to be evaluated unconditionally.

hhugo:
not sure about the performance, but one could use a shorter sytnax 
```
f.l >= 0 ? f.t : f.t=f.lenght
```

instead of

```
f.l === underfined ? f.l = f.lenght : f.l
```

runtime/stdlib.js
```diff
@@ -70,14 +94,40 @@ function caml_call_gen(f, args) {
...
+      var g = function (){
+        var extra_args = (arguments.length == 0)?1:arguments.length;
+        var nargs = new Array(argsLen + extra_args);
+        for(var i = 0; i < argsLen; i++ ) nargs[i] = args[i];
+        for(var i = 0; i < arguments.length; i++ )
+          nargs[argsLen + i] = arguments[i];
+        return caml_call_gen(f, nargs)
+      };
+    }}
+    g.l = d;
```
hhugo, line 129: the functions `g` defined above have arity 2 when d = 1, and arity 3 when d = 2

runtime/stdlib_modern.js
```diff
@@ -68,13 +92,39 @@ function caml_call_gen(f, args) {
...
+      var g = function (){
+        var extra_args = (arguments.length == 0)?1:arguments.length;
+        var nargs = new Array(argsLen + extra_args);
+        for(var i = 0; i < argsLen; i++ ) nargs[i] = args[i];
+        for(var i = 0; i < arguments.length; i++ )
+          nargs[argsLen + i] = arguments[i];
+        return caml_call_gen(f, nargs)
+      };
+    }}
+    g.l = d;
```
hhugo, line 127: Same issue here.

runtime/stdlib.js
vouillon: Indeed! Thanks!

runtime/stdlib_modern.js
```diff
@@ -33,13 +33,37 @@ function caml_call_gen(f, args) {
...
+    default: {
+      var g = function (){
+        var extra_args = (arguments.length == 0)?1:arguments.length;
+        var nargs = new Array(args.length+extra_args);
+        for(var i = 0; i < args.length; i++ ) nargs[i] = args[i];
+        for(var i = 0; i < arguments.length; i++ ) nargs[args.length+i] = arguments[i];
+        return caml_call_gen(f, nargs)
+      };
+    }}
+    g.l = d + 1;
```
hhugo, line 65:
```suggestion
    g.l = d
```

hhugo:
`caml_js_function_arity` should probably be updated as well to use (and set ?) the cache.

hhugo:
```javascript
function caml_call_gen(f, args) {
  if(f.fun)
    return caml_call_gen(f.fun, args);
  //FIXME, can happen with too many arguments
  if(typeof f !== "function") return f;
  var n = f.length | 0;
...
```
Shouldn't we also use the cache for computing `n` ?

TyOverby:
I pulled this PR in for prototyping and added @hhugo's suggestion

```diff
-  var n = f.length | 0;
+  var n = (f.l >= 0 ? f.l : f.l = f.length) | 0;
```

And the benchmarks I mentioned 

> smaller (2% - 3%) improvement in our benchmarks that attempt to replicate real-world applications and workloads

saw improvements in the 10% to 30% range!

vouillon:
> I pulled this PR in for prototyping and added @hhugo's suggestion
> 
> ```diff
> -  var n = f.length | 0;
> +  var n = (f.l >= 0 ? f.l : f.l = f.length) | 0;
> ```
> 
> And the benchmarks I mentioned
> 
> > benchmarks that attempt to replicate real-world applications and workloads saw smaller (2% - 3%) improvements
> 
> saw improvements in the 10% to 30% range!

@TyOverby Thanks a lot for your tests! I'm surprised this makes that much of a difference since this is on a slow path. Maybe that's the optimization for partially applied functions (33c3af845a642773d03f25b914ee4096dfa8f0fd) that makes such a difference?

TyOverby:
> I'm surprised this makes that much of a difference since this is on a slow path. Maybe that's the optimization for partially applied functions (https://github.com/ocsigen/js_of_ocaml/commit/33c3af845a642773d03f25b914ee4096dfa8f0fd) that makes such a difference?

I totally agree that it's almost all due to returning functions with specialized arity when partially applied.  I just wanted to mention that I cherry-picked a change that wasn't already in the PR.

TyOverby:
I traced a few apps and found this distribution for the `caml_call{n}` functions

| n | occurances |
| - | ---------- |
| 1 | 109        |
| 2 | 300        |
| 3 | 30         |
| 4 | 4          |
| 5 | <1         |

Based on this, I think I'm happy with only building specialized partially-applied functions for the the cases where only 1 or 2 args remain.

hhugo:
@vouillon, should we move this PR forward ?

hhugo:
I've rebased the PR, fix tests and rewrote history

PR #1359, opened 2022-12-15
CI: tune rules for effects

hhugo:


compiler/tests-ocaml/lib-effects/dune
```diff
@@ -1,4 +1,7 @@
 (env
+ (using-effects
```
hhugo, line 2: With this, we'll run the test with separate compilation as well.

toplevel/examples/lwt_toplevel/dune
```diff
@@ -83,15 +83,25 @@
...
+   (run ./effects_flags.exe))))
+
 (rule
  (targets toplevel.js)
  (action
   (run
    %{bin:js_of_ocaml}
    compile
    --pretty
-   --enable
```
hhugo, line 93: Only use effects with ocaml 5.

PR #1360, opened 2022-12-16
Lambda-lifting: ensure that functions are move to a more appropriate place

vouillon:
Functions are moved too high by lambda lifting, adding a huge number of parameters. This has a significant impact on code size.
 
This change ensures that lambda lifting moves function to a more appropriate level, which does not depend on compiling options.

With this change, the code generated for the Ocisgen Start demo is 55% larger  (instead of 80% larger). And when compressed with `bzip2` this code is only 4% larger (instead of 26% larger).

This is a temporary fix. A more robust solution will be to avoid a CPS transformation of the code at toplevel, which is not necessary since we know that it is not within an effect handler.


hhugo:
Can you explain a bit why the change fixes the issue ?

hhugo:
Maybe we should update the graph in `performance.wiki`  if the size difference is significant ?

vouillon:
Without this change, a block can be added at the beginning of the code (in particular for a standalone program):
```
==== 54705 () ====
   v41579 = "caml_fs_init"()
   branch 54704 ()
```
This branch is turned into a function call to a function containing basically the whole code, which is not inlined since it is so large. With a baseline of 1, functions are lifted outside of this large function.

One could have set a baseline of 2 instead. But with this change, functions are always lifted to the same position.

vouillon:
> Maybe we should update the graph in `performance.wiki` if the size difference is significant ?

The benchmarks are so small that no lambda lifting is performed.

Maybe I should add something in the documentation about the code size increase for large programs?

hhugo:
> Without this change, a block can be added at the beginning of the code (in particular for a standalone program):
> 
> ```
> ==== 54705 () ====
>    v41579 = "caml_fs_init"()
>    branch 54704 ()
> ```
> 
> This branch is turned into a function call to a function containing basically the whole code, which is not inlined since it is so large. With a baseline of 1, functions are lifted outside of this large function.
> 
> One could have set a baseline of 2 instead. But with this change, functions are always lifted to the same position.

I didn't realize we'were performing cps for every Branch. Couldn't we skip the transformation when there is not call and there is only one edge to the continuation ?  

> The benchmarks are so small that no lambda lifting is performed.

make -C benchmarks copy-extra-bc should give you bigger programs.

vouillon:
> I didn't realize we'were performing cps for every Branch. Couldn't we skip the transformation when there is not call and there is only one edge to the continuation ?

I'm working on it... Because of inlining, this is not really a problem.

> > The benchmarks are so small that no lambda lifting is performed.
> 
> make -C benchmarks copy-extra-bc should give you bigger programs.

I have updated the manual with these programs.

hhugo:
You haven't updated size-effects.png

vouillon:
> You haven't updated size-effects.png

Indeed! Thanks! This is done, now.

PR #1362, opened 2022-12-16
Alternative to #1360

hhugo:
#1360

### DIFF
--- a/compiler/lib/effects.ml
+++ b/compiler/lib/effects.ml
@@ -38,14 +38,20 @@
 open! Stdlib
 open Code
 
+type simple_branch =
+  | Simple
+  | No
+
 type graph =
   { succs : (Addr.t, Addr.Set.t) Hashtbl.t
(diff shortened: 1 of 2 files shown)


compiler/lib/effects.ml
```diff
@@ -289,13 +302,30 @@ let cps_block ~st ~k pc block =
...
+        then
+          let body =
+            List.map block.body ~f:(fun i -> cps_instr ~st i) @ alloc_jump_closures
+          in
+          body, block.branch
+        else
+          let last_instrs, last = cps_last ~st block.branch ~k in
+          let body =
+            List.map block.body ~f:(fun i -> cps_instr ~st i)
+            @ alloc_jump_closures
```
vouillon, line 325: I think that in this case `alloc_jump_closures` defines a single function whose body is the block we are jumping to.
```suggestion
```

vouillon:
That should work, but I would rather go with a simple fix for now

PR #1365, opened 2022-12-19
Test: new failing test + fix

hhugo:
Example taken from 
https://github.com/ocaml-multicore/effects-examples/blob/master/transaction.ml

There is suspicious stack overflow when runing in js.
```diff
$ dune runtest compiler/tests-jsoo/lib-effects/ --profile using-effects
Warning: overriding the purity of the primitive Base_am_testing: pure -> mutator
File "compiler/tests-jsoo/lib-effects/transaction.ml", line 1, characters 0-0:
------ compiler/tests-jsoo/lib-effects/transaction.ml
++++++ compiler/tests-jsoo/lib-effects/transaction.ml.corrected
File "compiler/tests-jsoo/lib-effects/transaction.ml", line 75, characters 0-1:
 |open Txn
 |
 |let%expect_test _ =
 |  atomically (fun () ->
 |      let r = ref 10 in
 |      printf "T0: %d\n" (!r);
 |      try atomically (fun () ->
 |          r := 20;
 |          r := 21;
 |          printf "T1: Before abort %d\n" (!r);
 |          raise (Res !r) |> ignore;
 |          printf "T1: After abort %d\n" (!r);
 |          r := 30)
 |      with
 |      | Res v -> printf "T0: T1 aborted with %d\n" v;
 |        printf "T0: %d\n" !r);
-|  [%expect {|
-|    T0: 10
-|    udpate
-|    restore
-|    udpate
-|    restore
-|    T1: Before abort 21
-|    before raise
-|    raise
-|    T0: T1 aborted with 21
-|    T0: 10 |}]
+|  [%expect.unreachable]
+|[@@expect.uncaught_exn {|
+|  ("Stack overflow")
+|  Trailing output
+|  ---------------
+|  T0: 10
+|  udpate
+|  restore
+|  udpate
+|  restore
+|  T1: Before abort 21
+|  before raise
+|  before raise |}]
```


hhugo:
@vouillon, can you take a look

vouillon:
@hhugo I have pushed a fix.
Thanks for testing!

hhugo:
Thanks.

Is there any check we can do in the runtime to detect some of the situations that are definitely wrong and print an error to the user ?

vouillon:
> Is there any check we can do in the runtime to detect some of the situations that are definitely wrong and print an error to the user ?

I don't see any simple test. We could check that `caml_call_gen` is only ever applied with a one-parameter function as last parameter. But we have to change `caml_callback` first, since the trampoline calls `caml_call_gen` without respecting this condition. And I think this would only catch internal errors, not errors from our users.

PR #1368, opened 2022-12-20
Misc: bump minimum ocaml version

hhugo:


hhugo:
We already depend on a version of cmdliner that only support ocaml 4.08 and higher.

PR #1371, opened 2022-12-22
Inlining: try...cactch is no longer an issue

vouillon:


hhugo:
We should update or remove the stale comment then.
```
                      At the moment, V8 doesn't optimize function
                      containing try..catch.  We disable inlining if the
                      inner and outer functions don't have the same
                      "try_catch" property *)
```

PR #1373, opened 2023-01-02
Js_assign: add debugging code

vouillon:
I have found this useful to debug escaping variables


hhugo:
Don't you also need to change the assert below ? 
```ocaml
  let color = function
    | V v ->
        let name = names.(Var.idx v) in
        assert (not (String.is_empty name));
        ident ~var:v name
    | x -> x
  in
```

vouillon:
I don't really care whether there is an error afterwards. But having the list of escaping variables and being able to see where they occur in the code is really useful.

PR #1374, opened 2023-01-02
Effects: improvements

vouillon:
This is in preparation for a partial CPS transform.

The CPS transform now introduce new functions only when necessary. (It was creating a function for each block, relying on an inlining phase to eliminate a lot of them.) We also handle function calls at toplevel especially (using `caml_callback`) to avoid nesting funtions unnecessarily. As a consequence, the lambda lifting phase does not have much work to do any longer.

I don't expect any change in the performance of the generated code. But the generated code is smaller and compilation times are lower.


compiler/lib/effects.ml
```diff
@@ -249,16 +301,22 @@ let cps_last ~st (last : last) ~k : instr list * last =
...
+        in
+        let body, branch = cps_branch ~st (pc, args) in
+        push_trap :: body, branch
+      else [], last
+  | Poptrap (pc', args) ->
+      if Addr.Set.mem (Hashtbl.find st.matching_exn_handler pc) st.blocks_to_transform
+      then
+        let exn_handler = Var.fresh () in
+        let body, branch = cps_branch ~st (pc', args) in
+        [ Let (exn_handler, Prim (Extern "caml_pop_trap", [])) ] @ body, branch
```
hhugo, line 320:
```ocaml
(Let (exn_handler, Prim (Extern "caml_pop_trap", [])) :: body), branch
```

compiler/lib/effects.ml
```diff
@@ -493,12 +494,24 @@ let cps_transform ~live_vars p =
           ; live_vars
           }
         in
-        let k = closure_continuation start in
+        let function_needs_cps =
+          Option.is_some name_opt || not (Addr.Set.is_empty blocks_to_transform)
+        in
+        if not function_needs_cps then wrap_toplevel := false;
```
hhugo, line 502: I would have expected the opposite.
I would make it clear that
`not function_needs_cps` can only be true when when looking at the toplevel closure  

```ocaml
let wrap_toplevel = ref false in
...
if toplevel && function_needs_cps then wrap_toplevel := true;
...
```

compiler/lib/effects.ml
```diff
@@ -493,12 +494,24 @@ let cps_transform ~live_vars p =
           ; live_vars
           }
         in
-        let k = closure_continuation start in
+        let function_needs_cps =
+          Option.is_some name_opt || not (Addr.Set.is_empty blocks_to_transform)
```
hhugo, line 500: `Option.is_some name_opt` would probably deserve a small comment.

compiler/lib/effects.ml
hhugo:
```ocaml
let function_need_cps =
  match name_opt with
  | None -> (* toplevel closure *)
    let need_cps =  not (Addr.Set.is_empty blocks_to_transform) in
    wrap_toplevel := needs_cps
    needs_cps
  | Some _ -> ..
```

compiler/lib/effects.ml
```diff
@@ -227,55 +221,61 @@ let cps_block ~st ~k pc block =
...
-            ])
+          (fun ~k ->
+            let x = Var.fresh () in
+
+            ( [ Let
+                  ( x
+                  , Prim (Extern "caml_perform_effect", [ Pv effect; Pc (Int 0l); Pv k ])
+                  )
+              ]
+            , Return x ))
```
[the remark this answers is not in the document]
vouillon: We would only use it at three places. Maybe I could define a `perform_effect` function instead that would be used both for `%perform` and `%reperform`?

compiler/lib/effects.ml
```diff
@@ -182,7 +210,17 @@ let tail_call ?(instrs = []) ~f ?(exact = true) args =
   let ret = Var.fresh () in
   instrs @ [ Let (ret, Apply { f; args; exact }) ], Return ret
 
-let cps_branch ~st (pc, args) = tail_call ~f:(closure_of_pc ~st pc) args
+let cps_branch ~st (pc, args) =
+  let args, instrs =
+    if List.is_empty args && Hashtbl.mem st.is_continuation pc
+    then
+      (* We are jumping to a block that is also used as a continuation.
+         We pass it a dummy argument. *)
```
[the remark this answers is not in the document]
vouillon: I will add some tests, indeed

compiler/lib/effects.ml
```diff
@@ -132,6 +228,10 @@ type st =
   ; blocks : Code.block Addr.Map.t
   ; jc : jump_closures
   ; closure_continuation : Addr.t -> Var.t
+  ; blocks_to_transform : Addr.Set.t
+  ; is_continuation : (Addr.t, [ `Param of Var.t | `Loop ]) Hashtbl.t
+  ; matching_exn_handler : (Addr.t, Addr.t) Hashtbl.t
+  ; live_vars : int array
```
hhugo, line 234:
```suggestion
  ; live_vars : Deadcode.variable_uses
```

compiler/lib/effects.ml
```diff
@@ -220,63 +344,96 @@ let cps_block ~st ~k pc block =
     match Addr.Map.find pc st.jc.closures_of_alloc_site with
     | to_allocate ->
         List.map to_allocate ~f:(fun (cname, jump_pc) ->
-            let jump_block = Addr.Map.find jump_pc st.blocks in
-            let fresh_params = List.map jump_block.params ~f:(fun _ -> Var.fresh ()) in
-            Let (cname, Closure (fresh_params, (jump_pc, fresh_params))))
+            let params =
+              let jump_block = Addr.Map.find jump_pc st.blocks in
+              if List.is_empty jump_block.params && Hashtbl.mem st.is_continuation jump_pc
```
hhugo, line 349: Can we have a small comment explaining what's going on here ?

vouillon: I have added a short comment above and updated the comment below

compiler/lib/effects.ml
```diff
@@ -364,6 +392,84 @@ let cps_transform p =
...
+let rewrite_toplevel_instr (p, accu) instr =
+  match instr with
+  | Let (x, Apply { f; args; _ }) -> wrap_call p x f args accu
+  | Let (x, (Prim (Extern ("%resume" | "%perform" | "%reperform"), _) as e)) ->
+      wrap_primitive p x e accu
+  | _ -> p, [ instr ] :: accu
+
+(* Wrap function calls inside [caml_callback] at toplevel to avoid
+   unncessary function nestings. This is not done inside loops since
+   using repeatedly [caml_callback] can be costly. *)
```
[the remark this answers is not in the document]
vouillon: I don't expect many loops at toplevel, so they should not have much impact.
The code is rather straightforward. We just have to rewrite some instructions without having to change the control flow graph. It considered wrapping loops within a `caml_callback` but this seems a lot more complicated. In particular, one has to deal with loops with multiple outgoing edges.

hhugo:
Thanks for adding the tests. Adding expect tests first would allow to review how they are affected by each commit.

compiler/tests-compiler/effects_continuations.ml
```diff
@@ -0,0 +1,174 @@
...
+             None
+
+         (* Conditional whose result is used *)
+         let cond1 b =
+           let ic = if b then open_in "toto" else open_in "titi" in
+           (ic , 7)
+
+         (* Conditional whose result is not used *)
+         let cond2 b =
+           if b then Format.eprintf "toto" else Format.eprintf "toto";
```
hhugo, line 47: Using `Printf` instead of `Format` should make the test pass with 4.14 and 5.0

compiler/lib/effects.ml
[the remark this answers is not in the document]
hhugo: One of the downside is that it makes the output harder to read.

The issue with multiple outgoing edges is annoying. For toplevel loops, it seems that we would want to find a unique block dominated by the all the outgoing edges that have a `Stop` in their successors (?)

Is `Tailcall.f` the only reason we can have loops with multiple exit edges ? if so, inlining less at toplevel could be a solution.

I'm happy to skip the optimisation for loops for now.

hhugo:
I think you've accidentally squashed two commits into one 
`Effects: merge the two functions` https://github.com/ocsigen/js_of_ocaml/pull/1374/commits/4a7c57d2891dd907c2e0666c5e19e6b85ba4bb44
now contains the code for wrapping toplevel calls into caml_callback

hhugo:
Should we update the performance wiki page if the size if significantly reduced ?

vouillon:
> Should we update the performance wiki page if the size if significantly reduced ?

Indeed. I have also updated the time results since the change is actually significant for some programs.

hhugo:
A changelog entry and go to go

Issue #1375, opened 2023-01-02
False positive `your program contains effect handlers` warning?


johnridesabike:
**Describe the bug**

I recently tested a project with OCaml 5 and JSOO 5 and it printed this warning:

`Warning: your program contains effect handlers; you should probably run js_of_ocaml with option '--enable=effects'`

I'm almost certain that this is incorrect. None of the project's code has changed since using OCaml 4.14, and it only uses a few small dependencies. I opened the JavaScript output and saw that the `Effect` module functions are included, but do not seem to be called anywhere.

**Expected behavior**

The repository for my project: https://github.com/johnridesabike/acutis

Expected behavior:
- Run `dune build`
- Everything builds with no warnings.

Actual behavior: the above warning prints.

**Versions**

ocamlc: 5.0.0
js_of_ocaml: 5.0.1

hhugo:
This is due to the use of separate compilation. When compiling the stdlib to JS, we don't know yet whether the Effect module will be used or not later. Also, as tracked by https://github.com/ocsigen/js_of_ocaml/issues/943, with separate compilation, we unconditionally link all modules of libraries.

In the short term, you can build with `--profile release` which disable separate compilation of JS. or change the compilation mode in the env stanza (see https://dune.readthedocs.io/en/stable/jsoo.html?highlight=separate#separate-compilation).

Once #943 is fixed, we should be able to improve the situation and only warn during the linking phase.

johnridesabike:
Thanks for the information! I can confirm that `--profile release` does indeed build without the warning. 

It may be useful to include that information in the documentation and/or readme. I can imagine that other people may see the warning and incorrectly assume that they need to use `--enable=effects` when they actually do not.

hhugo:
This was fixed in #1378. We still need to wait for a dune release including https://github.com/ocaml/dune/pull/6832 and then change the default value for `auto-link` in jsoo.

PR #1378, opened 2023-01-05
Separate compilation: Link only what we need

### LINKED ISSUE #943: [FEATURE REQUEST] Separate compilation: only link modules that are needed (like ocamlc)

### LINKED ISSUE #1316: Compiler: Remove mutable fields from Source_map.t

hhugo:
Fix #943 and #1375
- `js_of_ocaml link` now accept a `--linkall` flags.
-  cma files are now compiled one compilation unit / module at a time. We no longer perform cross-module optimization with separate compilation. All compilation units of a cma are still stored in a single file.
- We only warn about `Warning: your program contains effect handlers; you should probably run js_of_ocaml with option '--enable=effects' if the offending module end up being linked.
- The resulting files should be much reduced (sourcemap is shrinked as well)
- In addition, this PR fix sourcemap with separate compilation and fix #1316 

ISSUE 
- The dune rules for separate compilation currently completely ignore `-linkall` information. In practice, with this PR, it means that inline_tests are broken because the library containing the tests is no longer linked.

TODO
- [x] check that sourcemap are still correct.


hhugo:
Dune integration discussed in https://github.com/ocaml/dune/pull/6832

hhugo:
@pmwhite @TyOverby, could you check this PR ? You should look at artifact sizes and compilation times with separate compilation. Compilation rules will need to be updated to propagate `-linkall` to `js_of_ocaml link`

pmwhite:
This is an exciting patch. I can test this out, hopefully sometime today.

pmwhite:
I'm getting the following build warnings from this patch:
```
Warning: '--source-map' is enabled but the bytecode program was compiled with no debugging information.
Warning: Consider passing '-g' option to ocamlc.
```
I'm assuming this patch has changed something to cause those messages to be printed, since we weren't getting these warnings before. Is this new warning expected? If so, what is the right response? One obvious answer is to add `-g` like the message says, but that has its own set of trade-offs.

pmwhite:
Oh, it's worth saying that I didn't get to the point of adding the `linkall` flag to the build yet; maybe that error goes away if I add the flag?

hhugo:
> I'm getting the following build warnings from this patch:
> 
> ```
> Warning: '--source-map' is enabled but the bytecode program was compiled with no debugging information.
> Warning: Consider passing '-g' option to ocamlc.
> ```
> 
> I'm assuming this patch has changed something to cause those messages to be printed, since we weren't getting these warnings before. Is this new warning expected? If so, what is the right response? One obvious answer is to add `-g` like the message says, but that has its own set of trade-offs.

This warning is not new. it used to trigger on empty cma but it was fixed quite sometime ago. I suspect that it now triggers because we compile module from a cma individually and one of the module doesn't have debug info.

pmwhite:
> This warning is not new. it used to trigger on empty cma but it was fixed quite sometime ago. I suspect that it now triggers because we compile module from a cma individually and one of the module doesn't have debug info.

I see. If I understand right, this is a bug in this PR, since it is okay for individual modules to not have any debug info.

hhugo:
> > This warning is not new. it used to trigger on empty cma but it was fixed quite sometime ago. I suspect that it now triggers because we compile module from a cma individually and one of the module doesn't have debug info.
> 
> I see. If I understand right, this is a bug in this PR, since it is okay for individual modules to not have any debug info.

It's ok but could result in worse sourcemap. One could argue that the warning is correct. Is there a good reason for omitting the -g for some modules only ?

hhugo:
> > Is there a good reason for omitting the -g for some modules only ?
> 
> I'm assuming our build system doesn't selectively omit -g (I'll double-check this assumption; EDIT: assumption seems correct), so either
> 
> * none of the modules have debug-info, which happens when the build is in release mode. EDIT: actually, we always pass -g, even in release mode
> * all of the modules have debug-info, but some modules are so small that their debug-info is empty. Is that possible?

Compiling an empty module doesn't trigger the warning. Could you give details about the command triggering the warning ?

pmwhite:
Inside `compiler/tests-jsoo`, we run the command
```
../bin-js_of_ocaml/js_of_ocaml.exe -I . -o ./jsoo_testsuite.cma.js --enable with-js-error --source-map-inline --pretty ./jsoo_testsuite.cma
```
Which spits out the warning twice. A lot of other commands have the same warning, but this one is representative.

hhugo:
> Inside `compiler/tests-jsoo`, we run the command
> 
> ```
> ../bin-js_of_ocaml/js_of_ocaml.exe -I . -o ./jsoo_testsuite.cma.js --enable with-js-error --source-map-inline --pretty ./jsoo_testsuite.cma
> ```
> 
> Which spits out the warning twice. A lot of other commands have the same warning, but this one is representative.

Are you still running ocaml 4.12 ?

hhugo:
How is `./jsoo_testsuite.cma` compiled ?

hhugo:
@pmwhite, looking at old jenga-rules available on github, I suspect that you still don't pass `-g` when compiling `.ml-gen` to cmo.

pmwhite:
> Are you still running ocaml 4.12 ?

No, 4.14 now.

> How is `./jsoo_testsuite.cma` compiled ?

This is the slightly sanitize version of the command that the build system runs.
```
ocamlc.opt -w @a-4-29-40-41-42-44-45-48-52-56-58-59-66-67-69-70 -strict-sequence -short-paths -strict-formats -g -error-style short -bin-annot -compat-32 jsoo_testsuite__.cmo jsoo_testsuite__Calc_parser.cmo jsoo_testsuite__Calc_lexer.cmo jsoo_testsuite__Parser_1307.cmo jsoo_testsuite__Lexer_1307.cmo jsoo_testsuite__Gh_1307.cmo jsoo_testsuite__Test_bigarray.cmo jsoo_testsuite__Test_channel.cmo jsoo_testsuite__Test_floats.cmo jsoo_testsuite__Test_fma.cmo jsoo_testsuite__Test_ints.cmo jsoo_testsuite__Test_io.cmo jsoo_testsuite__Test_marshal.cmo jsoo_testsuite__Test_obj.cmo jsoo_testsuite__Test_parsing.cmo jsoo_testsuite__Test_rec_mod.cmo jsoo_testsuite.cmo -cclib -ljsoo_testsuite_stubs -a -o jsoo_testsuite.cma
```

hhugo:
> @pmwhite, looking at old jenga-rules available on github, I suspect that you still don't pass `-g` when compiling `.ml-gen` to cmo.

Check the command generating `jsoo_testsuite__.cmo`

hhugo:
@pmwhite, My last commit disable the feature by default so that the PR can be release without silently breaking inline_tests with dune. For testing you can enable it with the `--enable auto-link` flag to `js_of_ocaml link`.

pmwhite:
I realized I may not be understanding what this PR actually does. I was under the impression that we already were trimming unnecessary files, and that this moved the granularity down to the module level instead (this is just based on the description - I haven't read the code in the PR). However, now I'm thinking that this actually merely extends the tree-shaking from whole-program compilation into separate compilation as well. Is that the right way to think about it? Are the inclusion semantics for whole and separate compilation now exactly the same?

hhugo:
> Are the inclusion semantics for whole and separate compilation now exactly the same?

This is the indent 

> However, now I'm thinking that this actually merely extends the tree-shaking from whole-program compilation into separate compilation as well. Is that the right way to think about it?

Both whole program compilation and separate compilation perform some deadcode elimination based on what they know. The DCE should not remove any side-effects.

When doing whole program compilation, we rely on the link ouput of ocamlc which does drop unused compilations units present in cma files. This PR implement the same drop logic inside jsoo when doing separate compilation.

pmwhite:
Just as an update on the testing of this PR: I've not yet done all the testing I plan to, since it's a bit of work to get the build system to pass link-all in the right places. I'll update this thread when I eventually get to it.

hhugo:
> Just as an update on the testing of this PR: I've not yet done all the testing I plan to, since it's a bit of work to get the build system to pass link-all in the right places. I'll update this thread when I eventually get to it.

Any update?

pmwhite:
@hhugo No, not yet. It's on my stack of things to do, but I haven't gotten around to it yet.

Issue #1034, opened 2020-07-12
[FEATURE REQUEST] Better static JavaScript String compilation.


dbuenzli:
It seems we are not far from being able to specify toll-free JavaScript String Unicode literals from OCaml strings:

Consider this:
```ocaml
type jv
type jstr

external pure_js_expr : string -> jv = "caml_pure_js_expr"
let global = pure_js_expr "joo_global_object"

external apply : jv -> jv array -> jv = "caml_js_fun_call"
external get : jv -> jstr -> jv = "caml_js_get"
external jstr : string -> jstr = "caml_jsstring_of_string"

let () = ignore (apply (get global (jstr "npi")) [||])
```
If we look at the code: 

```shell
> ocamlc -no-check-prims lit.ml 
> js_of_ocaml --pretty a.out
> grep npi a.js
    joo_global_object.npi();
```

this is very nice: `js_of_ocaml` translated my OCaml string literal directly to a property name and no representation conversion occurs.

Now let's replace that `npi` by `npiπ`:
```ocaml
let () = ignore (apply (get global (jstr "npiπ")) [||])
```
recompile and look for that symbol:
```shell
> grep npi a.js
      [caml_jsstring_of_string(caml_string_of_jsbytes("npi\xcf\x80"))]
```
this is a little bit less nice, basically the UTF-8 to UTF-16 happens at runtime but given the previous compilation I don't think there should by any problem of doing that translation statically, I suspect it's a matter of being smarter on [that line](https://github.com/ocsigen/js_of_ocaml/blob/7266c08cd095216725a92c58d89308429ba09792/compiler/lib/specialize_js.ml#L130).

I'm not sure I fully grasped `js_of_ocaml`'s current naming scheme for the various string primitives but basically I think there should be two sets of primitives for converting OCaml strings to JavaScript strings which correspond to the two sane ways of interpreting OCaml strings: either as a sequence of bytes or as a sequence of UTF-8 encoded characters:

```
caml_jsstring_of_byte_string
caml_jsstring_to_byte_string
caml_jsstring_of_utf8_string
caml_jsstring_to_utf8_string
```

The first set of primitives is used for representing OCaml strings in JavaScript faithfully, that is, as sequences of bytes. It's mainly used for OCaml programs themselves.

[...]

PR #1381, opened 2023-01-10
Compiler: better handling of js string

hhugo:
Fixing #1034 (partially)

We could go further and rely on `Uucp.Id` to decide if a unicode string is a valid js identifier.
This would allowing to transform `x["npiπ"]` into `x.npiπ`. The main issue is that we need to update the js parser as well. (https://github.com/ocsigen/js_of_ocaml/tree/utf-ident)

cc @dbuenzli


hhugo:
@pmwhite, yet another PR to test on your codebase, if you can

dbuenzli:
That was too fast for me to have a look but thanks @hhugo!

CHANGES.md
```diff
@@ -5,6 +5,8 @@
 * Effects: improved CPS transform, resulting in lower compilation time and smaller generated code
 * Compiler: separate compilation can now drops unused units when linking (similar to ocamlc). (#1378)
   Feature is disabled by default while dune rules are being fixed. Enable with --enable=auto-link.
+* Compiler: specialize string to js-string conversion for all valid utf8 strings (previously just ascii)
```
dbuenzli, line 8: IIUC I think it would be worthwhile to add here something along the lines:
```
JavaScript files generated by `js_of_ocaml` are now UTF-8 encoded. 
You need to make sure your web server serves them with that encoding or integrate 
them in your HTML with <script src="…" charset="UTF-8"></script>`
```

hhugo: I was considering also adding an option to tune the behavior at compile time.
And I am now wondering whether we should have a runtime check that print an error on the console if the encoding is not the one expected. What do you think ?

dbuenzli: > I was considering also adding an option to tune the behavior at compile time.

In general configurability is worse for bugreports/maintenance so it's your call :-) I'd rather go for a well advertised change, if not a main version bump.

> And I am now wondering whether we should have a runtime check that print an error on the console if the encoding is not the one expected. What do you think ?

I don't understand how that would/could work. But in any case, IIUC the code has shipped to the client at that point so it's kind of too late.

hhugo: I was hoping that a test / a few test such as the following would work (place toward the top of the file).
```js
if("π" != "\u{3c0}") {
  console.error(..)
}
```

hhugo: Do you thing the generated file should contain a BOM ?

hhugo: > I was hoping that a test / a few test such as the following would work (placed toward the top of the file).
> 
> ```js
> if("π" != "\u{3c0}") {
>   console.error(..)
> }
> ```

That probably only work when the encoding used "include" ASCII.

pmwhite:
> @pmwhite, yet another PR to test on your codebase, if you can

I'll try to get to this next week.

pmwhite:
This PR appears to be working for us. I assume I don't need to pass an extra flag to turn this on.

hhugo:
Thanks for the report. ( No special flag needed)

PR #1383, opened 2023-01-13
UTF-8 encoding, continue #1381.

hhugo:
Follows #1381.
- ~~Emit the utf-8 BOM sequence~~
- ~~Add a runtime check to verify that the file was properly interpreted as utf8 and not as another ASCII compatible encoding (e.g. ISO-8859-1)~~
- Document that the compiler now emits utf-8 encoded files (as suggested by @dbuenzli)


hhugo:
@dbuenzli, I'll let you comment before merging

dbuenzli:
So I looked a bit the JavaScript source encoding business. TLDR I don't think it's going to pose a problem and I don't think most of the things of this PR should be done (will comment in the PR).

> * Emit the utf-8 BOM sequence

I would rather not add a BOM. MDN recommends [against](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#sect3), I worry a bit people are going e.g. to prefix a shebang to a script which will shift the bom in non initial position and create problems.

The file obviously won't have an UTF-16 BOM and the [encoding detection](https://www.rfc-editor.org/rfc/rfc9239.html#name-character-encoding-scheme-d) of the `text/javascript` MIME type defaults to UTF-8 so I think all is good.

> * Add a runtime check to verify that the file was properly interpreted as utf8 and not as another ASCII compatible encoding (e.g. ISO-8859-1)

Do you add this in all sources ? I would also rather not. 

In fact according to what I have read now it seems unlikely we are going to have that situation, I think that for that to happen your server would need to explicetly serve your file as `Content-Type: text/javascript; encoding=ISO-8859-1` which is rather unlikely. By default it seems webservers serve `.js` files as simply `text/javascript` which should trigger the encoding determination which defaults to UTF-8.

CHANGES.md
```diff
@@ -6,6 +6,9 @@
 * Compiler: separate compilation can now drops unused units when linking (similar to ocamlc). (#1378)
   Feature is disabled by default while dune rules are being fixed. Enable with --enable=auto-link.
 * Compiler: specialize string to js-string conversion for all valid utf8 strings (previously just ascii)
+* Compiler: JavaScript files generated by `js_of_ocaml` are now UTF-8 encoded.
+  You need to make sure your web server serves them with that encoding or integrate
+  them in your HTML with `<script src="…" charset="UTF-8"></script>`
```
dbuenzli, line 11: Given what I know now, in the end I think this is going to create more confusion. Let's keep that 

``Compiler: JavaScript files generated by `js_of_ocaml` are now UTF-8 encoded.``

compiler/lib/config.ml
```diff
@@ -96,6 +96,10 @@ module Flag = struct
 
   let header = o ~name:"header" ~default:true
 
+  let emit_BOM = o ~name:"BOM" ~default:true
+
+  let check_utf8_at_runtime = o ~name:"utf8-runtime-check" ~default:true
+
```
dbuenzli, line 102: I would drop these two features.

manual/overview.wiki
```diff
@@ -45,6 +45,10 @@ For alternatives, see <<a_manual chapter="install" |Install>>.
       js_of_ocaml cubes.byte
 }}}
 
+JavaScript files generated by js_of_ocaml are UTF-8 encoded.
```
dbuenzli, line 48: Let's keep only that line. Again the two lines I suggested are going to create more confusion than needed I think.

hhugo: done

CHANGES.md
hhugo: done

compiler/lib/config.ml
hhugo: commit dropped

hhugo:
> So I looked a bit the JavaScript source encoding business. TLDR I don't think it's going to pose a problem and I don't think most of the things of this PR should be done (will comment in the PR).
> 
> > * Emit the utf-8 BOM sequence
> 
> I would rather not add a BOM. MDN recommends [against](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#sect3), I worry a bit people are going e.g. to prefix a shebang to a script which will shift the bom in non initial position and create problems.
> 
> The file obviously won't have an UTF-16 BOM and the [encoding detection](https://www.rfc-editor.org/rfc/rfc9239.html#name-character-encoding-scheme-d) of the `text/javascript` MIME type defaults to UTF-8 so I think all is good.
> 
> > * Add a runtime check to verify that the file was properly interpreted as utf8 and not as another ASCII compatible encoding (e.g. ISO-8859-1)
> 
> Do you add this in all sources ? I would also rather not.
> 
> In fact according to what I have read now it seems unlikely we are going to have that situation, I think that for that to happen your server would need to explicetly serve your file as `Content-Type: text/javascript; encoding=ISO-8859-1` which is rather unlikely. By default it seems webservers serve `.js` files as simply `text/javascript` which should trigger the encoding determination which defaults to UTF-8.

Thanks for the review and the research. I've updated the PR following you comments

PR #1386, opened 2023-01-16
Compiler: Modernize the js lexer, now utf-8 aware

hhugo:
In particular, it allows to recognize and emit utf8 identifiers.
This lexer now uses sedlex. The implementation was taken from flow and cleaned to remove unused features.


dbuenzli:
I'm not familiar enough with the code base to go through these changes to approve them and don't know how much non US-ASCII entry points there are in the wild JavaScript world but the fact that needless string munging is avoided is very nice.

hhugo:
Minification and variable renaming is currently broken with this PR.
The optimisation rely of a free variable analysis, which rely on string equality of idents.

Idents are currently not normalized and even contain escape sequence verbatim.
We could either
- normalize ident
- disable the variable renaming for unicode ident (We would still need to resolve escape sequence)

dbuenzli:
Did you check what regular JavaScript minifier do ? I suspect they don't do these things. 

Note that there are technologies like `xml` that rely on balancing unicode identifiers and throw normalization out of the window and no-one seems to ever run in the problematic and puzzling cases :-)

hhugo:
> Did you check what regular JavaScript minifier do ? I suspect they don't do these things.
> 
> Note that there are technologies like `xml` that rely on balancing unicode identifiers and throw normalization out of the window and no-one seems to ever run in the problematic and puzzling cases :-)

I had a quick look at terser and flow. I didn't see any unicode normalization. However, they both decode escape sequence in ident.

hhugo:
The last commit decodes escape sequence in  ident

compiler/lib/flow_lexer.ml
```diff
@@ -450,6 +456,62 @@ let new_line env lexbuf =
...
+    | unicode_escape ->
+        let hex = sub_lexeme lexbuf 2 0 in
+        let code = int_of_string ("0x" ^ hex) in
+        if not (Uchar.is_valid code) then failwith "invalid unicode";
+        Buffer.add_utf_8_uchar buf (Uchar.of_int code);
+        id_char buf lexbuf
+    | codepoint_escape ->
+        let hex = sub_lexeme lexbuf 3 1 in
+        let code = int_of_string ("0x" ^ hex) in
+        if not (Uchar.is_valid code) then failwith "invalid unicode";
```
dbuenzli, line 477: You cannot really do that, if you get a high surrogate you need to combine it with the next low surrogate to make the final uchar character ([`UTF16SurrogatePairToCodePoint`](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#sec-utf16decodesurrogatepair) in ECMA). See for example the combination logic [here](https://github.com/ocaml/ocaml/blob/fb4e2acbd4e61fd376d3ef6bdf886ad6055b190f/stdlib/bytes.ml#L744-L759).

For example '🐫' can be represented by the escape sequence '\xD83D\xDC2B',

hhugo: I'm not sure non-scalar escape sequence are accepted in identifier. node doesn't accept `\u{d83b}\u{de62}` for instance.

dbuenzli: I suspect that are not in `\u{}` (even though nothing seems to prevent them in ECMA) but are in `\uXXXX` did you try ?

hhugo: note that I don't think 🐫(`U+1F42B`) is a valid identifier. '𞹢' (`U+1ee62`) seems to be and is accepted by node.

dbuenzli: Ah but that's the reason: 
```
> ucharinfo 🐫 -k id_continue
ID_Continue: false
```
However:

```
> ucharinfo U+1D4DD -k name -k id_continue -k UTF-16BE
Name: MATHEMATICAL BOLD SCRIPT CAPITAL N
ID_Continue: true
UTF-16BE: \xD8\x35\xDC\xDD
```

`var nat_is_\uD835\uDCDD = true` doesn't work but `var nat_is_\u{1D4DD} = true` does (e.g. in in the chrome console). But I can't reconcile that with my reading of: 

> In other words, if a \ [UnicodeEscapeSequence](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-UnicodeEscapeSequence) sequence were replaced by the [SourceCharacter](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#prod-SourceCharacter) it contributes, the result must still be a valid [IdentifierName](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-IdentifierName) that has the exact same sequence of [SourceCharacter](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#prod-SourceCharacter) elements as the original [IdentifierName](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-IdentifierName).

[here](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-identifier-names).

dbuenzli: Ah I think I get the convoluted logic :-)

The definition of [`SourceCharacter`](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#prod-SourceCharacter) is: 


> [SourceCharacter](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#prod-SourceCharacter) ::
any Unicode code point

(and not a *scalar value*). So going back to my previous quote: an escape of a surrogate contributes a surrogate code point `SourceCharacter`. But the `ID_Continue` property of surrogates is (I suppose, `ucharinfo` can only query scalar values) `false`, so it cannot be part of an identifier. So surrogate pairs escape sequences are excluded from the identifier syntax.

vouillon: it seems that surrogate pairs are not combined. So, `\u{12345}` is a valid identifier, but `\ud808\udf45` is not.

hhugo: I've updated the lexer to better handle escaped surrogate pairs (record errors and recover) and added expect tests.

compiler/lib/flow_lexer.ml
```diff
@@ -267,27 +271,71 @@ let decode_identifier =
...
+  in
+  let codepoint_escape_code lexbuf =
+    let hex = sub_lexeme lexbuf 3 1 in
+    let code = int_of_string ("0x" ^ hex) in
+    code
+  in
+  let is_lead_surrogate c = 0xD800 <= c && c <= 0xDBFF in
+  let is_trail_surrogate c = 0xDC00 <= c && c <= 0xDFFF in
+  let surrogate hi lo = (((hi land 0x3FF) lsl 10) lor (lo land 0x3FF)) + 0x10000 in
+  let trail_surrogate env loc buf lexbuf lead =
```
dbuenzli, line 287: Forgive me I'm not familiar with sedlex, I gather this always fails but I wonder why you still combine the chars and add them to buf if possible.

dbuenzli (approved):
Just had a look at the last commit. Don't claim I undersood everything of it but the tests look right :-)

compiler/lib/flow_lexer.ml
hhugo: This JS lexer always try to recover. It's the responsibility of the caller to decide what to do in case of errors. This design was inherited from the flow parser that I used as a starting point.
Here, recovering by combining surrogate pair seems to be the most natural behavior.

compiler/lib/flow_lexer.ml
```diff
@@ -267,27 +271,71 @@ let decode_identifier =
...
+    let code = int_of_string ("0x" ^ hex) in
+    code
+  in
+  let codepoint_escape_code lexbuf =
+    let hex = sub_lexeme lexbuf 3 1 in
+    let code = int_of_string ("0x" ^ hex) in
+    code
+  in
+  let is_lead_surrogate c = 0xD800 <= c && c <= 0xDBFF in
+  let is_trail_surrogate c = 0xDC00 <= c && c <= 0xDFFF in
```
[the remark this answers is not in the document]
hhugo: Updated

PR #1387, opened 2023-01-18
Cache function length correctly in stdlib_modern.js

pmwhite:
This fixes a bug in `stdlib_modern.js` that slipped in via https://github.com/ocsigen/js_of_ocaml/commit/23f4f71a9bdf77e4b2adac1ddee7d23471e7cbb4. Perhaps this repository should run the tests under both `stdlib.js` and `stdlib_modern.js`, so that bugs like this get caught earlier.

### DIFF
--- a/compiler/tests-compiler/gh1051.ml
+++ b/compiler/tests-compiler/gh1051.ml
@@ -26,6 +26,7 @@ let%expect_test _ =
   [%expect
     {|
 Warning: integer overflow: integer 0xffffffff (4294967295) truncated to 0xffffffff (-1); the generated code might be incorrect.
+Warning: integer overflow: integer 0xffffffff (4294967295) truncated to 0xffffffff (-1); the generated code might be incorrect.
 ffffffff |}];
   ()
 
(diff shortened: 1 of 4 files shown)


hhugo:
Thanks.
This PR would be a good place to add a test for stdlib_modern.js, could you add one here ?

pmwhite:
I'm happy to do this, though it'll take a bit longer. The plan I have in mind is to modify `compile_and_run` so that it compiles its input twice (once with and once without `stdlib_modern.js`) and compares the output to ensure they are the same. Thus, no test output should change, but the tests will be stronger.

To do the above, I think I need to modify the command-line interface, since that is how `compile_and_run` controls what happens in the compiler. For this, I was thinking of renaming the `-noruntime` flag to `-runtime` and have it accept one of "none", "default", and "modern".

Do you think the above plan is good?

hhugo:
I don't think you need to touch the compiler. I would expect that passing `+stdlib_modern.js` on the command line would be enough.

pmwhite:
> I don't think you need to touch the compiler. I would expect that passing `+stdlib_modern.js` on the command line is enough. (Buy maybe not)

Thanks; I didn't think of that. I assumed that a compiler modification was necessary we've been literally patching the compiler to include `stdlib_modern.js` - it's nice to know that we could stop doing that...

Anyway, I've added the change that makes the tests stronger. Let me know if you think the double warning message is sad enough that we should figure out a way to eliminate it.

hhugo:
Thanks for changing the tests, LGTM.

PR #1388, opened 2023-01-19
Make tests clean up files they create

pmwhite:
This PR upstreams some diffs we maintain internally to satisfy our build system. While these extra files don't bother dune, it seems better to me for tests to leave the filesystem the way they found it.

### DIFF
--- a/compiler/tests-jsoo/test_io.ml
+++ b/compiler/tests-jsoo/test_io.ml
@@ -99,4 +99,5 @@ let%expect_test _ =
   close_out oc;
   print_endline (file_contents fname);
   [%expect {| this is a test |}];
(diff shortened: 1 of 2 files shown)


compiler/tests-jsoo/test_io.ml
```diff
@@ -99,4 +99,5 @@ let%expect_test _ =
   close_out oc;
   print_endline (file_contents fname);
   [%expect {| this is a test |}];
+  Sys.remove "file2.txt";
```
hhugo, line 102:
```suggestion
  Sys.remove fname;
```

hhugo (commented):
LGTM modulo small suggestion

PR #1389, opened 2023-01-19
Clean-up caml_call_gen

vouillon:
- We don't need to check for `f.fun` since, when it is set, function `f` is a wrapper that calls `f.fun`.
- We only need to check whether we still have a function in the over application case (`d < 0').
- We don't need a special case for `n === 0` since we now explictly set `f.l` for under-applied functions


hhugo:
LGTM

PR #1390, opened 2023-01-19
Control flow analysis: fixes

vouillon:
We were not handling `Array.unsafe_get` correctly, and some annotations in the runtime were wrong.


TyOverby:
Is there a place that I can read more about these annotations on the functions?

hhugo:
There is a short and incomplete description here http://ocsigen.org/js_of_ocaml/latest/manual/linker

PR #1393, opened 2023-01-23
Fix compilation of some mutually recursive functions

vouillon:
Implement caml_alloc_dummy_infix

### DIFF
--- a/compiler/tests-jsoo/test_rec_fun.ml
+++ b/compiler/tests-jsoo/test_rec_fun.ml
@@ -1,0 +1,111 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                OCaml                                   *)
+(*                                                                        *)
+(*   Copyright 2019 Institut National de Recherche en Informatique et     *)
+(*     en Automatique.                                                    *)
+(*                                                                        *)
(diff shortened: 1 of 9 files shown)


hhugo:
Shouldn't we change the bytecode to also emit the arity ? and perform the same logic we already use for `caml_alloc_dummy_function` ?

compiler/tests-jsoo/test_rec_fun.ml
```diff
@@ -0,0 +1,98 @@
+let rec h =
```
hhugo, line 1: Missing header

vouillon:
> Shouldn't we change the bytecode to also emit the arity ? and perform the same logic we already use for `caml_alloc_dummy_function` ?

Ideally, yes. On the other hand, this does not seem to be widely used...

hhugo:
> > Shouldn't we change the bytecode to also emit the arity ? and perform the same logic we already use for `caml_alloc_dummy_function` ?
> 
> Ideally, yes. On the other hand, this does not seem to be widely used...

Patch available in https://github.com/hhugo/ocaml/commit/881d767cd4fda0f3fc6d1855063e1abe1771643b

(x-link https://github.com/ocaml/ocaml/pull/37)

PR #1394, opened 2023-01-24
Compiler: better deps tracking for expressions queue

hhugo:


compiler/lib/generate.ml
```diff
@@ -1218,64 +1224,68 @@ let rec translate_expr ctx queue loc in_tail_position e level : _ * J.statement_
...
       let clo = compile_closure ctx cont in
+      let deps = Code.Addr.Map.find pc ctx.free_vars in
       let clo =
         match clo with
         | (st, J.N) :: rem -> (st, J.U) :: rem
         | _ -> clo
       in
       let clo = J.EFun (None, List.map args ~f:(fun v -> J.V v), clo, loc) in
-      (clo, flush_p, queue), []
+      (clo, flush_p, deps, queue), []
```
vouillon, line 1267: I think that since we have a `flush_p` here, no variable in `deps` will remain in the queue. So, we could just as well put `Var.Set.empty` here.

vouillon:
It seems possible to minimize changes by putting `prop` and `deps` together in a pair and redefining `or_p`: 
```ocaml
let or_p (p, s) (q, t) = (max p q, Var.Set.union s t)
let const_p = (0, Var.Set.empty)
...
```

hhugo:
> It seems possible to minimize changes by putting `prop` and `deps` together in a pair and redefining `or_p`:
> 
> ```ocaml
> let or_p (p, s) (q, t) = (max p q, Var.Set.union s t)
> let const_p = (0, Var.Set.empty)
> ...
> ```
The diff is much smaller !!!

compiler/lib/generate.ml
vouillon: I wish we could remove the `flush_p` here. We would need to know the free variables of the closure or rather the ones that are in the queue. But it's a bit tricky to compute that efficiently (`Freevar` only computes the free variables that are bound in a loop to avoid having to deal with large sets of variables.)

vouillon:
> The diff is much smaller !!!

Much easier to review! This looks good to me.

I'm wondering whether we should use a monad at some point. This might make the code more readable. But that may have some performance impact.

compiler/lib/generate.ml
hhugo: Noted

Issue #594, opened 2017-06-15
Improved function arity analysis


vouillon:
At least forward propagation (through function return), so that we can perform optimized call to function inside functors.

hhugo:
https://github.com/ocsigen/js_of_ocaml/pull/56

hhugo:
We could reuse `global_flow.ml` from #1384

PR #1384, opened 2023-01-13
Effects: partial CPS transform

vouillon:
We identify functions that don't involve effects by analyzing the call graph and we keep then in direct style. This relies on a global control flow analysis to find which function might be called where.

The analysis is very effective on small / monomorphic programs.
- `hamming` is somewhat slower since it uses lazy values (we don't analyze mutable values).
- `nucleic` is faster since the global control flow analysis is used to avoid some slow function calls. This measurement was made before #1358 was merged. The gap is probably narrower now.
time-effects

The analysis is less effective on large programs. Higher-order functions such as `List.iter` are turned into CPS and then all functions that calls directly or indirectly such a function needs to be turned into CPS as well. There is also some horizontal contamination, where a function needs to be turned into CPS since it is used in a context which expects a CPS function, and then this impacts all other places it is called. Still, `ocamlc` is now only about ~10% slower (it is about 60% slower with the released version of Js_of_ocaml).

CAMLboy is less than 25% slower (650 FPS instead of 800 FPS).

The size of the generated code is less than 20% larger, a few percents larger when compressed. For a large Web app, I have a 44% increase of generated code (6% when compressed).
size-effects
size-bzip2-effects

Compiling `ocamlc` is about 25% slower.

PR #1397, opened 2023-02-01
Improved function arity analysis

vouillon:
Addresses #594

The analysis is a bit costly, so we want to see whether this should be enabled by default.

There is also the option to make a coarser analysis (propagation information about what the functions returns but not through function parameters). 

| currently | faster analysis    | full analysis     |
|:-----------------:|:----------------------:|:----------------------:|
| 5.24 sec          | 5.19 sec: 1.01x faster | 5.11 sec: 1.03x faster |


hhugo:
@vouillon, I've rebased the branch and force-pushed after merging #1384

hhugo:
I don't know how to read the timing in the description. What is it about ? Why is the "full analysis" faster that the "faster" one ?

vouillon:
> I don't know how to read the timing in the description. What is it about ? Why is the "full analysis" faster that the "faster" one ?

This is the timing of the generated code, when executing the following command:
```
node ocamlc.js -c ~/js_of_ocaml/benchmarks/sources/ml/*.ml
```
So, a more precise analysis yields a significant gain.

Compiling `ocamlc.byte` takes about 12s. The faster analysis adds about 0.35s. The full one adds about 0.8s.

hhugo:
How is the size affected ? how many extra direct call are generated ?

hhugo:
We could enable the full analysis with `-O3` only

vouillon:
> How is the size affected ? how many extra direct call are generated ?

The code is about the same size (0.25% larger since we use a somewhat verbose notation for best performances:`f[7].call(null, e1, ...,en)`).

There are not that many additional extra direct calls but I think they are more likely to be performed repeatedly.

|            | currently | faster analysis    | full analysis     |
|------------|:---------:|:------------------:|:-----------------:|
|Code size   | 1912637   | 1917311            | 1917226
|Direct calls | 25883     | +1409              |  + 1854

compiler/lib/specialize.ml
```diff
@@ -21,38 +21,44 @@ open! Stdlib
...
+  in
+  arity info x []
+
+let count = ref 0
 
-let specialize_instr info (acc, free_pc, extra) i =
+let specialize_instr function_arity (acc, free_pc, extra) i =
   match i with
-  | Let (x, Apply { f; args; _ }) when Config.Flag.optcall () -> (
+  | Let (x, Apply { f; args; exact }) when (not exact) && Config.Flag.optcall () -> (
```
hhugo, line 56:
```suggestion
  | Let (x, Apply { f; args; exact = false }) when Config.Flag.optcall () -> (
```

compiler/lib/global_flow.ml
```diff
@@ -550,6 +565,7 @@ let f p =
     ; possibly_mutable
     ; known_cases = Hashtbl.create 16
     ; applied_functions = Hashtbl.create 16
+    ; fast = not (Config.Flag.effects ())
```
hhugo, line 568: Making`fast` an argument of `f` would give more freedom to the caller (and maybe allow to use `~fast:false` when using `-O3`)

compiler/lib/global_flow.ml
```diff
@@ -621,3 +637,25 @@ let exact_call info f n =
...
+  | Values { known; others = false } -> (
+      match
+        Var.Set.fold
+          (fun g acc ->
+            match info.info_defs.(Var.idx g) with
+            | Expr (Closure (params, _)) -> (
+                let n = List.length params in
+                match acc with
+                | None -> Some (Some n)
+                | Some (Some n') when n <> n' -> Some None
```
hhugo, line 653: Would it make sense to return early here ? or would it be overkill ?

hhugo:
I've rebased your branch and make `fast` depend on `profile` (o1,o2,o3)

hhugo:
I've rebased the PR again

hhugo:
@vouillon, this is the last PR I'd like to sort out before the next release (see https://github.com/ocsigen/js_of_ocaml/milestone/10).
Should we merge or not ?

compiler/lib/global_flow.ml
vouillon: This is overkill. On average, you have few functions to consider.

compiler/lib/global_flow.ml
```diff
@@ -204,12 +212,15 @@ let expr_deps blocks st x e =
       (* If [f] is obviously a function, we can add appropriate
          dependencies right now. This speeds up the analysis
          significantly. *)
-      match st.defs.(Var.idx f) with
-      | Expr (Closure (params, _)) when List.length args = List.length params ->
-          Hashtbl.add st.applied_functions (x, f) ();
-          List.iter2 ~f:(fun p a -> add_assign_def st p a) params args;
-          Var.Set.iter (fun y -> add_dep st x y) (Var.Map.find f st.return_values)
-      | _ -> ())
+      if false
```
vouillon, line 215: I need to fix that...

vouillon:
@hhugo I think it is ready

compiler/lib/generate.ml
```diff
@@ -950,7 +950,14 @@ let parallel_renaming params args continuation queue =
 
 let apply_fun_raw ctx f params exact cps =
   let n = List.length params in
-  let apply_directly = J.call f params J.N in
+  let apply_directly =
+    (* Make sure we are performing a regular call, not a (slower)
+       method call *)
```
hhugo, line 955: @vouillon, can you provide some reference mentioning this optimization. I did some test and it seems to show the opposite.

vouillon: I don't have any reference, but without this, the improved analysis resulted in slower code.
I just tried is compiling `ocamlc.byte` (`dune exec -- js_of_ocaml --opt 3 `which ocamlc.byte` -o /tmp/ocamlc.js`).
And then running it on some ml source files:
```
time node /tmp/ocamlc.js -c ./benchmarks/sources/ml/*.ml ./benchmarks/sources/ml/*.ml ./benchmarks/sources/ml/*.ml
```
This optimization makes close to a 8% performance improvement on my machine.

hhugo: What ocaml version do you use ? it seems that ocamlc no longer works when compiled to js due to https://github.com/ocaml/ocaml/pull/11997 (since OCaml 5.1). The reason is that Reloc_literal can now contain floats that jsoo doesn't want to marshal.

vouillon: I'm using `python3 -m pyperf system tune` to reduce the noise.

PR #1402, opened 2023-02-09
Compiler: don't re-parse the builtin js every single time

hhugo:


hhugo:
This change cuts ~200/300ms from every `jsoo compile` which build up quite a bit when doing separate compilation.

Before
```
$ time dune runtest compiler/tests-compiler/ --cache disabled
                                         
real	1m16,151s
user	4m49,958s
sys	0m34,609s
```

After

```
$ time dune runtest compiler/tests-compiler/ --cache disabled
                                         
real	0m54,416s
user	2m56,506s
sys	0m31,809s
```

hhugo:
@pmwhite @TyOverby, You'll surely be interested by this change.

pmwhite:
Just tried it out internally. I was able to separate the time spent running js_of_ocaml jobs from all the other jobs, and the time dropped from 74 seconds down to 24 seconds. That's a good enough improvement that I'm forced to be suspicious of my methodology. I _would_ expect the improvement to be better than what you reported, though, since your timings include both the ocamlc and tests.

hhugo:
@pmwhite, if you look at https://discuss.ocaml.org/t/is-rescript-bsc-significantly-faster-than-jsoo/11372/4.
You can see that for small files, the improvement is big. compilation time going from 290ms to 20ms

Issue #747, opened 2019-02-16
[BUG] debug locations sometime wrong, affecting sourcemaps


jordwalke:
While modifying `parse_bytecode.ml` for [rehp](https://github.com/jordwalke/rehp), I might have stumbled upon some evidence that `parse_bytecode.ml` is responsible for source maps being
difficult to use in practice (they jump all over the place in very unintuitive ways).

Problem: For the following Reason code:
         
```reason
let greeting = "hello world";
print_endline(greeting);
let greeting = "hello world with unicode: Ɋ";
print_endline(greeting);       /* THIS IS LINE 4 */

let unicodeLength = String.length(String.make(Random.int(30), 'i'));
print_endline(
 "String.length(\"Ɋ\") should be two:" ++ string_of_int(unicodeLength),
);
print_endline(String.make(1, "Ɋ".[0]) ++ String.make(1, "Ɋ".[1]));
```
         
We get the following debug data: (**Notice the many things claiming to be on line `4` that aren't!
For example `_g_` (the number `30`)**)

```javascript
  /*<<rehack_tests/strings/strings.re 2 0>>*/  /*<<rehack_tests/strings/strings.re 2 0>>*/ print_endline(
   greeting
 );
  /*<<rehack_tests/strings/strings.re 4 0>>*/  /*<<rehack_tests/strings/strings.re 4 0>>*/ print_endline(
   greeting__0
 );
  /*<<rehack_tests/strings/strings.re 4 0>>*/ var  /*<<rehack_tests/strings/strings.re 4 0>>*/ _f_ =
   105,
   /*<<rehack_tests/strings/strings.re 4 0>>*/ _g_ = 30,
   /*<<rehack_tests/strings/strings.re 6 46>>*/ _h_ =
    /*<<rehack_tests/strings/strings.re 6 46>>*/ int__1(_g_),
   /*<<rehack_tests/strings/strings.re 6 34>>*/ unicodeLength =
    /*<<rehack_tests/strings/strings.re 6 34>>*/ caml_ml_string_length(
      /*<<rehack_tests/strings/strings.re 6 34>>*/ make__0(_h_, _f_)
   ),
   /*<<rehack_tests/strings/strings.re 8 44>>*/ _i_ =
    /*<<rehack_tests/strings/strings.re 8 44>>*/ string_of_int(unicodeLength),
   /*<<rehack_tests/strings/strings.re 8 2>>*/ _k_ =
[...]

PR #1407, opened 2023-02-11
Compiler: no longer split blocks at fun call to propagate location

hhugo:
Trying to address https://discuss.ocaml.org/t/dune-build-vs-makefile/11394/9.

It should also help with https://github.com/ocsigen/js_of_ocaml/issues/1300
and fix https://github.com/ocsigen/js_of_ocaml/issues/747

@vouillon, the PR is not ready yet but what do you think of this approach ?

@pmwhite, this could improve your compilation time significantly.


hhugo:
> Maybe we should use this instead of adding a field to every constructors of type `instr`?
> 
> ```ocaml
> type block =
>   { params : Var.t list
>   ; body : (instr * loc) list
>   ; branch : last * loc
>   }
> ```

Yeah, probably. Ill do the change and run a quick benchmark to make sure there performance profile is the same.

compiler/lib/parse_bytecode.ml
```diff
@@ -258,31 +258,21 @@ end = struct
...
       else
         let pos =
           if after
           then loc.Location.loc_end
-          else if before
-          then loc.Location.loc_start
           else
             match event.ev_kind with
-            | Event_after _ -> loc.Location.loc_end
+            | Event_after _ when false -> loc.Location.loc_end
```
vouillon, line 275: For an event that is after an expression, the relevant location should be at the end of the expression, shouldn't it?

hhugo:
Note that I'm still exploring the design space of this change. I'm not really happy with the mapping between debug event and instructions.

hhugo:
Building the toplevel with sourcemaps goes from 37.6s to 28.2s

hhugo:
> There is a mismatch between where OCaml put events, and where the JavaScript debuggers might stop. I'm not sure what to do about that.

Can you give more details ? are you just referring to the fact `debugger` is a statement ?

hhugo:
I've cleaned up the code a little and recovered few locations that initially disappeared. I'm still trying find a better way to map event to instruction.

hhugo:
@pmwhite, would you be able to test this and report on both compilation time and size of generated sourcemap. 
This PR both add and remove locations. I'd be curious to see if how the size of sourcemap is affected.

vouillon:
>

hhugo:
The PR seems to fix two issues already
- remove some completely unrelated location
- fix a compilation time issue when debug is enabled

Maybe this PR should be only do the right think to existing well understood scenario: events after a call and event before tail calls

hhugo:
@vouillon, can I request another round of review from you ?

PR #1408, opened 2023-02-13
support getModifierState in keyboard events

TyOverby:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState


hhugo:
LGTM, waiting for ci

PR #1410, opened 2023-02-18
Make phys_equal more like native

### LINKED ISSUE #1372: Use `Object.is` instead of `===` to implement phys_equal
There are a few places where the behavior of `phys_equal` differ between native OCaml and Jsoo: 

| a   | b    | native | jsoo  |
|-----|------|--------|-------|
| NaN | NaN  | true   | false |
| 0.0 | -0.0 | false  | true  |

The (relatively) new `Object.is` function implements those comparisons "correctly":
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

hhugo:
Fix #1372

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
   that follows the semantic of the backend (js or wasm)
 * Compiler: warn on joo_global_object
 * Compiler: revisit static env handling (#1708)
+* Compiler: Make phys_equal more like native (wrt NaN and +0/-0) #1410
 * Runtime: change Sys.os_type on windows (Cygwin -> Win32)
 * Runtime: backtraces are really expensive, they need to be be explicitly
   requested at compile time (--enable with-js-error) or at startup (OCAMLRUNPARAM=b=1)
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -289,14 +289,19 @@ end
 
 type cont = Addr.t * Var.t list
 
+type float_or_not =
+  | Float
+  | Not_float
+  | Unknown
+
 type prim =
   | Vectlength
   | Array_get
   | Extern of string
   | Not
   | IsInt
-  | Eq
-  | Neq
+  | Eq of float_or_not
+  | Neq of float_or_not
   | Lt
   | Le
   | Ult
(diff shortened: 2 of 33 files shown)


hhugo:
This change revealed some issues with some js stubs what were returning `-0` instead of `0`.

TyOverby (approved):

hhugo:
@ty, is there any per difference between the two ? Is `object.is` available everywhere?

TyOverby:
https://caniuse.com/mdn-javascript_builtins_object_is

It’s pretty widely supported!

TyOverby:
As for performance impacts, I don't have a good intuition for if it would be faster or slower.  A small microbenchmark that I just made seems to indicate that `Object.is` is faster 

image

hhugo:
I'm not really happy with the current implementation, mostly because the decision to use `===` or `Object.is` is syntactic.
@vouillon, any opinion ?

hhugo:
> I'm not really happy with the current implementation, mostly because the decision to use `===` or `Object.is` is syntactic. @vouillon, any opinion ?

I've changed the implementation. We now keep track of whether `Eq` and `Neq` compare float or non-float values.

TyOverby (approved):

vouillon:
I'm not sure how much of a difference this PR makes. In most cases, the physical equality will still not behave the same. For instance, this will still return `true` with Js_of_ocaml:
```ocaml
# 0. == 0.;;
- : bool = false
```

hhugo:
> I'm not sure how much of a difference this PR makes. In most cases, the physical equality will still not behave the same. For instance, this will still return `true` with Js_of_ocaml:
> 
> ```ocaml
> # 0. == 0.;;
> - : bool = false
> ```

The most interesting bit is "fixing" `phys_equal` between `0.` and `-0.`.
The rest (whether two identical float are equal or not) can been seen as  an implementation detail.

That said, looking at the documentation of `==`, `0. == -0.` is fine.
```
   On non-mutable types, the behavior of [( == )] is
   implementation-dependent; however, it is guaranteed that
   [e1 == e2] implies [compare e1 e2 = 0].
 ```

PR #1414, opened 2023-02-21
Support custom runtime events producer primitives

TheLortex:
https://github.com/ocaml/ocaml/pull/11474 introduces new primitives in the runtime that enables libraries to emit custom events in the runtime event tracing system. This feature will appear in the next release of OCaml.

For now the implementation is just a stub that does nothing with the event, so that linking with these libraries still works.

### DIFF
--- a/runtime/runtime_events.js
+++ b/runtime/runtime_events.js
@@ -1,0 +1,9 @@
+//Provides: caml_runtime_events_user_register
+function caml_runtime_events_user_register(event_name, event_tag, event_type) {
+  return 0;
+}
+
+//Provides: caml_runtime_events_user_write
+function caml_runtime_events_user_write(event, event_content) {
+
+}
--- a/runtime/runtime_events.js
+++ b/runtime/runtime_events.js
@@ -1,9 +1,50 @@
+
+//Provides: caml_custom_event_index
+var caml_custom_event_index = 0;
+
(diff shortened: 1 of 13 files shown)


hhugo:
aren't you missing `caml_runtime_events_user_resolve` ?

TheLortex:
`caml_runtime_events_user_resolve` is used on the consumer side (the function is used by `otherlibs/runtime_events/runtime_events_consumer.c`). Runtime events work by reading from and writing into a memory-mapped ring buffer, I don't know how it would work in Javascript, so my decision was not to implement the event consumer side primitive. 

It's possible to have a stub implementation by returning `None` instead. What do you prefer ?

TheLortex:
@hhugo I have added stubs for all externals.

runtime/runtime_events.js
```diff
@@ -0,0 +1,50 @@
...
+function caml_runtime_events_user_resolve() {
+  return 0;
+}
+
+//Provides: caml_runtime_events_start
+function caml_runtime_events_start() {
+  return 0;
+}
+
+//Provides: caml_runtime_events_stop
```
hhugo, line 27: Looking at the CI failure, it seems it's called caml_runtime_events_pause

hhugo:
Can you rebase your PR ?

TheLortex:
@hhugo should be good now.

PR #1417, opened 2023-02-23
Runtime: restore support for zero-arity functions in caml_call_gen

pmwhite:
We discovered a bug caused by `caml_call_gen` removing logic for supporting the situation when it is called with a function whose length is 0. Arguably, this is not really a problem with `caml_call_gen` because it is only meant to be given _OCaml_ functions and not general Javascript functions; however, `Js.export` was implemented in such a way that this invariant is not always upheld, even if the `js_of_ocaml` library is used in a type-safe way.

This PR fixes the bug by changing `caml_call_gen`, rather than `Js.export`, since it seems possible for other code to be relying on this condition. I've looked for a while to see if I could find functions other than `Js.export`, but I couldn't find any, so I _think_ it's just that one function, but it's hard to be sure.

### DIFF
--- a/lib/js_of_ocaml/import.ml
+++ b/lib/js_of_ocaml/import.ml
@@ -35,21 +35,21 @@ module Poly = struct
 end
 
 module Int_replace_polymorphic_compare = struct
-  let ( < ) (x : int) y = x < y
+  external ( < ) : int -> int -> bool = "%lessthan"
 
-  let ( <= ) (x : int) y = x <= y
+  external ( <= ) : int -> int -> bool = "%lessequal"
 
-  let ( <> ) (x : int) y = x <> y
+  external ( <> ) : int -> int -> bool = "%notequal"
 
-  let ( = ) (x : int) y = x = y
+  external ( = ) : int -> int -> bool = "%equal"
 
-  let ( > ) (x : int) y = x > y
+  external ( > ) : int -> int -> bool = "%greaterthan"
 
-  let ( >= ) (x : int) y = x >= y
+  external ( >= ) : int -> int -> bool = "%greaterequal"
 
-  let compare (x : int) y = compare x y
+  external compare : int -> int -> int = "%compare"
 
-  let equal (x : int) y = x = y
+  external equal : int -> int -> bool = "%equal"
 
   let max (x : int) y = if x >= y then x else y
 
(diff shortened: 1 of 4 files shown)


hhugo:
@pmwhite, could you split the commit and put the test on its own commit so that we can see the effect of the fix ?

@vouillon, can you look at this ?

lib/js_of_ocaml/import.ml
```diff
@@ -35,21 +35,21 @@ module Poly = struct
 end
 
 module Int_replace_polymorphic_compare = struct
-  let ( < ) (x : int) y = x < y
+  external ( < ) : int -> int -> bool = "%lessthan"
```
TyOverby, line 38: Why is this change in the PR?

hhugo: I've added the commit while working on the PR as it generate nicer code. But it doesn't have to be part of this PR.

lib/js_of_ocaml/js.ml
```diff
@@ -805,6 +805,8 @@ let export_js (field : js_string t) x =
     (Unsafe.pure_js_expr "jsoo_exports")
     field
     (if String.equal (Js.to_string (typeof (Obj.magic x))) "function"
+        (* function with arity/length equal to zero are already wrapped *)
+        && Unsafe.get (Obj.magic x) (Js.string "length") > 0
```
vouillon, line 809: I think this is the correct fix, and that we do not need to modify `caml_call_gen`.

runtime/stdlib.js
```diff
@@ -73,7 +73,7 @@ function caml_call_gen(f, args) {
   var n = (f.l >= 0)?f.l:(f.l = f.length);
   var argsLen = args.length;
   var d = n - argsLen;
-  if (d == 0) {
+  if (n === 0 || d === 0) {
```
vouillon, line 76: This is not going to work since the last element of `args` is a continuation function, which function `f` is certainly not expecting.

hhugo:
I've dropped the change for caml_call_gen but kept the test

lib/js_of_ocaml/js.ml
hhugo: done

pmwhite:
This all looks fine to me! (Was out of town the past few days, so just now catching up on the rest of the conversation).

PR #1426, opened 2023-03-10
Switch CI to ocaml-opam's sunset repo

dra27:


dra27:
I expect this to fail until https://github.com/ocaml/opam-repository/pull/23480 is merged, as ocamlfind.1.9.6 needs constraining away from Windows.

PR #1434, opened 2023-03-22
Clarify contents of LICENSE

nojb:
Hello!

We were going over the licenses of the open-source packages we use at LexiFi and we noticed that the LICENSE file of `js_of_ocaml` is a bit confusing:

1. It includes the text of the full GPL license, even though it is not used for any of the components of the distribution, and
2. For the "Compiler and Library" part it refers to the "Library" GPL, which nowadays is known as "Lesser" GPL (it is the same license). Moreover, the "Lesser" GPL is used for the "Examples" part.

This PR proposes to clarify the terms of the license by:

1) removing the text of the full GPL as it is not used; and
2) changing "Library" to "Lesser" for the "Compiler and Library" part.

What do you think?


LICENSE
```diff
@@ -8,11 +8,8 @@ and "the Examples" refers to all files marked "Copyright CNRS
 Université Paris Diderot" in directory "examples" and its
 sub-directories.
 
-The Compiler and Library are distributed under the terms of the GNU
-Lesser General Public License version 2 (included below).
-
-The Examples are distributed under the terms of the GNU Lesser
-General Public License version 2 (included below).
+Both the Compiler and Library and the Examples are distributed under the
```
hhugo, line 11: Both A and B and C reads weird to me.

nojb: I pushed another version: `The Compiler and Library, as well as the Examples, are distributed ..."

hhugo: LGTM

nojb:
Thanks!

PR #1442, opened 2023-03-30
JavaScript parser: fix parsing of labels

### LINKED ISSUE #1440: [BUG] New js parser fails to parse some js that old parser supported
**Describe the bug**
I tried upgrading Grain's binaryen.ml bindings to jsoo 5.1 and it crashes with failure to parse the binaryen.js file. I still need to create a reproduction that doesn't have the whole (large) js file to comb through but I wanted to preemptively open this issue. 

**Expected behavior**
Should be able to parse es5 code that was supported by 5.0

**Versions**
Ocaml 4.14
Jsoo 5.1

vouillon:
Identifiers such as 'of' and 'async' are valid labels.
Fixes #1440.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ## Bug fixes
 * Compiler: put custom header at the top of the output file (fix #1441)
+* Compiler (js parser): fix parsing of js labels (fix #1440)
 
 # 5.1.1 (2023-03-15) - Lille
 ## Bug fixes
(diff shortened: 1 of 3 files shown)


hhugo:
Maybe add a regression test in https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/tests-compiler/js_parser_printer.ml

hhugo:
@phated, would you have time to give this a try and report other blockers ?

PR #1445, opened 2023-04-04
Sourcemap: stop producing sourcemap mappings with negative lines and columns

pmwhite:
This patch began by changing the mapping type in compiler/lib/source_map.ml so that it more obviously encodes the possible cases for a particular mapping. All other changes were made in response to the type errors that ensued.

No new tests were added because this change is mostly behavior preserving, and the situation in which behavior changes is demonstrated by changes to existing tests.

The motivation for this change is that the sourcemaps produced by js_of_ocaml are rejected by the [Mozilla sourcemap library](https://github.com/mozilla/source-map). In particular, it's [this error message](https://github.com/mozilla/source-map/blob/4e304db915cecd42505feb007766a034395c6eb7/lib/source-map-consumer.js#L325) that gets triggered.

### DIFF
--- a/compiler/lib/source_map.ml
+++ b/compiler/lib/source_map.ml
@@ -19,13 +19,17 @@
 
 open! Stdlib
 
+type ori_location =
+  { source : int
+  ; line : int
+  ; col : int
+  ; name : int option
+  }
+
 type map =
   { gen_line : int
   ; gen_col : int
-  ; ori_source : int
-  ; ori_line : int
-  ; ori_col : int
-  ; ori_name : int option
+  ; ori_location : ori_location option
   }
 
 type mapping = map list
@@ -52,7 +56,11 @@ let empty ~filename =
 
 let map_line_number ~f =
   let f i = if i < 0 then i else f i in
-  fun m -> { m with ori_line = f m.ori_line; gen_line = f m.gen_line }
+  fun m ->
+    { m with
+      gen_line = f m.gen_line
+    ; ori_location = Option.map m.ori_location ~f:(fun l -> { l with line = f l.line })
+    }
 
 let string_of_mapping mapping =
   let mapping =
@@ -79,15 +87,7 @@ let string_of_mapping mapping =
     if i < len
     then
       let c = a.(i) in
-      if prev >= 0
-         && c.ori_source = a.(prev).ori_source
-         && c.ori_line = a.(prev).ori_line
-         && c.ori_col = a.(prev).ori_col
-      then (* We already are at this location *)
-        loop prev (i + 1)
-      else if i + 1 < len
-              && c.gen_line = a.(i + 1).gen_line
-              && c.gen_col = a.(i + 1).gen_col
+      if i + 1 < len && c.gen_line = a.(i + 1).gen_line && c.gen_col = a.(i + 1).gen_col
       then (* Only keep one source location per generated location *)
         loop prev (i + 1)
       else (
@@ -104,26 +104,25 @@ let string_of_mapping mapping =
         let l =
           (c.gen_col - !gen_col)
           ::
-          (if c.ori_source = -1
-           then []
-           else
-             (c.ori_source - !ori_source)
-             :: (c.ori_line - !ori_line)
-             :: (c.ori_col - !ori_col)
-             ::
-             (match c.ori_name with
-             | None -> []
-             | Some n ->
-                 let n' = !ori_name in
-                 ori_name := n;
-                 [ n - n' ]))
+          (match c.ori_location with
+          | None -> []
+          | Some ori ->
+              (ori.source - !ori_source)
+              :: (ori.line - !ori_line)
+              :: (ori.col - !ori_col)
+              ::
+              (match ori.name with
+              | None -> []
+              | Some n ->
+                  let n' = !ori_name in
+                  ori_name := n;
+                  [ n - n' ]))
         in
         gen_col := c.gen_col;
-        if c.ori_source <> -1
-        then (
-          ori_source := c.ori_source;
-          ori_line := c.ori_line;
-          ori_col := c.ori_col);
+        Option.iter c.ori_location ~f:(fun ori ->
+            ori_source := ori.source;
+            ori_line := ori.line;
+            ori_col := ori.col);
         Vlq64.encode_l buf l;
         loop i (i + 1))
   in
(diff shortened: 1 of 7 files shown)


hhugo:
Is a mapping ever useful when ori_location is none ?

pmwhite:
Maybe it isn't, but then why are those mappings even allowed? It seems strange for the spec to mention this kind of mapping if they aren't useful. That said, I can't think of a use for them, so maybe you're right.

dyedgreen:
Seeing that the ocaml compiler seems to generate these maybe we can ask someone who know more about that code why these are generated?

hhugo:
> Seeing that the ocaml compiler seems to generate these maybe we can ask someone who know more about that code why these are generated?

I don't think the OCaml compiler has anything to do with this. What makes you think that ?

 
> Maybe it isn't, but then why are those mappings even allowed? It seems strange for the spec to mention this kind of mapping if they aren't useful. That said, I can't think of a use for them, so maybe you're right.

I initially misread your patch.
Replying to myself, I think having a mapping when there are no source info is important. The logic was added by https://github.com/ocsigen/js_of_ocaml/commit/0e78d6dbe546bf70d373a00c54c465b9fbb3629f. It convey the information that the previous mapping no longer apply.

If you have the following list of token `M X X X` and have source location for `M` only and if you only emit a mapping for M, the program interpreting the sourcemap will consider that mapping also apply to `X X X`.

Do you know which part of this PR fixes your issue ? Have you measured the impact of boxing ori_location ?

compiler/lib/source_map.ml
```diff
@@ -79,15 +87,7 @@ let string_of_mapping mapping =
     if i < len
     then
       let c = a.(i) in
-      if prev >= 0
-         && c.ori_source = a.(prev).ori_source
-         && c.ori_line = a.(prev).ori_line
-         && c.ori_col = a.(prev).ori_col
-      then (* We already are at this location *)
```
hhugo, line 86: Why was this logic removed ?

compiler/lib/source_map.ml
```diff
@@ -198,14 +197,13 @@ let mapping_of_string str =
   readline 0 0 []
 
 let maps ~sources_offset ~names_offset x =
-  let gen_line = x.gen_line in
-  let ori_source = x.ori_source + sources_offset in
```
hhugo, line 202: It seems that this line was turning no source into an invalid source.

dyedgreen:
> I don't think the OCaml compiler has anything to do with this. What makes you think that ?
My understanding was that js_of_ocaml reads a source map produced for the ocaml byte code (by the ocaml compiler) and then translates the mappings to match the generated JavaScript, but this might be inaccurate?

Your explanation makes sense though!

pmwhite:
> If you have the following list of token `M X X X` and have source location for `M` only and if you only emit a mapping for M, the program interpreting the sourcemap will consider that mapping also apply to `X X X`.

Thanks for explaining; this makes sense to me.

> Do you know which part of this PR fixes your issue?

I think the line you pointed out in one of your review comments is indeed the location of bug. However, I'd argue there are several places where the bug _could_ have been. The whole patch fixes the issue for good because you can't forget to handle both cases.

> Have you measured the impact of boxing ori_location?

No, I have not. I think it's actually two extra boxes - one for the option and one for the new record. We could eliminate one of them by introducing a specialized option type where the `Some` case uses an inline record. We could eliminate _both_ boxes if we make `mapping` an abstract type with functions that don't let you access certain fields without checking that they aren't negative one; this seems annoying, though.

hhugo:
> We could eliminate both boxes if we make mapping an abstract type with functions that don't let you access certain fields without checking that they aren't negative one; this seems annoying, though.

You can also turn the mapping record type into a variant with 3 constructors, one for each case defined in the spec.
```
each segment is made up of 1,4 or 5 variable length fields.
```

pmwhite:
> > We could eliminate both boxes if we make mapping an abstract type with functions that don't let you access certain fields without checking that they aren't negative one; this seems annoying, though.
> 
> You can also turn the mapping record type into a variant with 3 constructors, one for each case defined in the spec.
> 
> ```
> each segment is made up of 1,4 or 5 variable length fields.
> ```

Yes, good point; that would elegantly avoid introducing more boxes. That was actually approach I initially considered, not because I wanted to avoid boxing, but just because I thought it a good idea to have the types match the spec. I didn't end up doing that because I thought it would cause the PR to be a bit too large.

I'd be happy to re-write the PR to follow your suggestion. Does that sound good to you? I think if we do that, we probably don't need to measure impact after all because we aren't adding or removing boxes. (aside: I'm not sure how I would get a definitive answer about how much impact the changes had, anyway; it seems like a hard measurement to make without putting a lot time into building good benchmarks)

pmwhite:
Actually, I just started trying that re-write, but I'm quickly discouraged from it because greatly increases the friction of accessing `gen_line` and `gen_col`, which are accessed rather frequently it seems. On top of the code friction (which could be reduced by using a helper function that matches on all the cases and extracts the relevant field), I would think that the cost of matching on the type constantly would be similar to the savings from not having an extra box. Would you be okay with sticking with this PR's approach?

hhugo:
@pmwhite, I've done a minimal fix  in #1449 and tried the 3 constructors approach in https://github.com/ocsigen/js_of_ocaml/pull/1450. It seems to improve both speed and memory consumption.

I run `js_of_ocaml link` with large js files containing sourcemaps. I did 4-5 runs and kept the best one against this PR, master, and #1450.

Here are the results 


master
```
4.52user 0.31system 0:04.84elapsed 99%CPU (0avgtext+0avgdata 1118328maxresident)k
0inputs+161560outputs (0major+319701minor)pagefaults 0swaps
```

This PR
```
4.73user 0.44system 0:05.18elapsed 99%CPU (0avgtext+0avgdata 1230924maxresident)k
0inputs+161568outputs (0major+347848minor)pagefaults 0swaps
```

#1450
```
4.02user 0.33system 0:04.36elapsed 100%CPU (0avgtext+0avgdata 917084maxresident)k
0inputs+161568outputs (0major+269390minor)pagefaults 0swaps
```

#1450 is 10 to 15% faster and consume 17% to 25% less.

pmwhite:
Oh, that _is_ significant. Thank you very much for doing this!

dyedgreen:
Thank you both for doing this!

PR #1447, opened 2023-04-07
Treat *.ml{i} in Linguist as OCaml

kevinji:
The Linguist stats on the bottom-right of the main page are incorrectly recognizing some files as SML; this fixes them to be OCaml instead.


hhugo:
Thank you

PR #1453, opened 2023-04-21
For debugging, print escaped variables before allocating names

OlivierNicole:
This is a fix to #1448. Currently, when some variables escape, we print their names after calling `Strategy.allocate`; as a result those escaped names no longer appear in the program and the debug output is unhelpful. This prints the program before allocating new names.

### DIFF
--- a/compiler/lib/js_assign.ml
+++ b/compiler/lib/js_assign.ml
@@ -294,7 +294,15 @@ module Preserve : Strategy = struct
         let _assigned =
           S.fold
             (fun var assigned ->
-              assert (String.is_empty names.(Var.idx var));
+              if not (String.is_empty names.(Var.idx var))
+              then (
+                assert (debug_shortvar ());
+                Format.eprintf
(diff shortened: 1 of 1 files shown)


hhugo:
> as a result those escaped names no longer appear in the program and the debug output is unhelpful. 

I have to disagree.

#1448 allocate variables first so that only unbound/unallocated variable appear  in the `<_eF_>` form;

OlivierNicole:
Then I am not sure where the problem lies exactly, but with #1448 I had a message `Some variables escaped: aF_e`, but the name `<aF_e>` was nowhere to be found in the printed program.

hhugo:
> ```diff
> +  IdentSet.iter
> +    (function
> +      | S _ -> ()
> +      | V x -> names.(Var.idx x) <- "")
> +    free;
> ```

I've pushed the diff on master

hhugo:
Closing. Please open a new issue or PR is you're experiencing more issues

PR #1454, opened 2023-04-21
Fix "clean" target for effect benchmarking

OlivierNicole:
The `clean` Makefile target is not removed by `make clean`, resulting in errors if you run in sequence `make clean` and `make graphseff` (or `make clean` and just `make`).

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -257,7 +257,7 @@ _noperf:
 
 
 clean: _noperf
-	rm -rf results build __precomp __run *.svg *.pdf *~ precomptmp __missingsizes __missingcompiletimes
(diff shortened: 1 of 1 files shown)


benchmarks/Makefile
```diff
@@ -257,7 +257,7 @@ _noperf:
 
 
 clean: _noperf
-	rm -rf results build __precomp __run *.svg *.pdf *~ precomptmp __missingsizes __missingcompiletimes
+	rm -rf results build __precomp __run *.svg *.pdf *~ precomptmp __missingsizes __missingcompiletimes __run_effects
```
hhugo, line 260:
```suggestion
	rm -rf results build __precomp __run __run_effects *.svg *.pdf *~ precomptmp __missingsizes __missingcompiletimes
```

PR #1455, opened 2023-04-21
Bench: revert f985d962db91c35bd4b4224edd9f89c51bf4a6d9

hhugo:


hhugo:
@OlivierNicole, since you're playing with benchmarks, can you test this change ?

This is a revert of an old commit of mine, plus some doc.

OlivierNicole:
I can run the benchmarks (it works), but I’m not sure if you want me to do something more.

hhugo:
> I can run the benchmarks (it works), but I’m not sure if you want me to do something more.

running the benchmarks and generate the report should be enough.
Can you try generating reports ?

OlivierNicole:
Yes, it works.

PR #1461, opened 2023-04-28
Effects: double translation of functions and dynamic switching between direct-style and CPS code

OlivierNicole:
This feature makes programs that use OCaml 5 effects run faster in Javascript, by running as little continuation-passing style (CPS) code as possible. Based on an initial suggestion by @lpw25, we generate two versions for each functions that may be called from inside an effect handler (according to the existing static analysis): a direct-style version and a CPS version. At runtime, direct-style versions are used, except when entering an effect handler, in which case only CPS is run until the outermost effect handler is exited. This approach trades speed for program size, since—because a number of functions are compiled to two versions—the generated programs are bigger. For this reason, the feature is opt-in behind the `--enable doubletranslate` flag[^1]. This is a joint work with @vouillon.

We encountered a design difficulty: when functions are transformed into pairs of functions, it is unclear how to deal with captured identifiers when the functions are nested. To avoid this problem, functions that must be transformed are lambda-lifted, and thus no longer have any free variables except toplevel ones.

The transform is rather successful in preserving the performance of small / monomorphic programs.

- ~~I hypothesize that `hamming` is slower for the same reason as on current master: it uses lazy values, which are an obstacle for the global flow analysis.~~ Edit: I am not able to reproduce a slowdown on hamming in my latest round of benchmarks.
- A number of micro-benchmarks are somewhat faster, maybe because the static analysis performed during the CPS transform is better at finding exact-arity calls.
- I am not sure why `fft` is slightly slower, however. The generated codes look very similar.

EDIT: Note that in the following benchmarks, `--enable=effects` means the vanilla CPS transform; `--enable=effects w. dyn. sw.` (with dynamic switching) means the CPS transform with double translation enabled.

image

The difference becomes negligible on large programs. CAMLboy is actually… 3 % faster with effects enabled (compared to 25 % slower previously): 520 FPS instead of 505 FPS, although the standard deviation is high at ~11 FPS, so it would be fair to say that the difference is not discernable.

`ocamlc` is not discernably slower, either (compared to 10 % slower previously).

https://github.com/ocsigen/js_of_ocaml/pull/1461#issuecomment-1889593948 breaks down which parts of program are actually made faster or slower, and why typical effect-using programs will be faster with this feature.

As some functions must be generated in two versions, the size of the generated code is larger (up to 76 % larger), a few percents larger when compressed.

image

image

Compiling `ocamlc` is about 70 % slower; the resulting file is 64 % larger when compressed.

A caveat of this approach is that all benefits are lost as soon as an effect handler is installed. This is an issue for scheduling libraries such as Eio, as they usually work by having an effect handler installed for the program’s entire lifetime. To mitigate this, we provide `Js_of_ocaml.Js.Effect.assume_no_perform : (unit -> 'a) -> 'a`. Evaluating `assume_no_perform f` runs the direct style version of `f`—the faster version. This also applies to transitive callees of `f` that do not use effect handlers. The programmer must ensure that these functions do not perform effects (not without installing a new effect handler).

[^1]: During the review process, this has been modified to `--effects=double-translation`.


OlivierNicole:
It looks like these two [effect handler benchmarks](https://github.com/kayceesrk/code-snippets/tree/master/eff_bench) are slower with this PR, 25 % and 16 % slower, respectively. I need to spend some time on it to understand why.

|            | 5.2.0      | this PR  |
| ---------- | ---------- | -------- |
| generators |    6.952 s |  8.714 s |
| chameneos  |    64.9 ms |    75 ms |

kayceesrk:
Chameneos runs are too short. You should increase the input size. It takes the input size as a command line argument. Something that runs for a second is more representative as it eliminates the noise due to  JIT and other constant time costs.

OlivierNicole:
Good point. I find that chameneos is 9,8 % slower with this PR, 3.428 s versus 3.753 s.

My theory is that effect handlers are slightly slower due to the fact that function calls in CPS code cost an additional field access (applying `f.cps` instead of just `f`). So these benchmarks that use effect handlers intensively are unfavorable. However, I expect that programs whose execution mixes more usual code with some effect handling (i.e., programs that do not spent all of their time in effect handlers) will see their performance much improved by this PR, like the non-effect-using programs above.

kayceesrk:
I agree with the reasoning and do not expect real programs to behave like `generators` or `chameneos`. The performance difference is small enough that I would consider the performance to be good enough for programs that heavily use effect handlers.

kayceesrk:
Btw, the original numbers aren't useful to understand the improvements brought about by this PR. For this, you need 3 variants:

1. default
2. --enable=effects on `master`
3. --enable=effects on this PR

I'd be interested to see the difference between (2) and (3) in addition to the current numbers which show the difference between (1) and (3).

vouillon:
> My theory is that effect handlers are slightly slower due to the fact that function calls in CPS code cost an additional field access (applying `f.cps` instead of just `f`). So these benchmarks that use effect handlers intensively are unfavorable. 

Note that `f.cps(x1,...,xn)` is a method call, which is somewhat slower than a plain function call. It might be faster to do the following instead:`f.cps.call(null,x1,...,xn)`

I had to do that in https://github.com/ocsigen/js_of_ocaml/pull/1397:
https://github.com/ocsigen/js_of_ocaml/blob/3b3f66ba0f2755ff96c5f0a5ae41d2fb2fb1f8ea/compiler/lib/generate.ml#L954-L959

OlivierNicole:
I believe that the form `f.cps.call(null, x1, ..., xn)` is already the one used.



> Btw, the original numbers aren't useful to understand the improvements brought about by this PR. For this, you need 3 variants:
> 
>     1. default
> 
>     2. --enable=effects on `master`
> 
>     3. --enable=effects on this PR
> 
> 
> I'd be interested to see the difference between (2) and (3) in addition to the current numbers which show the difference between (1) and (3).

Here are the graphs showing the difference between `--enable=effects` on `master` (revision 5.2.0) and `--enable=effects` on this PR:

image

image

image

kayceesrk:
Thanks. The execution time improvement is smaller than what I would have expected. Is that surprising to you or does it match your expectation? 

Also, it would be useful to have all the variants plotted in the same graph with `direct` as the baseline.

OlivierNicole:
It more or less matches my expectation. My reasoning is the following: on most of these small, monomorphic benchmarks, the static analysis will eliminate most CPS calls at compile time. Therefore, the dynamic switching will not change the run time a lot and maybe slightly worsen it. On benchmarks that heavily use effect handlers, I also expect the run time to be worse: most of the time is spent in CPS code anyway, the dynamic switching only adds overhead.

I therefore expect the biggest improvements to happen on larger programs, on which the static analysis does not work as well due to higher-order and mutability; and which do not spend most of their time in effect handlers.

If my hypothesis is verified, then the question is: is this trade-off acceptable? Keeping in mind that there might be ways to improve this PR to save more performance.

OlivierNicole:
> Also, it would be useful to have all the variants plotted in the same graph with `direct` as the baseline.

I have updated the PR message with new graphs showing all the variants.

OlivierNicole:
I tried to build a benchmark that uses Domainslib, as an example of a more typical effect-using program. But the linker complains that the primitive `caml_thread_initialize` missing.

I tried to add it in a new `runtime/thread.js` file but I doesn’t seem to be taken into account, I’m not sure what is the way to add a primitive.

Also, when I build js_of_ocaml I’m getting a lot of primitive-related messages that I don’t really understand:
```
$ dune exec -- js_of_ocaml
Entering directory '/home/olivier/jsoo/js_of_ocaml'
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
warning: overriding primitive "caml_call_gen"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:154
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:151
warning: overriding primitive "caml_call_gen_cps"
  old: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib.js:159
  new: /home/olivier/jsoo/js_of_ocaml/_build/default/runtime/stdlib_modern.js:156
```

OlivierNicole:
I think that this PR is ready for review. The only two problems that prevent the CI from being green are:

- the `Array.fold_left_map` function being available only from 4.13. What is the policy in this case, do we add it to `compiler/lib/stdlib.ml`or do we avoid using it?
- a stack overflow when running the testsuite with profile `using-effects`, which I have yet to investigate.

hhugo:
> the Array.fold_left_map function being available only from 4.13. What is the policy in this case, do we add it to compiler/lib/stdlib.ml or do we avoid using it?

Just add it to the stdlib module compiler/lib/stdlib.ml

compiler/lib/code.ml
```diff
@@ -152,7 +154,8 @@ end = struct
 
   let to_string ?origin i = Var_printer.to_string printer ?origin i
 
-  let print f x = Format.fprintf f "v%d" x
+  (*let print f x = Format.fprintf f "v%d" x*)
+  let print f x = Format.fprintf f "%s" (to_string x)
```
hhugo, line 158: What is this diff about ?

OlivierNicole: It was to clarify variable names in debugging. I had forgotten to remove it.

hhugo:
From a quick look at the PR, the benefit of such change is not clear, can you highlight examples where we see clear improvements ?

OlivierNicole:
> Lambda_lifting doesn't seem to be used anymore, is it expected ? Should Lambda_lifting_simple replace Lambda_lifting ?

As I discovered just today, `Lambda_lifting` is still relevant to avoid generating too deeply nested functions. I just pushed a commit that reinstates the post-CPS-transform `Lambda_lifting.f` pass. Therefore, `Lambda_lifting` is now used.

There are now two lambda lifting passes for two different reasons. `Lambda_lifting` and `Lambda_lifting_simple` do rather different things. The latter simply lifts functions to toplevel, but it takes as a parameters which functions to lift and returns some information about the lifted functions, to be used by the subsequent CPS transform. It also handles mutually recursive functions. `Lambda_lifting` does no such thing as they are not useful for its purpose, however the lifting threshold and baseline are configurable.

For this reason I am not convinced that there is an interest in merging the two modules.

> From a quick look at the PR, the benefit of such change is not clear, can you highlight examples where we see clear improvements ?

I am convinced that most real-world effect-using programs will benefit from this PR, for the reasons given in [my above message](https://github.com/ocsigen/js_of_ocaml/pull/1461#issuecomment-1534716088); but it’s hard to prove it, because we don’t have (yet) examples of such typical programs that work in Javascript. Programs using Domainslib don’t work well with js_of_ocaml (and are arguably not really relevant as JS is not a multicore language). Concurrency libraries like Eio are a more natural fit. I am currently trying to cook up a benchmark using the [experimental JS backend for Eio](https://github.com/ocaml-multicore/eio/pull/405/).

hhugo:
Given the size impact of this change, it would be nice to be able to disable (or control) this optimization. There are programs that would not benefit from this optimization, it be nice to not have to pay the size cost for no benefits.  

The two lambda_lifting modules are confusing, we should either merge them (allowing to share duplicated code) or at least find better names.

> Compiling ocamlc is about 70 % slower

Do you know where this come from ? does it mostly come from the new lambda lifting pass ? or are other passes heavily affected as well (generate, js_assign, ..)

compiler/tests-compiler/direct_calls.ml
```diff
@@ -123,53 +123,91 @@ let%expect_test "direct calls with --enable effects" =
            M1.f 1; M2.f 2
 |}
   in
-  print_fun_decl code (Some "test1");
-  print_fun_decl code (Some "test2");
-  print_fun_decl code (Some "test3");
-  print_fun_decl code (Some "test4");
+  print_fun_decl code (Some "test1$0");
+  print_fun_decl code (Some "test1$1");
+  print_var_decl code "test1";
```
hhugo, line 128: Maybe a helper could extract names to lookup automatically by looking up `test1`, and collecting arguments to `caml_cps_closure`.

compiler/tests-compiler/error.ml
```diff
@@ -25,14 +25,18 @@ let normalize x =
   |> Str.global_replace (Str.regexp "node\\(.exe\\)?") "%{NODE}"
 
 let%expect_test "uncaugh error" =
-  let prog = {| let _ = raise Not_found |} in
+  let prog =
+    {|
+(* Ignore OCAMLRUNPARAM=b to be reproducible *)
+Printexc.record_backtrace false;;
```
hhugo, line 31: I think we should do something at the level of `compile_and_run` instead of repeating this logic everywhere.

compiler/tests-compiler/lambda_lifting.ml
```diff
@@ -22,17 +22,23 @@ Printf.printf "%d\n" (f 3)
...
+                ? f(a0, a1)
+                : runtime.caml_call_gen(f, [a0, a1]);
+       }
+       runtime.caml_initialize_fiber_stack();
        var
-        runtime = globalThis.jsoo_runtime,
         global_data = runtime.caml_get_global_data(),
         Stdlib_Printf = global_data.Stdlib__Printf,
-        _b_ =
+        _e_ =
```
hhugo, line 35: I'm surprised by this diff, do you understand why it's here ?

compiler/tests-compiler/effects_toplevel.ml
```diff
@@ -39,60 +39,47 @@ let%expect_test "test-compiler/lib-effects/test1.ml" =
...
+        _b_ =
+          [0,
+           [11, caml_string_of_jsbytes("abc"), 0],
+           caml_string_of_jsbytes("abc")],
+        Stdlib_Printf = global_data.Stdlib__Printf;
+       function g$0(param){return caml_call1(Stdlib_Printf[2], _b_);}
+       function g$1(param, cont){
+        return caml_cps_call2(Stdlib_Printf[2], _b_, cont);
+       }
+       var g = runtime.caml_cps_closure(g$0, g$1);
```
hhugo, line 71: Why do we still generate the cps version of `g` ?

compiler/lib/subst.mli
```diff
@@ -39,3 +39,15 @@ val from_array : Var.t array -> Var.t -> Var.t
 val build_mapping : Var.t list -> Var.t list -> Var.t Var.Map.t
 
 val from_map : Var.t Var.Map.t -> Var.t -> Var.t
+
+(** The operations of this module also substitute the variables names that
+    appear on the left-hand-side of a {!constructor:Code.Let}, or as block
+    parameters, or as closure parameters, or are bound by an exception handler.
+    *)
+module Bound : sig
```
hhugo, line 47: I think we should move existing substitution function into another sub module to have a more regular API.  


```
val program : (Var.t -> Var.t) -> program -> program
val expr : (Var.t -> Var.t) -> expr -> expr
val instr : (Var.t -> Var.t) -> instr -> instr
val instrs : (Var.t -> Var.t) -> (instr * loc) list -> (instr * loc) list
val block : (Var.t -> Var.t) -> block -> block
val last : (Var.t -> Var.t) -> last * loc -> last * loc
val cont : (Var.t -> Var.t) -> int -> program -> program
```

OlivierNicole:
Thank you for the review and sorry for the response delay, I have been prioritizing another objective in the previous weeks.

One update is that there is no performance gain on programs that use Eio, which is a shame as it is expected to be one of the central uses of effects. More generally, when the program stays almost all the time within at least one level of effect handlers, there is essentially no performance gain as we run the CPS versions of every function. And I expect this programming pattern (installing a topmost effect handler at the beginning of the program) to be the most common with effects.

So it is unclear to me yet if the implementation of double translation can be adapted to accommodate this.

hhugo:
> when the program stays almost all the time within at least one level of effect handlers, there is essentially no performance gain as we run the CPS versions of every function.when the program stays almost all the time within at least one level of effect handlers, there is essentially no performance gain as we run the CPS versions of every function

The PR should speed up program that don't use effects but still compile with `-enable effects`.  I guess you've witnessed the difference using ocamlc ?

OlivierNicole:
Yes, but `ocamlc` doesn’t use effects. In its current state, this PR is only useful if there are many real-world programs that both use effects and are sped up by this.

hhugo:
Have you considered doing a static analysis to find out function calls that never perform effects (Like `List.map f l` when f doesn't doesn't perform effect). and use direct call for them, even inside an handler.

I though that this PR was doing exactly that, but  reading the description again, it seems it does not.

lpw25:
> when the program stays almost all the time within at least one level of effect handlers, there is essentially no performance gain as we run the CPS versions of every function.

I thought there was an analysis that determined some functions do not ever perform effects, presumably you always call the direct versions of those even if you are in a handler.

vouillon:
We already keep in direct style functions which are determined not to perform any effect through a data-flow analysis.

I experimented with using the locality mode to keep more functions in direct style, but this does not work well with a partial CPS transformation, since we have to go back and forth between direct and CPS style (depending on whether we were able to determine that a function could be in direct style or not), which is costly. That should work much better here since once we switch to direct style, we can keep on calling direct versions of the functions.

OlivierNicole:
> Have you considered doing a static analysis to find out function calls that never perform effects (Like List.map f l when f doesn't doesn't perform effect). and use direct call for them, even inside an handler.

As @vouillon said we do have a static analysis since https://github.com/ocsigen/js_of_ocaml/pull/1384, but currently it is unable to do much on higher-order functions like `List.map`.

OlivierNicole:
We discussed this with @vouillon and agreed that I would put the CPS transform with dynamic switching behind a command line flag, and default to the current CPS transform.

compiler/tests-compiler/lambda_lifting.ml
OlivierNicole: Isn’it because the generated program contains more variables, so all generated identifier names are “shifted”? And this effect is only visible on identifiers which do not correspond to an OCaml variable.

OlivierNicole: This is only a hypothesis; to be sure I would need to fully understand how these identifiers are generated.

hhugo:
@OlivierNicole, sorry for the radio silence. Would you be to rebase your PR on master.
Thanks

OlivierNicole:
@hhugo No need to apologize, the radio silence has been mine. I have been struggling to allocate time for the task of making this feature optional, but I finally have completed it in the last weeks. I only need to rebase—doing it now.

OlivierNicole:
Now rebased on master. In the updated version of this PR, the dynamic switching feature (a.k.a. double translation) is now guarded by the `--enable doubletranslate` flag (not set by default). This implies that some Javascript primitives of the runtime now exist in two versions, one for when double translation is enabled and one for when it is not.

This PR is best reviewed as a whole (not commit by commit). A lot of the diff comes from added tests.

Potential blockers:

- I have copied all tests of `tests-ocaml/lib-effects` into a new subdirectory, in order to perform the same tests with double translation enabled. However, there might be a way to perform these tests without copying the files, e.g. using Dune profiles.
- The generated Javascript has different variable names—for variables that are not mapped to an OCaml identifier. I am not sure yet what is the origin of this.
- When trying to compile `ocamlc` with double translation and run it, I get a `Fatal error: exception Stack overflow`. Very strangely, if I run it with `OCAMLRUNPARAM=b`, I don’t get the stack overflow. I’m investigating this. Do you have any idea what could cause it?

compiler/lib/subst.mli
OlivierNicole: https://github.com/ocsigen/js_of_ocaml/pull/1461/commits/0a8183c79c7fb92afd20a0fdc836e5aeb4fb81c4

compiler/tests-compiler/error.ml
OlivierNicole: I have removed this logic from the tests.

compiler/tests-compiler/direct_calls.ml
OlivierNicole: Done in https://github.com/ocsigen/js_of_ocaml/pull/1461/commits/2360daaae0be44740247de1a188302500b46f8af.

OlivierNicole:
When running the tests locally with `--profile using-effects-double` (a derivative of `using-effects` with double translation enabled), I get internal failures of Node.js on two tests:

```
File "compiler/tests-ocaml/lib-bigarray/dune", line 6, characters 0-136:
 6 | (rule
 7 |  (target bigarrays.referencejs)
 8 |  (deps bigarrays.bc.js)
 9 |  (action
10 |   (with-stdout-to
11 |    %{target}
12 |    (run node ./bigarrays.bc.js))))
node:internal/vm:73
  const result = compileFunction(
                 ^

RangeError: Maximum call stack size exceeded
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1178:20)
    at Module._compile (node:internal/modules/cjs/loader:1220:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.17.1
File "compiler/tests-jsoo/dune", line 22, characters 1-33:
22 |  (inline_tests
23 |   (modes js best))
node:internal/vm:73
  const result = compileFunction(
                 ^

RangeError: Maximum call stack size exceeded
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1178:20)
    at Module._compile (node:internal/modules/cjs/loader:1220:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.17.1
```

hhugo:
The CI is also complaining with `TypeError: b.caml_initialize_fiber_stack is not a function`

OlivierNicole:
It looks like my directory-local modifications to the environment—adding `--enable doubletranslate`—are not taken into account

OlivierNicole:
Actually, I think this is because with the profile `using-effects` the runtime is built with `--enable effects` but not `--enable doubletranslate`, which is incompatible with executables built with `--enable doubletranslate`.

An alternative would be, instead of a separate test directory for double translation, to have the CI run the tests in `tests-ocaml/lib-effects` with a new profile `using-effects-double-translation`.

OlivierNicole:
I had forgotten `(compilation_mode whole_program)`, which fixed it.

The remaining failures are due to a field of the `List` module changing indices between 4.14 and 5.0, which is a problem with the test rather than with my PR.

hhugo:
@vouillon, have you done a full review of this PR ? Do you think it is in a merge-able state ?

OlivierNicole:
I don’t think the whole code has been reviewed yet.

hhugo:
I've rebased the PR on master and did some small cleanup

OlivierNicole:
Thanks.

kayceesrk:
I'm trying to understand the results in this PR, but with the given the current results, the sense I get is that double translation  is 

1. slower than default and enabling effects with larger code size on programs that do not use effect handlers: https://github.com/ocsigen/js_of_ocaml/pull/1461#issue-1688951750
2. slower than enabling effects on programs that use effect handlers: https://github.com/ocsigen/js_of_ocaml/pull/1461#issuecomment-1528124024

Is this a correct reading? What is missing (or I am failing to see) are cases where the double translation is faster where it is expected to be. Do you have a (micro-)benchmark that exhibits the best-case scenario for double translation in terms of running time? Such a program would be useful to determine the upper limit of the performance benefit; any real programs will have an improvement that is smaller.

OlivierNicole:
> 1. slower than default and enabling effects with larger code size on programs that do not use effect handlers: https://github.com/ocsigen/js_of_ocaml/pull/1461#issue-1688951750

I don’t think the data points unilaterally in the direction of a slowdown. Microbenchmarks are sometimes faster, sometimes slower. Arguably, when they are slower, they can be up to 20 % slower, but this only happens on one benchmark. Macroscopic benchmarks that do not use effect handlers show no significant differences in run time. It is true that the code size is larger.

>Is this a correct reading? What is missing (or I am failing to see) are cases where the double translation is faster where it is expected to be. Do you have a (micro-)benchmark that exhibits the best-case scenario for double translation in terms of running time?

I expect the improvements to happen on larger programs which do not spend most of their time in effect handlers. Unfortunately, this does not seem to be a typical use case as of now, as e.g. Eio programs tend to install an effect handler at the start of the program, in which case dynamic switching will probably not help and will only increase the code size. I can try to write programs that would benefit from dynamic switching, I just fear they will not represent real-life use cases.

lpw25:
I think that in its current state you're probably not going to see much benefit on programs using effects for concurrency. However, you probably would see benefit in, say, a program that used effects to implement generators. Such programs use effects, but don't spend most of their execution time within the effect handler.

The real win with this approach requires some analysis that can tell you when a particular call of a function won't use effects. That would come naturally out of the type-based approaches that we'll be experimenting with at Jane Street, but you could also probably write other static analyses for obtain such information. For instance, you could look at a call to `List.map` and see that `List.map` itself doesn't use effects and that none of its arguments contain functions that use effects, and know that this specific call couldn't use effects.

kayceesrk:
> However, you probably would see benefit in, say, a program that used effects to implement generators.

Would this program make sense? https://github.com/ocaml-multicore/effects-examples/blob/master/generator.ml

The program benchmarks 3 different versions of iterating over a binary tree -- a plain iterator, a generator implemented by hand using CPS and one using effect handlers. The native code executable produces the following output:

```bash
% dune exec ./generator.exe
Iter: mean = 0.100558, sd = 0.000017
Gen_cps: mean = 0.336419, sd = 0.000009
Gen_eff: mean = 1.506840, sd = 0.000424
```

I would be curious to see the performance of this program with 

1. CPS but no static analysis (from https://github.com/ocsigen/js_of_ocaml/pull/1384; but I'm not sure if the option to enable/disable the analysis is behind a flag)
2. CPS but with the static analysis (current trunk)
3. This PR.

OlivierNicole:
Thank you for suggesting this benchmark.

Here is the output with only the static analysis (current state on master):
```
Iter: mean = 1.623000, sd = 0.007848
Gen_cps: mean = 0.928100, sd = 0.004282
Gen_eff: mean = 5.990300, sd = 0.046392
```

Here is the output with static analysis + double translation (this PR):
```
Iter: mean = 0.269500, sd = 0.001488
Gen_cps: mean = 0.667400, sd = 0.000239
Gen_eff: mean = 6.906500, sd = 0.158372
```

A first notable thing is that the effect-using code (`Gen_eff`) is about 17 % slower. A second notable thing is that code that functions that do not use effects at all (`Iter` and `Gen_cps`) are 500 % and 39 % _faster_, respectively.

How is this possible? It’s because all three benchmarks use a function, namely `Tree.iter`, which needs to be translated in CPS. Without double translation, `Iter` and `Gen_eff` are “contaminated” as they have to run `Tree.iter` in CPS. With double translation, not effect-using code is essentially unaffected.

The consequences of this “contamination” are aggravated  if we modify the benchmark to call `Tree.iter` more often. (This is not so unrealistic: think about a code that relies a lot on `List.map`, for instance, a function that easily gets contaminated.)

```diff
diff --git a/test_generator/generator.ml b/test_generator/generator.ml
index df7f2741f6..49e1a81073 100644
--- a/test_generator/generator.ml
+++ b/test_generator/generator.ml
@@ -17,6 +17,8 @@ module type TREE = sig
   (** [deep n] constructs a tree of depth n, in linear time, where every node at
       level [l] has value [l]. *)
 
+  val iter : ('a -> unit) -> 'a t -> unit
+
   val to_iter : 'a t -> ('a -> unit) -> unit
   (** Iterator function. *)
 
@@ -123,7 +125,12 @@ let t = Tree.deep n
 let iter_fun () = Tree.to_iter t (fun _ -> ())
 let m, sd = benchmark iter_fun 10
 let () = printf "Iter: mean = %f, sd = %f\n%!" m sd
-let rec consume_all f = match f () with None -> () | Some _ -> consume_all f
+let t' = Tree.deep 4
+let total = ref 0
+let rec consume_all f =
+  match f () with
+  | None -> ()
+  | Some n -> Tree.iter (fun m -> total := !total + m + n) t'; consume_all f
 
 let gen_cps_fun () =
   let f = Tree.to_gen_cps t in
```

Output with only the static analysis:
```
Iter: mean = 1.621400, sd = 0.013772
Gen_cps: mean = 32.865400, sd = 3.180572
Gen_eff: mean = 41.179400, sd = 6.977796
```

Output with static analysis + double translation:
```
Iter: mean = 0.254200, sd = 0.002514
Gen_cps: mean = 5.582200, sd = 0.113679
Gen_eff: mean = 13.450000, sd = 0.489052
```

Here, all three benchmarks are much faster with double translation, even the one using effects.

This is also the reason why double translation would combine very well with the locality modes of Jane Street for effects: as @vouillon’s experiments have shown, they allow many more functions to be kept in direct style, but—paradoxically—this forces to switch more often between direct style and CPS, which is costly (another case of contamination). With double translation, this is not the case: all these functions can be called in direct style.

kayceesrk:
Thanks for the benchmark results. The purpose of the PR is now apparent from the performance side.

OlivierNicole:
This PR is now rebased on master and ready again for review (cc @vouillon).

Of note: as previously, a number of expect tests underwent some variable name substitution, i.e., they are the same up to alpha equivalence. I have not looked closely at the cause; I assume that in the CPS transform, some new variables are now created and not always used. If this is a problem, I can take some time to look at it.

OlivierNicole:
I pushed a commit which adds a new primitive, `caml_assume_no_effects`, which allows to guarantee that a function is called in its (faster) direct-style version, for optimization purposes. See commit message for details.

OlivierNicole:
Coming back to this after spending time on the merge between js_of_ocaml and wasm_of_ocaml.

I’ve rebased this PR on master, re-aligned things where needed, and made the CI happy again. It’s ready for review.

cc @vouillon

vouillon:
In the message of the second commit, you write:
> As a consequence, performing an effect in a transitive callee
of [f] will raise `Effect.Unhandled`, regardless of any effect handlers
installed before the call to `caml_assume_no_effects`, unless a new
effect handler was installed in the meantime.

Maybe I'm missing something, but I don't see why this should be true. Performing an effect will result in calling function `caml_perform_effect`, which calls the function `uncaught_effect_handler` only if the fiber stack is currently empty.

In any case, you should add a test for that.

vouillon (changes requested):
Here is a first batch of comments

OlivierNicole:
> In the message of the second commit, you write:
> 
> > As a consequence, performing an effect in a transitive callee
> > of [f] will raise `Effect.Unhandled`, regardless of any effect handlers
> > installed before the call to `caml_assume_no_effects`, unless a new
> > effect handler was installed in the meantime.
> 
> Maybe I'm missing something, but I don't see why this should be true. Performing an effect will result in calling function `caml_perform_effect`, which calls the function `uncaught_effect_handler` only if the fiber stack is currently empty.
> 
> In any case, you should add a test for that.

Well spotted, thanks. I added a test and pushed a fixed, but for some reason I get an error with intra-runtime dependencies:

```
File "compiler/lib-runtime-files/dune", lines 6-14, characters 0-143:
 6 | (rule
 7 |  (target files.ml)
 8 |  (deps
 9 |   gen/gen.exe
10 |   (glob_files ../../runtime/js/*.js))
11 |  (action
12 |   (with-stdout-to
13 |    %{target}
14 |    (run %{deps}))))
Fatal error: exception Failure("missing dependency 'caml_uncaught_effect_handler'\n")
```

and I’m not sure how to fix it.

compiler/lib/lambda_lifting_simple.ml
```diff
@@ -0,0 +1,344 @@
...
+                match current_contiguous with
+                | [ (f, _, _, _) ] -> not (Var.Set.mem f to_lift)
+                | _ -> true);
+              match current_contiguous with
+              | [] -> (
+                  match l with
+                  | i :: rem ->
+                      let rem', st = rewrite_body [] st rem in
+                      i :: rem', st
+                  | [] -> [], st)
```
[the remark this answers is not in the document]
OlivierNicole: I have refactored the code to make it simpler and tail-recursive. Let me know if it seems acceptable to you or if you still want to have two separate passes for collecting functions to lift, and rewriting the closures.

compiler/lib/lambda_lifting_simple.ml
```diff
@@ -0,0 +1,344 @@
...
+                  , st )
+              | _ :: _ ->
+                  let rem, st =
+                    match l with
+                    | i :: rem ->
+                        let rem, st = rewrite_body [] st rem in
+                        i :: rem, st
+                    | [] -> [], st
+                  in
+                  ( List.map current_contiguous ~f:(fun (f, params, pc, args) ->
```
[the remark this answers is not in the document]
OlivierNicole: It should be fixed now.

compiler/lib/lambda_lifting_simple.ml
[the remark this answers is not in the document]
vouillon: This is much better.

vouillon (changes requested):
I'm not done reviewing `effects.ml`, but here is another batch of comments to keep you busy.

compiler/lib/effects.ml
```diff
@@ -355,34 +441,37 @@ let allocate_continuation ~st ~alloc_jump_closures ~split_closures pc x cont =
      closure to bind [x] if it is used in the loop body. In
      other cases, we can just pass the closure corresponding
      to the next block. *)
-  let pc', args = cont in
+  let _, args = cont in
```
vouillon, line 444: I think it would be clearer to remove the parameter `direct_pc`, and maybe rename `cont` into `direct_cont`. One would then write:
```suggestion
  let direct_pc, args = direct_cont in
```

compiler/lib/lambda_lifting_simple.ml
```diff
@@ -0,0 +1,363 @@
...
+            let tuple = Var.fresh_n "tuple" in
+            let rev_decl =
+              Let (tuple, Apply { f = f_tuple; args = List.map ~f:fst s; exact = true })
+              :: List.mapi current_contiguous ~f:(fun i (f, _, _, _) ->
+                     Let (f, Field (tuple, i, Non_float)))
+            in
+            ( (program, functions, lifters)
+            , rev_decl
+              @ Let (tuple, Apply { f = f_tuple; args = List.map ~f:fst s; exact = true })
+                :: acc_instr )
```
vouillon, line 297:
```suggestion
              @ acc_instr )
```

vouillon (changes requested):
Another batch of comments...

vouillon:
You can add this to `runtime/wasm/effect.wat`:
```wasm
   (import "obj" "caml_callback_1"
      (func $caml_callback_1
         (param (ref eq)) (param (ref eq)) (result (ref eq))))
```
```wasm
   (func (export "caml_assume_no_perform") (param $f (ref eq)) (result (ref eq))
      (call $caml_callback_1 (local.get $f) (ref.i31 (i32.const 0))))
```

compiler/lib/effects.ml
```diff
@@ -710,53 +940,136 @@ let cps_transform ~live_vars ~flow_info ~cps_needed p =
...
+             subst with fresh ones to avoid clashing with the definitions in the original
+             blocks (the actual substitution is done later). *)
+          if double_translate ()
+          then
+            if function_needs_cps && double_translate ()
+            then
+              Code.traverse
+                Code.{ fold = fold_children }
+                (fun pc () ->
+                  let block = Addr.Map.find pc blocks in
```
vouillon, line 1015:
```suggestion
                  let block = Addr.Map.find pc p.blocks in
```

compiler/lib/effects.ml
```diff
@@ -482,63 +557,86 @@ let cps_last ~st ~alloc_jump_closures pc (last : last) ~k : instr list * last =
...
+        (* Translated like the [Apply] case, with a unit argument *)
+        assert (
+          (* If this function is unknown to the global flow analysis, then it was
+             introduced by the lambda lifting and does not require CPS *)
+          Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
+          || Global_flow.exact_call st.flow_info f 1);
+        let unit = Var.fresh_n "unit" in
+        [ Let (unit, Constant (Int Targetint.zero))
+        ; Let (x, Apply { f; args = [ unit ]; exact = true })
+        ])
```
vouillon, line 625: This case cannot happen since the partial CPS analysis enforces a CPS transform for this primitive.

compiler/lib/effects.ml
```diff
@@ -557,9 +655,33 @@ let cps_block ~st ~k pc block =
...
+    | Prim (Extern "caml_assume_no_perform", [ Pv f ])
+      when (not (double_translate ())) && Var.Set.mem x st.cps_needed ->
+        (* Translated like the [Apply] case, with a unit argument *)
+        Some
+          (fun ~k ->
+            let exact =
+              (* If this function is unknown to the global flow analysis, then it was
+                 introduced by the lambda lifting and is exact *)
+              Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
+              || Global_flow.exact_call st.flow_info f 1
```
vouillon, line 674: Same comment as above: if we don't know the function, we must assume the call is not exact.

vouillon (changes requested):
Another batch of comments... The code looks much better already!

compiler/lib/effects.ml
```diff
@@ -482,67 +557,76 @@ let cps_last ~st ~alloc_jump_closures pc (last : last) ~k : instr list * last =
...
+            || Global_flow.exact_call st.flow_info f (List.length args)
+        | `Direct_style -> Global_flow.exact_call st.flow_info f (List.length args));
+      [ Let (x, Apply { f; args; exact = true }) ]
+  | `Cps, Let (_, (Apply _ | Prim (Extern ("%resume" | "%perform" | "%reperform"), _))) ->
+      (* Applications of CPS functions and effect primitives require more work
+         (allocating a continuation and/or modifying end-of-block branches) and
+         are handled in a specialized function below. *)
       assert false
-  | _ -> instr
+  | `Direct_style, Let (_, e) when effect_primitive_or_application e -> assert false
```
vouillon, line 601: I think you can move these two cases after the cases for `caml_assume_no_perform` and combine them.

compiler/lib/effects.ml
```diff
@@ -482,67 +557,76 @@ let cps_last ~st ~alloc_jump_closures pc (last : last) ~k : instr list * last =
...
-      Let (x, Apply { f; args; exact = true })
-  | Let (_, (Apply _ | Prim (Extern ("%resume" | "%perform" | "%reperform"), _))) ->
+      assert (
+        (* If this function is unknown to the global flow analysis, then it was
+           introduced by the lambda lifting and we don't have exactness info any more. *)
+        match target with
+        | `Cps ->
+            Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
+            || Global_flow.exact_call st.flow_info f (List.length args)
+        | `Direct_style -> Global_flow.exact_call st.flow_info f (List.length args));
```
vouillon, line 594:
```suggestion
            Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
            || Global_flow.exact_call st.flow_info f (List.length args)
```

compiler/lib/effects.ml
```diff
@@ -482,67 +557,72 @@ let cps_last ~st ~alloc_jump_closures pc (last : last) ~k : instr list * last =
...
-      assert (Global_flow.exact_call st.flow_info f (List.length args));
-      Let (x, Apply { f; args; exact = true })
-  | Let (_, (Apply _ | Prim (Extern ("%resume" | "%perform" | "%reperform"), _))) ->
+      assert (
+        (* If this function is unknown to the global flow analysis, then it was
+           introduced by the lambda lifting and we don't have exactness info any more. *)
+        Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
+        || Global_flow.exact_call st.flow_info f (List.length args));
+      [ Let (x, Apply { f; args; exact = true }) ]
+  | `Cps, Let (x, Prim (Extern "caml_assume_no_perform", [ Pv f ])) ->
```
vouillon, line 593: This case is not supposed to happen in direct style. So maybe we can just stop matching on the `target`?
Or we could move it to a separate function:
```ocaml
let cps_instr ~st instr =
  match instr with
  | Let (x, Prim (Extern "caml_assume_no_perform", [ Pv f ])) ->
      ...
  | _ ->
      [rewrite_instr ~st instr]
```

compiler/lib/effects.ml
```diff
@@ -355,35 +422,38 @@ let allocate_continuation ~st ~alloc_jump_closures ~split_closures pc x cont =
...
       if not split_closures
       then alloc_jump_closures, []
-      else if is_merge_node st.cfg pc'
+      else if is_merge_node st.cfg direct_pc
       then [], alloc_jump_closures
       else
-        List.partition
-          ~f:(fun i ->
-            match i with
-            | Let (_, Closure (_, (pc'', []))) -> dominates st.cfg st.idom pc' pc''
```
vouillon, line 384: We can revive this code, using this condition instead:
```ocaml
            | Let (_, Closure (_, (pc'', []))) -> pc'' = mk_cps_pc_of_direct ~st direct_pc
```
Indeed, we have `Hashtbl.find idom pc'' = src_pc` (by construction of `closures_of_alloc_site`). And `direct_pc` cannot dominate `src_pc`  since it comes afterwards. So, in the code below, the second case `dominates g idom direct_pc (Hashtbl.find idom pc'')` will always be false.
```ocaml
let rec dominates g idom pc pc' =
  pc = pc'
  || Hashtbl.find g.block_order pc < Hashtbl.find g.block_order pc'
     && dominates g idom pc (Hashtbl.find idom pc')
```

vouillon (changes requested):
We are almost there!

What about adding an option: `--effects {cps,double-translate,jspi}`. We would keep `--enable effects` for now for compatibility to mean `--effects cps`.

We would have something like this for wasm_of_ocaml in `bin-wasm_of_ocaml/cmd_args.ml`
```ocaml
  let effects =
    let doc =
      "Select an implementation of effect handlers. [$(docv)] should be one of $(b,jspi) \
       (the default) or $(b,cps)."
    in
    Arg.(
      value
      & opt (enum [ "jspi", None; "cps", Some CPS ]) None
      & info [ "effects" ] ~docv:"KIND" ~doc)
  in
```

And this in `bin-js_of_ocaml/cmd_args.ml`
```ocaml
  let effects =
    let doc =
      "Select an implementation of effect handlers. [$(docv)] should be one of $(b,cps) or $(b,double-translation). Effects are not allowed by default."
    in
    Arg.(
      value
      & opt (some (enum [ "cps", CPS; "double-translation", Double_translation ])) None
      & info [ "effects" ] ~docv:"KIND" ~doc)
  in
```
And one needs to update `build_info.ml` to set the option appropriately (so one should probably have a global parameter in `config.ml` like `target` together with some helper functions that would replace `Config.Flag.effects` and `Config.Flag.double_translation`).

compiler/lib/partial_cps_analysis.ml
```diff
@@ -96,15 +96,6 @@ let block_deps ~info ~vars ~tail_deps ~deps ~blocks ~fun_name pc =
...
               add_dep deps f x)
-      | Let (x, Prim (Extern "caml_assume_no_perform", _)) -> (
-          add_var vars x;
-          match fun_name with
-          | None -> ()
-          | Some f ->
-              add_var vars f;
-              (* If a function contains effect primitives, it must be
-                 in CPS *)
-              add_dep deps f x)
```
OlivierNicole: Couldn’t we exploit the fact that functions called under assume_no_perform may not need to be in CPS?

hhugo:
> We are almost there!
> 
> What about adding an option: `--effects {cps,double-translate,jspi}`. We would keep `--enable effects` for now for compatibility to mean `--effects cps`.
> 
> We would have something like this for wasm_of_ocaml in `bin-wasm_of_ocaml/cmd_args.ml`
> 
> ```ocaml
>   let effects =
>     let doc =
>       "Select an implementation of effect handlers. [$(docv)] should be one of $(b,jspi) \
>        (the default) or $(b,cps)."
>     in
>     Arg.(
>       value
>       & opt (enum [ "jspi", None; "cps", Some CPS ]) None
>       & info [ "effects" ] ~docv:"KIND" ~doc)
>   in
> ```
> 
> And this in `bin-js_of_ocaml/cmd_args.ml`
> 
> ```ocaml
>   let effects =
>     let doc =
>       "Select an implementation of effect handlers. [$(docv)] should be one of $(b,cps) or $(b,double-translation). Effects are not allowed by default."
>     in
>     Arg.(
>       value
>       & opt (some (enum [ "cps", CPS; "double-translation", Double_translation ])) None
>       & info [ "effects" ] ~docv:"KIND" ~doc)
>   in
> ```
> 
> And one needs to update `build_info.ml` to set the option appropriately (so one should probably have a global parameter in `config.ml` like `target` together with some helper functions that would replace `Config.Flag.effects` and `Config.Flag.double_translation`).

We also need to update dune rules `Jsoo_rules.Config.{t,of_flags,to_flags}`

compiler/lib/partial_cps_analysis.ml
vouillon: Not easily. At the moment, we are considering that all escaping functions need to be in CPS. But I'm planning to change that.

vouillon: We are not marking anywhere explicitly the function passed to `caml_assume_no_perform` as needing CPS.
Any value passed to a primitive is considered as escaping by the data flow analysis. And any escaping function is marked as needing CPS in function `cps_needed`.

OlivierNicole: Ah, I see.

compiler/lib/effects.ml
```diff
@@ -502,47 +579,41 @@ let cps_instr ~st (instr : instr) : instr =
...
   | _ -> instr
 
-let cps_block ~st ~k pc block =
+let cps_instr ~st (instr : instr) : instr list =
+  match instr with
+  | Let (x, Prim (Extern "caml_assume_no_perform", [ Pv f ])) when double_translate () ->
+      (* When double translation is enabled, we just call [f] in direct style.
+         Otherwise, the runtime primitive is used. *)
+      let unit = Var.fresh_n "unit" in
+      let exact = Global_flow.exact_call st.flow_info f 1 in
```
vouillon, line 601: We need to check everywhere that `Var.idx f < Var.Tbl.length st.flow_info.info_approximation`.
```suggestion
      let exact =
        Var.idx f < Var.Tbl.length st.flow_info.info_approximation
        && Global_flow.exact_call st.flow_info f 1
      in
```

dune
```diff
@@ -9,9 +9,9 @@
   (js_of_ocaml
    (compilation_mode separate)
    (flags
-    (:standard --enable effects))
+    (:standard --effects=cps))
```
vouillon, line 12: I think we need to keep `--enable effects` here, while `--effects=cps` is not supported by dune.

compiler/tests-ocaml/lib-effects/double-translation/dune
```diff
@@ -1,12 +1,12 @@
...
   (flags
    (:standard -w -38))
   (js_of_ocaml
    (flags
-    (:standard --effects=double-translation))
+    ((:standard \ --effects=cps) --effects=double-translation))
    (build_runtime_flags
-    (:standard --effects=double-translation))
+    ((:standard \ --effects=cps) --effects=double-translation))
    ;; separate compilation doesn't work when using
```
vouillon, line 10: We should update the comment to say that dune does not know about `--effects=double-translation`.

dune
OlivierNicole: I’m not sure to understand why Dune needs to know about Js_of_ocaml flags?

compiler/lib/effects.ml
OlivierNicole: Should I still apply `do_alloc_jump_closures` to both resulting lists?

OlivierNicole: Should I still apply `do_alloc_jump_closures` to both resulting lists?

vouillon: No. The point is that you can work on `alloc_jump_closures`.

dune
vouillon: With separate compilation, we don't want to compile libraries (cma files) multiple times when they are used by several executables. But we still have to account for the fact that some flags change the compilation scheme in incompatible way: you cannot link a library compiled with `--enable effects` with some code compiled without this flag, for instance.

Maybe one could add dune options instead of parsing the js_of_ocaml flags. This is [what is done for source maps](https://github.com/ocaml/dune/pull/10777).

compiler/lib/effects.ml
```diff
@@ -273,32 +291,56 @@ type trampolined_calls = Var.Set.t
...
 
+(* Provide the address of the CPS translation of a block *)
+let mk_cps_pc_of_direct ~st pc =
+  if double_translate ()
+  then (
+    try Hashtbl.find st.cps_pc_of_direct pc
+    with Not_found ->
+      let new_blocks = st.new_blocks in
+      let free_pc = st.free_pc in
+      st.new_blocks <- new_blocks;
```
vouillon, line 331:
```suggestion
      let free_pc = st.free_pc in
```

lib/js_of_ocaml/js.ml
```diff
@@ -843,6 +843,12 @@ let export_all obj =
 
 (****)
 
+module Effect = struct
```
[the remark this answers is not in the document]
vouillon: `Jsoo_runtime` feels like an internal library. It does not seem to be documented anywhere. But maybe we should change that?

vouillon (changes requested):
Some bound checks are missing. I only have a few cosmetic changes otherwise.

Regarding the documentation, we should probably still document `--enable effects` somehow while dune has not been updated, which might take a while.

lib/js_of_ocaml/js.ml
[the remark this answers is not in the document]
hhugo: Jsoo_runtime was created so that alternative lib (such as brr and gen_js_api) can rely on jsoo primitives without having to depend on the whole jsoo lib. The module should be re-exported and documented in js_ of_ocaml, the lib.

compiler/bin-js_of_ocaml/cmd_arg.ml
```diff
@@ -253,6 +254,18 @@ let options =
       & opt (some string) None
       & info [ "ofs" ] ~docs:filesystem_section ~docv:"FILE" ~doc)
   in
+  let effects =
+    let doc =
+      "Select an implementation of effect handlers. [$(docv)] should be one of $(b,cps) \
+       or $(b,double-translation). Effects are not allowed by default."
```
hhugo, line 260:
```suggestion
       or $(b,double-translation). Effects won't be supported if unspecified."
```

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -172,6 +173,12 @@ let run
   | `Name _, _ -> ());
   List.iter params ~f:(fun (s, v) -> Config.Param.set s v);
   List.iter static_env ~f:(fun (s, v) -> Eval.set_static_env s v);
+  (* For backward compatibility, consider that [--enable effects] alone means
+     [--effects cps] *)
+  Config.set_effects_backend
```
hhugo, line 178: Maybe move the set call inside the match branch

hhugo: Ignore the request. I misread the whole logic

hhugo: I think that this whole backward compatibility logic should be moved to the cmd_arg module.

OlivierNicole: It’s non-trivial because `Config.Flag.effects ()` is initialized by some separate logic, not in `Cmd_arg`. But I can try.

hhugo: Indeed, it is initialized by Jsoo_cmdline.Arg.eval.

OlivierNicole:
> Regarding the documentation, we should probably still document `--enable effects` somehow while dune has not been updated, which might take a while.

I tried to hit a compromise, keeping the docs about `--enable effects` for now, while documenting also `--effects=double-translation` and documenting possible caching issues with Dune.

manual/effects.wiki
```diff
@@ -11,6 +11,8 @@ The analysis is especially effective on monomorphic code. It is not so effective
 We hope to improve on this by trying alternative compilation
 strategies.
 
+An alternative CPS transform is provided under the {{--effects=double-translation}} option. It keeps a direct-style version of the transformed functions in addition to the CPS version. The choice of running the CPS version is delayed to run time. Since CPS code is usually slower, this can avoid degradations. In addition, one can ensure that some code is run in direct style by using {{Jsoo_runtime.Effect.assume_no_perform}}. A caveat is that Dune does not know about {{--effects=double-translation}} yet and may try to link together files built with {{--enable=double-translation}} and files built with only {{--enable=effects}}, which gives an error. As long as this is not fixed, running {{dune clean}} should help.
```
vouillon, line 14: How about this?
```suggestion
An alternative CPS transform is provided under the {{--effects=double-translation}} option. It keeps a direct-style version of the transformed functions in addition to the CPS version. The choice of running the CPS version is delayed to run time. Since CPS code is usually slower, this can avoid degradations. In addition, one can ensure that some code is run in direct style by using {{Jsoo_runtime.Effect.assume_no_perform}}. A caveat is that Dune does not know about {{--effects=double-translation}} yet and may try to link together files built with {{--enable=double-translation}} and files built with only {{--enable=effects}}, which gives an error. A work-around is to disable separate compilation by using the option {{(js_of_ocaml (compilation_mode whole_program))}}.
```

manual/effects.wiki
```diff
@@ -25,7 +27,7 @@ Put the following in a {{{dune}}} (or {{{dune-workspace}}}) file at the root of
 (env
  (_
   (js_of_ocaml
-   (flags (:standard --enable effects))
+   (flags (:standard --effects=cps))
```
vouillon, line 30:
```suggestion
   (flags (:standard --enable effects))
```

manual/effects.wiki
```diff
@@ -50,7 +52,7 @@ Then pass the rights {{{js_of_ocaml}}} flags to the executable stanza
 {{{
 (executable
   (name main)
-  (js_of_ocaml (flags (:standard --enable effects)))
+  (js_of_ocaml (flags (:standard --effects=cps)))
```
vouillon, line 55:
```suggestion
  (js_of_ocaml (flags (:standard --enable effects)))
```

manual/tailcall.wiki
```diff
@@ -3,7 +3,7 @@ JavaScript does not (yet) support tail call optimization.
 To circumvent this limitation, and mitigate stack overflows, the Js_of_ocaml
 compiler optimizes some common tail call patterns.
 Besides, all tail calls are optimized when you set the flag
-{{{--enable=effects}}}, at the cost of some performance degradation.
+{{{--effects=cps}}}, at the cost of some performance degradation.
```
vouillon, line 6:
```suggestion
{{{--enable=effects}}}, at the cost of some performance degradation.
```

vouillon (changes requested):
You forgot to revert some document changes.

vouillon (approved):

compiler/lib-wasm/generate.ml
```diff
@@ -21,6 +21,8 @@ open Code
 module W = Wasm_ast
 open Code_generation
 
+let effects () = Option.is_some (Config.effects ())
```
hhugo, line 24: Shouldn't this be called effects_cps ?

compiler/lib/config.mli
```diff
@@ -39,6 +40,7 @@ module Flag : sig
 
   val staticeval : unit -> bool
 
+  (* Deprecated in favor of toplevel function [effects] below *)
   val effects : unit -> bool
```
hhugo, line 44: remove this ?

compiler/lib/driver.ml
```diff
@@ -98,28 +98,34 @@ let ( +> ) f g x = g (f x)
...
+        if Config.Flag.globaldeadcode ()
+        then
+          let p = Global_deadcode.f p ~deadcode_sentinal info in
+          Deadcode.f p
+        else p, live_vars
+      in
+      p
+      |> Effects.f ~flow_info:info ~live_vars
+      |> map_fst
+           (match Config.effects () with
```
hhugo, line 118: You can match the inner of the option directly, just name the backend on line 102.

compiler/lib/driver.ml
```diff
@@ -98,28 +98,34 @@ let ( +> ) f g x = g (f x)
...
+           | Some Cps -> Lambda_lifting.f
+           | None -> assert false)
+  | None ->
+      ( p
+      , (Code.Var.Set.empty : Effects.trampolined_calls)
+      , (Code.Var.Set.empty : Effects.in_cps) )
 
 let exact_calls profile ~deadcode_sentinal p =
-  if not (Config.Flag.effects ())
+  if Option.is_none (Config.effects ())
```
hhugo, line 128: a `match .. with` would be clearer I think

compiler/lib/inline.ml
```diff
@@ -331,7 +331,7 @@ let times = Debug.find "times"
 let f p live_vars =
   let first_class_primitives =
     match Config.target () with
```
hhugo, line 333: match on both target and effect

compiler/lib/js_output.ml
```diff
@@ -2101,7 +2101,8 @@ let program ?(accept_unnamed_var = false) ?(source_map = false) f p =
     let accept_unnamed_var = accept_unnamed_var
   end) in
   PP.set_needed_space_function f need_space;
-  if Config.Flag.effects () then PP.set_adjust_indentation_function f (fun n -> n mod 40);
+  if Option.is_some (Config.effects ())
```
hhugo, line 2104: I would prefer an exhaustive `match .. with` here so that we reconsider the need for this every time one add a backend

compiler/lib/unit_info.ml
```diff
@@ -53,7 +53,7 @@ let of_cmo (cmo : Cmo_format.compilation_unit) =
   let requires = StringSet.of_list (Cmo_format.requires cmo) in
   let requires = StringSet.diff requires provides in
   let effects_without_cps =
-    (not (Config.Flag.effects ()))
+    Option.is_none (Config.effects ())
```
hhugo, line 56: Please use a `match .. with` here

compiler/tests-compiler/util/util.ml
```diff
@@ -291,15 +291,21 @@ let compile_to_javascript
     ?(flags = [])
     ?(use_js_string = false)
     ?(effects = false)
+    ?(doubletranslate = false)
```
hhugo, line 294: the effects argument could take a backend instead of a boolean.

compiler/tests-ocaml/lib-effects/double-translation/dune
hhugo: The comment is stale. separate compilation now works with `--enable use-js-srtring` or `--enable effects`.

runtime/js/jslib.js
```diff
@@ -146,8 +139,23 @@ function caml_jsoo_flags_use_js_string(unit) {
...
+function caml_jsoo_flags_effects(unit) {
+  return "cps";
+}
+
+//Provides: caml_jsoo_flags_effects
+//If: effects
+//If: doubletranslate
 function caml_jsoo_flags_effects(unit) {
-  return FLAG("effects");
+  return "double-translation";
```
hhugo, line 158: Maybe we could update `macro.ml` to handle this case. We could have a `CONFIG` macro.

hhugo (changes requested):

hhugo:
Do you already have a `dune` PR to reflect changes made in this PR ?

OlivierNicole:
I am preparing one at the moment.

OlivierNicole:
See ocaml/dune#11222.

compiler/tests-ocaml/lib-effects/double-translation/dune
OlivierNicole: > The comment is stale. separate compilation now works with --enable use-js-srtring or --enable effects.

> ```
> ;; separate compilation doesn't yet work when using '--effect=double-translation' since dune doesn't know it should compile a different version of the dependencies
> ``
> ```

Done.

compiler/lib/driver.ml
```diff
@@ -98,42 +98,47 @@ let ( +> ) f g x = g (f x)
...
-        Deadcode.f p
-      else p, live_vars
-    in
-    p |> Effects.f ~flow_info:info ~live_vars +> map_fst Lambda_lifting.f)
-  else
-    ( p
-    , (Code.Var.Set.empty : Effects.trampolined_calls)
-    , (Code.Var.Set.empty : Effects.in_cps) )
+  match Config.effects () with
+  | Some (_ as effects) ->
```
hhugo, line 102:
```suggestion
  | Some effects ->
```

OlivierNicole: Oops

OlivierNicole:
@hhugo Is there something remaining for this to be merged? Do you consider https://github.com/ocaml/dune/pull/11222 to be a necessary prerequisite?

hhugo:
> @hhugo Is there something remaining for this to be merged? Do you consider [ocaml/dune#11222](https://github.com/ocaml/dune/pull/11222) to be a necessary prerequisite?

I'd like to do another review with a proper screen. Reviewing on a phone is not ideal

compiler/lib/config.ml
```diff
@@ -195,3 +195,13 @@ let set_target (t : [ `JavaScript | `Wasm ]) =
...
   | `Wasm -> Targetint.set_num_bits 31);
   target_ := (t :> [ `JavaScript | `Wasm | `None ])
+
+type effects_backend =
+  | Cps
+  | Double_translation
+
+let effects_ : effects_backend option ref = ref None
+
+let effects () = !effects_
```
hhugo, line 205: What about doing something similar to `target` above.
```ocaml
type effects_backend =
  | `Cps
  | `Double_translation
  | `Jspi
  | `Not_supported
  | `None

let effects_ : effects_backend option ref = ref `None

let effects () = match !effects_ with
  | `None -> failwith "effect backend was not set"
  | (`Cps | `Double_translation | `Jspi | `Not_supported) as t -> t
```

hhugo: I don't like that `None` currently mean unsupported for jsoo and jspi for wasmoo

manual/overview.wiki
```diff
@@ -84,7 +84,11 @@ functions are optimized:
   * trampolines are used otherwise.
 <<a_manual chapter="tailcall" |More about tail call optimization>>.
 
-Effect handlers are fully supported with the {{{--enable=effects}}} flag. This is not the default for now since effects are not widely used at the moment and the generated code can be slower, larger and less readable.
+Effect handlers are fully supported with the
+{{{--enable=effects}}} flag. Effect support is disabled by
```
hhugo, line 88: This needs to be updated

runtime/js/jslib.js
```diff
@@ -61,13 +61,14 @@ function caml_trampoline(res) {
...
-function caml_trampoline_return(f, args) {
-  return { joo_tramp: f, joo_args: args };
+function caml_trampoline_return(f, args, direct) {
+  return { joo_tramp: f, joo_args: args, joo_direct: direct };
 }
 
 //Provides:caml_stack_depth
 //If: effects
-var caml_stack_depth = 0;
+var caml_stack_depth = 10; // Initialized to a non-zero value in case of
```
hhugo, line 70: Can you say more about this ? why 10 and not just 1 ?

hhugo:
> @hhugo Is there something remaining for this to be merged? Do you consider [ocaml/dune#11222](https://github.com/ocaml/dune/pull/11222) to be a necessary prerequisite?


[ocaml/dune#11222](https://github.com/ocaml/dune/pull/11222) is clearly not a prerequisite.

runtime/js/jslib.js
vouillon: I believe this change is no longer relevant and can be reverted.

runtime/js/effect.js
```diff
@@ -215,3 +254,62 @@ function caml_ml_condition_signal(t) {
...
+    caml_exn_stack = 0;
+    caml_fiber_stack = {
+      h: [0, 0, 0, { cps: caml_uncaught_effect_handler }],
+      r: { k: 0, x: 0, e: 0 },
+    };
+    var k = caml_resume_stack(stack, (x) => x);
+    /* Note: f is not an ordinary function but a (direct-style, CPS) closure pair */
+    var res = { joo_tramp: f, joo_args: [arg, k], joo_direct: 0 };
+    do {
+      caml_stack_depth = 40;
```
hhugo, line 276: can we have a comment explaining this ?

compiler/bin-js_of_ocaml/cmd_arg.ml
```diff
@@ -496,6 +523,18 @@ let options_runtime_only =
...
```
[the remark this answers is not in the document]
OlivierNicole: I’m not sure to understand what needs to be updated.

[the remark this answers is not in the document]
OlivierNicole: never mind, the order of comments is probably mixed in the Github interface.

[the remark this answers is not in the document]
hhugo:
```
let doc =
      "Select an implementation of effect handlers. [$(docv)] should be one of $(b,cps) \
       or $(b,double-translation). Effects won't be supported if unspecified."
```

manual/overview.wiki
OlivierNicole: Shouldn’t we wait for the new flag to be supported by Dune before documenting it? See https://github.com/ocsigen/js_of_ocaml/pull/1461#issuecomment-2545773411.

hhugo: Sounds good

runtime/wasm/effect.wat
```diff
@@ -303,7 +306,13 @@
          (ref.func $do_perform)
          (struct.new $pair (local.get $eff) (local.get $cont))))
 
+   (global $effect_allowed (mut i32) (i32.const 1))
+
    (func (export "%perform") (param $eff (ref eq)) (result (ref eq))
+      (if (i32.eqz (global.get $effect_allowed))
```
hhugo, line 312: I dont see any equivalent logic on the js side. I seems that caml_assume_no_perform completely forbid effect with wasm while effects could still happen with js after resuming a stack ? I must admit my understanding is poor in the area.

runtime/js/effect.js
OlivierNicole: This value is the same as in `caml_callback`. I guess it has been chosen because empirically, it avoids trampolining too often but still avoids overflow. I added a comment saying this.

runtime/wasm/effect.wat
OlivierNicole: In both JS and Wasm, performing under `caml_assume_no_perform` must raise `Effect.Unhandled`, expect after installing a handler, i.e. resuming a stack. On the js_of_ocaml side, it is ensured by having caml_assume_no_perform be a synonym of `caml_callback`, which switches to a blank fiber stack. On the wasm_of_ocaml side, it’s not possible to do it for a reason that @vouillon explained me but that I forgot. An alternative is to have this global boolean to track whether performing an effect should raise.

hhugo: So we have different behavior between wasm and js ?

OlivierNicole: No, the behavior is the same, as checked by the tests in `tests-ocaml/lib-effects/`.

OlivierNicole:
@rickyvetter If you want to run benchmarks on this branch and check that nothing behaves unexpectedly, I expect this to be close to the final version.

lib/js_of_ocaml/dune
```diff
@@ -1,7 +1,8 @@
 (library
  (name js_of_ocaml)
  (public_name js_of_ocaml)
- (libraries js_of_ocaml-compiler.runtime)
+ (libraries
+  (re_export js_of_ocaml-compiler.runtime))
```
hhugo, line 5: Why this diff ?

OlivierNicole: This is to export the module `Jsoo_runtime` from the js_of_ocaml library, as you suggested https://github.com/ocsigen/js_of_ocaml/pull/1461#discussion_r1882707548

hhugo: I see, I had something different in mind. I think we could introduce a module `Effect_js` in js_of_ocaml the lib (similar to `Sys_js`) that would expose the primitive from jsoo_runtime.

compiler/tests-ocaml/lib-effects/dune
```diff
@@ -47,3 +54,18 @@
     (run node %{test}))
    (run cat)))
  (modes js wasm))
+
+(tests
```
hhugo, line 58: I  try to keep compiler/tests-ocaml for tests coming from the ocaml repo. Could you move it to `compiler/tests-jsoo/lib-effecfs/` ?

(I need to add a readme here to explain the layout.)

Issue #1399, opened 2023-02-03
Code size optimizations


vouillon:
Some ideas to reduce the size of the generated code...

Turn conditional statements into expressions
-
More compact code, and may allow further optimizations
```
if(e1)e2 else e3
===>
e1?e2:e3
```
```
if(e1)e2
==>
e1&&e2
```
```
if(!e1)e2
==>
e1||e2
```
Eliminate expression statements
-
Then, we may be able to coalesce more variable declarations, or omit the return of an arrow function.
```
e1; var x = e2
===> var x = e1, e2
```
```
e1; return e2
==> return e1, e2
```
```
e1; if (e2) ...
==>
if (e1, e2) ...
```
```
e1; e2
==>
e1, e2
```
Improve handling of functions
-
We are currently generating assignments for variables defined before a function instead of inlining them, just in case they are bound in the function. Typically, we have
```
var x = 0; f(x, (y)=>e)
```
where we could have
```
f(0, (y)=>e)
```

hhugo:
> Turn conditional statements into expressions

We already do some of this in js_simpl.ml

OlivierNicole:
> Then, we may be able to coalesce more variable declarations, or omit the return of an arrow function.

Is that something that will happen magically when #1467 is merged, or will it required further work?

hhugo:
> > Then, we may be able to coalesce more variable declarations, or omit the return of an arrow function.
> 
> Is that something that will happen magically when #1467 is merged, or will it required further work?

Possibly automatic with `--enable es6`.
See https://github.com/ocsigen/js_of_ocaml/pull/1398

hhugo:
> Improve handling of functions

This is fixed by #1568

OlivierNicole:
To summarize, @micahcantor's implementation of the first two ideas has shown no benefits, and the third idea has been implemented. Should we close then?

PR #1467, opened 2023-06-06
#1399 Code Size Optimizations (WIP)

micahcantor:
#1399


OlivierNicole:
Thanks for this. Could we allow the CI to run?

compiler/tests-compiler/effects_continuations.ml
```diff
@@ -120,21 +120,22 @@ let%expect_test "test-compiler/lib-effects/test1.ml" =
...
+       caml_string_equal(s, cst)){
       var _z_ = Stdlib[8], raise = caml_pop_trap();
       return raise(caml_maybe_attach_backtrace(_z_, 1));
      }
      var _A_ = Stdlib[79];
      return caml_cps_call2
              (_A_,
               cst_toto,
-              function(_D_){caml_pop_trap(); return cont([0, [0, _D_, n, m]]);});
+              function(_D_){return caml_pop_trap(), cont([0, [0, _D_, n, m]]);});
```
hhugo, line 138: note that we don't save any space here.

compiler/tests-compiler/minify.ml
```diff
@@ -97,7 +97,7 @@ console.log("xx =", xx);
       3: f(){a=2;try{throw 1}catch(a){var
       4: a=3}return a}function
       5: g(){var
-      6: b=2;return b}console.log("xx =",xx);console.log("f() =",f());console.log("xx =",xx);console.log("g() =",g());console.log("xx =",xx);
+      6: b=2;return b}console.log("xx =",xx),console.log("f() =",f()),console.log("xx =",xx),console.log("g() =",g()),console.log("xx =",xx);
```
hhugo, line 100: not space save here.

compiler/tests-compiler/effects_continuations.ml
OlivierNicole: True, but Jérôme’s point in the issue was:

> Then, we may be able to coalesce more variable declarations, or omit the return of an arrow function.

So there can be a benefit. What’s not clear to me is whether these further compactions are already happening or whether they require further changes.

hhugo:
compiler/tests-full/stdlib.cma.js doesn't need to be committed

compiler/tests-full/stdlib.cma.expected.js
micahcantor, line 1: Can this be removed as well?

hhugo: no, this file is committed to track  the evolution of the code generation of the stdlib.

compiler/tests-compiler/space_optimization.ml
```diff
@@ -0,0 +1,130 @@
...
+let%expect_test "\"use strict\"; expression statement" =
+  with_temp_dir ~f:(fun () ->
+      let js_prog = "\"use strict\"; e2;" in
+      let js_file =
+        js_prog |> Filetype.js_text_of_string |> Filetype.write_js ~name:"test.js"
+      in
+      print_endline (Util.optimize_space js_file);
+      [%expect {|
+        "use strict"; e2;
+        //end|}])
```
OlivierNicole, line 130: Maybe add a check that `if (e1) e2; if (e3) e4 else e5; return e6` becomes
`return e1 && e2, e3 ? e4 : e5, e6`
and similarly for `var` instead of return?
That is, checking that more than one statements can be pushed into the one following.

compiler/tests-compiler/static_eval.ml
```diff
@@ -136,24 +137,4 @@ let%expect_test "static eval of string get" =
...
-       key$0 = param$0[1],
-       data$0 = param$0[2],
-       next$0 = param$0[3],
-       prec$0 = [0, key$0, data$0, next$0];
-      prec[3] = prec$0;
-      var prec = prec$0, param$0 = next$0;
-     }
-    }
-    //end |}]
+    not found |}]
```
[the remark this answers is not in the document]
OlivierNicole: `print_fun_decl` looks for a given function declaration in the program. Looks like `copy_bucketlist` was not found.

OlivierNicole (commented):
I suggested a few minor changes (and we need to address the weird `not found` in `static_eval.ml`), but otherwise this looks good to me.

compiler/tests-compiler/call_gen.ml
```diff
@@ -78,10 +78,7 @@ module M1 = struct
               });
     }
     //end
-    function k(a, b, c, d){
-     return caml_call1(Stdlib[44], ((a + b | 0) + c | 0) + d | 0);
-    }
-    //end
+    not found
```
micahcantor, line 81: Another not found to resolve here

compiler/lib/js_traverse.ml
```diff
@@ -1406,29 +1401,83 @@ class simpl =
...
+            (e1, (Return_statement (Some e2), _), Some (Return_statement (Some e3), _)) ->
+            [ Return_statement (Some (ECond (e1, e2, e3))), loc ]
+        (* if (e1) v1 = e2 else v1 = e3 --> v1 = e1 ? e2 : e3 *)
+        | If_statement
+            ( cond
+            , (Expression_statement (EBin (Eq, v1, e1)), _)
+            , Some (Expression_statement (EBin (Eq, v2, e2)), _) )
+          when Poly.(v1 = v2) ->
+            [ Expression_statement (EBin (Eq, v1, ECond (cond, e1, e2))), loc ]
+        (* if (e1) e2 else e3 --> if e1 ? e2 : e3 *)
```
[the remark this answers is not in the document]
micahcantor: [978eae9](https://github.com/ocsigen/js_of_ocaml/pull/1467/commits/978eae96983db9852f190d399539bd248ce3d09b)

compiler/tests-compiler/space_optimization.ml
micahcantor: [398a1b9](https://github.com/ocsigen/js_of_ocaml/pull/1467/commits/398a1b9fbc16ac66b00a8833f6e40a082f089640)

compiler/tests-compiler/static_eval.ml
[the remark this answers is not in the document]
micahcantor: It looks like these changes are causing the function declaration in this case to be adjoined to the previous variable declarations. The result is that it evades the `find_function_declaration` method used by the `print_func_decl` function. Here's the (abridged) before and after of the generated program:

 Before:

```js
(function(globalThis){
    "use strict";
    var
      runtime = globalThis.jsoo_runtime,
      global_data = runtime.caml_get_global_data();
      global_data.Assert_failure;
    function copy_bucketlist(param){
      ...
      }
      var Test = [0, copy_bucketlist];
      runtime.caml_register_global(2, Test, "Test");
      return;
    }
  (globalThis));
//end
```

After:

```js
(function(globalThis){
    "use strict";
    var
      runtime = globalThis.jsoo_runtime,
      copy_bucketlist =
        (global_data.Assert_failure,
          function(param){
           ...
          }),
      Test = [0, copy_bucketlist];
      runtime.caml_register_global(2, Test, "Test");
      return;
    }
  (globalThis));
//end
```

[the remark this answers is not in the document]
vouillon: I guess you can just check for `ESeq (_, EFun (_, fun_decl))` as well here:
https://github.com/ocsigen/js_of_ocaml/blob/df91fb798de5f0bd4b56857b8fa59847e79682d8/compiler/tests-compiler/util/util.ml#L468

compiler/tests-compiler/effects_continuations.ml
vouillon: With `--enable es6`, this really makes a difference, indeed.

compiler/lib/js_traverse.ml
```diff
@@ -1406,29 +1401,83 @@ class simpl =
...
+                (* e1; e2; --> e1, e2; *)
+                | Expression_statement e2 ->
+                    (Expression_statement (ESeq (e1, e2)), loc) :: rem
+                | _ -> (st, loc) :: (st2, loc2) :: rem))
+        | _ -> (st, loc) :: rem
+      in
+      (* First simplify single statements *)
+      List.concat_map ~f:simplify_statement s
+      (* Then try to combine expression statements *)
+      |> List.fold_right ~f:eliminate_expression_statements ~init:[]
```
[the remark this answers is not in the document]
hhugo: Note that it could be nice to be able keep optimisations while having pretty printing when debugging.
Look at what's done for `shortvar` and `share` in `lib-cmdline/arg.ml`

micahcantor:
When we began benchmarking these changes, we ran into a surprising result. We used `ocamlc` as a readily available large OCaml project to test. With the changes, the generated JS file for `ocamlc` was about 4.5kb smaller, or an improvement of about 0.25%. But after compression with `bzip2`, the result artifact was about 2kb (or 0.47%) bigger than it would be without the changes! So when combined with compression, these changes may backfire and produce larger code.

We aren't exactly sure the reason for this, but it's possible that some transformations are adding parentheses that worsen compression, or they are interfering with other optimizations, such as combining consecutive variable declarations, that compress better. I'll do some more testing to see if any improvement can be salvaged, but otherwise we may conclude the changes are ineffective (unless there is a use-case for which uncompressed file size is important).

PR #1472, opened 2023-06-16
Support conversion of Uint8ClampedArray typed arrays to bigarrays.

dbuenzli:
These arrays are used for example by [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData) objects.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 ## Features/Changes
 * Misc: Bump magic number for ocaml 5.1
 * Misc: changes to stay compatible with the next version of ppx_expect
+* Runtime: support conversion of Uint8ClampedArray typed arrays to bigarrays.
+
 
 ## Bug fixes
 * Compiler: fix location for parsing errors when last token is a virtual semicolon
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,7 @@
 ## Features/Changes
 * Misc: Bump magic number for ocaml 5.1
 * Misc: changes to stay compatible with the next version of ppx_expect
-* Runtime: support conversion of Uint8ClampedArray typed arrays to bigarrays.
-
+* Runtime: support conversion of Uint8ClampedArray typed arrays to bigarrays (#1472)
 
 ## Bug fixes
 * Compiler: fix location for parsing errors when last token is a virtual semicolon
@@ -12,6 +11,7 @@
 * Compiler: consise body should allow any expression but object literals
 * Compiler: preserve [new] without arguments [new C] (vs [new C()]
 * Compiler: remove invalid rewriting of js (#1471, #1469)
+* Runtime: fix int32 values returned from bigarrays when wrapping Uint32Array objects (#1472)
 
 # 5.2.0 (2023-04-28) - Lille
 ## Features/Changes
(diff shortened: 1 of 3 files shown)


dbuenzli:
Did you mean something else ? On `caml_ba_from_typed_array` there should be no typed array creation.

dbuenzli:
Ah I think I got what you said. I'm a bit unsure though. If these clamped arrays are supposed to be written clamped then what happens you if you don't clamp them in Js ?

dbuenzli:
I think the answer is no. This is not about respecting the OCaml bigarray semantics it's about respecting the JavaScript semantics of the typed array that was given to you.

hhugo:
@vouillon, I didn't find any doc regarding the semantic of Bigarray.set in presence of overflow.

I think there is currently an issue with caml_ba_from_typed_array of Uint32Array because Bigarray.get doesn't normalize the returned  value.

```ocaml
let a = new%js Typed_array.uint32Array 10;;
val a : Js_of_ocaml.Typed_array.uint32Array Js_of_ocaml.Js.t = <abstr>
let b = Typed_array.to_genarray a;;
val b : (int32, Bigarray.int32_elt, Bigarray.c_layout) Bigarray.Genarray.t =
  <abstr>
Typed_array.set a 0 (-1l);;
- : unit = ()
Bigarray.Genarray.get b [|0|];;
- : int32 = 4294967295l
 ```

**Comment by hhugo** [comment]
My understanding is that turning Uint32Array into a Int32Array in `caml_ba_from_typed_array` would be fine and would fix the issue mentioned above.
I'm not sure that changing  `Uint8ClampedArray` into `Uint8Array` would give us much (clamp vs wrap around when setting values in the array).

**Comment by vouillon** [comment]
Since the semantics of `Bigarray.set` is not specified, this PR looks good indeed.

Regarding Uint32Array, we have an issue indeed. Another possible fix would be to modify `Ml_Bigarray.prototype.get`, but it seems better to make the change in `caml_ba_from_typed_array`.

**Comment by dbuenzli** [comment]
Ok I'll add what you suggested above for UInt32

**Comment by dbuenzli** [comment]
I tried to do that in [6304e1b](https://github.com/ocsigen/js_of_ocaml/pull/1472/commits/6304e1b2394f730664c31eaf452e40ce190583fb) but it doesn't seem to work.

**Comment by dbuenzli** [comment]
> ```js
>   new Int32Array(ta.buffer,ta.byteOffset,ta.byteLength):ta,
> ```

Fixed it in [ad6d9db](https://github.com/ocsigen/js_of_ocaml/commit/ad6d9db3085c43a05bed7fa2426a9ede67a43585) this had to be `ta.length` not `ta.byteLength`. 

The following code succeeds:
```
let ta = Brr.Tarray.create Int32 10
let ba = Brr.Tarray.to_bigarray1 ta
let () = Brr.Tarray.set ta 0 (-1l)
let () =
  Brr.Console.(log [Bigarray.Array1.get ba 0]);
  assert (Bigarray.Array1.get ba 0 = -1l)
```

hhugo:
I've pushed two small commit.
Will merge if CI is happy.

hhugo:
Thanks

PR #1475, opened 2023-06-21
Small runtime fixes

vouillon:


runtime/mlBytes.js
```diff
@@ -225,7 +225,7 @@ function caml_string_get (s, i) {
 //Requires: caml_string_unsafe_get, caml_string_bound_error
 //Requires: caml_ml_string_length
 function caml_string_get16(s,i) {
-  if (i >>> 0 >= caml_ml_string_length(s) - 1) caml_string_bound_error();
+  if (i < 0 || i + 1 >= caml_ml_string_length(s)) caml_string_bound_error();
```
hhugo, line 228: isn't is equivalent ?

runtime/gc.js
```diff
@@ -38,7 +38,7 @@ function caml_gc_set(_control) {
 
 //Provides: caml_gc_get
 function caml_gc_get(){
-  return [0,0,0,0,0,0,0,0,0]
+  return [0,0,0,0,0,0,0,0,0,0,0,0]
```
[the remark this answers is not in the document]
hhugo: Maybe we should add expect test for this. 
```
print_int (Obj.size (Obj.repr (Gc.get())));
```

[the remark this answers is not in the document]
hhugo: same thing for gc_stat and quick_stat

runtime/sys.js
```diff
@@ -230,18 +230,9 @@ function caml_sys_time_include_children(b) {
...
-      return [0,a[0]];
-    } else if(globalThis.crypto.randomBytes === 'function'){
-      // Nodejs
-      var buff = globalThis.crypto.randomBytes(4);
-      var a = new Uint32Array(buff);
-      return [0,a[0]];
-    }
+  if (globalThis.crypto && globalThis.crypto.getRandomValues) {
+     var a = globalThis.crypto.getRandomValues(new Int32Array(4));
+     return [0, a[0], a[1], a[2], a[3]];
```
hhugo, line 235: > I think using function globalThis.crypto.randomBytes did not work since that
should have been: require('crypto').randomBytes.

The randomBytes codepath was working fine with node14

```
$ node
Welcome to Node.js v14.18.3.
Type ".help" for more information.
> crypto.randomBytes(4)
<Buffer 5e d7 a2 45>
```

runtime/mlBytes.js
vouillon: Oh, indeed!

runtime/gc.js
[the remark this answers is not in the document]
vouillon: I'm not sure we really care if there are too many fields...
I have added some tests.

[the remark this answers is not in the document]
hhugo: I would at least mention this inside a comment next to the implem.

compiler/tests-ocaml/lib-gc/test_gc.ml
```diff
@@ -0,0 +1,4 @@
+let () =
+  assert ((Gc.get ()).custom_minor_max_size = 0);
+  assert ((Gc.stat ()).forced_major_collections = 0);
+  assert ((Gc.quick_stat ()).forced_major_collections = 0)
```
hhugo, line 4: I would be nice to have a test that will fail next time one add another field on the ocaml side.

hhugo: this file is missing a header.

PR #1479, opened 2023-06-27
Upgrade to ocamlformat 0.26.0

Julow:
The classes are formatted completely differently, which might be problematic.

The aim of this commit is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,4 @@
 # ocamlformat bump
 3c98cd3e0c4e1191983b602a8e38a1c92d259891
+# Upgrade to OCamlformat 0.26.0
+5586996a5e8f11981ec5337916e24e0449c13a97
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -856,35 +856,35 @@ struct
     | CoverInitializedName (e, _, _) -> early_error e
 
   and method_ : 'a. _ -> (PP.t -> 'a -> unit) -> 'a -> method_ -> unit =
-    fun (type a) f (name : PP.t -> a -> unit) (n : a) (m : method_) ->
-     match m with
-     | MethodGet (k, l, b, loc') | MethodSet (k, l, b, loc') ->
-         (match k with
-         | { async = false; generator = false } -> ()
-         | _ -> assert false);
-         let prefix =
-           match m with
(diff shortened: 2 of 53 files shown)


compiler/lib/js_output.ml
```diff
@@ -856,35 +856,35 @@ struct
...
-               PP.string f "async";
-               PP.space f
-           | { async = true; generator = true } ->
-               PP.string f "async*";
-               PP.space f);
-           name f n
-         in
-         function_declaration f "" fpn (Some ()) l b loc'
+   fun (type a) f (name : PP.t -> a -> unit) (n : a) (m : method_) ->
+    match m with
```
hhugo, line 860: Is the 1 char indent on purpose ?

hhugo (commented):
After a very very quick look, it looks mostly fine.
The size of the diff regarding indentation in classes makes the whole diff complicated the evaluate. Anyway to split the diff in two commits ? putting the indentation of classes on iits own ?

compiler/lib/js_output.ml
Julow: Yes. An indentation of 1 here was the default but due to a bug, the indentation was 2 when there were a type annotation. This is a consistency change.

Julow:
I splitted the changes to classes into a new commit. This commit is best viewed with `git diff -w`.

hhugo:
Thanks a lot. LGTM

Julow:
Thanks for the feedback. Let's not merge this until OCamlformat is released.

Julow:
OCamlformat 0.26.0 has been released! This PR can be merged (not squashed due to `.git-blame-ignore-revs`) or closed if it's a bit too early to upgrade.

.git-blame-ignore-revs
```diff
@@ -1,2 +1,3 @@
 # ocamlformat bump
-3c98cd3e0c4e1191983b602a8e38a1c92d259891
\ No newline at end of file
+3c98cd3e0c4e1191983b602a8e38a1c92d259891# Upgrade to OCamlformat 0.26.0
```
hhugo, line 2: This comment seems to be on the wrong line

Julow: Fixed.

PR #1483, opened 2023-07-04
Minor fixes for OCaml trunk

hhugo:


compiler/tests-ocaml/lib-format/dune
```diff
@@ -2,7 +2,7 @@
  (names pr6824 print_if_newline pp_print_custom_break print_seq tformat)
  (libraries)
  (flags
-  (:standard \ -strict-formats))
+  (:standard -no-strict-formats \ -strict-formats))
```
OlivierNicole, line 5: I was looking at this file and I’m wondering: what is the `\` for here?

OlivierNicole: And also, what is the semantics of setting these two opposite flags?

hhugo:
``` 
(<sets1> \ <sets2>) is the set composed of elements of <sets1> that do not appear in <sets2>
```

The line you're looking at aims at removing `-strict-formats` (if present) from `:standard` and enable `-no-sttrict-format`.
The reason is that I wanted these flags to be consistent regardless of the OCaml version

OlivierNicole: I understand now. Thank you!

PR #1492, opened 2023-07-25
Improve compatibility with Wasm_of_ocaml

vouillon:
[Wasm_of_ocaml](https://github.com/ocaml-wasm/wasm_of_ocaml) needs additional bindings:
- float conversions are no longer deprecated
- add conversion between OCaml int32 and nativeint and JavaScript numbers
- add binding for JavaScript strict equality (and better expose the regular equality)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Runtime: fix Dom_html.onIE (#1493)
 * Compiler: fix global flow analysis (#1494)
+* Runtime: add conversion functions + strict equality for compatibility with Wasm_of_ocaml (#1492)
 
 # 5.4.0 (2023-07-06) - Lille
 
--- a/lib/js_of_ocaml/js.mli
+++ b/lib/js_of_ocaml/js.mli
@@ -217,15 +217,35 @@ val nfkd : normalization t
 val nfkc : normalization t
 (** Compatibility Decomposition, followed by Canonical Composition *)
 
+(** Specification of Javascript number objects. *)
+
+type number_t = float
+
+class type number = object
+  method toString : js_string t meth
+
+  method toString_radix : int -> js_string t meth
+
+  method toLocaleString : js_string t meth
+
+  method toFixed : int -> js_string t meth
+
+  method toExponential : js_string t meth
+
+  method toExponential_digits : int -> js_string t meth
+
+  method toPrecision : int -> js_string t meth
+end
+
 (** Specification of Javascript string objects. *)
-class type js_string = object
+and js_string = object
   method toString : js_string t meth
 
   method valueOf : js_string t meth
 
   method charAt : int -> js_string t meth
 
-  method charCodeAt : int -> float meth
+  method charCodeAt : int -> number_t meth
 
   (* This may return NaN... *)
   method concat : js_string t -> js_string t meth
@@ -245,7 +265,7 @@ class type js_string = object
 
   method lastIndexOf_from : js_string t -> int -> int meth
 
-  method localeCompare : js_string t -> float meth
+  method localeCompare : js_string t -> number_t meth
 
   method _match : regExp t -> match_result_handle t opt meth
 
@@ -358,7 +378,7 @@ class type ['a] js_array = object
 
   method slice_end : int -> 'a js_array t meth
 
-  method sort : ('a -> 'a -> float) callback -> 'a js_array t meth
+  method sort : ('a -> 'a -> number_t) callback -> 'a js_array t meth
 
   method sort_asStrings : 'a js_array t meth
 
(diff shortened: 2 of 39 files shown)


lib/js_of_ocaml/js.mli
```diff
@@ -813,6 +827,24 @@ external to_bytestring : js_string t -> string = "caml_string_of_jsbytes"
       Javascript string should only contain UTF-16 code points below
       255.) *)
 
+external float : float -> float = "caml_js_from_float"
```
hhugo, line 830: The signature looks wrong. Why don't we have `float -> number t` ?

vouillon: This is so that our users can write code that works with this version of the library, where we are using `float` throughout for JavaScript numbers, and a version of the library where the type of JavaScript numbers is `number t`.

By the way, I'm using `float t` for now, since I was concerned of a possible performance impact of using an object type. I need to check...

vouillon:
> What's the plan for wasm_of_ocaml ? Is the goal to eventually merge it with jsoo? Will it stay as a fork of jsoo? There are other projects providing other backends to jsoo, https://github.com/jordwalke/rehp, should we think of a way to share common logic, without merging everything into this repo.

I think it would be simpler to merge it with Jsoo. Though, since jsoo already exposes the compiler as a library, and only `driver.ml` depends on Wasm-specific files, it should not be too hard to keep thinks separate.

CHANGES.md
```diff
@@ -6,6 +6,7 @@
 * Runtime: fix Gc.stat, Gc.quick_stat, Gc.get (#1475)
 * Compiler: fix some miscompilation, probably introduced in jsoo 5.0.0,
   revealed by OCaml 5.0 ?
+* Runtime: add coercion functions + strict equality for compability with Wasm_of_ocaml (#1492)
```
hhugo, line 9: needs to move to a `#dev` section

lib/js_of_ocaml/js.mli
hhugo: I'd be easier to read if we had an alias.

Also, if I understand correctly, this is just a temporary solution if we end up merging jsoo and wasm_of_ocaml.
We'll eventually need this library to no use float directly. Is that correct ?

lib/js_of_ocaml/js.mli
```diff
@@ -813,6 +827,24 @@ external to_bytestring : js_string t -> string = "caml_string_of_jsbytes"
...
+(** Conversion of OCaml floats to Javascript numbers. *)
+
+external to_int32 : float -> int32 = "caml_js_to_int32"
+(** Conversion of Javascript numbers to OCaml 32-bits. *)
+
+external nativeint : nativeint -> float = "caml_js_from_nativeint"
+(** Conversion of OCaml 32-bits integers to Javascript numbers. *)
+
+external to_nativeint : float -> nativeint = "caml_js_to_nativeint"
+(** Conversion of Javascript numbers to OCaml native integers. *)
```
hhugo, line 846: same here

lib/js_of_ocaml/js.mli
vouillon: Yes, this is only temporary, to make the switch smoother. 

I can add a `number_t` type (or do you have a better name?) and use it throughout the Js_of_ocaml library, if you like.

hhugo: > I can add a number_t type (or do you have a better name?) and use it throughout the Js_of_ocaml library, if you like.

Sounds good, it makes the code a bit clearer and should also help with the transition.

hhugo: It seems that examples have not been updated in the PRs.
Can you also check the manual ? Or do you prefer to leave it for later ?

vouillon: The main objective of this PR is to undeprecate the float conversion functions and to add missing primitives.
I need to check the manual at some point, indeed. But maybe that can wait.

hhugo: Ok. The diff looks good. I'll let you rebase and merge

hhugo:
I've been trying to make a new release quickly but I forgot about this change. It breaks many packages. I wonder if we should make a major release and/or help other packages with the transition. @vouillon, any opinion ? would you be able to help with this release ?

vouillon:
@hhugo , the following line was meant to be `type number_t = float`. I used an opaque type at some point to see were I needed to make changes but forgot to revert it afterwards.
https://github.com/ocsigen/js_of_ocaml/blob/3e27aad041b2f1b59e37d6c8cb79a66d5a59d489/lib/js_of_ocaml/js.mli#L222

PR #1494, opened 2023-07-26
Fix global flow analysis

vouillon:
In fast mode, we don't track function arguments, so we should consider them as escaping.


hhugo:
Would you be able to come up with a regression test for this fix ? Do you have a source that exhibit an invalid optimization?

vouillon:
I have added a regression test.

It is currently difficult to get incorrect code because of this bug since we are using the analysis only to optimize function calls. But @micahcantor is working on an improved dead code elimination that uses this analysis and he got wrong results because of this bug.

PR #1496, opened 2023-07-27
Compiler: refactor control flow compilation

### LINKED ISSUE #1336: Compiler: rewrite control flow logic in generate.ml
Maybe we should consider rewriting this code based on this nice paper from last ICFP: [Beyond Relooper: recursive translation of unstructured control flow to structured control flow, Norman Ramsey](https://dl.acm.org/doi/abs/10.1145/3547621)

hhugo:
Instead of using additional variables, use labelled and break statements.
Fix #1336 

This PR includes a copy of wa_structture.ml from wasm_of_ocaml.

~~This PR just a proof of concept for now.~~
cc @vouillon

### DIFF
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -63,6 +63,10 @@ let list_group f g l =
 
 (****)
 
+type fall_through =
+  | Block of Addr.t
+  | Return
+
 type application_description =
   { arity : int
   ; exact : bool
@@ -302,6 +306,12 @@ module Ctx = struct
     ; cps_calls
     }
 end
+
+type edge_kind =
+  | Loop
+  | Exit_loop of bool ref
+  | Exit_switch of bool ref
+  | Forward
 
 let var x = J.EVar (J.V x)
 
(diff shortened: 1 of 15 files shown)


compiler/lib/generate.ml
```diff
@@ -1461,6 +1461,8 @@ and compile_block_no_loop st queue (pc : Addr.t) ~fall_through scope_stack =
     Structure.get_edges st.dom pc
     |> Addr.Set.elements
     |> List.filter ~f:(fun pc' -> is_switch pc' || Structure.is_merge_node st.graph pc')
+    |> List.sort ~cmp:(fun a b ->
+           compare (Structure.block_order st.graph a) (Structure.block_order st.graph b))
```
hhugo, line 1465: @vouillon, I don't understand why you don't have similar sorting in wasm_of_ocaml.

vouillon: I would export from `Structure` a comparison function, or a sorting function, rather than a `block_order` function.

compiler/lib/structure.mli
```diff
@@ -0,0 +1,32 @@
...
+val dominator_tree : control_flow_graph -> (int, int) Hashtbl.t
+
+val dominates : control_flow_graph -> (int, int) Hashtbl.t -> int -> int -> bool
+
+val is_merge_node : control_flow_graph -> int -> bool
+
+val is_loop_header : control_flow_graph -> int -> bool
+
+val dominance_frontier :
+  control_flow_graph -> (int, int) Hashtbl.t -> (int, Code.Addr.Set.t) Hashtbl.t
```
vouillon, line 32: By reversing the dominator tree from within this module (`let dominator_tree g = reverse_tree (reversed_dominator_tree g)`), I ended up with this interface:
```ocaml
type graph

val get_edges : graph -> Code.Addr.t -> Code.Addr.Set.t

type control_flow_graph

val build_graph : Code.block Code.Addr.Map.t -> Code.Addr.t -> control_flow_graph

val dominator_tree : control_flow_graph -> graph

val is_loop_header : control_flow_graph -> Code.Addr.t -> bool

val is_merge_node : control_flow_graph -> Code.Addr.t -> bool

val is_backward : control_flow_graph -> Code.Addr.t -> Code.Addr.t -> bool

val sort_in_post_order : control_flow_graph -> Code.Addr.t list -> Code.Addr.t list
```

compiler/lib/generate.ml
hhugo: done

compiler/lib/structure.mli
hhugo: did the same

hhugo:
@pmwhite, this PR changes how we translate control flow to javascript. It can affect compilation time, size of generated files, and runtime performance. Let me know if you have time to test this and report positive and negative impact.

hhugo:
@vouillon, I think the diff is ready for review, commit history is not and will be squashed.

compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -586,10 +585,9 @@
             (oc);
     /*<<stdlib.ml:390:49>>*/ }
    function close_out_noerr(oc){
-     /*<<stdlib.ml:392:2>>*/ try{
-      /*<<stdlib.ml:392:7>>*/  /*<<stdlib.ml:392:7>>*/ caml_ml_flush(oc);
-    }
-    catch(_u_){}
+     /*<<stdlib.ml:392:2>>*/ b:
+    try{ /*<<stdlib.ml:392:7>>*/  /*<<stdlib.ml:392:7>>*/ caml_ml_flush(oc);}
+    catch(_u_){break b;}
```
vouillon, line 590: This looks unnecessary.

compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -5008,15 +5031,16 @@
...
     if(_P_ >= 0){
      var i = _P_;
      for(;;){
       r[1] =
         /*<<bytes.ml:256:9>>*/ caml_call2
         (f, caml_bytes_unsafe_get(a, i), r[1]);
        /*<<bytes.ml:255:2>>*/  /*<<bytes.ml:255:2>>*/ var _Q_ = i - 1 | 0;
-      if(0 !== i){var i = _Q_; continue;}
-      break;
+      if(0 === i) break b;
```
vouillon, line 5042: Here, `break` should be enough.

pmwhite:
@hhugo I should have time to test this out this week, hopefully tomorrow.

compiler/tests-full/stdlib.cma.expected.js
vouillon: Yes, probably. I don't understand why we don't have the same issue with the current implementation.

compiler/tests-full/stdlib.cma.expected.js
hhugo: I've fixed this case by introducing two "scopes"  when compiling the body of a for loop.
```ocaml
          let scope_stack = (pc, (lab, lab_used, Loop)) :: scope_stack in
          let scope_stack =
            match fall_through with
            | Block fall_through ->
                (fall_through, (lab, lab_used, Exit_loop exit_branch_used)) :: scope_stack
            | Return -> scope_stack
          in
```

compiler/tests-full/stdlib.cma.expected.js
hhugo: In the current implementation, we have the following piece of code
```ocaml
    (* Create an artificial frontier when we pop an exception handler *)
    let rec keep_front pc =
      if Hashtbl.find preds pc > 1
      then (* already part of a merge node *) false
      else
        match Addr.Map.find pc blocks with
        | { body = []; branch = Return _, _; _ } -> false
        | { body = []; branch = Stop, _; _ } -> false
        | { body = []; branch = Branch (pc', _), _; _ } -> keep_front pc'
        | _ -> true
    in
```
Which drops the poptrap frontier if there is a sequence of empty blocks ending in a merge node.

vouillon: So, I suppose we can do the same?

compiler/tests-full/stdlib.cma.expected.js
vouillon: Yes, this makes a lot of sense.

compiler/tests-full/stdlib.cma.expected.js
hhugo: Yes, I've pushed a change

hhugo: Can this be used in wasm_of_ocaml ?

hhugo:
@vouillon, I've cleaned up the commit history.

hhugo:
@pmwhite, any news about testing this ?
@vouillon, do you want to do more review on this ?

vouillon:
This looks good to me.

pmwhite:
Sorry, I have not gotten to testing this yet. I will let you know the results when I do.

PR #1498, opened 2023-07-29
Compiler: lower level switch

### LINKED ISSUE #1495: Compiler: Lower level Code.Switch
`Code.Switch` currently deal with both immediate and block/tagged values.
I feel we could simplify things a bit by adding an explicit condition on `is_int` and have `Switch` just dispatch on a variable immediate value

In other words, I means turning 
```
B1: Switch(x,A1,A2)
```

Into
```
B1: Cond(isInt x, B2, B3)
B2: Switch(x, A1)
B3: let x' = tag x; Switch (x', A2)
```

hhugo:
fix #1495

### DIFF
--- a/compiler/lib/eval.ml
+++ b/compiler/lib/eval.ml
@@ -189,6 +189,34 @@ let is_int info x =
   | Pc (Int _) -> Y
   | Pc _ -> N
 
+type case_of =
+  | CConst of int
+  | CTag of int
+  | Unknown
+
+let the_case_of info x =
+  match x with
+  | Pv x ->
+      get_approx
+        info
+        (fun x ->
+          match info.info_defs.(Var.idx x) with
+          | Expr (Constant (Int i)) -> CConst (Int32.to_int i)
+          | Expr (Block (j, _, _)) ->
+              if Var.ISet.mem info.info_possibly_mutable x then Unknown else CTag j
+          | Expr (Constant (Tuple (j, _, _))) -> CTag j
+          | _ -> Unknown)
+        Unknown
+        (fun u v ->
+          match u, v with
+          | CTag i, CTag j when i = j -> u
+          | CConst i, CConst j when i = j -> u
+          | _ -> Unknown)
+        x
+  | Pc (Int i) -> CConst (Int32.to_int i)
+  | Pc (Tuple (j, _, _)) -> CTag j
+  | _ -> Unknown
+
 let eval_instr info ((x, loc) as i) =
   match x with
   | Let (x, Prim (Extern ("caml_js_equals" | "caml_equal"), [ y; z ])) -> (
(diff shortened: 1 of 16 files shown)


compiler/lib/global_flow.ml
```diff
@@ -243,14 +243,9 @@ let program_deps st { blocks; _ } =
...
           cont_deps blocks st cont1;
           cont_deps blocks st ~ignore:x cont2
-      | Switch (x, a1, a2) ->
+      | Switch (x, a1) ->
           Array.iter a1 ~f:(fun cont -> cont_deps blocks st cont);
-          Array.iter a2 ~f:(fun cont -> cont_deps blocks st cont);
           let h = Hashtbl.create 16 in
-          Array.iteri
-            ~f:(fun i (pc, _) ->
-              Hashtbl.replace h pc (i :: (try Hashtbl.find h pc with Not_found -> [])))
```
hhugo, line 252: @vouillon, how important is the optimization about known_cases ? do we need to restore it ?

compiler/lib/eval.ml
```diff
@@ -341,14 +341,13 @@ let eval_branch info (l, loc) =
...
           | Unknown -> b)
-    | Switch (x, const, tags) as b -> (
+    | Switch (x, const) as b -> (
         (* [the_case_of info (Pv x)] might be meaningless when we're inside a dead code.
            The proper fix would be to remove the deadcode entirely.
            Meanwhile, add guards to prevent Invalid_argument("index out of bounds")
            see https://github.com/ocsigen/js_of_ocaml/issues/485 *)
         match the_case_of info (Pv x) with
         | CConst j when j >= 0 && j < Array.length const -> Branch const.(j)
-        | CTag j when j >= 0 && j < Array.length tags -> Branch tags.(j)
```
hhugo, line 351: This knowledge could be used when evaluating "%direct_obj_tag"

compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -15075,85 +15078,93 @@
...
+            switch$1 = 1; break;
            case 11:
-            switch$0 = 1; break;
+            switch$0 = 1; switch$1 = 1; break;
            case 12:
-            switch$0 = 2; break;
+            switch$0 = 2; switch$1 = 1; break;
            case 13:
-            switch$0 = 3; break;
+            switch$0 = 3; switch$1 = 1; break;
```
hhugo, line 15153: The diff in this file may look like a regression (in term of code size) but it looks much better once rebased on top of #1496

compiler/lib/eval.ml
hhugo: done

compiler/lib/global_flow.ml
hhugo: I've tried to restore the optimization but it appears that our test-suite doesn't capture/test this.
Would you be able to contribute an expect test for this ?

vouillon: I'm not sure it is really useful in practice. It makes the analysis more precise and does not cost much, but it's not clear to me whether it make much of a difference.

vouillon: Your change looks fine. I need to think of a test.

Issue #595, opened 2017-06-15
Improved dead code elimination


vouillon:
- cross-function analysis (though probably not through function arguments?)
- eliminate unused components of blocks

One motivation is to be able to eliminate unused code in functors (in tyxml, for instance).

PR #1503, opened 2023-08-03
Global Deadcode Elimination

micahcantor:
This PR adds a new optimization pass to perform a global and block-aware liveness analysis. The primary purpose of this is to be able to remove unused code from functors that are instantiated more than once, a limitation of the current deadcode elimination algorithm. This issue has been known for some time, see https://github.com/ocsigen/js_of_ocaml/issues/595. 

Since this change is somewhat involved, I will split this comment into a few sections to make it easier to review. The commit history is messy, but the PR should be ready to review file-by-file. I've worked closely with @vouillon and @OlivierNicole on these changes, but there are still some remaining questions to address.

## New changes

The primary new contribution is found in `deadcode_dgraph.ml`. There are four main steps to the analysis it performs:

1. Collect all variable definitions and build the variable dependency graph (the `usages` function). A variable `x` is considered used in a variable `y` if either `x` appears in the definition of `y`, or `x` is applied as a block or closure argument to parameter `y`.
2. An initial liveness analysis is performed by traversing the program AST (the `liveness` function). Here we mark a variable `x` as `Top` if it's used in an impure expression or instruction (more details are given in the doc comment). Otherwise it is marked as dead. This pass uses information from `Global_flow` to determine whether a return value is used at its callsites.
3. The initial liveness state is propagated through the dependency graph using the flow graph solver (the `solver` and `propagate` functions). For each variable `x` in the graph, its liveness is defined by a `join` of it's current liveness and the contribution of each of it's usages `y`. More detail is given in the comment for `contribution`, but here we can determine if `x` is used only in a single field `i` of a block `y`, in which case `x` depends only on that field. Then it's marked as `Live {i}`.
4. Using the completed liveness table, the dead variables and unused block fields are "zeroed" out. This is done by adding a sentinal variable to the beginning of the program defined by the JS value `undefined`. Any dead variables are replaced by the sentinal. This means the existing deadcode elimination should be able to remove these usages from the code, reducing the size. At this stage, we also truncate blocks that end with one or more sentinal values to the last non-sentinal variable.

Outside of `deadcode_dgraph.ml`, there are a few changes in the driver and a few other functions to integrate this pass into the optimization pipeline. The interface to the pass `Deadcode_dgraph.f` is called in the function `exact_calls` in `driver.ml`. This function runs after most other optimizations, which is important since global flow will break if it is ran beforehand with a different number of variables than it expects. Also, it ensures that this pass, which doesn't expose much opportunity for further optimization, is run only once before a final elimination pass.

We also added an optimization that deletes sentinal fields in arrays. Since the sentinal variable has the value `undefined`, the following JS transformation is valid:

```js
[a, sentinal, b] -> [a, ,b]
```

This helps save a few extra bytes in the generated code. There were a few changes I made to the interface in `generate.ml`
and `mlvalue.ml` to facilitate this optimization. There are a few other minor changes that expose information from global flow and expose an `undefined` primitive that it needs.

## Results and benchmarking

The overall effect of this process is that deadcode elimination is now sensitive to the liveness of individual fields within a block (rather than a block being entirely live or dead), and it can mark this liveness in the inputs and outputs of block functions. In the IR, functors are represented as functions from one block to another, where the member functions constitute the elements of these blocks. With this new pass, we can mark which member functions are live and remove the rest.

In practice, this means that if you instantiate a functor (like `Set.Make` in the stdlib) and use just a few functions from its interface (like `add`, `find`, etc) then the other 30 functions provided by the functor will be eliminated from the JS. This already occurs if the functor is instantiated only once, since in that case the compiler can specialize the block function to just a block and eliminate unused code inside. However, this could not occur if the functor is instantiated more than once.

Here's a minimal example demonstrating the effect. This program instantiates integer and string sets, and uses a few functions from the `Set` interface.

```ocaml
module Int_set = Set.Make (Int)
module String_set = Set.Make (String)

let int_set = Int_set.singleton 1 in
let string_set = String_set.empty in
let string_set = String_set.add "hello" string_set in
print_endline (string_of_int (Int_set.find 1 int_set) ^ (String_set.find "hello" string_set))
```

If we compile this to JS with and without the new pass, we get the following results:
| Name | Size | Size Difference | Size % Change|
| --- | ---- | ---- | ---- |
| sets.js | 32.6 KiB | 0.0 B | 0.0% |
| sets-gdc.js | 23.6 KiB | -9.0 KiB | -27.66% | 

We see that a large portion of this small program was taken up by the definitions of all `Set` interface functions, which can now be removed. Indeed, we expect to see the most significant changes from this optimization when the input program is small and uses large functor interfaces.

It should be noted that the size of the code removed grows in relation to the number of functors used and (inversely) to the number of functions used from those interfaces. In a small program like this one, the 5-10kb removed by this optimization for each functor can be significant, but for larger programs the percent change will be much smaller. 

For instance, another benchmark we used is the [catala_web_interpreter](https://github.com/CatalaLang/catala/blob/master/compiler/catala_web_interpreter.ml):

| Name | Size | Size Difference | Size % Change|
| --- | ---- | ---- | ---- |
| catala_web_interpreter.js | 3.9 MiB | 0.0 B | 0.0% | 
| catala_web_interpreter-gdc.js | 3.9 MiB | -8.8 KiB | -0.22% |

In this case we can remove 5.5kb, or just 0.22% of the program. 

We also observe a modest decrease in size can be seen in another benchmark on toplevel code using `lwt`. The source can be found in `/toplevel/examples/lwt_toplevel/toplevel.ml`:

| Name | Size | Size Difference | Size % Change| 
| --- | ---- | ---- | ---- |
| toplevel.js | 3.8 MiB | 0.0 B | 0.0% | 
| toplevel-gdc.js | 3.7 MiB | -28.2 KiB | -0.73% |

Here we remove 28kb of code or a little less than 1% of the total size.

One real-world example that we saw encouraging results for was using the library [ocamlgraph](https://github.com/backtracking/ocamlgraph/), which exposes a large functor interface to graph algorithms abstracted over the graph data structure. We compiled the [demo](https://github.com/backtracking/ocamlgraph/blob/master/examples/demo.ml) found in the library source, and obtained the following:

| Name | Size | Size Difference | Size % Change| 
| --- | ---- | ---- | ---- |
| ocamlgraph-demo.js | 151.1 KiB | 0.0 B | 0.0% | 
| ocamlgraph-demo-gdc.js | 137.2 KiB | -13.9 KiB | -9.2% | 

Here we're able to remove about 10kb or 9.2% of the program, which instantiates several of the functors provided by the library. 

## Conclusion

Overall, we expect this optimization to be useful for small web programs that want to use a functor interface like `Set`, `Map` or `ocamlgraph` without unnecessarily increasing the code size by 5-20kb. Larger programs may see a significant change if they internally expose many functor interfaces where they don't use many of the provided functions. 

This change may also cause a small increase in compile times for some programs. I tested this by compiling `ocamlc` and the toplevel example using the [hyperfine](https://github.com/sharkdp/hyperfine) benchmarking program, and these were the results:

```
Benchmark 1: js_of_ocaml `which ocamlc.byte` -o ocamlc.js
  Time (mean ± σ):      9.020 s ±  1.069 s    [User: 8.845 s, System: 0.149 s]
  Range (min … max):    8.319 s … 11.888 s    10 runs

Benchmark 1: js_of_ocaml --enable globaldeadcode `which ocamlc.byte` -o ocamlc.js
  Time (mean ± σ):      9.542 s ±  0.883 s    [User: 9.319 s, System: 0.169 s]
  Range (min … max):    8.872 s … 11.885 s    10 runs

Benchmark 1: js_of_ocaml ./bc/toplevel.bc -o toplevel.js
  Time (mean ± σ):     18.680 s ±  1.354 s    [User: 18.312 s, System: 0.324 s]
  Range (min … max):   17.916 s … 22.493 s    10 runs

Benchmark 1: js_of_ocaml --enable globaldeadcode ./bc/toplevel.bc -o toplevel.js
  Time (mean ± σ):     18.875 s ±  1.124 s    [User: 18.530 s, System: 0.313 s]
  Range (min … max):   18.327 s … 21.996 s    10 runs
```

So on `ocamlc` the pass adds about 0.5s and on toplevel it adds about .2s on average.

## Future Work

We found during testing that the optimization can fail to remove code from nested functors, (i.e. functors that take other functors as arguments), such as in the interfaces exposed by `tyxml`. We made progress on implementing a fix for this, but we didn't finish, so decided not to include that here.


compiler/tests-compiler/global_deadcode.ml
```diff
@@ -0,0 +1,75 @@
...
+    function add(x, t){
+     if(! t) return [0, 0, x, 0, 1];
+     var r = t[3], v = t[2], l = t[1], c = caml_call2(Ord[1], x, v);
+     if(0 === c) return t;
+     if(0 <= c){var rr = add(x, r); return r === rr ? t : bal(l, v, rr);}
+     var ll = add(x, l);
+     return l === ll ? t : bal(ll, v, r);
+    }
+    //end |}];
+  (* Expect: does not compile unused functor member *)
```
hhugo, line 46: Would't it be easier to just print `Set.Make` function  ?

compiler/tests-compiler/global_deadcode.ml
```diff
@@ -0,0 +1,75 @@
...
+        let (u, _, v) = t in
+        (u, v)
+      in print_int (fst (f true 1) + snd (f false 2))
+      |}
+  in 
+  (* Expect second field in each triple to be replaced by a sentinal variable. *)
+  print_fun_decl program (Some "f");
+  [%expect {|
+    function f(b, x){
+     var t = b ? [0, 1, _c_, x] : [0, 3, _c_, 4], v = t[3], u = t[1];
```
hhugo, line 67: in blocks, undefined fields can be omitted 
`var t = b ? [0, 1, , x] : [0, 3, , 4]

hhugo:
Thanks a lot for the contribution. Can you say a word about the impact on compilation speed ?

hhugo:
Can you also report the size impact on another example, the toplevel in `toplevel/examples/lwt_toplevel/toplevel.js`

hhugo:
You should rebase and fix your work  on top of the recently merged #1495

compiler/tests-compiler/global_deadcode.ml
micahcantor: Probably! Good idea

micahcantor:
> Thanks a lot for the contribution. Can you say a word about the impact on compilation speed ?

And thanks for the initial review! It's definitely slower, and I was meaning to add speed benchmarks on my machine but was focusing on if it worked first. I will add that information here though.

> Can you also report the size impact on another example, the toplevel in toplevel/examples/lwt_toplevel/toplevel.js

Sure, and actually some of the other size benchmarks still have room to improve since we're still refining how it takes advantage of the information from global flow, right now it's still a bit conservative. Because of that, and since Jerome and Olivier will be off the next two weeks, I'm going to move this into draft status. Though what's here now shouldn't break anything besides being a bit slower.

compiler/tests-compiler/global_deadcode.ml
vouillon: @micahcantor I would do this in `generate.ml`. You can register in the context which variables are set to `undefined`. I think you can add a case in this match for that:
https://github.com/ocsigen/js_of_ocaml/blob/2b429e532915a82181be28527864971c3d3efc58/compiler/lib/generate.ml#L1323

micahcantor: @vouillon  I'm not sure if I understand exactly, since here I think we would need to output some `Javascript.expression` that represents empty strings for the undefined variables, which we can't do. Since there's no "empty" expression, I would think we'd either need to add this to the JS AST or instead do this at the printing/final generation level.

vouillon: In file `javascript.ml`, an `array_literal` is an `element list`, where an `element` can be an `ElementHole` which corresponds to an omitted  undefined field.

micahcantor: Done in ca4642592ab4545d8d9fb9384d41f64ec2963f35

compiler/tests-compiler/global_deadcode.ml
```diff
@@ -0,0 +1,75 @@
...
+     return [0, u, v];
+    }
+    //end |}];
+  (* And that variable is defined as `undefined` *)
+  print_var_decl program "_c_";
```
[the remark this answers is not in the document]
micahcantor: Done in db8cbbf85a9e193f0ff254cc7de7eae56be86b77

compiler/tests-compiler/global_deadcode.ml
micahcantor: I could do this, but it's a bit fragile since that doesn't have a stable name. It's given the name `_e_` when I run it, but that could change.

micahcantor:
Change log since last review:
1. Made sentinal variable a stable name
2. Added omitting unused fields
3. Turned optimization off by default
4. Started, but then reverted, changes to make liveness analysis recursive since they were unfinished

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+    Array.blit ~src:vars ~src_pos:0 ~dst:compacted ~dst_pos:0 ~len:(!i + 1);
+    compacted
+  in
+  let is_live v =
+    match Var.Tbl.get live_table v with
+    | Dead -> false
+    | _ -> true
+  in
+  let zero_var x = if is_live x then x else sentinal in
+  let zero_cont ((pc, args) : cont) =
```
vouillon, line 331: I don't think it makes any difference to zero continuation arguments, so I think you can just remove this code.

compiler/lib/driver.ml
```diff
@@ -104,9 +104,22 @@ let exact_calls profile p =
...
+        Specialize.f ~function_arity:(fun f -> Global_flow.function_arity info f) p
+      in
+      p, Some sentinal
+    else
+      let info = Global_flow.f ~fast p in
+      let p =
+        Specialize.f ~function_arity:(fun f -> Global_flow.function_arity info f) p
+      in
+      p, None
+  else p, None
```
vouillon, line 122: Maybe you can move the call to `Specialize.f` outside of the conditional:
```suggestion
    let (p, info, sentinal) =
      if Config.Flag.globaldeadcode ()
      then
        let p, sentinal = Deadcode_dgraph.add_sentinal p in
        let info = Global_flow.f ~fast p in
        let p = Deadcode_dgraph.f p sentinal info in
        (p, info, Some sentinal)
      else
        let info = Global_flow.f ~fast p in
        (p, info, None)
    in
    let p =
      Specialize.f ~function_arity:(fun f -> Global_flow.function_arity info f) p
    in
    p, Some sentinal
  else p, None
```

compiler/lib/generate.ml
vouillon, line 1: You could generate just `return` for `Return x` where `x` is the sentinal.

compiler/lib/generate.ml
```diff
@@ -285,12 +285,14 @@ module Ctx = struct
     ; should_export : bool
     ; effect_warning : bool ref
     ; cps_calls : Effects.cps_calls
+    ; sentinal : Var.t option
```
vouillon, line 288: You should use a more explicit name outside of `Deadcode_dgraph`.

compiler/lib/deadcode_dgraph.ml
vouillon, line 1: I would rename this file `global_deadcode.ml`.

compiler/lib/driver.ml
```diff
@@ -579,8 +593,11 @@ let full ~standalone ~wrap_with_fun ~profile ~linkall ~source_map formatter d p
        | O2 -> o2
        | O3 -> o3)
     +> exact_calls profile
-    +> effects
-    +> map_fst (Generate_closure.f +> deadcode')
+    +> (fun (p, sentinal) ->
+         let p, cps_calls = effects p in
+         p, cps_calls, sentinal)
+    +> fun (p, cps_calls, sentinal) ->
+    deadcode' (Generate_closure.f p), cps_calls, sentinal
```
vouillon, line 600:
```suggestion
    +> map_fst (effects +> map_fst (Generate_closure.f +> deadcode'))
```

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+  let pure_funs = Pure_fun.f p in
+  let live_vars = liveness nv p pure_funs global_info in
+  (* Propagate liveness to dependencies *)
+  let vars = variables uses in
+  let live_table = solver vars uses defs live_vars in
+  (* Zero out dead fields *)
+  let p = zero p sentinal live_table in
+  if debug ()
+  then (
+    Code.Print.program (fun _ _ -> "") p;
```
vouillon, line 454: You should move this before the call to `zero p sentinal live_table`, or use a different variable for the program before and after zeroing out dead fields.

vouillon (changes requested):
This looks pretty good!

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+      | Pushtrap (cont, _, cont_h, _) ->
+          add_cont_deps cont;
+          add_cont_deps cont_h
+      | Poptrap cont -> add_cont_deps cont)
+    prog.blocks;
+  uses
+
+(** Return the set of variables used in a given expression *)
+let expr_vars e =
+  let vars = Var.ISet.empty () in
```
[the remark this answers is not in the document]
micahcantor: [4df5399](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/4df539962d03ca39c01162184dbfcbe8c0d0fafd)

compiler/lib/deadcode_dgraph.ml
micahcantor: [075c6f3](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/075c6f34c5e4794524e23fa14f99f8fa0acd9f1f)

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+        | Live fields -> (
+            match defs.(Var.idx y) with
+            | Expr (Block (_, vars, _)) ->
+                let found = ref false in
+                Array.iteri
+                  ~f:(fun i v ->
+                    if Var.equal v x && IntSet.mem i fields then found := true)
+                  vars;
+                if !found then Top else Dead
+            | _ -> Top)
```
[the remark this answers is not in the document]
micahcantor: [075c6f3](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/075c6f34c5e4794524e23fa14f99f8fa0acd9f1f)

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+    | Let (_x, e) ->
+        if not (pure_expr pure_funs e)
+        then (
+          let vars = expr_vars e in
+          Var.ISet.iter add_top vars;)
+        else (match e with
+          | Apply { f; args; _} ->
+              add_top f;
+              List.iter ~f:(fun x -> if variable_may_escape x global_info then add_top x) args
+          | _ -> ())
```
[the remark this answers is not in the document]
micahcantor: [075c6f3](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/075c6f34c5e4794524e23fa14f99f8fa0acd9f1f)

compiler/lib/driver.ml
micahcantor: [a343dc9](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/a343dc9a6d11d4ab4159f399980168b81aadc082)

compiler/lib/deadcode_dgraph.ml
micahcantor: [f0d9506](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/f0d95068e5a373f1ba2a54990ea5b2e4753c00d5)

compiler/lib/driver.ml
micahcantor: This change seems to cause a type error with the (very confusing) input type to `emit` from essentially different associations in a product type

compiler/lib/generate.ml
micahcantor: I think [3523cf5](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/3523cf5255adf23663866b8af2464cf0a59618fe) does it, but will add an explicit test to make sure this works

compiler/lib/deadcode_dgraph.ml
```diff
@@ -0,0 +1,461 @@
...
+ *)
+let zero prog sentinal live_table =
+  let compact_vars vars =
+    let i = ref (Array.length vars - 1) in
+    while !i >= 0 && Var.equal vars.(!i) sentinal do
+      i := !i - 1
+    done;
+    let compacted = Array.make (!i + 1) sentinal in
+    Array.blit ~src:vars ~src_pos:0 ~dst:compacted ~dst_pos:0 ~len:(!i + 1);
+    compacted
```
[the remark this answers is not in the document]
micahcantor: [3d656eb](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/3d656ebe8d64f36499aea178dbc7d876aa5c62fd)

compiler/lib/deadcode_dgraph.ml
micahcantor: [5864030](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/58640303562f2f6680fdbaa0fe3f70c0ef5bd1e9)

compiler/lib/generate.ml
micahcantor: [8e9f0c2](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/8e9f0c243b5dfbc727284260c76ddba6d34a553b)

compiler/lib/generate.ml
micahcantor: [266221e](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/266221e88f53338cc07a5b0f013163978c6b1b62) adds a test for this case.

compiler/lib/driver.ml
vouillon: Sure. You need to adjust function `generate` as well: `(((p, live_vars), cps_calls), sentinal)` instead of `((p, live_vars), cps_calls, sentinal)`.

micahcantor: Okay, [9ab2476](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/9ab2476a25d2fe8c055af90c57ef0b0a753e9270)

hhugo:
@vouillon, could you "resolve conversation" if you\re happy with how they were addressed ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,445 @@
...
+  match e with
+  | Apply { f; args; _ } ->
+      let vars = Var.Set.add f vars in
+      List.fold_left ~f:(fun acc x -> Var.Set.add x acc) ~init:vars args
+  | Block (_, params, _) ->
+      Array.fold_left ~f:(fun acc x -> Var.Set.add x acc) ~init:vars params
+  | Field (z, _) -> Var.Set.add z vars
+  | Constant _ -> vars
+  | Closure (params, _) ->
+      List.fold_left ~f:(fun acc x -> Var.Set.add x acc) ~init:vars params
```
vouillon, line 191:
```suggestion
      vars
```
We want the set of variables which are defined before the expression and used in the expression. This does not include the closure parameters.
The closure may use some variables defined outside, but this will be handled when visiting its body.

compiler/lib/generate.ml
vouillon: Cool! We could omit the `return` when it is not followed by anything, but this is beyond the scope of this PR.

compiler/lib/deadcode_dgraph.ml
vouillon: What I meant is that you are calling `Code.Print.program (fun _ _ -> "") p` twice with the same `p`.
I would expect one occurrence of this expression to be before `zero p sentinal live_table` and the other one after.

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,445 @@
...
+          let vars = expr_vars e in
+          Var.Set.iter add_top vars;
+          match e with
+          | Apply { f; args; _ } ->
+              add_top f;
+              List.iter
+                ~f:(fun x -> if variable_may_escape x global_info then add_top x)
+                args
+          | _ -> ())
+        else ()
```
vouillon, line 237:
```suggestion
        if not (pure_expr pure_funs e)
        then (
          match e with
          | Apply { f; args; _ } ->
              add_top f;
              List.iter
                ~f:(fun x -> if variable_may_escape x global_info then add_top x)
                args
          | _ ->
              let vars = expr_vars e in
              Var.Set.iter add_top vars)
```

vouillon (changes requested):
We are almost there.
Do you think we could enable this optimization by default?

compiler/lib/deadcode_dgraph.ml
micahcantor: Oh okay I misunderstood the first time. And I agree, it was like this earlier but I changed it at some point. Is [d44390f](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/d44390fe2292f06d49313425d689f3b4fae41f54) what you meant?

compiler/lib/global_deadcode.ml
micahcantor: Make sense, thanks. [238b5de](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/238b5dee9c01120a4d1a55e821efe6d57fda51b6)

compiler/lib/global_deadcode.ml
micahcantor: [bbc3932](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/bbc3932060959168bebdca882c765631e384fed5)

micahcantor:
Okay, so after making the most recent round of changes, I re-ran the benchmarks, and found that the running with `globaldeadcode`, the program sizes are mostly the same, but it now runs significantly faster! Here are the results of running with and without the pass on `ocamlc` and `toplevel` with [hyperfine](https://github.com/sharkdp/hyperfine):

```
Benchmark 1: js_of_ocaml `which ocamlc.byte` -o ocamlc.js
  Time (mean ± σ):      9.020 s ±  1.069 s    [User: 8.845 s, System: 0.149 s]
  Range (min … max):    8.319 s … 11.888 s    10 runs

Benchmark 1: js_of_ocaml --enable globaldeadcode `which ocamlc.byte` -o ocamlc.js
  Time (mean ± σ):      9.542 s ±  0.883 s    [User: 9.319 s, System: 0.169 s]
  Range (min … max):    8.872 s … 11.885 s    10 runs

Benchmark 1: js_of_ocaml ./bc/toplevel.bc -o toplevel.js
  Time (mean ± σ):     18.680 s ±  1.354 s    [User: 18.312 s, System: 0.324 s]
  Range (min … max):   17.916 s … 22.493 s    10 runs

Benchmark 1: js_of_ocaml --enable globaldeadcode ./bc/toplevel.bc -o toplevel.js
  Time (mean ± σ):     18.875 s ±  1.124 s    [User: 18.530 s, System: 0.313 s]
  Range (min … max):   18.327 s … 21.996 s    10 runs
```

Those results make me confident that enabling this pass shouldn't cause a major speed regression. Now, it doesn't make a large size difference on these programs, but the gains on other examples haven't changed: Down by ~30% on `sets`, by ~10% on [ocamlgraph/demo](https://github.com/backtracking/ocamlgraph/blob/master/examples/demo.ml) and by  ~11% on [tyxml/basic_website](https://github.com/ocsigen/tyxml/blob/master/examples/basic_website/site_html.ml).

testing/dead_field.js
```diff
@@ -0,0 +1,29 @@
+// Generated by js_of_ocaml
```
hhugo, line 1: What's this file ? Maybe committed by mistake ?

hhugo: compiler/tests-full/stdlib.cma.js was likely also committed by mistake

compiler/lib/global_deadcode.mli
```diff
@@ -0,0 +1,21 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+val add_sentinal : Code.program -> Code.program * Code.Var.t
```
hhugo, line 19: Do we really need `add_sentinal` to be exposed ? Can't it be done inside `f` directly ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,445 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
```
hhugo, line 17: The comment should be split.
- first header 
- then doc

testing/dead_field.js
micahcantor: Oops, yes I added those two accidentally in the most recent commit

micahcantor: Should be fixed now.

compiler/lib/global_deadcode.mli
micahcantor: It can't be done inside `f`, since `Deadcode.f` has to run after `Global_flow.f` runs, but the sentinal needs to be added before global flow runs (otherwise it messes up the index math when accessing variable arrays produced by global flow). This is the order that we have in `exact_calls` (which is a bit of a confusing place to run this, but it's where global flow is right now and unwinding these dependencies would require more refactoring):

```ocaml
let p, deadcode_sentinal = Global_deadcode.add_sentinal p in
let info = Global_flow.f ~fast p in
let p = Global_deadcode.f p deadcode_sentinal info in
p, info, Some deadcode_sentinal
```

Maybe this dependency could be documented better?

compiler/lib/global_deadcode.ml
micahcantor: [658dcb0](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/658dcb0dcadbfc6ec7db1d4824b797e8c4ff06bc) moved this to a doc comment in the `.mli` file which I think is right but I'm not an `odoc` expert.

compiler/lib/global_flow.mli
```diff
@@ -31,10 +31,17 @@ type approx =
...
 
+type escape_status =
+  | Escape
+  | Escape_constant (* Escapes but we know the value is not modified *)
+  | No
+
 type info =
   { info_defs : def array
   ; info_approximation : approx Var.Tbl.t
   ; info_may_escape : Var.ISet.t
```
hhugo, line 42: Do we need to keep info_may_escape ? Or can't we just change its type ?

compiler/lib/generate.ml
```diff
@@ -285,12 +285,14 @@ module Ctx = struct
     ; should_export : bool
     ; effect_warning : bool ref
     ; cps_calls : Effects.cps_calls
+    ; deadcode_sentinal : Var.t option
```
hhugo, line 288: Having this not an option should decrease the number of comparison. The sentinal can be set to a fresh variable if not provided.

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    + They are returned; or
+    + They are applied to a function. 
+ *)
+let zero prog sentinal live_table =
+  let compact_vars vars =
+    let i = ref (Array.length vars - 1) in
+    while !i >= 0 && Var.equal vars.(!i) sentinal do
+      i := !i - 1
+    done;
+    Array.sub vars ~pos:0 ~len:(!i + 1)
```
hhugo, line 315: Can this optimization be observed at runtime using `Obj.size` ?

micahcantor: Do you mind elaborating? I see from [here](https://discuss.ocaml.org/t/what-does-obj-size-return/12053/5) that `Obj.size` will return the number of fields in the array but I don't see how to use that to count the number of sentinal fields at the tail?

compiler/lib/global_flow.mli
micahcantor: @vouillon

vouillon: We need both.
`info_may_escape` are a set of _values_ that may escape. It is used for the CPS transformation (functions that escape must be in CPS).
`info_variable_may_escape` is a set of _variables_ whose contents may escape. For dead code elimination, we use it to tell whether the arguments of a function may escape.

compiler/lib/global_deadcode.ml
vouillon: @hhugo is wondering whether shortening the block can be observed using function `Obj.size`, and so whether applying dead code elimination might change the behavior of a program.

I don't think so. If the array is passed to `Obj.size`, we are going to consider that all its fields are used, since we do not know what `Obj.size` does.

compiler/lib/global_flow.mli
hhugo: Ok, I mis-read that `info_may_escape` was derived from `variable_may_escape` but it's in fact derived from `may_escape`.

compiler/lib/global_deadcode.ml
micahcantor: I see, that makes sense. Is there a runtime primitive for `Obj.size` to check if it's marked as pure?

compiler/lib/generate.ml
micahcantor: [6867324](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/68673240462c4d700ed3d0613bfec51d8e043f95)

compiler/tests-compiler/global_deadcode.ml
hhugo: I've changed the test to first find the name of the functor and then print its body.

There might be a bug to fix later. I would expect this variable to be stable/named ("Make"), @vouillon do you think it can be a missing debug event ?

compiler/lib/global_deadcode.ml
hhugo: I think it uses the instruction `Code.Vectlength`.

compiler/lib/global_deadcode.mli
hhugo: Are you sure this is true ?

When I apply the following diff, the testsuite still run fine
```diff
diff --git a/compiler/lib/driver.ml b/compiler/lib/driver.ml
index 172bdcb324..6a481e3508 100644
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -107,8 +107,8 @@ let exact_calls profile p =
     let p, info, deadcode_sentinal =
       if Config.Flag.globaldeadcode ()
       then
-        let p, deadcode_sentinal = Global_deadcode.add_sentinal p in
         let info = Global_flow.f ~fast p in
+        let p, deadcode_sentinal = Global_deadcode.add_sentinal p in
         let p = Global_deadcode.f p deadcode_sentinal info in
         p, info, deadcode_sentinal
       else
```

I don't expect your pass to access global_flow for the sentinal.

Or maybe one just need to create/allocate the sentinal before
```ocaml
        let deadcode_sentinal = Code.Var.fresh () in
        let info = Global_flow.f ~fast p in
        Global_deadcode.f p deadcode_sentinal info in
```

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    match l1, l2 with
+    | _, Top | Top, _ -> Top
+    | Live f1, Live f2 -> Live (IntSet.union f1 f2)
+    | Dead, Live f | Live f, Dead -> Live f
+    | Dead, Dead -> Dead
+end
+
+module Solver = G.Solver (Domain)
+
+let pure_expr pure_funs e = Pure_fun.pure_expr pure_funs e && Config.Flag.deadcode ()
```
hhugo, line 61: Why is this conditioned on `deadcode` ? Shouldn't the whole pass be conditioned instead ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+module G = Dgraph.Make_Imperative (Var) (Var.ISet) (Var.Tbl)
+
+module Domain = struct
+  type t = live
+
+  let equal l1 l2 =
+    match l1, l2 with
+    | Top, Top | Dead, Dead -> true
+    | Live f1, Live f2 -> IntSet.equal f1 f2
+    | _ -> false
```
hhugo, line 46: Avoid using catch all. It will help future refactoring.
```ocaml
 | Top, (Dead | Live _)
 | Live _, (Dead | Top)
 | Dead, (Live _ | Top)
```

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    as [Propagate] if [x] is applied as a closure or block argument the parameter [y]. 
+    
+    We use information from global flow to try to add edges between function calls and their return values
+    at known call sites. *)
+let usages nv prog (global_info : Global_flow.info) : usage_kind Var.Map.t array =
+  let uses = Array.make nv Var.Map.empty in
+  let add_use kind x y = uses.(Var.idx y) <- Var.Map.add x kind uses.(Var.idx y) in
+  let add_arg_dep params args =
+    try List.iter2 ~f:(fun x y -> add_use Propagate x y) params args
+    with Invalid_argument _ -> ()
```
hhugo, line 100: Can this ever happen ? if so, why ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+        | Values { known; _ } ->
+            Var.Set.iter (* For each known closure value of f *)
+              (fun k ->
+                (* 1. Look at return values, and add edge between x and these values. *)
+                (* 2. Add an edge pairwise between the parameters and arguments *)
+                match global_info.info_defs.(Var.idx k) with
+                | Expr (Closure (params, _)) ->
+                    (* If the function is under/over-applied then global flow will mark arguments and return value as escaping.
+                       So we only need to consider the case when there is an exact application. *)
+                    if List.length params = List.length args
```
hhugo, line 121: List.compare_lenghts params args = 0

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+  | Field (z, _) -> Var.Set.add z vars
+  | Prim (_, args) ->
+      List.fold_left
+        ~f:(fun acc v ->
+          match v with
+          | Pv v -> Var.Set.add v vars
+          | Pc _ -> acc)
+        ~init:vars
+        args
+  | Constant _ | Closure (_, _) -> vars
```
hhugo, line 187: Maybe add a comment explaining why it's ok to ignore Closure

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+  let live_vars = Array.make nv Dead in
+  let add_top v =
+    let idx = Var.idx v in
+    live_vars.(idx) <- Top
+  in
+  let add_live v i =
+    let idx = Var.idx v in
+    match live_vars.(idx) with
+    | Live fields -> live_vars.(idx) <- Live (IntSet.add i fields)
+    | _ -> live_vars.(idx) <- Live (IntSet.singleton i)
```
hhugo, line 209: `| (Top | Dead) -> `

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+  in
+  let live_instruction i =
+    match i with
+    | Let (_x, e) -> (
+        if not (pure_expr pure_funs e)
+        then
+          match e with
+          | Apply { f; args; _ } ->
```
hhugo, line 217: Why Apply and not Prim ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+  | Propagate  (** values of y propagate to x *)
+
+(** Compute the adjacency list for the dependency graph of given program. An edge between
+    variables [x] and [y] is marked [Compute] if [x] is used in the definition of [y]. It is marked
+    as [Propagate] if [x] is applied as a closure or block argument the parameter [y]. 
+    
+    We use information from global flow to try to add edges between function calls and their return values
+    at known call sites. *)
+let usages nv prog (global_info : Global_flow.info) : usage_kind Var.Map.t array =
+  let uses = Array.make nv Var.Map.empty in
```
hhugo, line 96: Shouldn't this use `Var.Tbl.make () Var.Map.empty` ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    | Let (_x, e) -> (
+        if not (pure_expr pure_funs e)
+        then
+          match e with
+          | Apply { f; args; _ } ->
+              add_top f;
+              List.iter
+                ~f:(fun x -> if variable_may_escape x global_info then add_top x)
+                args
+          | _ ->
```
hhugo, line 222: Make the pattern exhaustive

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+              let vars = expr_vars e in
+              Var.Set.iter add_top vars)
+    | Assign (_, _) -> ()
```
hhugo, line 225: Can you add a comment explaining why we don't do anything here ?

compiler/lib/global_deadcode.ml
hhugo, line 315: We can avoid the array allocation if `!i + 1` = `length`

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    let i = ref (Array.length vars - 1) in
+    while !i >= 0 && Var.equal vars.(!i) sentinal do
+      i := !i - 1
+    done;
+    Array.sub vars ~pos:0 ~len:(!i + 1)
+  in
+  let is_live v =
+    match Var.Tbl.get live_table v with
+    | Dead -> false
+    | _ -> true
```
hhugo, line 320: Make it exhaustive

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+                let vars =
+                  Array.mapi
+                    ~f:(fun i v -> if IntSet.mem i fields then v else sentinal)
+                    vars
+                  |> compact_vars
+                in
+                let e = Block (start, vars, is_array) in
+                Let (x, e)
+            | _ -> instr)
+        | Apply ap ->
```
hhugo, line 339: Follow up, could we extend the analysis to primitives ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+            Let (x, Apply { ap with args })
+        | _ -> instr)
+    | _ -> instr
+  in
+  let zero_block block =
+    (* Analyze block instructions *)
+    let body = List.map ~f:(fun (instr, loc) -> zero_instr instr, loc) block.body in
+    (* Analyze branch *)
+    let branch =
+      let last, loc = block.branch in
```
hhugo, line 350: I would match on `block.branch` directly and return block.branch if it's not a `Return`

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+    Var.Tbl.iter
+      (fun v l -> Format.eprintf "%a: %s\n" Var.print v (live_to_string l))
+      live_table
+end
+
+(** Add a sentinal variable declaration to the IR. The fresh variable is assigned to `undefined`. *)
+let add_sentinal p =
+  let sentinal = Var.fresh_n "sentinal" in
+  let undefined = Prim (Extern "%undefined", []) in
+  let instr, loc = Let (sentinal, undefined), Before 0 in
```
hhugo, line 403: Should this be `noloc` ?

compiler/lib/global_deadcode.ml
vouillon: `Assign` is handled like passing an argument to a block parameter, which is handled in function `usages`

compiler/lib/global_deadcode.ml
vouillon: Because the global flow analysis tells us something about functions, but not about primitives.

compiler/lib/global_deadcode.ml
vouillon: It does not really matter whether it is pure or not. What matters is that the liveness of its arguments are going to be set to Top if they are used.

compiler/lib/global_deadcode.ml
vouillon: I think we can just remove the condition, since we have a different flag. This line was copied from `deadcode.ml`.
```suggestion
let pure_expr pure_funs e = Pure_fun.pure_expr pure_funs e
```

compiler/lib/global_deadcode.ml
vouillon: Probably not

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,435 @@
...
+          args
+  in
+  Addr.Map.iter
+    (fun _ block ->
+      (* Add deps from block body *)
+      List.iter
+        ~f:(fun (i, _) ->
+          match i with
+          | Let (x, e) -> add_expr_uses x e
+          | Assign (x, y) -> add_use Compute x y
```
vouillon, line 151: Assign just propages values
```suggestion
          | Assign (x, y) -> add_use Propagate x y
```

compiler/lib/global_deadcode.ml
micahcantor: [34edd45](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/34edd4540048866a4c530a8bb8febdc30cf9ed27)

compiler/lib/global_deadcode.ml
micahcantor: [cd3eb5a](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/cd3eb5a365d949951696482a424e29531db92e68)

compiler/lib/global_deadcode.ml
micahcantor: Yes and [646b216](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/646b216fade85c7567e0b559fedd661b1c972503) refactors to use Var.Tbl instead of Array in other places too

compiler/lib/global_deadcode.ml
micahcantor: https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/646b216fade85c7567e0b559fedd661b1c972503

compiler/lib/global_deadcode.ml
micahcantor: [fc846de](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/fc846de5ff436100be11ec8f869f90cdc62f444a)

compiler/lib/global_deadcode.ml
micahcantor: No, this was leftover from when there was a bug in an earlier version that allowed these to be different lengths. [fc846de](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/fc846de5ff436100be11ec8f869f90cdc62f444a)

compiler/lib/global_deadcode.ml
micahcantor: [5dfd1ca](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/5dfd1caed57d2d0c85606063347b77238cffdb77)

compiler/lib/global_deadcode.ml
micahcantor: [5dfd1ca](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/5dfd1caed57d2d0c85606063347b77238cffdb77)

compiler/lib/global_deadcode.ml
micahcantor: [dfb0c22](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/dfb0c22e92cd9faf8f2a91d4a281b1d92c5f53ae)

compiler/lib/global_deadcode.ml
micahcantor: [be14b11](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/be14b110213814c4d81b0fb7dca439c561f0b5c2)

compiler/lib/global_deadcode.ml
micahcantor: [be14b11](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/be14b110213814c4d81b0fb7dca439c561f0b5c2)

compiler/lib/global_deadcode.ml
micahcantor: I'm not sure. Right now that definition is inserted as the first line of the program. I think that makes sense in case we need to zero out anything following that.

compiler/lib/global_deadcode.ml
micahcantor: If the primitive is impure it doesn't matter since all the arguments will be marked top. @vouillon do you think it could make a difference for pure primitives?

compiler/lib/global_deadcode.ml
micahcantor: [0214876](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/02148767e80b11152e311c0f437f8d5ba59f386f)

compiler/lib/global_deadcode.ml
micahcantor: [0214876](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/02148767e80b11152e311c0f437f8d5ba59f386f)

compiler/lib/global_deadcode.mli
micahcantor: No, you're right. I think this dependency existed in an earlier version due to a bug that has since been fixed. I changed this in https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/07c37bdcb8dfa0229c30b42e5ca28aea47b7bee3.

micahcantor:
Okay I added changes and responded to the most recent round reviews. I resolved simple changes but still a few conversations left unresolved. Thanks!

compiler/lib/global_deadcode.ml
hhugo: I''ve pushed a commit to use `noloc` (instead of `Before 0`)

compiler/lib/driver.ml
```diff
@@ -105,8 +109,14 @@ let exact_calls profile p =
       | O1 | O2 -> true
     in
     let info = Global_flow.f ~fast p in
-    Specialize.f ~function_arity:(fun f -> Global_flow.function_arity info f) p
-  else p
+    let p =
+      if Config.Flag.globaldeadcode ()
```
hhugo, line 113: We don't perform the optimization if `Config.Flag.effects ()` is true, is it expected ?

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,445 @@
...
+    match l1, l2 with
+    | _, Top | Top, _ -> Top
+    | Live f1, Live f2 -> Live (IntSet.union f1 f2)
+    | Dead, Live f | Live f, Dead -> Live f
+    | Dead, Dead -> Dead
+end
+
+module Solver = G.Solver (Domain)
+
+let pure_expr pure_funs e = Pure_fun.pure_expr pure_funs e
```
hhugo, line 61: This alias is not necessary

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,445 @@
...
+let add_sentinal p sentinal =
+  let undefined = Prim (Extern "%undefined", []) in
+  let instr, loc = Let (sentinal, undefined), noloc in
+  Code.prepend p [ instr, loc ]
+
+(** Run the liveness analysis and replace dead variables with the given sentinal. *)
+let f p ~deadcode_sentinal global_info =
+  let t = Timer.make () in
+  (* Add sentinal variable *)
+  let p = add_sentinal p deadcode_sentinal in
```
hhugo, line 419: Is this variable kept in the js file if unused ? I'm assuming deadcode won't be able to remove it because the external `%undefined"`is not known as pure. Can you maybe fix this and add a test ?

compiler/lib/global_deadcode.ml
micahcantor: [48fccd4](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/48fccd4502c0114a26b0c3595dc37773a74a6c42)

compiler/lib/driver.ml
micahcantor: I really only put it here since this is where `Global_flow.f` is called in `driver`, but I'm not sure if it's correct beyond that.

compiler/lib/global_deadcode.ml
micahcantor: I may be misunderstanding your question, but no? I tried this program:

```ocaml
(* testing/nodead.ml *)
let f x = x + 1
in print_int (f 1 + f 2 + f 3)
```
And ran this:
```
ocamlc -c testing/nodead.ml -o testing/nodead.cmo
dune exec -- js_of_ocaml --enable globaldeadcode --pretty testing/nodead.cmo
```
The result is:
```js
(function
  (globalThis){
   "use strict";
   var runtime = globalThis.jsoo_runtime;
   function caml_call1(f, a0){
    return (f.l >= 0 ? f.l : f.l = f.length) == 1
            ? f(a0)
            : runtime.caml_call_gen(f, [a0]);
   }
   var
    Stdlib = global_data.Stdlib;
   caml_call1(Stdlib[44], 9);
   var Nodead = [0];
   runtime.caml_register_global(1, Nodead, "Nodead");
   return;
  }
  (globalThis));
```

hhugo: It turns out that `undefined`, if unused, is removed by `Js_traverse.clean`. More precisely, by 

```ocaml
    method statements l =
      let l = super#statements l in
      List.filter l ~f:(function
          | (Empty_statement | Expression_statement (EVar _)), _ -> false
          | _ -> true)
```

compiler/lib/driver.ml
hhugo: global_flow is also called in effects.ml

We should probably move some "driver" logic outside effects.ml into driver.ml

compiler/lib/global_deadcode.ml
hhugo: Let's leave that for later.

hhugo:
When turning the optimization on by default, I see tests failure in the following places:
- compiler/tests-ocaml/lib-printf/
- compiler/tests-ocaml/lib-format/
- toplevel/test

They seem to all involve printf/format

```diff
diff --git a/toplevel/test/test_toplevel.reference b/toplevel/test/test_toplevel.reference
index 2ab06fc0ce..0298111ab7 100644
--- a/toplevel/test/test_toplevel.reference
+++ b/toplevel/test/test_toplevel.reference
@@ -3,7 +3,6 @@ external parseInt : float -> int = "parseInt"
 let f = 3.14
 let () = Printf.printf "parseInt(%f) = %d\n" f (parseInt f);;
 Dynlink: looking for symbol parseInt
-parseInt(3.140000) = 3
 external parseInt : float -> int = "parseInt"
 val f : float = 3.14
```

compiler/lib/driver.ml
vouillon: We have not looked at that so far since it seemed a bit tricky.
But maybe it just work there as well if the deadcode elimination phase is performed right after the global flow analysis.

compiler/lib/global_deadcode.ml
```diff
@@ -0,0 +1,443 @@
...
+              let vars = expr_vars e in
+              Var.Set.iter add_top vars)
+    | Set_field (x, i, y) ->
+        add_live_field x i;
+        add_top y
+    | Array_set (x, y, z) ->
+        add_top x;
+        add_top y;
+        add_top z
+    | Offset_ref (x, i) -> add_live_field x i
```
vouillon, line 235: `Offset_ref (x, i)` adds `i` to the field `0` of the value of `x`
```suggestion
    | Offset_ref (x, _) -> add_live_field x 0
```

micahcantor: Okay I'm not getting printf failures with this change! I never paid much attention to this expression...what is it?

micahcantor: [Fix offset_ref liveness](https://github.com/ocsigen/js_of_ocaml/pull/1503/commits/1c492f09dd150a815d059b85d8c76d41d467cf85)

compiler/lib/driver.ml
hhugo: I've tried to naively apply the global DCE pass when effects are enabled.

Doing DCE before the effect logic result in a bunch of diffs as the following one
```diff
 |       return caml_callback
 |               (function(cont){
 |                 var
+|                  _c_ = undefined,
 |                  global_data = runtime.caml_get_global_data(),
 |                  Stdlib_Printf = global_data.Stdlib__Printf,
 |                  _a_ =
 |                    [0,
 |                     [11, caml_string_of_jsbytes("abc"), 0],
 |                     caml_string_of_jsbytes("abc")];
 |                 function g(param, cont){
-|                  return caml_cps_call2(Stdlib_Printf[2], _a_, cont);
+|                  var _g_ = Stdlib_Printf[2];
+|                  return caml_cps_call2
+|                          (_g_, _a_, function(_h_){return cont(_c_);});
 |                 }
-|                 caml_callback(g, [0]);
+|                 caml_callback(g, [_c_]);
```

Doing DCE after the effect logic gives me better diff but also triggers compilation and runtime errors.
```diff
------ compiler/tests-compiler/lambda_lifting.ml
++++++ compiler/tests-compiler/lambda_lifting.ml.corrected
File "compiler/tests-compiler/lambda_lifting.ml", line 18, characters 0-1:
 |  let prog =
 |    {|
 |let f x =
 |  let g y =
 |    let h z = x + y + z in
 |    h 7
 |  in
 |  g 5
 |in
 |Printf.printf "%d\n" (f 3)
 |    |}
 |  in
 |  let flags =
 |    [ "--no-inline"; "--set=lifting-threshold=1"; "--set=lifting-baseline=0" ]
 |  in
 |  Util.compile_and_run ~effects:true ~flags prog;
-|  [%expect {|15 |}];
+|  [%expect {|
+|    /tmp/build_259527_dune/jsoo-test4651ac/test.js:2741
+|         throw err;
+|         ^
+|
+|    TypeError: Cannot read properties of undefined (reading 'opened')
+|        at caml_ml_output_ta (/tmp/build_259527_dune/jsoo-test4651ac/test.js:1837:15)
+|        at caml_ml_output_bytes (/tmp/build_259527_dune/jsoo-test4651ac/test.js:1876:12)
+|        at caml_ml_output (/tmp/build_259527_dune/jsoo-test4651ac/test.js:1879:12)
+|        at output_string (/tmp/build_259527_dune/jsoo-test4651ac/test.js:3146:5)
+|        at /tmp/build_259527_dune/jsoo-test4651ac/test.js:5523:27
+|        at _cj_ (/tmp/build_259527_dune/jsoo-test4651ac/test.js:5540:41)
+|        at output_acc (/tmp/build_259527_dune/jsoo-test4651ac/test.js:5511:58)
+|        at caml_cps_exact_call3 (/tmp/build_259527_dune/jsoo-test4651ac/test.js:2792:15)
+|        at _cj_ (/tmp/build_259527_dune/jsoo-test4651ac/test.js:5568:16)
+|        at output_acc (/tmp/build_259527_dune/jsoo-test4651ac/test.js:5511:58)
+|
+|    Node.js v20.0.0
+|
+|    process exited with error code 7
+|     node test.js |}];
```

and 

```
  _build/default/compiler/bin-js_of_ocaml/js_of_ocaml.exe: Error: Invalid_argument("List.iter2")
```

vouillon: The global DCE does not preserve tail calls. We don't care normally since JavaScript does not support tail calls, but this makes a difference for the CPS transformation.

Maybe we can disable this transformation when effects are enabled?
https://github.com/ocsigen/js_of_ocaml/blob/8d1841f748ab5ef4a4eb37d4b1e13dddc934b132/compiler/lib/global_deadcode.ml#L355

micahcantor: Right now we only call `Global_flow` when effects are disabled, so I'm a little confused where the proposed change for calling global deadcode would be in the case that effects are enabled.

hhugo: Global_flow is called inside effect.ml

micahcantor: Thanks, missed that one

micahcantor: Okay I also tried a naive effort to call Global_flow with effects [here](https://github.com/micahcantor/js_of_ocaml/pull/1/files) with @vouillon 's suggestion, but, it's still crashing a bunch of effects tests. I don't really know anything about that pass so it's hard to debug.

micahcantor: It's hard for me to gauge how large of a change is required for this, from tweaking a few lines to a bigger refactor of the driver. If it's more like the latter, maybe we should split this into a separate PR?

micahcantor: I ask only because I have an article I'm waiting to post until this is merged, but haven't had time to take a closer look at this issue. I also don't know of any examples they would benefit from effects + gdc (but there might be!)

micahcantor: @vouillon @hhugo

compiler/lib/global_deadcode.ml
vouillon: This is a shortcut for `x := !x + i`, corresponding to the `OFFSETREF` bytecode instruction. I'm not sure we really need it, but well...

vouillon (approved):

compiler/lib/driver.ml
vouillon: If @hhugo, I think we can merge this PR, indeed.

vouillon: It seems it's missing a changelog entry, though.

compiler/tests-compiler/gh1007.ml
```diff
@@ -467,49 +463,45 @@ let ()  = M.run ()
...
+           if(2 < n >>> 0) return 1 - (1 - odd(n - 1 | 0));
+           switch(n){
+             case 0:
+              var
+               f = function(param){return caml_call2(Stdlib_Printf[2], _c_, i);};
+              delayed[1] = [0, f, delayed[1]];
+              f(0);
+              return 1;
+             case 1:
+              return 1 - (1 - odd(0));
```
hhugo: This double boolean not (` 1-X`) is surprising, I'll investigate

compiler/lib/driver.ml
hhugo: There are still a few points I'd like to look at before merging.
I ll try to do it today

micahcantor: Thanks. I'll add a PR for this effects change too

hhugo:
Thanks a lot for such a big contribution.

micahcantor:
Thank you for all the help getting this merged!!

OlivierNicole:
Congratulations @micahcantor for the merge! This is a great feature. Responding to the reviewers took substantial work, so thank you for spending time on this.

hhugo:
> The global DCE does not preserve tail calls. We don't care normally since JavaScript does not support tail calls, but this makes a difference for the CPS transformation.
> 
> Maybe we can disable this transformation when effects are enabled?
> 
> https://github.com/ocsigen/js_of_ocaml/blob/8d1841f748ab5ef4a4eb37d4b1e13dddc934b132/compiler/lib/global_deadcode.ml#L355

It seems that doing so can be incorrect because it could keep live some dead code that uses other zero-ed values. I'll this change,

hhugo:
cc @vouillon @micahcantor, see my comment above

PR #1505, opened 2023-08-07
Compiler: exit loop early

hhugo:
initially based on #1333 and #1334, then rebased on #1496.

The current implementation exits a loop as soon as possible. The generated code would be better if we allowed part of the continuations to remains inside the for-loop.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## Features/Changes
 * Compiler: global dead code elimination (Micah Cantor, #1503)
 * Compiler: change control-flow compilation strategy (#1496)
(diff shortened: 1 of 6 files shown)


hhugo:
Check https://github.com/ocaml-wasm/wasm_of_ocaml/commit/f517bbcafd81de486cf9951f401bd45a69ec4ae8

hhugo:
@vouillon, I've cherry-picked your implementation from wasm_of_ocaml. Do you think the same heuristic is appropriate for jsoo ? In particular, the is_small part ?

hhugo:
We should probably try to not nest loop if unnecessary.

PR #1507, opened 2023-09-03
Download attribute

slegrand45:
Hi,

Tiny patch to add `download` attribute in `<a>`:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download
https://caniuse.com/download


hhugo:
thanks

PR #1510, opened 2023-09-12
Dead code elimination of some unused references

vouillon:
If we have some code like the following where `f` and `g` are not used otherwise,  neither `f` nor `g` have to be considered live.
```ocaml
let f = ref None
....
let () = f := Some g
```
More generally, an assignment `Set_field (x, _, y)` will not mark `x` as live and will mark `y` as live only if `x` is live.

### DIFF
--- a/compiler/lib/deadcode.ml
+++ b/compiler/lib/deadcode.ml
@@ -28,6 +28,11 @@ open Code
 type def =
   | Expr of expr
   | Var of Var.t
+  | Field_update of Var.t
+
+let add_def defs x i =
+  let idx = Var.idx x in
+  defs.(idx) <- i :: defs.(idx)
 
 type variable_uses = int array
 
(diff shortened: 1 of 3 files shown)


hhugo:
x-ref https://github.com/ocaml-community/yojson/pull/168 which is working around the current limitation.

compiler/lib/deadcode.ml
```diff
@@ -109,8 +123,8 @@ and mark_reachable st pc =
 let live_instr st i =
   match i with
   | Let (x, e) -> st.live.(Var.idx x) > 0 || not (pure_expr st.pure_funs e)
-  | Assign (x, _) -> st.live.(Var.idx x) > 0
-  | Set_field _ | Offset_ref _ | Array_set _ -> true
+  | Assign (x, _) | Set_field (x, _, _) -> st.live.(Var.idx x) > 0
+  | Offset_ref _ | Array_set _ -> true
```
hhugo, line 127: Why don't we have the same logic for `Array_set` ? (when we statically know that bounds are ok)

vouillon: Using a forward reference is a common pattern which is worthwhile to take into account. I'm not sure there is much to gain with `Array_set` and this adds some complexity.

PR #1515, opened 2023-10-03
The async keyword should not be followed by a newline

### LINKED ISSUE #1514: [BUG] Compiling using `--profile=release` leads to runtime errors
**Describe the bug**
At @funkywork, we're working on a collection of tools to interact with Tezos. We recently realized that when using our libraries, producing JavaScript artifacts with the `release` profile generates incorrect JavaScript output. For example, in this project (I think it's pretty minimal):  https://github.com/funkywork/tezos-dapps-examples/tree/main.
Using `dune build; dune exec bin/server/dapps.exe` and going to http://localhost:8888/dapps/connect launch properly the apps. But if you run `dune build --profile=release; dune exec bin/server/dapps.exe` you will have an error console: 

```
12:10:45.847 Uncaught SyntaxError: missing ) after argument list client.bc.js:4635
12:10:45.848 Uncaught ReferenceError: nightmare_js is not defined
    <anonymous> http://localhost:8888/dapps/connect:2
connect:2:358
12:10:45.848 Uncaught ReferenceError: nightmare_js is not defined
    <anonymous> http://localhost:8888/dapps/connect:2
connect:2:465
```

And here is the client.bc.js generated: https://gist.githubusercontent.com/xvw/d679d490d085e671f926e106cfff37b5/raw/69c2c552065c0466a93a4c05b323ccf75c6ef01a/client.bc.js

From a development point of view, this isn't a problem, but it does make it difficult to put things into production, as the non-release bundle weighs in at ~16mb, compared with ~2mb for release.

If I can provide any additional information, or create a smaller example, please don't hesitate to ask!

**Versions**
- `ocaml`: `5.0.0`
- `js_of_ocaml`: `5.4.0+git-e0d000e-dirty`

vouillon:
Fixes #1514

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,8 +9,9 @@
 
 ## Bug fixes
 * Runtime: fix Dom_html.onIE (#1493)
-* Compiler: fix global flow analysis (#1494)
 * Runtime: add conversion functions + strict equality for compatibility with Wasm_of_ocaml (#1492)
+* Compiler: fix global flow analysis (#1494)
+* COmpiler: fix js parser/printer wrt async functions (#1515)
 
 # 5.4.0 (2023-07-06) - Lille
 
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -444,7 +444,7 @@ struct
         (match k with
         | { async = true; generator = false } ->
             PP.string f "async";
-            PP.space f
+            PP.non_breaking_space f
         | { async = false; generator = false } -> ()
         | { async = true | false; generator = true } -> assert false);
         PP.break f;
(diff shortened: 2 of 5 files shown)


xvw:
Hi @vouillon 
I have try your fix on https://github.com/funkywork/yourbones/tree/main/examples and I still have an issue related to that kind of line:

```
this.connection.open(a),a=this.connection),this.connection=this.setConnection(a),await
```

in https://gist.github.com/xvw/767a6e2ad72bff091b594a542d3cc744
So I guess that the fix should be applied to `await` too.

vouillon:
@xvw I missed a case. An `async` at the end of the line is not recognized as a keyword, and thus we get an error at the following `await`.

hhugo:
@vouillon, I've added some tests added more fixes. I also rebased  your branch.

hhugo:
I see a different line for line number 4875
```js
window&&void
```

xvw:
Ok, using a fixed commit (930234343cc18cd3d640a51656f43f741f3ef728) it is working! Thanks a lot!

PR #1516, opened 2023-10-15
Compiler: reduce memory consumption

### LINKED ISSUE #1444: [BUG] Latest master takes up too much memory and fails with Fatal error: out of memory
**Describe the bug**
While trying to test out #1442, I noticed that JSOO took up 10gb of RAM on my mac while building libbinaryen. It completed; however, when running in CI, we're getting fatal OoM errors. See https://github.com/grain-lang/libbinaryen/pull/78 for the CI runs.

**Expected behavior**
Expected to take up less memory and/or not crash with OoM.

**Versions**
OCaml: 4.13.1 and 4.14.0
JSOO: master
OS: Linux and Windows (the Mac runners might be larger??)

hhugo:
fix #1444

### DIFF
--- a/compiler/bin-js_of_ocaml/cmd_arg.ml
+++ b/compiler/bin-js_of_ocaml/cmd_arg.ml
@@ -365,7 +365,7 @@ let options =
   let t =
     Term.(
       const build_t
-      $ Jsoo_cmdline.Arg.t
+      $ Lazy.force Jsoo_cmdline.Arg.t
       $ set_param
       $ set_env
       $ dynlink
@@ -604,7 +604,7 @@ let options_runtime_only =
   let t =
     Term.(
       const build_t
-      $ Jsoo_cmdline.Arg.t
+      $ Lazy.force Jsoo_cmdline.Arg.t
       $ toplevel
       $ no_cmis
       $ set_param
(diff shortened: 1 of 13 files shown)


hhugo:
@phated, can you try this one ?

phated:
On my local computer, this has reduced the memory of each JSOO execution to 2.6gb 🎉  I've updated https://github.com/grain-lang/grain/pull/1909 to test out some runs against CI runners.

phated:
@hhugo did something else change in JSOO that would make the execution of our tests extremely slow? Ubuntu has been running for almost 2 hours and they usually complete in 3 minutes.

hhugo:
> @hhugo did something else change in JSOO that would make the execution of our tests extremely slow? Ubuntu has been running for almost 2 hours and they usually complete in 3 minutes with v5.0.1

We'd need to investigate, here are two changes probably affecting the runtime profile. 

- Compiler: change memory representation of OCaml strings to use js ones.
- Compiler: change control-flow compilation strategy (#1496)

hhugo:
I would be nice to understand what's the problematic pattern though

hhugo:
@hhugo, I'm looking a the grain CI and now wonder if I messed my testing (and conclusion) up

PR #1519, opened 2023-10-24
fix: add missing return for `stat` Unix functions

brandonspark:
## What:
This PR fixes the JS implementation of the `caml_unix_lstat_64`and `caml_unix_stat_64` calls.

## Why:
Per the OCaml documentation (https://github.com/ocaml/ocaml/blob/f9371a2ea294f75da793451ef7be5dc69aad6b53/otherlibs/unix/unix_unix.ml#L388), the `caml_unix_lstat_64` external function has type signature `string -> stats`. In particular, it is meant to return a value of type `stats`.

In the Js_of_ocaml implementation of that external function, however, it returns nothing at all. This results in an `undefined`, which will then proceed to wreak havoc.

`caml_unix_stat_64` is similar.

## How:
Added back the `return` for the `r` object.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 * Runtime: fix Dom_html.onIE (#1493)
 * Runtime: add conversion functions + strict equality for compatibility with Wasm_of_ocaml (#1492)
 * Runtime: Dynlink should be able to find symbols in jsoo_runtime #1517
+* Runtime: fix Unix.lstat, Unix.LargeFile.lstat (#1519)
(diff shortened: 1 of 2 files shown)


hhugo:
Same issue with `caml_unix_stat_64`, do you mind fixing this as well?

brandonspark:
@hhugo done!

hhugo:
Thanks for your contribution. Just missing a changelog entry

brandonspark:
ah, should have read the `CONTRIBUTING.md`. Added now! @hhugo

PR #1523, opened 2023-11-10
Compiler: check stack compatibility when jumping

hhugo:


compiler/lib/effects.ml
```diff
@@ -926,6 +926,8 @@ let remove_empty_blocks ~live_vars (p : Code.program) : Code.program =
 let f (p, live_vars) =
   let t = Timer.make () in
   let p = remove_empty_blocks ~live_vars p in
+  (* [remove_empty_blocks] can affect [Deadcode.variable_uses] *)
+  let p, live_vars = Deadcode.f p in
```
hhugo, line 930: I think the change in `parse_bytecode.ml` exposed a bug in `effects.ml`. @vouillon, can you correct the fix is correct ?

hhugo:
This change breaks TCO, we need something different

PR #1525, opened 2023-11-11
Compiler: fix for global deadcode

hhugo:


hhugo:
@micahcantor,  I think this bug is causing trouble when applying global DCE with effects.

micahcantor:
Whoops, good catch, that could definitely cause issues.

Issue #1161, opened 2021-10-21
RFC:  ecmascript modules as primary build artifact


cdaringe:
## context

JSOO has long been the sole, defacto provider mapping OCaml input to JavaScript output. This project has existed for well over a decade, and has the primary function of taking users' ML and converting it to immediately invocable or effectful JavaScript. This continues to be a great accomplishment.

Over the passed few years, the ECMAScript spec has evolved and developed [a proper module system](https://tc39.es/ecma262/#sec-modules).

This module system development is of interest, because its stabilization enables a greater opportunity for ML to participate in the JavaScript ecosystem.

## proposal

Provide a first class compile target to ESM. 

## justification

- **JS bundle peformance**
  - status-quo: the current compiler (bundler) costs >75kb of javascript, just for hello world (with --profile=release). [ref project](https://github.com/patricoferris/jsoo-reload-example)
    - 75kb is a large baseline
    - the full runtime is compiled, even if only a subset is used
  - future: like `elm`'s compiler, the more `ocaml` features that user code consumes, the more of the runtime we can incrementally include in the output
- **Improved portability**
  - ESM output enables OCaml modules to be participate in JavaScript norms. Such norms include participation with common javascript tooling (browsers (e.g. `<script src="..." module></script>`), bundlers, analyzers, dead-code analyzers, minifiers, pretty printers--the works!)

## hypothetical

The following hypothetical cases may be completely bogus. Consider these my temporary, envisioned target state, even if they are not necessarily achievable.

### case - empty

Input:

```ml
(* main.ml *)
(* no content *)
```

Output:

```js
// main.js
// no content
```

No source, no dist! 75k worth of savings, vs status quo :)

### case - hello world

Input:

```ml
(* main.ml *)
let () = print_endline "Hello, world!"
```

Output (a):

```js
// main.js
console.log("Hello, world!")
```

Output (b):

```js
// main.js
import { print_endline } from "@jsoo/std";

print_endline("Hello, world!")
```

### case - reduce

Input:

```ml
(* main.ml *)
let rec sum = function
  | [] -> 0
  | head::tail -> head + (sum tail)
```

Output:

```js
// main.js
import { fn, match_case, match } from "@jsoo/fn";
import { pattern_length } from "@jsoo/lists";

export const sum = fn(
  match_case(match(pattern_length(0, true)), () => 0),
  match_case(match(pattern_length(1, false), [head, ...tail] => head + (sum tail)))
);
[...]

Issue #551, opened 2016-12-06
Node.js compatible modules


vouillon:
It would be great to be able to generate Node.js style modules.

One issue is that we currently access the elements of a module by index. It would be much nicer to use their names instead.

When we have debugging informations, we have the name of all elements except for values that are declared as `external` in the `.ml` file and exported as values in an `.mli` file, due to the way the corresponding code is currently generated by the OCaml compiler. This information could be passed from module to module as suggested in #550.

Alternatively, we could rely on `.cmi` files. But what if we do not have them all?

correnson:
Adding 2cents of my own to this issue : it would be great to (optionally) make `js_of_ocaml` generating one `.js` file per OCaml module, in order to allow for hot module replacement via web-pack.

PR #1529, opened 2023-11-15
Compiler: support for es6 import/export

### LINKED ISSUE #1396: [FEATURE REQUEST] ES6 module support (import/export)
Following #1391, we should add support for `import` and `export` in javascript.

This is a necessary step for #551 and #1161.

- [x] update the ast in javascript.ml
- [x] update parser

hhugo:
fix #1396 

- [x] fix variable scopes (during renaming)
- [ ] ~~allow import in jsoo stubs~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 * Compiler: change control-flow compilation strategy (#1496)
 * Compiler: Dead code elimination of unused references (#2076)
 * Compiler: reduce memory consumption (#1516)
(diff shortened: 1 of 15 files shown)


cometkim:
Is this not available yet? I tried `--enable=es6` but it had no effect.

hhugo:
The pr just add the support in the js parser/printer but the jsoo compiler doesn't use it. Maybe #551 #1161 is what you are after

PR #1535, opened 2023-11-27
Runtime: fix for OCaml.5.1.1

### LINKED ISSUE #1534: Support for OCaml 5.1.1
The upcoming version 5.1.1 of the compiler introduces a new primitive `caml_initialize_zstd` which is not yet supported `js_of_ocaml`.

This primitive aims to switch the compiler marshall library to compressed mode explicitly to circumvent the frailty of dead symbol elimination in native linker.

I have the impression (see https://github.com/ocsigen/js_of_ocaml/commit/bfaa9649e4dbff625eadf41bca6e2e2ed5e24dd1) that `js_of_ocaml` doesn't need such circonvolutions and can implement the new primitive with 
```js
 var zstd_initialize =  (function () {})
```

hhugo:
fix #1534


hhugo:
@octachron, the following two points are about the ocaml codebase:
- no sure why caml_compression_available exists but is not used/exposed, is it a leftover ? 
- the return value of caml_zstd_initialize needs to be fixed.

PR #1538, opened 2023-12-02
Update GitHub Actions

smorimoto:


hhugo:
Why is it not updated on windows?

smorimoto:
Windows does not yet support opam 2.2. (on setup-ocaml side)

smorimoto:
The compiler setup time is from 2 minutes to 1 minute.

PR #1539, opened 2023-12-05
Compiler: restore TCO of mutually recursive functions

hhugo:
@vouillon. @micahcantor https://github.com/ocsigen/js_of_ocaml/pull/1503 breaks tail call as mentioned by @vouillon

>  The global DCE does not preserve tail calls. We don't care normally since JavaScript does not support tail calls, but this makes a difference for the CPS transformation.

Except the `Generate_closure` pass does care about tail calls as it rewrite mutually recursive functions using a trampoline.

Do you see any issue with the change proposed here ?

### DIFF
--- a/compiler/lib/global_deadcode.ml
+++ b/compiler/lib/global_deadcode.ml
@@ -352,7 +352,14 @@ let zero prog sentinal live_table =
     let branch =
       (* Zero out return values in last instruction, otherwise do nothing. *)
       match block.branch with
-      | Return x, loc -> Return (zero_var x), loc
+      | Return x, loc ->
+          let tc =
+            (* We don't want to break tailcalls. *)
+            match List.last body with
+            | Some (Let (x', Apply _), _) when Code.Var.equal x' x -> true
+            | Some _ | None -> false
+          in
+          if tc then Return x, loc else Return (zero_var x), loc
       | Raise (_, _), _
       | Stop, _
       | Branch _, _
--- a/compiler/tests-compiler/direct_calls.ml
+++ b/compiler/tests-compiler/direct_calls.ml
@@ -57,7 +57,7 @@ let%expect_test "direct calls without --enable effects" =
   [%expect
     {|
     function test1(param){
-     function f(g, x){caml_call1(g, x); return;}
+     function f(g, x){return caml_call1(g, x);}
      var _d_ = 7;
      f(function(x){return x + 1 | 0;}, _d_);
      var _e_ = 4.;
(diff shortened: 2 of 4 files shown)


hhugo:
This issue was spotted by a single package (ezjsonm) during the opam release.
See https://github.com/ocaml/opam-repository/pull/24888

compiler/lib/global_deadcode.ml
```diff
@@ -352,7 +352,14 @@ let zero prog sentinal live_table =
     let branch =
       (* Zero out return values in last instruction, otherwise do nothing. *)
       match block.branch with
-      | Return x, loc -> Return (zero_var x), loc
+      | Return x, loc ->
+          let tc =
+            match List.last body with
+            | Some (Let (x', Apply _), _) when Code.Var.equal x' x -> true
+            | Some _ -> false
+            | None -> false
```
OlivierNicole, line 360:
```suggestion
            | Some (Let (x', Apply _), _) when Code.Var.equal x' x -> true
            | Some _ | None -> false
```

compiler/lib/global_deadcode.ml
```diff
@@ -352,7 +352,14 @@ let zero prog sentinal live_table =
     let branch =
       (* Zero out return values in last instruction, otherwise do nothing. *)
       match block.branch with
-      | Return x, loc -> Return (zero_var x), loc
+      | Return x, loc ->
```
OlivierNicole, line 355: Maybe the comment should be updated as well to reflect the tail call special case?

OlivierNicole (commented):
I don’t exactly understand how global DCE broke tail calls and why this fixes it, so I won’t comment beyond the surface level.

vouillon:
The call to `Generate_closure.f` should be after the CPS transformation.
I think the second commit is enough to fix the issue.

hhugo:
> The call to `Generate_closure.f` should be after the CPS transformation. I think the second commit is enough to fix the issue.

Can you explain the reasons ?
Moving it before seems to detect more deadcode (as show in https://github.com/ocsigen/js_of_ocaml/pull/1539/commits/0fe4aab7306c0ffeabfdfad43a921742f3a209bf). It this invalid ?

hhugo:
> > The call to `Generate_closure.f` should be after the CPS transformation. I think the second commit is enough to fix the issue.
> 
> Can you explain the reasons ? Moving it before seems to detect more deadcode (as show in [0fe4aab](https://github.com/ocsigen/js_of_ocaml/commit/0fe4aab7306c0ffeabfdfad43a921742f3a209bf)). It this invalid ?

Obviously, it removes the deadcode introduced by `generate_closure`.

vouillon:
> > The call to `Generate_closure.f` should be after the CPS transformation. I think the second commit is enough to fix the issue.
> 
> Can you explain the reasons ? Moving it before seems to detect more deadcode (as show in [0fe4aab](https://github.com/ocsigen/js_of_ocaml/commit/0fe4aab7306c0ffeabfdfad43a921742f3a209bf)). It this invalid ?

`Generate_closure.rewrite_mutable` is only needed for functions within loops. The partial CPS transformation will turn some loops into mutual function calls, making it unnecessary.

hhugo:
I've opened #1540 to get rid of rewrite_mutable. Is this the only reason ?

vouillon:
I think so. `rewrite_tc` is disabled in this case.

PR #1541, opened 2023-12-07
No longer rely on iife for scoping inside loops

### LINKED ISSUE #1540: [FEATURE REQUEST] rewrite closure compilation inside for loops
Use `let/const` instead of iife

hhugo:
Fix #1540

### DIFF
--- a/compiler/tests-compiler/gh1007.ml
+++ b/compiler/tests-compiler/gh1007.ml
@@ -158,166 +158,162 @@ let ()  = M.myfun M.x
      var x$0 = x;
      for(;;){
       if(! x$0) return 0;
-      var
-       next = x$0[1],
-       sort =
-         function(n, l){
-          if(2 === n){
-           if(l){
-            var match = l[2];
-            if(match){
-             var
-              tl = match[2],
-              x2 = match[1],
-              x1 = l[1],
-              s =
-                0 < caml_int_compare(x1, x2)
-                 ? [0, x2, [0, x1, 0]]
-                 : [0, x1, [0, x2, 0]];
-             return [0, s, tl];
-            }
-           }
-          }
-          else if(3 === n && l){
-           var _d_ = l[2];
-           if(_d_){
-            var match$2 = _d_[2];
-            if(match$2){
-             var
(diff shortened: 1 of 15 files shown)


compiler/tests-compiler/gh1007.ml
```diff
@@ -158,166 +158,162 @@ let ()  = M.myfun M.x
...
-                   : 0
-                     < caml_int_compare(x2$0, x3)
-                     ? [0, x2$0, [0, x3, [0, x1$0, 0]]]
-                     : [0, x3, [0, x2$0, [0, x1$0, 0]]];
-             return [0, s$0, tl$1];
-            }
-           }
-          }
+      var next = x$0[1];
+      function sort(n, l){
```
vouillon, line 162: It seems that the function order got reversed, which makes thinks harder to compare.

compiler/tests-compiler/mutable_closure.ml
```diff
@@ -120,47 +120,39 @@ let%expect_test _ =
...
+        var counter$1 = counter + 1 | 0;
+        return f$0(counter$1, _h_);
+       }
+       if(0 === n) return i$0;
+       var _i_ = n - 1 | 0;
+       if(counter >= 50) return caml_trampoline_return(f$0, [0, _i_]);
+       var counter$0 = counter + 1 | 0;
+       return f$0(counter$0, _i_);
+      }
+      function f(n){return caml_trampoline(f$0(0, n));}
```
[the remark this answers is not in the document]
hhugo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_let_const_class_or_function_declaration_in_strict_mode

compiler/tests-compiler/gh1007.ml
hhugo: The ordering is altered by compiler/lib/strongly_connected_components.ml.
It should fix it both on master and in this PR.

hhugo: now fixed

PR #1555, opened 2024-01-16
added 'ellipse' method to canvasRenderingContext2D

### LINKED ISSUE #1552: [FEATURE REQUEST] Ellipse method for canvas API
Hi! I am currently designing a creative coding library for OCaml, and I'd like it to be able to target the browser via the HTML canvas API. I've got a prototype working but I've hit a bit of a wall with rendering ellipses specifically: the ellipse method of the rendering context is missing. I would be happy to do the work to add it myself, but would probably need some guidance.

FayCarsons:
fix #1552

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## Features/Changes
 * Mics: fix support for OCaml 5.2
 * Compiler: no longer rely on IIFE for scoping variable inside loops
+* Lib: add ellipse to canvasRenderingContext2D (@FayCarsons, #1555)
 
 # 5.6.0 (2024-01-02) - Lille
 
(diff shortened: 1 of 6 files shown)


hhugo:
Thanks, I've added two commits to make the CI happy, I'll merge after it runs.

(I've noticed that ellipse was used in the `examples/hyperbolic/hypertree.ml` example and implemented as

```ocaml
let ellipse_arc c cx cy rx ry start fin clock_wise =
  c##save;
  c##translate (Js.float cx) (Js.float cy);
  c##scale (Js.float rx) (Js.float ry);
  c##arc
    (Js.float 0.)
    (Js.float 0.)
    (Js.float 1.)
    (Js.float start)
    (Js.float fin)
    clock_wise;
  c##restore
``` 

One could cleanup this example now that we have ellipse binding.

PR #1560, opened 2024-02-02
Explicitly void unused c variable

rickyvetter:
When importing jsoo into a project where the c compiler errors on warnings, this becomes problematic. Since void is a noop, this seems like a reasonable addition to upstream and make this file more portable.


OlivierNicole (approved):
Clearly correct.

hhugo:
thanks

Issue #1559, opened 2024-01-30
[BUG] regression causing some definitions in recursive functions to be out of scope


rickyvetter:
**Describe the bug**
I'm seeing an issue where a very specific set of side effects and recursion used together causes a runtime "value is not defined" error in strict mode and can't be executed in most environments.

A file I've been able to reproduce this with looks like:

```ocaml
let my_ref = ref 1

module _ : sig end = struct
  type 'a thing =
    | Thing of 'a
    | No

  let f2 t =
    match t with
    | Thing 1 -> true
    | Thing _ | No -> false
  ;;

  let length = function
    | Thing i -> i
    | No -> -1
  ;;

  let () =
    let init = Thing 1 in
    let nesting = 1 in
    let rec handle_state t =
      let this_will_be_undefined () = if f2 t then 1 else 2 in
      match length t with
      | 0 -> this_will_be_undefined ()
      | 1 -> if Stdlib.Int.equal nesting 0 then nesting else this_will_be_undefined ()
      | _ -> handle_state (Thing 0)
    in
    print_endline (Int.to_string (handle_state init))
  ;;

  let _ : _ thing = No
end

let () = my_ref := 2
```

<details><summary>Full expect test with output</summary>

```ocaml
let%expect_test "recursive function not defined" =
  with_temp_dir ~f:(fun () ->
    let name = "test.ml" in
    Filetype.write_file
      name
      {|
      let my_ref = ref 1

      module _ : sig end = struct
        type 'a thing =
          | Thing of 'a
          | No

        let f2 t =
          match t with
          | Thing 1 -> true
          | Thing _ | No -> false
        ;;

        let length = function
          | Thing i -> i
          | No -> -1
        ;;

        let () =
          let init = Thing 1 in
          let nesting = 1 in
          let rec handle_state t =
            let this_will_be_undefined () = if f2 t then 1 else 2 in
            match length t with
            | 0 -> this_will_be_undefined ()
            | 1 -> if Stdlib.Int.equal nesting 0 then nesting else this_will_be_undefined ()
            | _ -> handle_state (Thing 0)
          in
          print_endline (Int.to_string (handle_state init))
        ;;

        let _ : _ thing = No
      end

      let () = my_ref := 2
    |};
    let file = Filetype.ocaml_file_of_path name in
    let cmo = compile_ocaml_to_cmo file in
    let lib = compile_lib [ cmo ] "test.cmo" in
    let js = compile_cmo_to_javascript lib in
[...]

PR #1561, opened 2024-02-04
Fix 1559

hhugo:
Fixing #1559

Exiting a for loop too early might result in invalid scope for functions declared inside the for-loop.
A function declaration inside inside a block is scoped to that block (in strict mode) and won't be visible outside.

A quick fix is to simply disable the exit-loop-early optimization until one find a proper fix for it.

### DIFF
--- a/compiler/lib/freevars.ml
+++ b/compiler/lib/freevars.ml
@@ -209,8 +209,43 @@ let free_variables vars in_loop p =
 let f p =
   Code.invariant p;
   let t = Timer.make () in
+  let bound = Code.Var.ISet.empty () in
+  let visited = BitSet.create' p.free_pc in
+  let free_vars =
+    Code.fold_closures_innermost_first
+      p
+      (fun _name_opt params (pc, args) acc ->
+        let free = ref Var.Set.empty in
+        let using x =
+          if Code.Var.ISet.mem bound x then () else free := Var.Set.add x !free
+        in
+        let rec traverse pc =
+          if not (BitSet.mem visited pc)
+          then (
+            BitSet.set visited pc;
+            let block = Addr.Map.find pc p.blocks in
+            List.iter params ~f:(fun x -> Code.Var.ISet.add bound x);
+            iter_block_bound_vars (fun x -> Code.Var.ISet.add bound x) block;
+            List.iter args ~f:using;
+            iter_block_free_vars using block;
+            List.iter block.body ~f:(function
+                | Let (_, Closure (_, (pc_clo, _))), _ ->
+                    Code.Var.Set.iter using (Code.Addr.Map.find pc_clo acc)
+                | _ -> ());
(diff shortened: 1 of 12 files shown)


compiler/lib/structure.ml
```diff
@@ -188,6 +188,44 @@ let mark_loops g =
     g.preds;
   in_loop
 
+let rec iter_used_var blocks f pc =
+  let block = Addr.Map.find pc blocks in
+  Freevars.iter_block_free_vars f block;
+  List.iter block.body ~f:(fun (i, _) ->
+      match i with
+      | Let (_, Closure (_, (pc', _))) -> iter_used_var blocks f pc'
```
hhugo, line 196: This might cause problems on code rewritten in cps (with effects)

hhugo:
@vouillon, do you have time to comment on the current approach ?

rickyvetter:
I tested 3556f362d5ad9a871dfed8306b376b248415bcaf locally and it fixes the issue with the actual code as well as the example I posted. Thank you! Happy to test other revisions as well if there are other ideas for solving.

compiler/lib/structure.ml
```diff
@@ -188,6 +188,44 @@ let mark_loops g =
     g.preds;
   in_loop
 
+let rec iter_used_var blocks f pc =
+  let block = Addr.Map.find pc blocks in
+  Freevars.iter_block_free_vars f block;
```
hhugo, line 193: we also need to traverse all children of pc.

hhugo:
I'd like to have  a single pass to compute all free variables of closures.

compiler/lib/structure.ml
hhugo: this is now stale

compiler/lib/structure.ml
hhugo: this is slate

hhugo:
@rickyvetter, I've pushed an alternative fix.

compiler/lib/freevars.ml
```diff
@@ -207,10 +207,45 @@ let free_variables vars in_loop p =
...
+          if Code.Var.ISet.mem bound x then () else free := Var.Set.add x !free
+        in
+        let rec traverse pc =
+          if not (BitSet.mem visited pc)
+          then (
+            BitSet.set visited pc;
+            let block = Addr.Map.find pc p.blocks in
+            List.iter params ~f:(fun x -> Code.Var.ISet.add bound x);
+            iter_block_bound_vars (fun x -> Code.Var.ISet.add bound x) block;
+            List.iter args ~f:using;
```
vouillon, line 229: Shouldn't we move these two lines outside of function `traverse`?
```ocaml
            List.iter params ~f:(fun x -> Code.Var.ISet.add bound x);
            List.iter args ~f:using;
```

hhugo: Indeed, fixed in #1565

PR #1565, opened 2024-02-12
Compiler: fix closure deps

hhugo:


compiler/lib/generate.ml
```diff
@@ -1062,6 +1062,7 @@ let rec translate_expr ctx queue loc x e level : _ * J.statement_list =
       (Mlvalue.Block.field cx n, or_p px mutable_p, queue), []
   | Closure (args, ((pc, _) as cont)) ->
       let loc = source_location ctx ~force:After (After pc) in
+      let fv = Addr.Map.find pc ctx.freevars in
```
vouillon, line 1065: Does `ctx.freevars` list all free variables, or just the ones bound inside a loop?

compiler/lib/generate.ml
```diff
@@ -1077,7 +1078,7 @@ let rec translate_expr ctx queue loc x e level : _ * J.statement_list =
         | _ -> clo
       in
       let clo = J.EFun (None, J.fun_ (List.map args ~f:(fun v -> J.V v)) clo loc) in
-      (clo, flush_p, queue), []
+      (clo, (fst mutable_p, fv), queue), []
```
vouillon, line 1081: Wouldn't this work?
```suggestion
      (clo, (fst const_p, fv), queue), []
```

hhugo: I think it does, I committed the wrong version of my experiments

hhugo:
This PR is incorrect, we need to flush all free variables of funcitons from the queue

PR #1568, opened 2024-02-23
Compiler: only flush the necessary env for closure 

hhugo:
following the previous broken attempt in #1565.

Also
- It revert the fix in https://github.com/ocsigen/js_of_ocaml/pull/1561 which is no longer needed. @rickyvetter, can you confirm the PR still works for you ?


rickyvetter:
This looks like it also works. Thanks!

PR #1575, opened 2024-02-29
Compiler: remove last argument of Pushtrap

hhugo:


OlivierNicole:
What was the motivation for the change?

hhugo:
Simplification of the code by removing an invariant from the IR that is annoying to maintain.
I think recomputing the information when needed is not too costly.

OlivierNicole:
Okay, thanks.

PR #1576, opened 2024-03-01
Lib: fix paragraph construction and coercion

gpetiot:
Some annoying bug I found while trying to coerce some elt with `Html.CoerceTo.p`.


hhugo:
Thanks

PR #1577, opened 2024-03-01
Don't hold on to channels after they've been closed

tkluck-janestreet:
In my use case, this issue is responsible for O(100MiB) of memory usage, because I'm reading one file per named time zone, and each file has a 65KiB buffer that doesn't get released.

The file io.js provides the same runtime as the C runtime from [2]. There's a global array of all existing channels, and this never gets released even if a channel is closed. In the C runtime, this does eventually happen in a call to [caml_finalize_channel] which is called by the GC, but I don't think we have that ability in js_of_ocaml.

This feature makes [chanid] an object rather than an integer, so we can let the javascript garbage collector take care of it.

The object [caml_ml_channels] continues to exist, but its only remaining function is to provide "override" functionality in the way that ppx_expect needs it.

Status:
 - [ ] get feedback from js_of_ocaml developers
 - [ ] when everyone is happy with the direction, translate to an older/more compatible version of javascript (avoiding `class`)
 - [ ] merge

Some useful references:
 - [standard library calling the runtime][1]
 - [the c runtime][2]
 - [ppx_expect calling into our runtime][3]

[1]: https://v2.ocaml.org/releases/5.1/api/Stdlib.html#2_Outputfunctionsonstandardoutput
[2]: https://github.com/ocaml/ocaml/blob/trunk/runtime/io.c#L163
[3]: https://github.com/janestreet/ppx_expect/blob/master/runtime/runtime.js

### DIFF
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -87,31 +87,77 @@ function caml_sys_open (name, flags, _perms) {
 
 // ocaml Channels
 
-//Provides: caml_ml_set_channel_name
-//Requires: caml_ml_channels
-function caml_ml_set_channel_name(chanid, name) {
-  var chan = caml_ml_channels[chanid];
-  chan.name = name;
-  return 0;
-}
+/* The [ChanId] class and the [caml_ml_channels] object are for backwards compatibility
+   towards ppx_expect. The [caml_ml_channels] used to map from an integer id to a
+   channel descriptor, but that leaks memory. (In particular, it leaks the buffer
+   which can be quite large.) We now replaced the integer id by a [ChanId] object,
+   which, after garbage collection, means we can release its resources.
+
+   So the [channel] object can be garbage collected like anything else. The only
+   small hiccup is that ppx_expect wants to be able to "shadow"/"override"
+   channels by manipulating the [caml_ml_channels] array. Fortunately, after the
+   channel is closed we don't need to allow for that anymore. So we keep channels in
+   the global object until then.
+
+   A future version could provide an actual shadowing API for ppx_expect, and get rid
+   of the last remaining potential memory leak (a channel being discarded before
+   closing it) entirely.
+*/
 
 //Provides: caml_ml_channels
-var caml_ml_channels = new Array();
+var caml_ml_channels = new Object();
+//Provides: caml_ml_open_out_channels
+var caml_ml_open_out_channels = new globalThis.Set();
+
+//Provides: ChanId
+//Requires: caml_ml_channels, caml_ml_open_out_channels
+var ChanId = class {
+  constructor(channel) {
+    this.channel = channel;
+    caml_ml_channels[this] = this;
+    if(channel.out) caml_ml_open_out_channels.add(this);
+  }
+  close() {
+    delete caml_ml_channels[this];
+    caml_ml_open_out_channels.delete(this);
+  }
+  toString() {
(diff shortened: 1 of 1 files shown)


hhugo:
The tests are currently broken, is this expected ? do we need a modified version of `ppx_expect` ?

There are at least two other packages using similar tricks
- https://github.com/mirage/alcotest (https://github.com/mirage/alcotest/blob/main/src/alcotest/runtime.js)
- https://github.com/hhugo/capture-output (https://github.com/hhugo/capture-output/blob/main/src/runtime.js)

I'm not convinced by the override logic. We should be able to solve the leak using a WeakMap instead of an array for `caml_ml_channels`. We just need to track the opened channels somehow.

hhugo:
I've worked on an alternative in https://github.com/ocsigen/js_of_ocaml/pull/1578. Let me know what you think.

hhugo:
The CI failure seems to be due to a piece of code in `parsing.js` assuming that a channel is an integer.

runtime/io.js
```diff
@@ -171,68 +216,64 @@ function caml_ml_open_descriptor_out_with_flags(fd, flags){
...
+//Requires: caml_ml_flush
 //Requires: caml_sys_close
 function caml_ml_close_channel (chanid) {
-  var chan = caml_ml_channels[chanid];
+  var chan = chanid.get();
+  if (!chan.opened) { return 0; }
   chan.opened = false;
-  caml_sys_close(chan.fd)
+  caml_sys_close(chan.fd);
+  chanid.close();
```
hhugo, line 250: What happen if you close an "overridden" channel twice ?
Don't you end up closing both the override and the original channel ?

tkluck-janestreet:
I'll look into the test failures; it's embarrassing but I'm having a hard time running this in its own tree because "vanilla" dune is being shadowed by a company-specific build. I'm sure I'll be able to sort that out.

hhugo:
The CI comes from `parsing.js`, `caml_ml_output` take a "chanid" but is called with `2` in the following code snippet.   
```js
  function log(x) {
    var s = caml_string_of_jsbytes(x + "\n");
    caml_ml_output(2, s, 0, caml_ml_string_length(s));
```

hhugo:
> Alternatively, you could swap out the fd; that's what the C runtime does. But that would need moving the offset property to the file descriptor object as well.

Can you elaborate a bit why one need to move the offset ?

tkluck-janestreet:
> Can you elaborate a bit why one need to move the offset ?

In the C runtime, `channel->offset` tracks the underlying file descriptors offset: it's initialized using
```c
  channel->offset = lseek(fd, 0, SEEK_CUR);
```
and from then on we keep it in sync using
```c
written = caml_write_fd(channel->fd, channel->flags,
                            channel->buff, towrite);
channel->offset += written;
```
In the JS runtime, `chan.offset` is actually the source of truth, and e.g. the `write` function takes the offset as a parameter:
```js
chan.file.write(chan.offset, chan.buffer, 0, chan.buffer_curr);
chan.offset += chan.buffer_curr;
```
To be honest, the C runtime seems a bit brittle when it comes to swapping out file descriptors, but evidently it works in practice.

In the JS runtime we'd have to override the `offset` together with the `fd` to get the correct behavior. Alternatively, we'd have to move `offset` to `chan.file`.

hhugo:
replaced by #1578

PR #1578, opened 2024-03-03
Alternative to #1577

hhugo:
Cons: ~~breaks 2/3 packages (ppx_expect, alcotest)~~
Pro: no override logic

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 
 * Compiler: fix variable renaming for property binding and assignment target
 * Compiler: fix separate compilation of toplevels (broken since 5.7.0)
+* Runtime: no longer leak channels
 
 # 5.7.2 (2024-04-03) - Lille
 
--- a/runtime/io.js
+++ b/runtime/io.js
@@ -100,45 +100,52 @@ function caml_ml_set_channel_name(chanid, name) {
   var chan = caml_ml_channels[chanid];
   chan.name = name;
   return 0;
-                             }
+}
 
-  //Provides: caml_ml_channels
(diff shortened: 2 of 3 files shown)


hhugo:
@tkluck-janestreet, would you have time to look at this PR. The latest version no longer break compatibility with ppx_expect thanks to js `Proxy`. Can you confirm it works for you ?

tkluck-janestreet:
Hi @hhugo it's looking good so far. I'm running it on our continuous integration and will report the results here on Monday. I'd be surprised if it's going to uncover any problems.

tkluck-janestreet:
I was too optimistic; unfortunately using `Symbol`s as keys to a `WeakMap` is a [relatively new feature](https://github.com/nodejs/node/issues/49135) and it doesn't work on the version of node that I'm using.

Not saying I wouldn't be happy if you merged this. Do you have a policy about which javascript versions are supported?

hhugo:
> I was too optimistic; unfortunately using `Symbol`s as keys to a `WeakMap` is a [relatively new feature](https://github.com/nodejs/node/issues/49135) and it doesn't work on the version of node that I'm using.
> 
> Not saying I wouldn't be happy if you merged this. Do you have a policy about which javascript versions are supported?

No policy, I just try to be reasonable.
So it seems that this doesn't work with node 18 which is still in maintenance mode until April 2025. I think it would be unfortunate to break such setup. Let's try something else, this PR was mostly a hack to avoid breaking 2 packages after all.

I think it's a better long term solution to break and fix ppx_expect and alcotest and come up with a better abstraction/api/helper inside jsoo.

PR #1581, opened 2024-03-04
Runtime: reduce memory leak caused by channels

hhugo:


tkluck-janestreet:
This was my first version too 😇

I'm looking back in review history and I think we ended up with #1577 because it separates the code we want to keep (with references being traced by the GC and nothing really depends on closing fds) from the code we eventually don't (overrides that can't be traced for ppx_expect compatibility).

But that depends on a possibly faulty model for how you'd like this code to develop. If you'd rather go with #1581 and/or with moving overrides to the fd, then I think that's all a very reasonable way forward.

hhugo:
Merging this one first, giving more time to properly fix the leak.

tkluck-janestreet:
Thanks @hhugo!

Issue #357, opened 2015-08-18
Better inlining ?


dbuenzli:
Somehow it is annoying that self-recursive functions that go through a simple error monad do blow the stack. See for example here https://github.com/dbuenzli/otfm/commit/4ea6a34b884c5a680fa4333b79f6fd8117a0b8c0

hhugo:
https://github.com/ocsigen/js_of_ocaml/issues/154

hhugo:
@dbuenzli, 9 years later .. would you want to give #1584 a try ?

dbuenzli:
I'm afraid I don't have a test case at hand. But thanks for the work !

PR #1584, opened 2024-03-10
Compiler: Rework inlining 

hhugo:
Should fix https://github.com/ocsigen/js_of_ocaml/issues/357

@vouillon, I've rebased an old branch of yours (`aggressive-inlining`) and changed the logic a bit so that
- it can replace the previous "simple" optimization (inlining closure of "size" one).
- it only duplicates other small closures if they call one of their arguments in tail position (this is an attempt at addressing #357)

### DIFF
--- a/compiler/lib/inline.ml
+++ b/compiler/lib/inline.ml
@@ -26,21 +26,82 @@ type prop =
   ; optimizable : bool
   }
 
-let optimizable blocks pc _ =
+type closure_info =
+  { cl_params : Var.t list
+  ; cl_cont : int * Var.t list
+  ; cl_prop : prop
+  ; cl_simpl : (Var.Set.t * int Var.Map.t * bool * Var.Set.t) option
+  }
+
+let block_size { branch; body; _ } =
+  List.length body
+  +
+  match fst branch with
+  | Cond _ -> 2
+  | Switch (_, a1) -> Array.length a1
+  | _ -> 0
+
+let simple_function blocks size name params pc =
+  let bound_vars =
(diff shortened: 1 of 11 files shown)


hhugo:
@pmwhite, @TyOverby, you might want to try this PR as it could affect performance and size of artifacts.

compiler/lib/inline.ml
```diff
@@ -26,21 +26,72 @@ type prop =
...
+        List.iter block.body ~f:(fun (i, _loc) ->
+            match i with
+            | Let (_, Closure _) -> raise Exit
+            | _ -> ());
+        Freevars.iter_block_bound_vars
+          (fun x -> bound_vars := Var.Set.add x !bound_vars)
+          block;
+        Freevars.iter_block_free_vars
+          (fun x ->
+            if not (Var.Set.mem x !bound_vars) then raise Exit;
```
hhugo, line 78: @vouillon, could you explain why we have this restriction ?

vouillon: To inline a function, we have to make sure that all its free variables are still in scope. This is a simple way to ensure that.

hhugo: Thinking more about it, I fail to see how free variables could be out of scope. Could you elaborate a bit more ?

Removing the restriction would require that we keep live_vars up to date when duplicate free vars usage.

vouillon: Oh, indeed this is probably not necessary since we are only inlining functions that are defined in the same scope or in an outer scope.

PR #1590, opened 2024-03-26
Runtime: abort instead of exit on unimplemented js primitives

### LINKED ISSUE #677: [FEATURE REQUEST] Server-side uses of unimplemented primitives are difficult to track down
I'm opening this a bit late and I may have forgotten part of the context. Please forgive me if this is the case.

Currently an unimplemented javascript primitive triggers an `exit(1)` (server-side iirc). This makes it difficult to find which code path is faulty. Mixing server-side and client-side should not happen but mistakes do.

I've used the following change to `generate_stubs.sh` locally:

> -echo "#define D(f) void f () { fprintf(stderr, \"Unimplemented Javascript primitive %s!\\n\", #f); exit(1); }";
> +echo "#define D(f) void f () { fprintf(stderr, \"Unimplemented Javascript primitive %s!\\n\", #f); caml_failwith(\"Unimplemented Javascript primitive: \" # f); exit(1); }";

I wasn't able to get a full backtrace but at least I was able to ocamldebug it.

Maybe a dedicated exception could be useful, especially because Invalid_argument might be caught and silenced.

hhugo:
fix #677

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
 * Compiler: codegen: more specialization for %int_add, %int_sub
 * Compiler: recognize and optimize String.concat
 * Compiler: more inlining - duplicate small function.
+* Runtime: abort instead of exit when calling unimplemented
+  js primitives in bytecode/native. It should help if one tries
+  to understand the source of the call with gdb (see #677)
 
 ## Bug fixes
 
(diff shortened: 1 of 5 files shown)


dbuenzli:
I think it would be better to use `caml_fatal_error` and eschew the `fprintf`. This has would hook into the support provided by the runtime system (notably you can easily define a `caml_fatal_error_hook` to do what you want with the error reporting). See the discussions [here](https://github.com/ocaml/ocaml/pull/8630).

hhugo:
Thanks, changed in https://github.com/ocsigen/js_of_ocaml/commit/28743eeb8a6428e9d82f4051512b9fb708e5af29

dbuenzli:
Thank you!

Issue #550, opened 2016-12-06
Separate compilation: generate and use cross module information regarding function arity


vouillon:
We could greatly improve the code generated when compiling `cmo` files if we knew the arity of the functions exported by each module.
This information could be stored in a separate file for each module compiled when the `--dynlink` option is used, or when compiling `cmo` and `cma` files.

hhugo:
:+1:
Open question: should we share this format with bucklescript
https://github.com/bloomberg/bucklescript/blob/master/jscomp/core/js_cmj_format.mli

vouillon:
There is no reason to share the same format, as the generated codes are not compatible.

hhugo:
@vouillon, any idea on how to distinguish immutable (module) block  from other possibly mutable ones? We could limit the optimisation to "compilation unit" modules and ignore sub-module but it would be a bit sad

vouillon:
One can get some type information from the debugging events.

hhugo:
I took a look at debug events. I think there are not enough.
We can recover some module info when the module appears on the `ce_stack` and the corresponding `Env_module` appears in the env summary but
- submodules that are constrained/coerced by an mli don't seem to appear on `ce_stack`
- toplevel modules / compilation units don't appear on ce_stack. (Not a real issue as we can look for SETGLOBAL to spot them).

I would be convenient to retrieve information about Immutable blocks.
I wonder if we can store that information inside debug events or it we need some other mechanism.

vouillon:
We can get some information from the cmi files. But I'm not sure we want to go this way. (One may have to expand some module types, and thus load cmi files corresponding to other toplevel modules.)

The information in `ce_stack` could still be useful internally to improve our analyses.

Maybe the module `Bytegen` could emit some information about block mutability, in a way similar to how debug events are produced. There is some more information in the lambda code which is lost when generating the bytecode. It might make sense to preserve it as well. For instance, int32/in64 comparisons are translated into generic comparisons. Or we loose any information on the kind and layout of bigarrays.

vouillon:
@hhugo See ocaml/ocaml#13438

PR #1594, opened 2024-04-08
PoC: Propagate arity between compilation units  

hhugo:
PoC for #550
 
- Store immutability property of blocks 
- Try to recover some immutability info from debug events. (currently assuming modules are immutable, need to check what happens with recursive modules
- Use this info to compute an approximated shape for registered globals.
- Use shape info to compute function arity across compilation units (this is currently only done when compiling cma files as we know the shape of previously compiled units inside the library)

TODO:
- [ ] check the behavior with recursive modules
- [ ] update global_flow ?
- [ ] better mechanism for shape propagation ?


hhugo:
@vouillon, feel free to comment, review and even commit the branch.

compiler/lib/specialize.ml
```diff
@@ -23,29 +23,36 @@ open Flow
...
-        match u, v with
-        | Some n, Some m when n = m -> u
-        | _ -> None)
-      x
+    match Shape.get x with
+    | Some shape -> (
+        match shape with
+        | Function { arity; _ } -> Some arity
+        | Block _ | Bot _ -> None)
+    | None ->
```
vouillon, line 31: That should be moved a few lines below, to take advantage of the flow analysis:
```ocaml
get_approx info (fun x -> match Shape.get x with ...)
```

compiler/lib/flow.ml
```diff
@@ -392,6 +397,29 @@ let direct_approx (info : Info.t) x =
         y
   | _ -> None
 
+let rec the_shape_of info x =
+  get_approx
+    info
+    (fun x ->
+      if Var.ISet.mem info.info_possibly_mutable x
```
vouillon, line 404: You don't need this here.
`possibly_mutable` means that the value might be modified, but you can still rely on the shape

compiler/lib/flow.ml
```diff
@@ -392,6 +397,29 @@ let direct_approx (info : Info.t) x =
...
+      then Shape.Bot "possibly_mutable"
+      else
+        match info.info_defs.(Var.idx x) with
+        | Expr (Block (_, a, _, true)) ->
```
vouillon, line 408: I think that could be the following, though it probably does not make the result any better: since the block escapes, we know that it is not possibly mutable only since it is immutable.
```suggestion
        | Expr (Block (_, a, _, _)) when not (Var.ISet.mem info.info_possibly_mutable x) ->
```

compiler/lib/flow.ml
```diff
@@ -392,6 +397,29 @@ let direct_approx (info : Info.t) x =
...
   | _ -> None
 
+let rec the_shape_of info x =
+  get_approx
+    info
+    (fun x ->
+      if Var.ISet.mem info.info_possibly_mutable x
+      then Shape.Bot "possibly_mutable"
+      else
+        match info.info_defs.(Var.idx x) with
```
vouillon, line 407: You should use the information from `Shape.get x` as well.

compiler/lib/shape.ml
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+open! Stdlib
+
+type t =
+  | Bot of string
```
vouillon, line 23: We usually use `Top` for unknown information.

compiler/lib/specialize.ml
hhugo: This doesn't work, there are no `known_origins` computed in flow and `get_approx` simply returns `top`

compiler/lib/flow.ml
hhugo: It does improve the result

hhugo: see https://github.com/ocsigen/js_of_ocaml/pull/1594/commits/c7c009b1198bb85658300a87057ba25e818b44b4

compiler/lib/flow.ml
hhugo: Looking again, I think we can't rely on the shape. The arity of a function could change depending on how it's constructed.

vouillon: But we know how it's constructed. Otherwise `info_maybe_unknown` would be `true` and we would not arrive here.

hhugo: It seems that my change in block_escape was incorrect.
I've reverted the change and my experiment now looks good with your suggestion.

compiler/lib/flow.ml
hhugo: It no longer improve the result after reverting the change in `block_escape`

compiler/lib/specialize.ml
vouillon: I have changed the flow analysis to update `known_origins`  and `maybe_unknown` so that it works.

compiler/lib/flow.ml
vouillon: Indeed, it seemed strange to me that it would make any difference.

compiler/lib/specialize.ml
hhugo: great

hhugo:
It increases to the time to build the toplevel in whole program compilation by ~10%. But we don't benefit from shapes in such case.

vouillon:
The following command spends a lot of time in `Flow.the_shape_of` (called from `Specialize.function_arity`).
```
js_of_ocaml `which ocamlc.byte` 
```

hhugo:
@vouillon, I've rebased and updated the PR.

The compilation time of partial-render-table increases by 20% because it no longer stabilizes after 7 rounds, it needs 13 but stops after 10 (because of `--opt 2`). Additional rounds are a sequence of optcall and inline optimizations. I haven't looked at it more than that.

I think the PR is ready for another round of review.

TyOverby:
@hhugo is this only for whole-program compilation?

hhugo:
> @hhugo is this only for whole-program compilation?

This is mostly for separate compilation but it improves whole_program compilation as well it seems.

hhugo:
The regression for partial-render-table is now fixed

hhugo:
@TyOverby, @rickyvetter, Would you mind testing this PR  as well ?

hhugo:
> @TyOverby, @rickyvetter, Would you mind testing this PR as well ?

@TyOverby, @rickyvetter, ping

rickyvetter:
We've imported this feature but haven't yet been able to import the build rules changes. Running js_of_ocaml with the same invocation and no build rules changes, it looks like we see some minor changes to both bundle size (+/-1%) and to runtime perf (+/-10%), but nothing that stands out as notable. We're working on getting the build rules imported and will report back on the changes there.

hhugo:
> You don't need this here. `possibly_mutable` means that the value might be modified, but you can still rely on the shape

The build rules change is not finished as far as I remember and would only useful for separate compilation.
For now, I'm more interested in whole program compilation and making sure the PR is not introducing slowdown.

Without touching your build rules, the optimization should trigger inside individual libraries (x-module propagation between modules of the same library).

compiler/tests-full/shapes.cma.expected.js
```diff
@@ -72,14 +77,14 @@
...
-             (_a_, param){
+             (a, param){
                /*<<compiler/tests-full/m3.ml:1:59>>*/ return Shapes_M2[1].call
                       (null, 0) /*<<compiler/tests-full/m3.ml:1:66>>*/ ;
              };
    }
    var
-    x =  /*<<compiler/tests-full/m3.ml:4:8>>*/ f(0)(0, 0),
-    Shapes_M3 =  /*<<compiler/tests-full/m3.ml:4:18>>*/ [0, f, x];
+    x =  /*<<compiler/tests-full/m3.ml:3:8>>*/ caml_call2(f(0), 0, 0),
```
vouillon, line 86: Why is this not optimized?

compiler/lib-wasm/link.ml
```diff
@@ -345,6 +345,7 @@ let info_from_sexp info =
   let build_info =
     info |> member "build_info" |> mandatory (single Build_info.from_sexp)
   in
+
```
vouillon:
```suggestion
```

compiler/lib/shape.ml
```diff
@@ -0,0 +1,183 @@
...
+  | Top, (Function _ | Block _) | Function _, (Top | Block _) | Block _, (Top | Function _)
+    -> false
+
+let rec merge (u : t) (v : t) =
+  match u, v with
+  | ( Function { arity = a1; pure = p1; res = r1 }
+    , Function { arity = a2; pure = p2; res = r2 } ) ->
+      if a1 = a2 then Function { arity = a1; pure = p1 && p2; res = merge r1 r2 } else Top
+  | Block b1, Block b2 ->
+      if List.length b1 = List.length b2 then Block (List.map2 b1 b2 ~f:merge) else Top
```
vouillon:
```suggestion
      if List.compare_lengths b1 b2 = 0 then Block (List.map2 b1 b2 ~f:merge) else Top
```

vouillon (changes requested):

compiler/tests-full/shapes.cma.expected.js
vouillon: Oh, this is because shapes are not computed recursively when specializing function calls.

hhugo: Yes, it was somehow too costly for very little gain.

PR #1601, opened 2024-04-15
Don't leak channels.

hhugo:
Alternative to #1578 and #1577.
This PR requires ppx_expect and alcotest to be changed.

Because the jsoo codebase rely on ppx_expect for its tests, we have a Chicken-and-egg situation.

I can make a minor release of jsoo that would include new runtime function while keeping the existing storage mechanism.
Once jsoo is released, ppx_expect can rely on the new runtime functions.
After ppx_expect is released, we can merge this PR. 

@tkluck-janestreet, how does that sound ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 ## Bug fixes
 * Fix small bug in global data flow analysis (#1768)
+* Runtime: no longer leak channels
 
 # 5.9.1 (02-12-2024) - Lille
 
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -181,7 +181,7 @@ let run
     | None -> None
     | Some file ->
         if not (Sys.file_exists file)
-        then failwith (Printf.sprintf "export file %S does not exists" file);
+        then failwith (Printf.sprintf "export file %S does not exist" file);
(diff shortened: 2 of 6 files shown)


TyOverby:
What are the required ppx_expect changes?

hhugo:
> What are the required ppx_expect changes?

https://github.com/janestreet/ppx_expect/pull/54

dkalinichenko-js:
Hi, you should be able to merge this after updating ppx_expect to v0.17.1.

hhugo:
> Hi, you should be able to merge this after updating ppx_expect to v0.17.1. (see [ocaml/opam-repository#26695](https://github.com/ocaml/opam-repository/pull/26695))

@dkalinichenko-js, I probably won't be able to upgrade to v0.17.1 because it will probably still be broken on windows. See  https://github.com/janestreet/ppx_expect/pull/57

dkalinichenko-js:
I see. I'll submit another release of `ppx_expect` with the Windows fix Monday.

hhugo:
@dkalinichenko-js, the new blocker is that the new release of ppx_expect is not available for ocaml 4.14, preventing us to run our expect_tests with OCaml 4.14

hhugo:
https://github.com/ocaml/opam-repository/pull/27065

PR #1603, opened 2024-04-18
Compiler: track block mutability

hhugo:


OlivierNicole:
That looks great! Does it improve e.g. on deadcode elimination? (I assume that modules are compiled to immutable blocks)

hhugo:
This was implemented for https://github.com/ocsigen/js_of_ocaml/pull/1594. I'm not sure it improves dead code. We still miss information about whether a block is a module or not.

OlivierNicole:
Ah, yes, I had forgotten that issue.

PR #1607, opened 2024-04-27
Fix lint errors

smorimoto:


dune-project
```diff
@@ -96,7 +96,7 @@
   (graphics :with-test)
   (num :with-test)
   (ppx_expect (and (>= v0.14.2) :with-test))
-  (ppxlib (and (>= 0.15) :with-test))
```
hhugo, line 99: Why this change ?

smorimoto: @talex5 @moyodiallo How can we find out why opam-dune-lint is complaining about this?

moyodiallo: By looking further, `ppxlib` is used by a private library `ppx_optcomp_light` in js_of_ocaml/compiler/ppx/dune.  As `js_of_ocaml_compiler` a public library in js_of_ocaml/compiler/lib/dune uses `ppx_optcomp_light` that makes `ppxlib` a build lib. This is why opam-dune-lint is complaining.

moyodiallo: `opam-dune-lint` is not supposed to complains because `js_of_ocaml_compiler` is a package/public-library it's where `ppxlib` is treated.

I'll try to fix it, Thanks.

hhugo: @smorimoto, I'm tempted to disable the checks in the CI for now, until the fix is released.
Also, I don't understand what makes `dune fmt` now complain. Is it just bumping the compiler to 5.1 ? or is it unrelated and just caused by a new version of dune ?

moyodiallo: It's because of the new version of dune which handle utf8 since 3.15.0.

https://github.com/ocaml/dune/releases/tag/3.15.0 
 - Dune file formatting: output utf8 if input is correctly encoded (https://github.com/ocaml/dune/pull/10113.

moyodiallo:
The issue about `ppxlib` turns out to be correct. The command `opam install js_of_ocaml-toplevel` works because `ppxlib` depends on `js_of_ocaml-compiler` without `:with-test`, in that case `ppxlib` is transitively resolved.

`opam-dune-lint` complains about the `ppxlib` coming from `compiler/ppx` which is not a test and `js_of_ocaml-toplevel` relies on it during the build, so the linting works because `opam-dune-lint` treats individually the public libraries.
```
$ cd js_of_ocaml
$ opam-dune-lint
js_of_ocaml-compiler.opam: OK
js_of_ocaml-lwt.opam: OK
js_of_ocaml-ppx.opam: OK
js_of_ocaml-ppx_deriving_json.opam: OK
js_of_ocaml-toplevel.opam: changes needed:
  "ppxlib"                                 [from compiler/ppx] (remove {with-test})
js_of_ocaml-tyxml.opam: OK
js_of_ocaml.opam: OK
Write changes? [y]
``` 

I hope this explanation is more clearer.

hhugo:
Sounds good, #1646 has been merged

PR #1609, opened 2024-05-02
Add Typed_array.Bytes module

johnridesabike:
This adds an API in `Typed_array` for converting uint8Array values into bytes. Because the runtime uses uint8Array internally to represent bytes, this doesn't need to copy anything. Modifying the resulting bytes will also modify the original uint8Array.

### DIFF
--- a/lib/js_of_ocaml/js_of_ocaml_stubs.c
+++ b/lib/js_of_ocaml/js_of_ocaml_stubs.c
@@ -1,5 +1,9 @@
 #include <caml/misc.h>
 
+void caml_bytes_of_array () {
+  caml_fatal_error("Unimplemented Javascript primitive caml_bytes_of_array!");
+}
+
 void caml_js_error_of_exception () {
   caml_fatal_error("Unimplemented Javascript primitive caml_js_error_of_exception!");
 }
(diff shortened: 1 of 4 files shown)


lib/js_of_ocaml/typed_array.mli
```diff
@@ -254,3 +254,11 @@ module String : sig
 
   val of_uint8Array : uint8Array Js.t -> string
 end
+
+module Bytes : sig
```
hhugo, line 258: Maybe add documentations explaining that theses functions do not copy their inputs (most of the time) and that modifying the input/output might modify the corresponding output/input

johnridesabike: I just added a short paragraph that explains the behavior with a warning.

lib/js_of_ocaml/js_of_ocaml_stubs.c
```diff
@@ -1,5 +1,9 @@
 #include <caml/misc.h>
 
+void caml_bytes_of_array () {
+  caml_fatal_error("Unimplemented Javascript primitive caml_bytes_of_array!");
+}
+
```
OlivierNicole, line 6: I’m a bit lost, can someone explain to me why using the API will not just raise this error?

vouillon: This is just to keep the Ocaml bytecode compiler happy. This code is executed instead:
https://github.com/ocsigen/js_of_ocaml/blob/a03f32c5a462b21ab84840fc4858632e1540004f/runtime/mlBytes.js#L512-L517

OlivierNicole: Ah, I see, thanks!

PR #1611, opened 2024-05-03
CI: partial revert of recent changes

hhugo:


hhugo:
@smorimoto, I'm partially reverting https://github.com/ocsigen/js_of_ocaml/commit/20364ae7b58f5fa1a6eb909a397dc2821f56fbe7 here. I'd prefer that we use PRs instead of pushing to the master branch directly because it gives more visibility to what's going on and provides a "space" to discuss a change, even after it's merged.

PR #1620, opened 2024-06-06
Fix marshalling test (compression issue was fixed in 5.1.1)

OlivierNicole:
See https://ocaml.org/releases/5.1.1. Currently this test fails with 5.1.1.

### DIFF
--- a/compiler/tests-jsoo/test_marshal.ml
+++ b/compiler/tests-jsoo/test_marshal.ml
@@ -139,17 +139,19 @@ let%expect_test _ =
     "\132\149\166\189\r\022\206\021\001\147F\137d(\181/\253\000Xm\000\0000\n\
      \000\000'\016c\001\000\012\135\007E"
   in
-  let ocaml_5_1 =
+  let ocaml_5_1_0 =
     match String.split_on_char '.' Sys.ocaml_version with
-    | major :: minor :: _ ->
-        let major = int_of_string major and minor = int_of_string minor in
-        major > 5 || (major = 5 && minor >= 1)
+    | major :: minor1 :: minor2 :: _ ->
+        let major = int_of_string major
+        and minor1 = int_of_string minor1
+        and minor2 = int_of_string minor2 in
+        major > 5 || (major = 5 && minor1 = 1 && minor2 = 0)
     | _ -> assert false
   in
   let s =
     (* This would only work on OCaml 5.1.0 if we were not linking
        against compiler-libs *)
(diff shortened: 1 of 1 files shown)


hhugo:
Why is the CI not complaining about this ?

OlivierNicole:
I’m not sure; maybe its 5.1.1 compiler is configured differently from mine?

hhugo:
Ah yes, I guess we could condition the test on `Compression.compression_supported`

hhugo:
This is the only test checking compression with marshal. I'd prefer to keep it enabled on ocaml > 5.1.0 as it runs and pass in the CI. We could try to come up with a better condition to decide whether compression is enabled though.

OlivierNicole:
After thinking about it more, I don’t think that is the right decision. The situation at present is the following:

- The test passes on 5.1.0, because 5.1.0 has mandatory zstd integration. But this was changed in 5.1.1 because it caused dependency problems and it has lead to the release of 5.1.1;
- On 5.1.1 and later, the test passes or fail depending on how the OCaml install was configured. What happens is that the condition `ocaml_5_1` is true and so the test attempts to unmarshal compressed data. It works if OCaml was built with ZSTD support _and_ compiler-libs is linked in (which is the case here).

But the building of OCaml with ZSTD support happens implicitly, _if_ zlib is available at configure time, which is not something the programmer necessarily thinks about. To wit, on my NixOS zlib is not available unless I explicitly ask for it in my shell, and so my switch was compiled without ZSTD without me even thinking about it. As a consequence, this test has been constantly failing on my machine on all OCaml versions except the (more or less deprecated) 5.1.0, and it took me several hours to find out exactly why. It is nice if all tests pass on the CI but the experience is not very good for us JSOO developers.

For this reason I propose the following: remove this test from the file and move it to a separate test, that will be enabled only on 5.1.1 or later and will test `Compression.compression_supported`. (I don’t see much worth in testing on 5.1.0.)

What do you think?

hhugo:
> For this reason I propose the following: remove this test from the file and move it to a separate test, that will be enabled only on 5.1.1 or later and will test `Compression.compression_supported`. (I don’t see much worth in testing on 5.1.0.)
> 
> What do you think?

Sounds good

OlivierNicole:
I pushed this to the branch. Could you re-open the PR?

hhugo:
Can't re-open it, can you create a new one

hhugo:
reopened

PR #1624, opened 2024-06-17
Runtime: os_type: Cygwin -> Win32

hhugo:


hhugo:
@phated, would you mind checking if this PR works for `grain` ? It should only affect windows.

phated (commented):
@hhugo sorry for the delay. We actually override the Cygwin value to Win32 already. See https://github.com/grain-lang/grain/blob/416936b08015d4bf0e945860b646bfa00e2f14b3/compiler/src/utils/hacks.js#L4

So this should be a good change for Grain.

hhugo:
Cool, do you need the rest of the stubs ? I thought they were present in jsoo already.

phated:
I think they are overrides related to `\` path separators instead of `/`

hhugo:
This PR now accepts both `/` and `\`. Could you try dropping your overrides with jsoo master and report if something still needs to be fixed ?

phated:
I'd love to try it, but we never solved the performance problems so our test suite doesn't complete. I forget which older version we needed to pin.

hhugo:
Can you try again and reopen an issue for the perf issue ?

PR #1632, opened 2024-06-27
Add msvc to the ci

hhugo:


hhugo:
@dra27, what would be the way to install gmp on msvc ?

dra27:
I think that this has never been possible with opam-repository-mingw (but this is something I haven't actually checked thoroughly). In theory, the .a files created by gcc are usable from an MSVC program (.a and .lib are an identical format, and both the .o and .obj files are COFF objects).

I've been musing whether we can do a similar trick with vcpkg (which [packages gmp](https://vcpkg.link/ports/gmp)) as @RyanGibb and I have been working on with nixpkgs, where we might add a second repository for enabling vcpkg support. However, I'm just at the "musing on it" stage at the moment. I know that DkML has integration with vcpkg already, but I haven't yet looked at the details of it.

Ages ago (2016), I started https://github.com/dra27/windows-native where the intention was to be able to build some of these depexts from sources in the switch - I had some work in progress on adding gmp (and mpir, I think) to this idea - but again, it's not something that I'm actively working on at the moment.

PR #1639, opened 2024-07-22
Misc: yojson is no longer optional

hhugo:


hhugo:
@nojb, would that be an issue for lexifi ?

nojb:
> @nojb, would that be an issue for lexifi ?

No, it is not a problem: we already depend on yojson due to some other dependencies.

Thanks for the ping!

PR #1617, opened 2024-05-23
Optimize sourcemap processing

OlivierNicole:
This improves the efficiency of sourcemap processing, chiefly during linking, but also in serializing and deserializing sourcemaps.

Currently, when linking with source maps enabled, js_of_ocaml will decode the source maps for each input file completely into an in-memory explicit mapping, then merge them, then re-encode the merged sourcemap. The merging algorithm involves superlinear sorting operations. However, the format of sourcemaps allows to avoid sorting for the concatenation of generated files.

In addition, all operations performed by linking (concatenation of files, removal and addition of lines) can be performed directly on the encoded form, which further avoids a full parsing to an in-memory representation.

When compiling js_of_ocaml with itself, the final linking step is accelerated by a speedup of about 2.3x with OCaml 5.2.0+fp:

| recent trunk (bfbeb692c) | this PR |
|--------------------|------------|
| 331.8 ms ±   6.1 ms | 143.3 ms ±   1.1 ms |

fix https://github.com/ocsigen/js_of_ocaml/issues/1446

PR #1640, opened 2024-07-22
extracted from #1617

hhugo:
#1617

### DIFF
--- a/compiler/lib/source_map.ml
+++ b/compiler/lib/source_map.ml
@@ -19,6 +19,16 @@
 
 open! Stdlib
 
+module Source_content = struct
+  type t = Sc_as_Stringlit of string
+
+  let create s = Sc_as_Stringlit (Yojson.Safe.to_string (`String s))
+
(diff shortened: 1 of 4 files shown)


hhugo:
cc @OlivierNicole

compiler/lib/source_map.ml
```diff
@@ -19,6 +19,16 @@
 
 open! Stdlib
 
+module Source_content = struct
```
OlivierNicole, line 22: Nit: naming it `Sources_content` (with an S to Sources) would match the sourcemap field name `sourcesContent`.

OlivierNicole (approved):
LGTM, thanks for extracting this.

compiler/lib/source_map.ml
hhugo: The sourcemap field store multiple sources but `Source_content.t` is about a single one.

OlivierNicole: Ah, true. Never mind me then.

PR #1647, opened 2024-07-26
Compiler: compact indirect call

hhugo:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#direct_and_indirect_eval

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,9 +5,12 @@
 * Misc: CI uses opam.2.2 and no longer use sunset repo
 * Misc: yojson is no longer optional
 * Compiler: speedup global_flow/global_deadcode pass on large bytecode
+* Compiler: make indirect call using sequence instead of using the call method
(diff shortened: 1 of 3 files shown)


vouillon:
Have you benchmarked this? From what I remember, it was slower.
See https://github.com/ocsigen/js_of_ocaml/pull/1397#issuecomment-1420979002

hhugo:
> Have you benchmarked this? From what I remember, it was slower. See [#1397 (comment)](https://github.com/ocsigen/js_of_ocaml/pull/1397#issuecomment-1420979002)

I've done some quick benchmark and it was slower to use `E.call(null)`. Maybe we need more testing here.

hhugo:
@vouillon, in case you want to play with some smallbenchmarks https://jsperf.app/test-call-vs-apply/93

PR #1648, opened 2024-07-29
Effects: keep track of CPS calls and distinguish CPS calls and trampolined calls

OlivierNicole:
This makes a more precise distinction between CPS calls and trampolined calls, and keeps track of both. It is used in wasm_of_ocaml to handle CPS calls.

This PR is the first in a series of PRs intending to reduce the diff between js_of_ocaml and wasm_of_ocaml, in order to merge the two repos eventually, or at least to ease integration of JSOO updates in WSOO.

### DIFF
--- a/compiler/lib/effects.ml
+++ b/compiler/lib/effects.ml
@@ -247,7 +247,9 @@ let jump_closures blocks_to_transform idom : jump_closures =
     idom
     { closure_of_jump = Addr.Map.empty; closures_of_alloc_site = Addr.Map.empty }
 
-type cps_calls = Var.Set.t
+type trampolined_calls = Var.Set.t
+
+type in_cps = Var.Set.t
 
 type st =
   { mutable new_blocks : Code.block Addr.Map.t * Code.Addr.t
@@ -263,7 +265,8 @@ type st =
   ; block_order : (Addr.t, int) Hashtbl.t
   ; live_vars : Deadcode.variable_uses
   ; flow_info : Global_flow.info
-  ; cps_calls : cps_calls ref
+  ; trampolined_calls : trampolined_calls ref
+  ; in_cps : in_cps ref
   }
 
 let add_block st block =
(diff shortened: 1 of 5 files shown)


compiler/lib/effects.ml
```diff
@@ -927,7 +944,7 @@ let f ~flow_info ~live_vars p =
   let cps_needed = Partial_cps_analysis.f p flow_info in
   let p, cps_needed = rewrite_toplevel ~cps_needed p in
   let p = split_blocks ~cps_needed p in
-  let p, cps_calls = cps_transform ~live_vars ~flow_info ~cps_needed p in
+  let p, trampolined_calls, in_cps = cps_transform ~live_vars ~flow_info ~cps_needed p in
```
hhugo, line 947: What's the difference between `cps_needed` and `in_cps` ?

OlivierNicole: I’m not entirely sure, to be honest, I’d rather wait for the author @vouillon to answer.

vouillon: Basically, we turn:
```
x = f(x1, ..., xn)
cont
```
into:
```
y = f(x1, ..., xn, (x) => cont)
return y
```
when `x` is in `cps_needed`.
And then, `y` is put into `in_cps`.

You can see that we use a fresh variable here:
https://github.com/ocsigen/js_of_ocaml/blob/f47b4d5456b36e1164f428ccfa1ec4d2740a393d/compiler/lib/effects.ml#L288-L290

vouillon:
CPS calls are calls that take a continuation as the last parameter. We also have direct calls which may make use of the trampoline: they are used to implement loops. In the current implementation, we don't really distinguish them since these direct calls are exact, and the JS implementation behaves the same for direct calls whether in CPS or not (just apply the function to all its arguments).
This is no longer true with `wasm_of_ocaml` where the two kind of functions have different types.
We also need to make the distinction for the double translation (#1461).

PR #1649, opened 2024-07-29
Make number representations compatible with Wasm

OlivierNicole:
This is best reviewed commit by commit. It adds a number of missing conversions, and other changes that have been performed in Wasm_of_ocaml in order to expose number-manipulating functions that are compatible with both the JS and Wasm backends.

This is part of a series of PRs intending to reduce the diff between js_of_ocaml and wasm_of_ocaml.


compiler/lib/effects.ml
```diff
@@ -390,7 +390,7 @@ let cps_last ~st ~alloc_jump_closures pc ((last, last_loc) : last * loc) ~k :
                         ( x'
                         , Prim
                             ( Extern "caml_maybe_attach_backtrace"
-                            , [ Pv x; Pc (Int (if force then 1l else 0l)) ] ) )
+                            , [ Pv x; Pc (Int (Regular, if force then 1l else 0l)) ] ) )
```
hhugo, line 393: Shouldn't this be native ?

compiler/lib/effects.ml
```diff
@@ -466,11 +466,12 @@ let cps_instr ~st (instr : instr) : instr =
       Let (x, Closure (params @ [ k ], cont))
   | Let (x, Prim (Extern "caml_alloc_dummy_function", [ size; arity ])) -> (
       match arity with
-      | Pc (Int a) ->
+      | Pc (Int (_, a)) ->
           Let
```
hhugo, line 470: Maybe assert that kind is Regular

compiler/lib/effects.ml
hhugo: ignore this

compiler/lib/eval.ml
```diff
@@ -33,21 +33,21 @@ module Int = Int32
 
 let int_binop l f =
   match l with
-  | [ Int i; Int j ] -> Some (Int (f i j))
+  | [ Int (_, i); Int (_, j) ] -> Some (Int (Regular, f i j))
```
hhugo, line 36: Why don't we check equality of kind and forward it ?

compiler/lib/eval.ml
```diff
@@ -33,21 +33,21 @@ module Int = Int32
...
   | _ -> None
 
 let shift l f =
   match l with
-  | [ Int i; Int j ] -> Some (Int (f i (Int32.to_int j land 0x1f)))
+  | [ Int (_, i); Int (_, j) ] -> Some (Int (Regular, f i (Int32.to_int j land 0x1f)))
```
hhugo, line 41: Same here, we could check equality of kind and forward it to the rhs

compiler/lib/code.ml
```diff
@@ -358,14 +358,19 @@ module Native_string = struct
...
+
 type constant =
   | String of string
   | NativeString of Native_string.t
   | Float of float
   | Float_array of float array
   | Int64 of int64
   | Tuple of int * constant array * array_or_not
-  | Int of int32
+  | Int of int_kind * int32
```
hhugo, line 373: Have you considered just adding two new constructors: Int32 and NativeInt to constant ? I feel the end result would be nicer and less error prone.

vouillon: That might be nicer indeed!

So, we need to change `Parse_bytecode.Constants.parse` and `Ocaml_compiler.constant_of_const` so that when producing JavaScript code, we convert `int32` and `nativeint` into regular `int` early and then only deal with regular integer. Then, functions `int_bin_op` and `shift` only have to deal with regular integers.

@OlivierNicole Do you want to give it a try?

vouillon: We can have a commit that just adds these two constructors here. And, then, the target specific code would be in #1655.

OlivierNicole: > @OlivierNicole Do you want to give it a try?

@vouillon Sure, I’ll do it.

OlivierNicole: @hhugo I have modified the relevant commit to implement your suggestion.

compiler/lib/code.mli
```diff
@@ -181,9 +181,11 @@ type constant =
   | NativeString of Native_string.t
   | Float of float
   | Float_array of float array
+  | Int of int32
+  | Int32 of int32  (** Only produced when compiling to WebAssembly. *)
```
hhugo, line 185: I was expecting that Int32 and NativeInt would be used regardless of the backend and that generate would translate all to int.

OlivierNicole: That is an option. One disadvantage of doing it that way is that then, functions like `int_bin_op` and `shift` in `Eval` have to deal with three types of integers instead of one.

On the other hand that is a rather small disadvantage. @vouillon Do you see other reasons not to produce `Int32` and `NativeInt` constructors when producing Javascript?

compiler/lib/eval.ml
vouillon: With JSOO, integer conversion functions are turned into the identity. So, a binary operation can have any kind of integer as argument. But the output is definitively a regular integer.

compiler/lib/code.mli
vouillon: I think it's a good thing to make it explicit that all these different integers share the same representation in JSOO.
In particular, it is clear that `int_bin_op` is correct and handles all the possible cases. Otherwise, since integer conversion functions are the identity, this function would have to deal with the three types of integers indeed, which could be somewhat confusing.

OlivierNicole: I tend to be convinced by @vouillon's argument.

compiler/lib/eval.ml
```diff
@@ -332,10 +342,12 @@ let the_cond_of info x =
     info
     (fun x ->
       match Flow.Info.def info x with
-      | Some (Constant (Int 0l)) -> Zero
+      | Some (Constant (Int 0l | Int32 0l | NativeInt 0n)) -> Zero
```
hhugo, line 345: Do we really want to treat `Int32 0` and `NativeInt 0` as zero ?

compiler/lib/parse_bytecode.ml
```diff
@@ -515,7 +515,7 @@ end = struct
     | Float_array _ -> false
     | Int64 _ -> false
     | Tuple _ -> false
-    | Int _ -> true
+    | Int _ | Int32 _ | NativeInt _ -> true
```
hhugo, line 518: Do we really want to inline Int32 and NativeInt in wasmoo ?

compiler/lib/parse_bytecode.ml
```diff
@@ -1554,7 +1554,17 @@ and compile infos pc state instrs =
...
         if debug_parser () then Format.printf "%a = %a[%d]@." Var.print x Var.print y n;
-        compile infos (pc + 2) state ((Let (x, Field (y, n)), loc) :: instrs)
+        compile
+          infos
+          (pc + 2)
+          state
+          (( Let
+               ( x
+               , Prim
+                   ( Extern "caml_floatarray_unsafe_get"
```
hhugo, line 1564: I wouldn't it be cleaner to extend the `Field` constructor  with some additional information ? Float | NonFloat ?

compiler/lib/parse_bytecode.ml
```diff
@@ -1628,7 +1638,13 @@ and compile infos pc state instrs =
           infos
           (pc + 2)
           (State.pop 1 state)
-          ((Let (x, const 0l), loc) :: (Set_field (y, n, z), loc) :: instrs)
+          (( Let
+               ( x
+               , Prim
+                   ( Extern "caml_floatarray_unsafe_set"
```
hhugo, line 1644: same here, we could extend set_field with additional information.

compiler/lib/parse_bytecode.ml
OlivierNicole: That would be cleaner. I’m on it, it will take me a bit of work.

OlivierNicole: Done

hhugo:
Thanks. Looks good.
Can you fix the CI and squash the fixup ?

hhugo:
> Done. It looks like the failing test will only work if I integrate the single commit from #1650, which redefines `Js.number_t` as being `Js.number Js.t`.
> 
> This may break programs that do not perform explicit float conversions yet.

I removed the last commit, and fixed the hypertree example to use number_t

OlivierNicole:
Thank you

PR #1650, opened 2024-07-29
Stop exposing numbers as floats in the Js_of_ocaml library

OlivierNicole:
This is for compatibility with wasm_of_ocaml, in which Javascript numbers and floats cannot be manipulated using the same functions and require conversions.

I open this as a draft PR because it will likely break a number of packages; in particular, before merging, we should wait for Jane Street to release a version of their packages with the missing explicit float conversions.


hhugo:
included in #1724

PR #1652, opened 2024-07-30
Make the tests for uncaught exceptions more portable

OlivierNicole:
This is part of a series of PRs intending to reduce the diff between js_of_ocaml and wasm_of_ocaml (see ocaml-wasm/wasm_of_ocaml#47).


hhugo:
Can you try to keep the bytecode mode as well ? It is useful to check js vs non-js behavior

OlivierNicole:
Sorry about that. Done

runtime/stdlib.js
```diff
@@ -131,7 +131,16 @@ var caml_named_values = {};
 //Provides: caml_register_named_value (const,mutable)
 //Requires: caml_named_values, caml_jsbytes_of_string
 function caml_register_named_value(nm,v) {
-  caml_named_values[caml_jsbytes_of_string(nm)] = v;
+  nm = caml_jsbytes_of_string(nm);
+  if (!caml_named_values.hasOwnProperty(nm)) caml_named_values[nm] = v;
```
hhugo, line 135: Why such change ? This is not what the upstream runtime does https://github.com/ocaml/ocaml/blob/48d3c5a065838e41388b18ea8f33d57703ef4337/runtime/callback.c#L403

runtime/stdlib.js
```diff
@@ -131,7 +131,16 @@ var caml_named_values = {};
...
+  return 0;
+}
+
+//Provides: caml_unregister_named_value (const)
+//Requires: caml_named_values, caml_jsbytes_of_string
+function caml_unregister_named_value(nm) {
```
hhugo, line 141: This function doesn't have to live in the main runtime and could just live in the test directory. These tests are the only users of such primitive.

Also, you need to implement the c counter-part if you want the byte version to work.

vouillon: Something like this should work:
```C
value caml_unregister_named_value(value nm) {
   return caml_register_named_value(nm, (value) NULL);
}

runtime/stdlib.js
vouillon: I don't remember why I did that. Maybe I misread the C code? I think we can remove this change.
I will need to fix the Wasm runtime as well.

OlivierNicole:
Oops, unassigned myself by mistake.

vouillon:
> Can you try to keep the bytecode mode as well ? It is useful to check js vs non-js behavior

We were only building the code, not running it, though...

hhugo:
> > Can you try to keep the bytecode mode as well ? It is useful to check js vs non-js behavior
> 
> We were only building the code, not running it, though...

I'm pretty sure I've run the byte version at the time to check the behavior. I don't remember why I didn't implement a js-vs-bytecode test though

hhugo:
It's 

> > > Can you try to keep the bytecode mode as well ? It is useful to check js vs non-js behavior
> > 
> > 
> > We were only building the code, not running it, though...
> 
> I'm pretty sure I've run the byte version at the time to check the behavior. I don't remember why I didn't implement a js-vs-bytecode test though

It's possible that I only ran the byte version without argument, (not calling unregister)

PR #1654, opened 2024-07-30
Small changes in the internal APIs for Wasm

OlivierNicole:
This PR consists in a number of changes in the APIs of `compiler/lib/` that @vouillon needed while developing wasm_of_ocaml.

It is best reviewed commit by commit.

### DIFF
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -680,46 +680,6 @@ let is_empty p =
       | _ -> false)
   | _ -> false
 
-let fold_children blocks pc f accu =
-  let block = Addr.Map.find pc blocks in
-  match fst block.branch with
-  | Return _ | Raise _ | Stop -> accu
-  | Branch (pc', _) | Poptrap (pc', _) -> f pc' accu
-  | Pushtrap ((pc', _), _, (pc_h, _)) ->
-      let accu = f pc' accu in
-      let accu = f pc_h accu in
-      accu
-  | Cond (_, (pc1, _), (pc2, _)) ->
-      let accu = f pc1 accu in
-      let accu = f pc2 accu in
-      accu
-  | Switch (_, a1) ->
-      let accu = Array.fold_right ~init:accu ~f:(fun (pc, _) accu -> f pc accu) a1 in
-      accu
-
-type 'c fold_blocs = block Addr.Map.t -> Addr.t -> (Addr.t -> 'c -> 'c) -> 'c -> 'c
-
-type fold_blocs_poly = { fold : 'a. 'a fold_blocs } [@@unboxed]
-
-let rec traverse' { fold } f pc visited blocks acc =
-  if not (Addr.Set.mem pc visited)
-  then
-    let visited = Addr.Set.add pc visited in
-    let visited, acc =
-      fold
-        blocks
-        pc
-        (fun pc (visited, acc) ->
-          let visited, acc = traverse' { fold } f pc visited blocks acc in
-          visited, acc)
-        (visited, acc)
-    in
-    let acc = f pc acc in
-    visited, acc
-  else visited, acc
-
-let traverse fold f pc blocks acc = snd (traverse' fold f pc Addr.Set.empty blocks acc)
-
 let poptraps blocks pc =
   let rec loop blocks pc visited depth acc =
     if Addr.Set.mem pc visited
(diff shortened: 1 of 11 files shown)


compiler/lib/source_map.ml
```diff
@@ -301,7 +301,7 @@ let merge = function
 
 (* IO *)
 
-let json t =
+let json ?replace_mappings t =
```
hhugo, line 304: Can you explain the usecase for such api ?

compiler/lib/source_map.mli
```diff
@@ -62,6 +62,13 @@ val empty : filename:string -> t
 
 val to_string : t -> string
 
-val to_file : t -> string -> unit
-
 val of_string : string -> t
+
+val of_file_no_mappings : string -> t * string option
```
hhugo, line 67: It seems that #1617 should fix the issue by delaying the parsing of mapping until necessary

compiler/lib/unit_info.mli
```diff
@@ -41,3 +41,7 @@ val prefix : string
 val to_string : t -> string
 
 val parse : t -> string -> t option
+
+val to_sexp : t -> Sexp.t list
```
hhugo, line 45: Does this really need to live inside jsoo ? It seems to me that you have everything you need to implement it in a separate module.

hhugo:
@OlivierNicole, I wonder if we could move faster by not grouping unrelated commits too much.
In this PR, some commits are trivial and could be merged very quickly, while some other will need much more discussion.

compiler/lib/source_map.ml
OlivierNicole: In wasm_of_ocaml, when linking, it was observed that parsing the `mappings` field of sourcemaps takes a big toll on the overall performance, and was unnecessary since the mappings are not modified. So this is a workaround to sidestep this parsing, the cleaner way being in #1617.

compiler/lib/source_map.mli
OlivierNicole: Yes. I considered not submitting these `Source_map` API changes, but it does make merging JSOO and WSOO a bit harder. If they are not acceptable as a temporary measure until #1617 is in proper shape, I will take them out and I will have to find another way to reconcile changes during the transitory period.

OlivierNicole:
Sure, I was worried of the opposite effect of bombarding the repo with too many PRs, hence the grouping. Let me know what you think should be separate PRs.

OlivierNicole:
Done. In fact I think I won’t do a PR for the Source_map changes for now, hopefully I can make progress on #1617 soon enough that it won’t be needed for long.

compiler/lib/code.ml
```diff
@@ -697,6 +730,23 @@ let fold_children blocks pc f accu =
       let accu = Array.fold_right ~init:accu ~f:(fun (pc, _) accu -> f pc accu) a1 in
       accu
 
+let fold_children_skip_try_body blocks pc f accu =
```
hhugo, line 733: This function seems to exist in Tailcall and Inline as `fold_children`. We should remove this duplication.

compiler/lib/stdlib.ml
```diff
@@ -986,6 +1022,10 @@ module String = struct
         | _ -> false
     in
     loop (length b - 1) b 0
+
+  let fold_left ~f ~init s = Bytes.fold_left ~f ~init (Bytes.unsafe_of_string s)
```
hhugo, line 1026: `Bytes.unsafe_of_string` is difficult to reason about. Even if it is correctly used here, I'd prefer to not have `Bytes.unsafe_of_string` in the codebase. You can just copy the implementation from Bytes.

compiler/lib/fs.mli
```diff
@@ -21,3 +21,9 @@ val find_in_path : string list -> string -> string option
 val absolute_path : string -> string
 
 val read_file : string -> string
+
+val write_file : name:string -> contents:string -> unit
+
+val gen_file : string -> (string -> 'a) -> 'a
+
+val with_intermediate_file : string -> (string -> 'a) -> 'a
```
hhugo, line 29: What is this function doing ?  I don't see any intermediate_file in the implementation. How is it used ? Maybe the name is misleading ?

compiler/lib/freevars.mli
```diff
@@ -25,4 +25,10 @@ val iter_block_bound_vars : (Code.Var.t -> unit) -> Code.block -> unit
 
 val f_mutable : Code.program -> Code.Var.Set.t Code.Addr.Map.t
 
+val iter_instr_free_vars : (Code.Var.t -> unit) -> Code.instr -> unit
```
hhugo, line 28: can you move the `iter_*` functions next the ones above

hhugo: Or rather, move `f_mutable` next to `f`

compiler/bin-js_of_ocaml/check_runtime.ml
```diff
@@ -43,6 +43,7 @@ let print_groups output l =
               output_string output (Printf.sprintf "%s\n" name)))
 
 let f (runtime_files, bytecode, target_env) =
+  Generate.init ();
```
hhugo, line 46: `Linker.reset ();` might be even better here. It calls`Generate.init` .

compiler/lib/code.ml
vouillon: Oh, right! That makes even more sense to add it to `Code`. then.

compiler/lib/fs.mli
vouillon: We are calling external tools (the Binaryen toolchain) and use temporary intermediate files for that.

Function `with_intermediate_file` is used to make sure that a temporary file is deleted in case of failure. It is similar to `Filename.with_open` but passes the temporary file name instead of a channel. I use it this way:
```ocaml
Fs.with_intermediate_file (Filename.temp_file "wasm-merged" ".wasm")
@@ fun temp_file ->
...
```

compiler/lib/code.ml
OlivierNicole: Done

compiler/lib/stdlib.ml
OlivierNicole: Done

compiler/lib/freevars.mli
OlivierNicole: Done

compiler/bin-js_of_ocaml/check_runtime.ml
OlivierNicole: Done

hhugo:
The CI is not happy. @vouillon mentioned elsewhere that it could be due to a missing `Generate.init ()` possibly in js_of_ocaml_dynlink.ml

compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
```diff
@@ -26,6 +26,7 @@ let normalize_bytecode code =
       Bytes.to_string b
 
 let () =
+  Linker.reset ();
```
hhugo, line 29: It needs to come after the `Config.Flag.set`

OlivierNicole:
I still get this error:

```
 --- a/_build/default/compiler/tests-toplevel/test_toplevel.reference
+++ b/_build/default/compiler/tests-toplevel/test_toplevel.referencebcjs
@@ -4,4 +4,7 @@ Line 4, characters 2-4:
 Error: Syntax error
 Line 5, characters 0-16:
 Error: Unbound module "Missing_module"
-val y : float = 0.333333333333333315
+Dynlink: looking for symbol caml_float_of_int
+Dynlink: looking for symbol caml_float_of_int
+Line 1:
+Error: The external function "caml_float_of_int" is not available
```

in spite of the `Linker.reset ()` that I added in `js_of_ocaml_compiler_dynlink.ml`. I’m not sure why.

hhugo:
> I still get this error:
> 
> ```
>  --- a/_build/default/compiler/tests-toplevel/test_toplevel.reference
> +++ b/_build/default/compiler/tests-toplevel/test_toplevel.referencebcjs
> @@ -4,4 +4,7 @@ Line 4, characters 2-4:
>  Error: Syntax error
>  Line 5, characters 0-16:
>  Error: Unbound module "Missing_module"
> -val y : float = 0.333333333333333315
> +Dynlink: looking for symbol caml_float_of_int
> +Dynlink: looking for symbol caml_float_of_int
> +Line 1:
> +Error: The external function "caml_float_of_int" is not available
> ```
> 
> in spite of the `Linker.reset ()` that I added in `js_of_ocaml_compiler_dynlink.ml`. I’m not sure why.

You need to initialize the thing in compiler/bin-js_of_ocaml/link.ml inside `f`.

I think I'd be more confortable moving the explicit initialization to its own PR so that we can discuss alternative and correctness without holding on the rest of this PR.

OlivierNicole:
Done. I have opened #1662 for explicit initialization.

OlivierNicole:
Thanks!

PR #1655, opened 2024-07-30
Target-specific code

OlivierNicole:
**Note:** This PR depends on #1649 and #1654 and is based on top of them. ~~The proposed feature is contained only in the last commit.~~ edit: These PRs have been merged.

This PR contains @vouillon's work to have two distinct targets: Javascript and WebAssembly.

Dune support for the new Wasm backend has been proposed at ocaml/dune#10695.

### DIFF
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -196,12 +196,12 @@ let run
           in
           let code = Code.prepend one.code instr in
           Driver.f
+            ~target:(JavaScript fmt)
             ~standalone
             ?profile
             ~link
             ~wrap_with_fun
             ?source_map
-            fmt
             one.debug
             code
       | `File, fmt ->
@@ -220,12 +220,12 @@ let run
           let code = Code.prepend one.code instr in
           let res =
             Driver.f
+              ~target:(JavaScript fmt)
               ~standalone
               ?profile
               ~link
               ~wrap_with_fun
               ?source_map
-              fmt
               one.debug
               code
           in
@@ -285,7 +285,7 @@ let run
   | `None ->
       let prims = Linker.list_all () |> StringSet.elements in
       assert (List.length prims > 0);
-      let code, uinfo = Parse_bytecode.predefined_exceptions () in
+      let code, uinfo = Parse_bytecode.predefined_exceptions ~target:`JavaScript in
       let uinfo = { uinfo with primitives = uinfo.primitives @ prims } in
       let code : Parse_bytecode.one =
         { code
@@ -331,6 +331,7 @@ let run
           let linkall = linkall || toplevel || dynlink in
           let code =
             Parse_bytecode.from_exe
+              ~target:`JavaScript
               ~includes:include_dirs
               ~include_cmis
               ~link_info:(toplevel || dynlink)
@@ -363,6 +364,7 @@ let run
           let t1 = Timer.make () in
           let code =
             Parse_bytecode.from_cmo
+              ~target:`JavaScript
               ~includes:include_dirs
               ~include_cmis
               ~debug:need_debug
@@ -419,6 +421,7 @@ let run
               let t1 = Timer.make () in
               let code =
                 Parse_bytecode.from_cmo
+                  ~target:`JavaScript
                   ~includes:include_dirs
                   ~include_cmis
                   ~debug:need_debug
@@ -450,6 +453,7 @@ let run
                 let t1 = Timer.make () in
                 let code =
                   Parse_bytecode.from_cmo
+                    ~target:`JavaScript
                     ~includes:include_dirs
                     ~include_cmis
                     ~debug:need_debug
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -285,7 +285,7 @@ let run
   | `None ->
       let prims = Linker.list_all () |> StringSet.elements in
       assert (List.length prims > 0);
-      let code, uinfo = Parse_bytecode.predefined_exceptions ~target:`JavaScript in
+      let code, uinfo = Parse_bytecode.predefined_exceptions in
       let uinfo = { uinfo with primitives = uinfo.primitives @ prims } in
       let code : Parse_bytecode.one =
         { code
@@ -331,7 +331,6 @@ let run
           let linkall = linkall || toplevel || dynlink in
           let code =
             Parse_bytecode.from_exe
-              ~target:`JavaScript
               ~includes:include_dirs
               ~include_cmis
               ~link_info:(toplevel || dynlink)
@@ -364,7 +363,6 @@ let run
           let t1 = Timer.make () in
           let code =
             Parse_bytecode.from_cmo
-              ~target:`JavaScript
               ~includes:include_dirs
               ~include_cmis
               ~debug:need_debug
@@ -421,7 +419,6 @@ let run
               let t1 = Timer.make () in
               let code =
                 Parse_bytecode.from_cmo
-                  ~target:`JavaScript
                   ~includes:include_dirs
                   ~include_cmis
                   ~debug:need_debug
@@ -453,7 +450,6 @@ let run
                 let t1 = Timer.make () in
                 let code =
                   Parse_bytecode.from_cmo
-                    ~target:`JavaScript
                     ~includes:include_dirs
                     ~include_cmis
                     ~debug:need_debug
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -186,7 +186,7 @@ let run
     let init_pseudo_fs = fs_external && standalone in
     let sm =
       match output_file with
-      | `Stdout, fmt ->
+      | `Stdout, formatter ->
           let instr =
             List.concat
               [ pseudo_fs_instr `create_file one.debug one.cmis
@@ -196,15 +196,15 @@ let run
           in
           let code = Code.prepend one.code instr in
           Driver.f
-            ~target:(JavaScript fmt)
             ~standalone
             ?profile
             ~link
             ~wrap_with_fun
             ?source_map
+            ~formatter
             one.debug
             code
-      | `File, fmt ->
+      | `File, formatter ->
           let fs_instr1, fs_instr2 =
             match fs_output with
             | None -> pseudo_fs_instr `create_file one.debug one.cmis, []
@@ -220,12 +220,12 @@ let run
           let code = Code.prepend one.code instr in
           let res =
             Driver.f
-              ~target:(JavaScript fmt)
               ~standalone
               ?profile
               ~link
               ~wrap_with_fun
               ?source_map
+              ~formatter
               one.debug
               code
           in
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -285,7 +285,7 @@ let run
   | `None ->
       let prims = Linker.list_all () |> StringSet.elements in
       assert (List.length prims > 0);
-      let code, uinfo = Parse_bytecode.predefined_exceptions in
+      let code, uinfo = Parse_bytecode.predefined_exceptions () in
       let uinfo = { uinfo with primitives = uinfo.primitives @ prims } in
       let code : Parse_bytecode.one =
         { code
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -90,6 +90,7 @@ let run
     } =
   let include_cmis = toplevel && not no_cmis in
   let custom_header = common.Jsoo_cmdline.Arg.custom_header in
+  Config.set_target `JavaScript;
   Jsoo_cmdline.Arg.eval common;
   Linker.reset ();
   (match output_file with
--- a/compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
+++ b/compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
@@ -26,7 +26,9 @@ let normalize_bytecode code =
       Bytes.to_string b
 
 let () =
-  Config.set_target `JavaScript;
+  (match Sys.backend_type with
+  | Sys.Other "js_of_ocaml" -> Config.set_target `JavaScript
+  | Sys.(Native | Bytecode | Other _) -> failwith "Expected backend `js_of_ocaml`");
   let global = J.pure_js_expr "globalThis" in
   Config.Flag.set "use-js-string" (Jsoo_runtime.Sys.Config.use_js_string ());
   Config.Flag.set "effects" (Jsoo_runtime.Sys.Config.effects ());
--- a/compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
+++ b/compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
@@ -26,6 +26,7 @@ let normalize_bytecode code =
       Bytes.to_string b
 
 let () =
+  Config.set_target `JavaScript;
   let global = J.pure_js_expr "globalThis" in
   Config.Flag.set "use-js-string" (Jsoo_runtime.Sys.Config.use_js_string ());
   Config.Flag.set "effects" (Jsoo_runtime.Sys.Config.effects ());
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -327,10 +327,10 @@ type constant =
   | NativeString of Native_string.t
   | Float of float
   | Float_array of float array
-  | Int of int32
-  | Int32 of int32
-  | Int64 of int64
-  | NativeInt of nativeint
+  | Int of Int32.t
+  | Int32 of Int32.t
+  | Int64 of Int64.t
+  | NativeInt of Int32.t (* Native int are 32bit on all known backend *)
   | Tuple of int * constant array * array_or_not
 
 module Constant = struct
@@ -354,7 +354,7 @@ module Constant = struct
           !same
     | Int a, Int b | Int32 a, Int32 b -> Some (Int32.equal a b)
     | Int64 a, Int64 b -> Some (Int64.equal a b)
-    | NativeInt a, NativeInt b -> Some (Nativeint.equal a b)
+    | NativeInt a, NativeInt b -> Some (Int32.equal a b)
     | Float_array a, Float_array b -> Some (Array.equal Float.ieee_equal a b)
     | Float a, Float b -> Some (Float.ieee_equal a b)
     | String _, NativeString _ | NativeString _, String _ -> None
@@ -500,7 +500,7 @@ module Print = struct
     | Int i -> Format.fprintf f "%ld" i
     | Int32 i -> Format.fprintf f "%ldl" i
     | Int64 i -> Format.fprintf f "%LdL" i
-    | NativeInt i -> Format.fprintf f "%ndn" i
+    | NativeInt i -> Format.fprintf f "%ldn" i
     | Tuple (tag, a, _) -> (
         Format.fprintf f "<%d>" tag;
         match Array.length a with
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -870,6 +870,14 @@ let invariant { blocks; start; _ } =
         assert (not (Var.ISet.mem defs x));
         Var.ISet.add defs x)
     in
+    let check_prim_arg = function
+      | Pc (NativeInt _ | Int32 _) ->
+          assert (
+            match Config.target () with
+            | `Wasm -> true
+            | _ -> false)
+      | Pc _ | Pv _ -> ()
+    in
     let check_expr = function
       | Apply _ -> ()
       | Block (_, _, _, _) -> ()
@@ -877,8 +885,13 @@ let invariant { blocks; start; _ } =
       | Closure (l, cont) ->
           List.iter l ~f:define;
           check_cont cont
+      | Constant (NativeInt _ | Int32 _) ->
+          assert (
+            match Config.target () with
+            | `Wasm -> true
+            | _ -> false)
       | Constant _ -> ()
-      | Prim (_, _) -> ()
+      | Prim (_, args) -> List.iter ~f:check_prim_arg args
       | Special _ -> ()
     in
     let check_instr (i, _loc) =
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -870,13 +870,18 @@ let invariant { blocks; start; _ } =
         assert (not (Var.ISet.mem defs x));
         Var.ISet.add defs x)
     in
-    let check_prim_arg = function
-      | Pc (NativeInt _ | Int32 _) ->
+    let check_constant = function
+      | NativeInt _ | Int32 _ ->
           assert (
             match Config.target () with
             | `Wasm -> true
             | _ -> false)
-      | Pc _ | Pv _ -> ()
+      | String _ | NativeString _ | Float _ | Float_array _ | Int _ | Int64 _
+      | Tuple (_, _, _) -> ()
+    in
+    let check_prim_arg = function
+      | Pc c -> check_constant c
+      | Pv _ -> ()
     in
     let check_expr = function
       | Apply _ -> ()
@@ -885,12 +890,7 @@ let invariant { blocks; start; _ } =
       | Closure (l, cont) ->
           List.iter l ~f:define;
           check_cont cont
-      | Constant (NativeInt _ | Int32 _) ->
-          assert (
-            match Config.target () with
-            | `Wasm -> true
-            | _ -> false)
-      | Constant _ -> ()
+      | Constant c -> check_constant c
       | Prim (_, args) -> List.iter ~f:check_prim_arg args
       | Special _ -> ()
     in
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -856,6 +856,7 @@ let with_invariant = Debug.find "invariant"
 let check_defs = false
 
 let invariant { blocks; start; _ } =
+  let target = Config.target () in
   if with_invariant ()
   then (
     assert (Addr.Map.mem start blocks);
@@ -873,7 +874,7 @@ let invariant { blocks; start; _ } =
     let check_constant = function
       | NativeInt _ | Int32 _ ->
           assert (
-            match Config.target () with
+            match target with
             | `Wasm -> true
             | _ -> false)
       | String _ | NativeString _ | Float _ | Float_array _ | Int _ | Int64 _
--- a/compiler/lib/config.ml
+++ b/compiler/lib/config.ml
@@ -183,7 +183,7 @@ end
 
 (****)
 
-let target_ : [ `JavaScript | `Wasm ] option ref = ref (Some `JavaScript)
+let target_ : [ `JavaScript | `Wasm ] option ref = ref None
 
 let target () =
   match !target_ with
--- a/compiler/lib/config.ml
+++ b/compiler/lib/config.ml
@@ -183,8 +183,11 @@ end
 
 (****)
 
-let target_ : [ `JavaScript | `Wasm ] ref = ref `JavaScript
+let target_ : [ `JavaScript | `Wasm ] option ref = ref (Some `JavaScript)
 
-let target () = !target_
+let target () =
+  match !target_ with
+  | Some t -> t
+  | None -> failwith "target was not set"
 
-let set_target t = target_ := t
+let set_target t = target_ := Some t
--- a/compiler/lib/config.ml
+++ b/compiler/lib/config.ml
@@ -183,11 +183,12 @@ end
 
 (****)
 
-let target_ : [ `JavaScript | `Wasm ] option ref = ref None
+let target_ : [ `JavaScript | `Wasm | `None ] ref = ref `None
 
 let target () =
   match !target_ with
-  | Some t -> t
-  | None -> failwith "target was not set"
+  | `None -> failwith "target was not set"
+  | (`JavaScript | `Wasm) as t -> t
 
-let set_target t = target_ := Some t
+let set_target (t : [ `JavaScript | `Wasm ]) =
+  target_ := (t :> [ `JavaScript | `Wasm | `None ])
--- a/compiler/lib/config.ml
+++ b/compiler/lib/config.ml
@@ -164,7 +164,7 @@ module Param = struct
     p
       ~name:"tc"
       ~desc:"Set tailcall optimisation"
-      (enum [ "trampoline", TcTrampoline; (* default *) "none", TcNone ])
+      (enum [ "trampoline", TcTrampoline (* default *); "none", TcNone ])
 
   let lambda_lifting_threshold =
     (* When we reach this depth, we start looking for functions to be lifted *)
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -691,9 +691,9 @@ let optimize ~profile p =
     +> exact_calls ~deadcode_sentinal profile
     +> effects ~deadcode_sentinal
     +> map_fst
-         (match Config.target () with
-         | `JavaScript -> if Config.Flag.effects () then Fun.id else Generate_closure.f
-         | `Wasm -> Fun.id)
+         (match Config.target (), Config.Flag.effects () with
+         | `JavaScript, false -> Generate_closure.f
+         | `JavaScript, true | `Wasm, _ -> Fun.id)
     +> map_fst deadcode'
   in
   if times () then Format.eprintf "Start Optimizing...@.";
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -23,6 +23,14 @@ let debug = Debug.find "main"
 
 let times = Debug.find "times"
 
+type optimized_result =
+  { program : Code.program
+  ; variable_uses : Deadcode.variable_uses
+  ; trampolined_calls : Effects.trampolined_calls
+  ; in_cps : Effects.in_cps
+  ; deadcode_sentinal : Code.Var.t
+  }
+
 type profile =
   | O1
   | O2
@@ -194,14 +202,13 @@ let generate
     ~exported_runtime
     ~wrap_with_fun
     ~warn_on_unhandled_effect
-    ~deadcode_sentinal
-    ((p, live_vars), trampolined_calls, _) =
+    { program; variable_uses; trampolined_calls; deadcode_sentinal; in_cps = _ } =
   if times () then Format.eprintf "Start Generation...@.";
   let should_export = should_export wrap_with_fun in
   Generate.f
-    p
+    program
     ~exported_runtime
-    ~live_vars
+    ~live_vars:variable_uses
     ~trampolined_calls
     ~should_export
     ~warn_on_unhandled_effect
@@ -670,14 +677,6 @@ let link_and_pack ?(standalone = true) ?(wrap_with_fun = `Iife) ?(link = `No) p 
   |> coloring
   |> check_js
 
-type optimized_result =
-  { program : Code.program
-  ; variable_uses : Deadcode.variable_uses
-  ; trampolined_calls : Effects.trampolined_calls
-  ; in_cps : Effects.in_cps
-  ; deadcode_sentinal : Code.Var.t
-  }
-
 let optimize ~profile p =
   let deadcode_sentinal =
     (* If deadcode is disabled, this field is just fresh variable *)
@@ -704,22 +703,14 @@ let optimize ~profile p =
   { program; variable_uses; trampolined_calls; in_cps; deadcode_sentinal }
 
 let full ~standalone ~wrap_with_fun ~profile ~link ~source_map ~formatter d p =
-  let { program; variable_uses; trampolined_calls; in_cps; deadcode_sentinal } =
-    optimize ~profile p
-  in
+  let optimized_code = optimize ~profile p in
   let exported_runtime = not standalone in
   let emit formatter =
-    generate
-      d
-      ~exported_runtime
-      ~wrap_with_fun
-      ~warn_on_unhandled_effect:standalone
-      ~deadcode_sentinal
+    generate d ~exported_runtime ~wrap_with_fun ~warn_on_unhandled_effect:standalone
     +> link_and_pack ~standalone ~wrap_with_fun ~link
     +> output formatter ~source_map ()
   in
-  let source_map = emit formatter ((program, variable_uses), trampolined_calls, in_cps) in
-  source_map
+  emit formatter optimized_code
 
 let full_no_source_map ~formatter ~standalone ~wrap_with_fun ~profile ~link d p =
   let (_ : Source_map.t option) =
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -670,7 +670,19 @@ let link_and_pack ?(standalone = true) ?(wrap_with_fun = `Iife) ?(link = `No) p 
   |> coloring
   |> check_js
 
-let optimize ~profile ~deadcode_sentinal p =
+type optimized_result =
+  { program : Code.program
+  ; variable_uses : Deadcode.variable_uses
+  ; trampolined_calls : Effects.trampolined_calls
+  ; in_cps : Effects.in_cps
+  ; deadcode_sentinal : Code.Var.t
+  }
+
+let optimize ~profile p =
+  let deadcode_sentinal =
+    (* If deadcode is disabled, this field is just fresh variable *)
+    Code.Var.fresh_n "dummy"
+  in
   let opt =
     specialize_js_once
     +> (match profile with
@@ -687,16 +699,14 @@ let optimize ~profile ~deadcode_sentinal p =
   in
   if times () then Format.eprintf "Start Optimizing...@.";
   let t = Timer.make () in
-  let r = opt p in
+  let (program, variable_uses), trampolined_calls, in_cps = opt p in
   let () = if times () then Format.eprintf " optimizations : %a@." Timer.print t in
-  r
+  { program; variable_uses; trampolined_calls; in_cps; deadcode_sentinal }
 
 let full ~standalone ~wrap_with_fun ~profile ~link ~source_map ~formatter d p =
-  let deadcode_sentinal =
-    (* If deadcode is disabled, this field is just fresh variable *)
-    Code.Var.fresh_n "dummy"
+  let { program; variable_uses; trampolined_calls; in_cps; deadcode_sentinal } =
+    optimize ~profile p
   in
-  let r = optimize ~profile ~deadcode_sentinal p in
   let exported_runtime = not standalone in
   let emit formatter =
     generate
@@ -708,7 +718,7 @@ let full ~standalone ~wrap_with_fun ~profile ~link ~source_map ~formatter d p =
     +> link_and_pack ~standalone ~wrap_with_fun ~link
     +> output formatter ~source_map ()
   in
-  let source_map = emit formatter r in
+  let source_map = emit formatter ((program, variable_uses), trampolined_calls, in_cps) in
   source_map
 
 let full_no_source_map ~formatter ~standalone ~wrap_with_fun ~profile ~link d p =
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -658,13 +658,34 @@ let configure formatter =
   Code.Var.set_pretty (pretty && not (Config.Flag.shortvar ()));
   Code.Var.set_stable (Config.Flag.stable_var ())
 
-let full ~standalone ~wrap_with_fun ~profile ~link ~source_map formatter d p =
-  let exported_runtime = not standalone in
+type 'a target =
+  | JavaScript : Pretty_print.t -> Source_map.t option target
+  | Wasm
+      : (Deadcode.variable_uses * Effects.in_cps * Code.program * Parse_bytecode.Debug.t)
+        target
+
+let link_and_pack ?(standalone = true) ?(wrap_with_fun = `Iife) ?(link = `No) p =
   let export_runtime =
     match link with
     | `All | `All_from _ -> true
     | `Needed | `No -> false
   in
+  p
+  |> link' ~export_runtime ~standalone ~link
+  |> pack ~wrap_with_fun ~standalone
+  |> coloring
+  |> check_js
+
+let full
+    (type result)
+    ~(target : result target)
+    ~standalone
+    ~wrap_with_fun
+    ~profile
+    ~link
+    ~source_map
+    d
+    p : result =
   let deadcode_sentinal =
     (* If deadcode is disabled, this field is just fresh variable *)
     Code.Var.fresh_n "dummy"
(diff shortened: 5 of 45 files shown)


hhugo:
Turning into a draft until underlying PRs are merged

hhugo:
@OlivierNicole, can you rebase, fix and undraft this PR

OlivierNicole:
> @OlivierNicole, can you rebase, fix and undraft this PR

Done.

compiler/lib/eval.ml
```diff
@@ -170,14 +187,18 @@ type is_int =
...
-let is_int info x =
+let is_int ~target info x =
   match x with
   | Pv x ->
       get_approx
         info
         (fun x ->
           match Flow.Info.def info x with
           | Some (Constant (Int _)) -> Y
+          | Some (Constant (NativeInt _ | Int32 _)) -> (
```
hhugo, line 198: I though it NativeInt and Int32 were not used with the js backend ..

OlivierNicole: Well spotted. I replaced these with assertions that the backend is Wasm instead.

compiler/lib/eval.ml
```diff
@@ -84,20 +84,32 @@ let eval_prim x =
...
+      | "%int_or", _ -> int_binop l wrap Int.logor
+      | "%int_xor", _ -> int_binop l wrap Int.logxor
+      | "%int_lsl", _ -> shift l wrap Fun.id Int.shift_left
+      | "%int_lsr", _ ->
+          shift
+            l
+            wrap
+            (match target with
+            | `JavaScript -> Fun.id
+            | `Wasm -> fun i -> Int.logand i 0x7fffffffl)
```
hhugo, line 110: Why do we need this logand ?

compiler/lib/eval.ml
```diff
@@ -330,7 +353,14 @@ let eval_instr info ((x, loc) as i) =
       | None -> [ i ])
   | Let (x, Prim (Extern "caml_sys_const_backend_type", [ _ ])) ->
       let jsoo = Code.Var.fresh () in
-      [ Let (jsoo, Constant (String "js_of_ocaml")), noloc
+      [ ( Let
+            ( jsoo
+            , Constant
+                (String
+                   (match target with
```
hhugo, line 360: I would lift this match up next to the `let jsoo = ...` above to improve overall formatting

compiler/lib/linker.ml
```diff
@@ -437,7 +437,7 @@ let list_all ?from () =
     provided
     StringSet.empty
 
-let load_fragment ~target_env ~filename (f : Fragment.t) =
+let load_fragment ~ignore_always_annotation ~target_env ~filename (f : Fragment.t) =
```
hhugo, line 440: Can you explain the usecase for ignore always annotations?

compiler/lib/eval.ml
OlivierNicole: I’m not sure. @vouillon, could you explain why?

compiler/lib/eval.ml
OlivierNicole: Done.

hhugo:
Do we really need to pass target around ? It creates a lot of noise for few uses only. It could be turned into an global config like `use_js_string` or `effect`. Alternatively, we could move all configs into a single config  value and pass it around explicitly.

compiler/lib/linker.ml
OlivierNicole: @vouillon

OlivierNicole:
What I like about passing `target` around is that it makes it explicit in the type of functions that they depend on the target, and which backends they support.

The readability does not seem too much impaired to me, but it’s not a strong opinion, if there’s opposition to it I’m OK with a global config parameter, since it’s not a value that is expected to change during a run.

TyOverby:
Wouldn’t a global flag make it harder to use the jsoo compiler as a library, which could be invoked many times with different targets?

TyOverby:
For example, an online ocaml repl

compiler/lib/eval.ml
vouillon: The number `-1` is represented as the 32-bit integer `0xffffffff`, with all bits sets. If we shift it to the right by one, we should get the integer `0x3fffffff` (30 bits remain set), but without clearing the sign bit first we get the incorrect value `0x7fffffff`.

compiler/lib/linker.ml
vouillon: We are currently linking the JavaScript runtime. But we don't want to include the last lines of `runtime/sys.js`

I'm not sure whether this really make sense, though... Maybe we should just link a file with the primitives we care about (`caml_js_on_ie`, `caml_js_html_escape` and `caml_js_html_entities`).

compiler/lib/eval.ml
hhugo: Maybe we should introduce a new module Int_target and use it inside constant `Int of Int_target.t

compiler/lib/eval.ml
```diff
@@ -31,14 +31,14 @@ let get_static_env s = try Some (Hashtbl.find static_env s) with Not_found -> No
 
 module Int = Int32
```
hhugo, line 32: This could be moved inside eval_prim as a first class module based on backend using Int31 when backend is wasm. One would have to complete Int31. Having such a standalone implementation will help with testing.

compiler/lib/linker.ml
hhugo: Do we really want to enforce this behavior? Another way to do this would be to add a new annotations to the js stubs `//Target: Wasm`.

compiler/lib/eval.ml
OlivierNicole: > The number `-1` is represented as the 32-bit integer `0xffffffff`, with all bits sets.

It seems to be one of the possible choices. I would be curious to know the rationale for it.

vouillon: If you leave the sign bit cleared, some operations, like `(asr)`, `(/)`, or simply printing an integer, get more complicated.

compiler/lib/eval.ml
```diff
@@ -31,14 +31,14 @@ let get_static_env s = try Some (Hashtbl.find static_env s) with Not_found -> No
...
   match l with
-  | [ Int i; Int j ] -> Some (Int (f i j))
+  | [ Int i; Int j ] -> Some (Int (w (f i j)))
   | _ -> None
 
-let shift l f =
+let shift l w t f =
   match l with
-  | [ Int i; Int j ] -> Some (Int (f i (Int32.to_int j land 0x1f)))
+  | [ Int i; Int j ] -> Some (Int (w (f (t i) (Int32.to_int j land 0x1f))))
```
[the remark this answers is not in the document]
OlivierNicole: I got it, it avoids an exception when the offset is not in [0, 31].

compiler/lib/eval.ml
OlivierNicole: Done.

compiler/lib/linker.ml
vouillon: I think we should just remove this change.
See https://github.com/ocaml-wasm/wasm_of_ocaml/pull/72

hhugo:
@OlivierNicole, the CI is unhappy, and the diff has extra lines because of formatting. Can you run `dune fmt` ?

OlivierNicole:
There was a subtyping relation not being automatically used under 4.12. The CI should be happy now.

compiler/lib/linker.ml
vouillon: Actually, we need to parse the JavaScript runtime files to get the primitive annotations.
I think I'm going to add a flag `//If: wasm`

vouillon: See #1674

compiler/lib/specialize_js.ml
```diff
@@ -23,47 +23,49 @@ open Code
...
+  match i, Config.target () with
+  | Let (x, Prim (Extern "caml_format_int", [ y; z ])), `JavaScript -> (
       match the_string_of info y with
       | Some "%d" -> (
           match the_int info z with
           | Some i -> Let (x, Constant (String (Int32.to_string i)))
           | None -> Let (x, Prim (Extern "%caml_format_int_special", [ z ])))
       | _ -> i)
-  | Let (x, Prim (Extern "%caml_format_int_special", [ z ])) -> (
+  | Let (x, Prim (Extern "%caml_format_int_special", [ z ])), `JavaScript -> (
```
hhugo, line 34: Why is this disabled in wasm ?

compiler/lib/specialize_js.ml
```diff
@@ -23,47 +23,49 @@ open Code
 open Flow
 
 let specialize_instr info i =
-  match i with
-  | Let (x, Prim (Extern "caml_format_int", [ y; z ])) -> (
+  match i, Config.target () with
+  | Let (x, Prim (Extern "caml_format_int", [ y; z ])), `JavaScript -> (
```
hhugo, line 27: Why is this disabled ?

compiler/lib/driver.mli
```diff
@@ -20,16 +20,22 @@
 
 type profile
 
+type 'a target =
```
hhugo, line 23: I'm not convinced by the use of GADT here.
Can't you just export a new function with what you need ?

vouillon: Indeed, I think one can define a function `optimize` that corresponds to the initial part of function `full` and use it to define a Wasm-specific funtion.

compiler/lib/driver.ml
```diff
@@ -677,56 +679,59 @@ let full ~standalone ~wrap_with_fun ~profile ~link ~source_map formatter d p =
...
+  let t = Timer.make () in
+  let r = opt p in
+  let () = if times () then Format.eprintf " optimizations : %a@." Timer.print t in
+  r
+
+let full ~standalone ~wrap_with_fun ~profile ~link ~source_map ~formatter d p =
+  let deadcode_sentinal =
+    (* If deadcode is disabled, this field is just fresh variable *)
+    Code.Var.fresh_n "undef"
+  in
```
vouillon, line 698: This should be in function `optimize`.

compiler/lib/driver.ml
```diff
@@ -677,56 +679,59 @@ let full ~standalone ~wrap_with_fun ~profile ~link ~source_map formatter d p =
        | O3 -> o3)
     +> exact_calls ~deadcode_sentinal profile
     +> effects ~deadcode_sentinal
-    +> map_fst (if Config.Flag.effects () then fun x -> x else Generate_closure.f)
+    +> map_fst
+         (match target with
```
vouillon, line 683:
```suggestion
         (match Config.target () with
```

compiler/lib/driver.ml
OlivierNicole: Are you sure? It’s used as an argument to generate below, as well.

compiler/lib/driver.ml
OlivierNicole: Done

compiler/lib/config.mli
```diff
@@ -104,3 +105,13 @@ module Param : sig
 
   val lambda_lifting_baseline : unit -> int
 end
+
+(****)
+
+(** {2 Parameters that are constant across a program run} *)
+
+(** These parameters should be set at most once at the beginning of the program. *)
```
hhugo, line 113: Maybe we should make `target` fail if `set_target` was not called first. That way we can spot uninitialized codepath.

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -285,7 +285,7 @@ let run
   | `None ->
       let prims = Linker.list_all () |> StringSet.elements in
       assert (List.length prims > 0);
-      let code, uinfo = Parse_bytecode.predefined_exceptions () in
+      let code, uinfo = Parse_bytecode.predefined_exceptions ~target:`JavaScript in
```
hhugo, line 288: Why do we pass `~target` explicitly here instead of looking at `Config.target` ? Is `Parse_bytecode` different from other modules ? If we end up deciding to keep it this way, we could define/bind target at the top of this file

compiler/lib/specialize_js.ml
```diff
@@ -23,47 +23,49 @@ open Code
 open Flow
 
 let specialize_instr info i =
-  match i with
-  | Let (x, Prim (Extern "caml_format_int", [ y; z ])) -> (
+  match i, Config.target () with
```
hhugo, line 26: We should probably call `Config.target ()` once per call to `f ()`

compiler/lib/eval.ml
```diff
@@ -84,21 +126,26 @@ let eval_prim x =
...
-      | "%int_neg", [ Int i ] -> Some (Int (Int.neg i))
+      | "%direct_int_div", _ -> Int.int_binop l Int.div
+      | "%direct_int_mod", _ -> Int.int_binop l Int.rem
+      | "%int_and", _ -> Int.int_binop l Int.logand
+      | "%int_or", _ -> Int.int_binop l Int.logor
+      | "%int_xor", _ -> Int.int_binop l Int.logxor
+      | "%int_lsl", _ -> Int.shift_op l Int.shift_left
+      | "%int_lsr", _ -> Int.shift_op l Int.shift_right_logical
+      | "%int_asr", _ -> Int.shift_op l Int.shift_right
+      | "%int_neg", _ -> Int.int_unop l Int.neg
```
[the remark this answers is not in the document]
hhugo: and `to_int`

compiler/lib/eval.ml
```diff
@@ -143,7 +190,12 @@ let eval_prim x =
           | Some env -> Some (String env)
           | None -> None)
       | "caml_sys_const_word_size", [ _ ] -> Some (Int 32l)
-      | "caml_sys_const_int_size", [ _ ] -> Some (Int 32l)
+      | "caml_sys_const_int_size", [ _ ] ->
+          Some
+            (Int
```
hhugo, line 195: numbit should live in the `Int` module

compiler/lib/eval.ml
```diff
@@ -359,14 +422,17 @@ let eval_instr info ((x, loc) as i) =
...
                 , Prim
                     ( prim
-                    , List.map2 prim_args prim_args' ~f:(fun arg c ->
-                          match c with
-                          | Some ((Int _ | Float _ | NativeString _) as c) -> Pc c
-                          | Some (String _ as c) when Config.Flag.use_js_string () -> Pc c
-                          | Some _
+                    , List.map2 prim_args prim_args' ~f:(fun arg (c : constant option) ->
+                          match c, Config.target () with
+                          | ( Some ((Int _ | Int32 _ | NativeInt _ | NativeString _) as c)
```
hhugo, line 427: Do we really want to have Int32 and NativeInt in this branch ?

compiler/lib/eval.ml
```diff
@@ -84,21 +126,26 @@ let eval_prim x =
   | Ult, [ Int i; Int j ] -> bool (Int32.(j < 0l) || Int32.(i < j))
   | Extern name, l -> (
       let name = Primitive.resolve name in
+      let (module Int : Int) =
+        match Config.target () with
```
hhugo, line 130: `Config.target ()` could be called once per call to `f ()`

compiler/lib/driver.ml
vouillon: Maybe it should be returned by `optimize`, then.

compiler/lib/driver.mli
OlivierNicole: Done

compiler/lib/driver.ml
OlivierNicole: Done

compiler/lib/config.mli
OlivierNicole: Good idea, done.

compiler/bin-js_of_ocaml/compile.ml
OlivierNicole: This was an oversight on my part. Fixed.

compiler/lib/specialize_js.ml
OlivierNicole: Done

compiler/lib/eval.ml
OlivierNicole: Done

compiler/lib/eval.ml
OlivierNicole: Fixed

compiler/lib/eval.ml
OlivierNicole: Done

compiler/lib/eval.ml
```diff
@@ -127,8 +137,15 @@ let eval_prim x =
...
-        | `JavaScript -> (module Int32)
+        match target with
+        | `JavaScript ->
+            (module struct
+              include Int32
+
+              let of_int32_warning_on_overflow = Fun.id
+
+              let to_int32 = Fun.id
+            end)
```
vouillon, line 148: One should define this module at toplevel rather than allocating a module each time.

compiler/lib/parse_bytecode.ml
```diff
@@ -3099,7 +3134,7 @@ let from_channel ic =
           `Exe
       | _ -> raise Magic_number.(Bad_magic_number (to_string magic)))
 
-let predefined_exceptions () =
+let predefined_exceptions =
```
hhugo, line 3137: The unit needs to stay

compiler/lib/eval.ml
```diff
@@ -158,7 +175,9 @@ let eval_prim x =
       | "caml_mul_float", _ -> float_binop l ( *. )
       | "caml_div_float", _ -> float_binop l ( /. )
       | "caml_fmod_float", _ -> float_binop l mod_float
-      | "caml_int_of_float", [ Float f ] -> Some (Int (Int32.of_float f))
+      | "caml_int_of_float", [ Float f ] ->
+          Some
+            (Int (Int32.of_float f |> Int.of_int32_warning_on_overflow |> Int.to_int32))
       | "to_int", [ Float f ] -> Some (Int (Int32.of_float f))
```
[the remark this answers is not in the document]
OlivierNicole: The `Int` constructor takes an `int32`, to be able to hold both JavaScript 32-bit integers and Wasm int31s

compiler/lib/eval.ml
OlivierNicole: Done

compiler/lib/parse_bytecode.ml
OlivierNicole: Done

compiler/lib/eval.ml
```diff
@@ -29,17 +29,80 @@ let set_static_env s value = Hashtbl.add static_env s value
...
+    match l with
+    | [ Int i; Int j ] ->
+        Some (Int (to_int32 (f (of_int32_warning_on_overflow i) (Int32.to_int j))))
+    | _ -> None
+
+  let numbits = 31
+end
+
+let m_int : (module Int) =
+  match Config.target () with
```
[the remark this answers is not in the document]
OlivierNicole: Ah, yes, that will not be evaluated in the right order

OlivierNicole:
I just pushed a commit that fixes `Flow.constant_identical` with with Wasm backend.

compiler/lib/config.ml
```diff
@@ -183,8 +183,11 @@ end
 
 (****)
 
-let target_ : [ `JavaScript | `Wasm ] ref = ref `JavaScript
+let target_ : [ `JavaScript | `Wasm ] option ref = ref (Some `JavaScript)
```
OlivierNicole, line 186: Oops, this should be None. Fixed.

compiler/lib/config.ml
```diff
@@ -183,8 +183,11 @@ end
 
 (****)
 
-let target_ : [ `JavaScript | `Wasm ] ref = ref `JavaScript
+let target_ : [ `JavaScript | `Wasm ] option ref = ref None
```
hhugo, line 186:
```suggestion
let target_ : [ `JavaScript | `Wasm | `None ]  ref = ref `None
```

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -90,6 +90,7 @@ let run
     } =
   let include_cmis = toplevel && not no_cmis in
   let custom_header = common.Jsoo_cmdline.Arg.custom_header in
+  Config.set_target `JavaScript;
```
hhugo, line 93: set_target is already called in `js_of_ocaml.ml`

compiler/lib/eval.ml
hhugo: It doesn't seem to be committed

compiler/lib/eval.ml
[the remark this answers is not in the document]
hhugo: What I mean is that a conversion from Float to Int is different for 31 and 32 bits integer. 

That said, I don't know why we have branches for `"to_int"`, maybe @vouillon knows ?

compiler/lib/eval.ml
```diff
@@ -170,14 +233,17 @@ type is_int =
   | N
   | Unknown
 
-let is_int info x =
+let is_int ~target info x =
```
hhugo, line 236: passing `target` here doesn't look very useful, we use it to check for an invariants that could be moved elsewhere. (e.g. `Code.invariant`)

compiler/lib/ocaml_compiler.ml
```diff
@@ -35,9 +39,12 @@ let rec constant_of_const : _ -> Code.constant =
...
       Float_array (Array.of_list l)
   | ((Const_pointer i) [@if ocaml_version < (4, 12, 0)]) ->
-      Int (Int32.of_int_warning_on_overflow i)
+      Int
+        (match Config.target () with
+        | `JavaScript -> Int32.of_int_warning_on_overflow i
+        | `Wasm -> Int31.(of_int_warning_on_overflow i |> to_int32))
   | Const_block (tag, l) ->
-      let l = Array.of_list (List.map l ~f:constant_of_const) in
+      let l = Array.of_list (List.map l ~f:(fun c -> constant_of_const c)) in
```
hhugo, line 47:
```suggestion
      let l = Array.of_list (List.map l ~f:constant_of_const) in
```

compiler/lib/ocaml_compiler.ml
```diff
@@ -35,9 +39,12 @@ let rec constant_of_const : _ -> Code.constant =
       let l = List.map ~f:(fun f -> float_of_string f) sl in
       Float_array (Array.of_list l)
   | ((Const_pointer i) [@if ocaml_version < (4, 12, 0)]) ->
-      Int (Int32.of_int_warning_on_overflow i)
```
hhugo, line 38: (Does wasm_of_ocaml support ocaml < 4.12 ? the opam file says yes, but the CI doesn't test it.)

compiler/lib/parse_bytecode.ml
```diff
@@ -764,7 +767,24 @@ let access_global g i =
...
+  then (
+    let name =
+      match g.named_value.(i) with
+      | None -> assert false
+      | Some name -> name
+    in
+    Code.Var.name (access_global g i) name;
+    ( Let
+        ( Var.fresh ()
+        , Prim (Extern "caml_set_global", [ Pc (String name); Pv (access_global g i) ]) )
```
hhugo, line 784: Can we try to unify the use of `caml_set_global` and `caml_register_global` ?

compiler/lib/parse_bytecode.ml
```diff
@@ -789,18 +809,33 @@ let get_global state instrs i loc =
...
         x, state, instrs)
+      else
+        match g.named_value.(i) with
+        | None -> assert false
+        | Some name ->
+            let x, state = State.fresh_var state loc in
+            if debug_parser () then Format.printf "%a = get_global(%s)@." Var.print x name;
+            ( x
+            , state
+            , (Let (x, Prim (Extern "caml_get_global", [ Pc (String name) ])), loc)
```
hhugo, line 837: Do I understand correctly that we don't want to share access to globals ?

compiler/lib/specialize_js.ml
```diff
@@ -22,49 +22,51 @@ open! Stdlib
...
-      , Prim
-          ( Extern (("caml_js_var" | "caml_js_expr" | "caml_pure_js_expr") as prim)
-          , [ (Pv _ as y) ] ) )
+  | ( Let
+        ( x
+        , Prim
+            ( Extern (("caml_js_var" | "caml_js_expr" | "caml_pure_js_expr") as prim)
+            , [ (Pv _ as y) ] ) )
+    , _ )
     when Config.Flag.safe_string () -> (
```
hhugo, line 45: Do you want this with wasm ?

compiler/lib/specialize_js.ml
```diff
@@ -22,49 +22,51 @@ open! Stdlib
...
-      match the_string_of info y with
+      match the_string_of ~target info y with
       | Some s -> Let (x, Prim (Extern prim, [ Pc (String s) ]))
       | _ -> i)
-  | Let (x, Prim (Extern ("caml_register_named_value" as prim), [ y; z ])) -> (
-      match the_string_of info y with
+  | Let (x, Prim (Extern ("caml_register_named_value" as prim), [ y; z ])), `JavaScript
```
hhugo, line 49: Do you need to drop this feature with wasm ?

compiler/lib/specialize_js.ml
```diff
@@ -109,51 +111,52 @@ let specialize_instr info i =
...
       | Some _ | None -> i)
-  | Let (x, Prim (Extern "caml_jsbytes_of_string", [ y ])) -> (
-      match the_string_of info y with
+  | Let (x, Prim (Extern "caml_jsbytes_of_string", [ y ])), _ -> (
+      match the_string_of ~target info y with
       | Some s -> Let (x, Constant (NativeString (Native_string.of_bytestring s)))
       | None -> i)
-  | Let (x, Prim (Extern "%int_mul", [ y; z ])) -> (
-      match the_int info y, the_int info z with
+  | Let (x, Prim (Extern "%int_mul", [ y; z ])), `JavaScript -> (
```
hhugo, line 136: Is this optimisation useless or does one just need to update the checks below for wasm ?

compiler/lib/stdlib.ml
```diff
@@ -361,6 +405,50 @@ module Int31 = struct
...
+  let of_int32_warning_on_overflow n =
+    Int32.convert_warning_on_overflow
+      ~to_int32:(fun i -> wrap i)
+      ~of_int32:Fun.id
+      ~equal:Int32.equal
+      ~to_dec:(Printf.sprintf "%ld")
+      ~to_hex:(Printf.sprintf "%lx")
+      n
+
+  let neg = Int32.neg
```
hhugo, line 418: I don't think this is correct when the argument is min_int

compiler/lib/eval.ml
hhugo: Looking more at this, I wonder if it would be simpler to 
- shift_right by 1
- do the Int32 op
- shift_left by 1

compiler/lib/ocaml_compiler.ml
OlivierNicole: In principle, yes, I think so.

compiler/lib/stdlib.ml
```diff
@@ -341,7 +341,51 @@ module Int32 = struct
...
+
+  val shift_left : t -> int -> t
+
+  val shift_right : t -> int -> t
+
+  val shift_right_logical : t -> int -> t
+end
+
+module Int31 : sig
+  type t = private int32
```
hhugo, line 373: Do we really need to expose the private here ?

compiler/lib/eval.ml
hhugo: Similar to how Int63.t is emulated using Int64.t in https://github.com/janestreet/base/blob/master/src/int63_emul.ml, 
we could have Int31.t implemented as a int32 shilfted to the left.

compiler/lib/specialize_js.ml
vouillon: The issue in JavaScript is that when using `*` for multiplying integers, we get a float; and if the float is large enough, we can lose some bits.  So, in the general case, we have to use `Math.imul`. There is no such issue in Wasm.

compiler/lib/parse_bytecode.ml
vouillon: Right. This is compiled to a single `global.get` instruction, which is very cheap.

compiler/lib/parse_bytecode.ml
vouillon: With `wasm_of_ocaml`, `caml_set_global` set a global variable in the generated code, while `caml_register_global` register a value into the runtime system (this is used for exceptions).
I may reconsider this if we try to support dynamic linking, but for now they are doing very different things.

compiler/lib/specialize_js.ml
OlivierNicole: Do you mean that this flag should only matter when emitting JS?

compiler/lib/specialize_js.ml
OlivierNicole: I don’t see why. @vouillon ?

vouillon: I have just tested, and I think we can just drop this change.

vouillon: I mean, we can allow this case when generating Wasm code as well.

compiler/lib/specialize_js.ml
hhugo: Make sense, I think we should document (inside a comment) theses small difference in behavior

compiler/lib/stdlib.ml
OlivierNicole: Well spotted, thanks. Fixed

compiler/lib/stdlib.ml
```diff
@@ -361,6 +405,50 @@ module Int31 = struct
...
+      ~to_int32:(fun i -> wrap i)
+      ~of_int32:Fun.id
+      ~equal:Int32.equal
+      ~to_dec:(Printf.sprintf "%ld")
+      ~to_hex:(Printf.sprintf "%lx")
+      n
+
+  let neg = Int32.neg
+
+  let int_binop wrap f x y = wrap (f x y)
```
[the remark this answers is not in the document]
OlivierNicole: Done

compiler/lib/stdlib.ml
OlivierNicole: Not really. I removed it.

compiler/lib/specialize_js.ml
OlivierNicole: Fixed

compiler/lib/specialize_js.ml
OlivierNicole: Done

compiler/lib/specialize_js.ml
OlivierNicole: Done

compiler/lib/eval.ml
vouillon: This does not look that much simpler to me. There are still a lot of special cases for `(*)`, `(/)`, `(lsr)`, `(asr)`, ...

compiler/lib/specialize_js.ml
hhugo: @OlivierNicole, can you also add a comment explaining why it's disabled with wasm ?

compiler/lib/eval.ml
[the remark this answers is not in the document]
vouillon: It seems there are here from the start (PR #22). I have no idea why.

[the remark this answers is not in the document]
vouillon: I think we can just drop these cases.

compiler/lib/specialize_js.ml
hhugo: I think we don't care at all about safe_string here. `Flow.the_string_of` is supposed to do the work already.

compiler/lib/ocaml_compiler.ml
vouillon: Probably not, actually, since there have been some runtime changes since then.
It should not be difficult to support these older versions of the compiler, but I'm not sure it is worth it. I should just bump the version in the opam file.
It's hard to test with the CI since it installs a version of `ppx_expect` that depends on ocaml 4.14.

vouillon: Anyway, it does not cost much to do the right thing here.

compiler/lib/specialize_js.ml
OlivierNicole: Done

compiler/lib/eval.ml
hhugo: > , with all bits sets. If we shift it to the right by one, we should get the integer `0x3fffffff` (30 bits remain set), but without clearing the sign bit first we get the incorrect value `0x7fffffff`.

I somehow find reasoning about this (2 complement logic) much harder than the understanding theses special cases.
Let's move the discussion next the new implementation in stdlib.

compiler/lib/specialize_js.ml
OlivierNicole: (Tangent: Is `safe_string` about string mutability? And if yes, since it has been dropped since 4.02, maybe we could drop this flag as well?)

compiler/lib/eval.ml
OlivierNicole: Sorry. Now it is.

compiler/lib/stdlib.ml
```diff
@@ -341,7 +341,51 @@ module Int32 = struct
...
+
+  val shift_left : t -> int -> t
+
+  val shift_right : t -> int -> t
+
+  val shift_right_logical : t -> int -> t
+end
+
+module Int31 : sig
+  type t
```
OlivierNicole, line 373: Regarding the implementation of `Int31`: the two’s complement implementation seems rather simple to me. IMO the it should ideally come with randomized property-based tests, regardless of the implementation.

compiler/lib/eval.ml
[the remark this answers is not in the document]
OlivierNicole: Cases dropped.

> What I mean is that a conversion from Float to Int is different for 31 and 32 bits integer.

Given that `Int` here is either `Int32` or `Int31`, I think the result of these three functions is the expected one (i.e. if the backend is Wasm and the float does not fit on 31 bits, the middle function will complain).

compiler/lib/specialize_js.ml
hhugo: I think it's 4.10

> OCaml 4.10.0 (21 February 2020)
> -------------------------------
> ### Language features
> ...
> * #1859, #9117: enforce safe (immutable) strings by removing
  the -unsafe-string option by default. This can be overridden by
  a configure-time option (available since 4.04 in 2016):
  --disable-force-safe-string since 4.08, -no-force-safe-since
  between 4.07 and 4.04.
  In the force-safe-string mode (now the default), the return type of the
  String_val macro in C stubs is `const char*` instead of
  `char*`. This change may break C FFI code.
  (Kate Deplaix)

hhugo: or even OCaml 5 for 
> * #10893: Remove configuration options --disable-force-safe-string and
  DEFAULT_STRING=unsafe as well as --enable-force-safe-string and
  DEFAULT_STRING=safe which are now the default unconditionally.
  (Kate Deplaix, review by Gabriel Scherer and David Allsopp)

compiler/lib/specialize_js.ml
```diff
@@ -22,49 +22,53 @@ open! Stdlib
...
-          , [ (Pv _ as y) ] ) )
-    when Config.Flag.safe_string () -> (
-      match the_string_of info y with
+  | ( Let
+        ( x
+        , Prim
+            ( Extern (("caml_js_var" | "caml_js_expr" | "caml_pure_js_expr") as prim)
+            , [ (Pv _ as y) ] ) )
+    , target )
+    when Poly.equal target `Wasm || Config.Flag.safe_string () -> (
```
hhugo, line 48: just a reminder (because I'm a bit lost with all the threads), the whole guard can probably be dropped.

compiler/lib/eval.ml
```diff
@@ -359,14 +432,21 @@ let eval_instr info ((x, loc) as i) =
...
                     ( prim
-                    , List.map2 prim_args prim_args' ~f:(fun arg c ->
-                          match c with
-                          | Some ((Int _ | Float _ | NativeString _) as c) -> Pc c
-                          | Some (String _ as c) when Config.Flag.use_js_string () -> Pc c
-                          | Some _
+                    , List.map2 prim_args prim_args' ~f:(fun arg (c : constant option) ->
+                          match c, target with
+                          | Some ((Int _ | NativeString _) as c), _ -> Pc c
+                          | Some ((Int32 _ | NativeInt _) as c), `Wasm -> Pc c
```
hhugo, line 438:
```suggestion
                          | Some ((Int32 _ | NativeInt _) as c), `Wasm -> arg
```

compiler/lib/specialize_js.ml
OlivierNicole: Done

compiler/lib/eval.ml
OlivierNicole: Why not use the statically evaluated constant in Wasm?

compiler/lib/parse_bytecode.ml
vouillon: I have added a comment.

compiler/lib-dynlink/js_of_ocaml_compiler_dynlink.ml
```diff
@@ -26,6 +26,7 @@ let normalize_bytecode code =
       Bytes.to_string b
 
 let () =
+  Config.set_target `JavaScript;
```
hhugo, line 29: Maybe this should match on Sys.backend_type and fail if not `js_of_ocaml`

compiler/lib/code.ml
```diff
@@ -909,15 +909,28 @@ let invariant { blocks; start; _ } =
...
     let check_expr = function
       | Apply _ -> ()
       | Block (_, _, _, _) -> ()
       | Field (_, _, _) -> ()
       | Closure (l, cont) ->
           List.iter l ~f:define;
           check_cont cont
+      | Constant (NativeInt _ | Int32 _) ->
+          assert (
+            match Config.target () with
```
hhugo, line 929: This could be shared with a check_constant helper

compiler/lib/driver.ml
```diff
@@ -677,31 +691,39 @@ let full ~standalone ~wrap_with_fun ~profile ~link ~source_map formatter d p =
...
-    +> check_js
+    +> link_and_pack ~standalone ~wrap_with_fun ~link
     +> output formatter ~source_map ()
   in
-  if times () then Format.eprintf "Start Optimizing...@.";
-  let t = Timer.make () in
-  let r = opt p in
-  let () = if times () then Format.eprintf " optimizations : %a@." Timer.print t in
-  emit r
+  let source_map = emit formatter ((program, variable_uses), trampolined_calls, in_cps) in
```
hhugo, line 721: The nested tuples doesn't make sense anymore, maybe one should just pass an optimized result.

compiler/lib/eval.ml
hhugo: Wouldn't that duplicate the allocation ?

compiler/lib/flow.ml
```diff
@@ -343,30 +343,31 @@ let the_def_of info x =
...
-  | Float a, Float b -> Float.bitwise_equal a b
-  | NativeString a, NativeString b -> Native_string.equal a b
-  | String a, String b -> Config.Flag.use_js_string () && String.equal a b
-  | Int _, Float _ | Float _, Int _ -> false
+let constant_identical ~(target : [ `JavaScript | `Wasm ]) a b =
+  match a, b, target with
+  | Int i, Int j, _ -> Int32.equal i j
+  | Float a, Float b, `JavaScript -> Float.bitwise_equal a b
+  | Float _, Float _, `Wasm -> false
+  | NativeString a, NativeString b, `JavaScript -> Native_string.equal a b
```
hhugo, line 351: Can you add a ``` | NativeString a, NativeString b, `Wasm ``` branch to make it clear what is going on with Wasm ? same string for String.

hhugo: and Int32 and NativeInt

compiler/lib/eval.ml
OlivierNicole: Ah, you’re right. Done.

compiler/lib/code.ml
OlivierNicole: Done

compiler/lib/driver.ml
OlivierNicole: It’s indeed better. Done.

compiler/lib/flow.ml
OlivierNicole: Done

compiler/lib/eval.ml
```diff
@@ -359,14 +432,24 @@ let eval_instr info ((x, loc) as i) =
...
                 , Prim
                     ( prim
-                    , List.map2 prim_args prim_args' ~f:(fun arg c ->
-                          match c with
-                          | Some ((Int _ | Float _ | NativeString _) as c) -> Pc c
-                          | Some (String _ as c) when Config.Flag.use_js_string () -> Pc c
-                          | Some _
+                    , List.map2 prim_args prim_args' ~f:(fun arg (c : constant option) ->
+                          match c, target with
+                          | Some ((Int _ | NativeString _) as c), _ -> Pc c
```
hhugo, line 437: NativeString with the Wasm backend should probably be excluded as well if I understand right.

compiler/lib/ocaml_compiler.ml
```diff
@@ -18,11 +18,15 @@
...
   let open Lambda in
   let open Asttypes in
-  function
-  | Const_base (Const_int i) -> Int (Int32.of_int_warning_on_overflow i)
+  match c with
+  | Const_base (Const_int i) ->
+      Int
+        (match Config.target () with
+        | `JavaScript -> Int32.of_int_warning_on_overflow i
+        | `Wasm -> Int31.(of_int_warning_on_overflow i |> to_int32))
```
hhugo, line 29: `Const_base (Const_int32 i) -> Int i` and `| Const_base (Const_nativeint i) ` below need to be updated.

compiler/lib/parse_bytecode.ml
```diff
@@ -507,7 +507,10 @@ end = struct
       else assert false
     else
       let i : int = Obj.magic x in
-      Int (Int32.of_int_warning_on_overflow i)
```
hhugo, line 510: `| Some name when same_ident name ident_32 -> Int (Obj.magic x : int32)` needs an update

hhugo: `| Some name when same_ident name ident_native ->` as well

compiler/lib/parse_bytecode.ml
hhugo: Ok, I've pushed some small refactoring, can you check I didn't change the expected semantic ?

compiler/lib/eval.ml
vouillon: We have changed the OCaml physical equality `(==)` to use the JavaScript equality `===` on JavaScript values, so this should actually be correct.

compiler/lib/stdlib.ml
hhugo: @OlivierNicole, can you take care of adding tests ?

compiler/lib/ocaml_compiler.ml
OlivierNicole: True. Done

compiler/lib/parse_bytecode.ml
OlivierNicole: True. Done.

compiler/lib/eval.ml
hhugo: Reading `Flow.constant_identical`, is seems that NativeString are boxed. Inlining it here could duplicate allocation similar to Int32 and NativeInt. Am i missing something ?

OlivierNicole: That seems right. I restricted the inlining to Javascript.

compiler/lib/stdlib.ml
OlivierNicole: The tests I have in mind would add a dependency to `qcheck` though. Is that OK?

compiler/lib/parse_bytecode.ml
hhugo: Force is used to select which branch to follow.

compiler/lib/stdlib.ml
hhugo: We could make the test conditional on qcheck.

compiler/lib/parse_bytecode.ml
OlivierNicole: If the backend is JS and this branch is taken, `force` can have any value without changing the result.

vouillon: When registering a predefined exception, we set `force` to true so that this branch is taken.

hhugo: To make sure we read the same thing, here is the branch ```| true, _, (`JavaScript as target) | false, true, ((`Wasm | `JavaScript) as target) ->```

The branch is taken if exported or force is true. If force is true, it means we're registering an exception

compiler/lib/stdlib.ml
OlivierNicole: I’m not sure how to do that.

OlivierNicole: @hhugo I added property-based tests in `compiler/tests-num/test_int31.ml`. They actually caught a bug in my implementation of `Int31.shift_right_logical`.

hhugo: Thanks, I'm now very confused but let's move the discussion elsewhere.

compiler/lib/stdlib.ml
```diff
@@ -447,8 +447,7 @@ end = struct
 
   let shift_right = shift_op Int32.shift_right
 
-  let shift_right_logical a b =
-    shift_op Int32.shift_right_logical (Int32.logand a 0x7fffffffl) b
+  let shift_right_logical a b = shift_op Int32.shift_right_logical a b
```
hhugo, line 450: https://github.com/ocsigen/js_of_ocaml/pull/1655#discussion_r1751937120 explains the logand. Are you sure your tests are correct ?

compiler/tests-num/dune
```diff
@@ -1,12 +1,21 @@
...
  (modes
   js
   (best exe))
  (flags
   (:standard -linkall -w -3-7-33-35-37 -safe-string -no-strict-sequence)))
 
+(library
+ (name test_int31)
+ (modules test_int31)
+ (inline_tests)
```
hhugo, line 14:
```suggestion
 (inline_tests)
 (enabled_if %{lib-available:qcheck})
```

OlivierNicole: Should I then remove qcheck as a `:with-test` dependency?

compiler/lib/stdlib.ml
OlivierNicole: You’re right, my test was wrong for value −1 for instance. I believe my test is more correct now.

compiler/tests-num/dune
hhugo: Keep it for now

OlivierNicole: Looks like the Dune language version is too old. Is it OK to upgrade it? (See last commit)

hhugo: yes

hhugo:
What's the semantics of nativeInt with the wasm backend?

OlivierNicole:
They’re on 32 bits.

hhugo:
Then we should use   | NativeInt of Int32.t

OlivierNicole:
Done.

hhugo:
@OlivierNicole, @vouillon, I'm tempted to introduce a dedicated type & module for target int. What do you think ? 
See https://github.com/ocsigen/js_of_ocaml/pull/1693

OlivierNicole:
Is the idea to have `Targetint.t` be an int31 in Wasm and int32 in JavaScript?

Sure, it sounds better, but IMO this looks like an improvement over the present PR that could be merged separately.

vouillon:
That's a good idea, indeed.

hhugo:
@OlivierNicole @vouillon I've rebased and adapted https://github.com/ocsigen/js_of_ocaml/pull/1693.

OlivierNicole:
Thank you for the exceptional review work.

hhugo:
@OlivierNicole, what's the next step for https://github.com/ocaml-wasm/wasm_of_ocaml/issues/47

OlivierNicole:
I am in the process of merging the JSOO `master` into WSOO cleanly. Once that is done, my hope is to merge WSOO `main` into this repo as soon as possible, to avoid a new divergence and the amount of work involved in resorbing it. But that last step is blocked on https://github.com/ocaml/dune/pull/10695.

PR #1656, opened 2024-07-30
Update typing of typed arrays to support Wasm, and test Typed_array.Bytes

OlivierNicole:
**Note:** This PR depends on #1649 and is based on top of it. ~~The feature change consists in the last two commits.~~ edit: #1649 has been merged.

This:

- Updates the typing of `Typed_array` so that they work with the Wasm backend as well
- Adds tests for the new `Typed_array.Bytes` module.

### DIFF
--- a/lib/js_of_ocaml/typed_array.ml
+++ b/lib/js_of_ocaml/typed_array.ml
@@ -20,7 +20,9 @@
 open! Import
 open Js
 
-type uint32 = float
+type int32 = number_t
+
+type uint32 = number_t
 
 class type arrayBuffer = object
   method byteLength : int readonly_prop
@@ -40,7 +42,7 @@ class type arrayBufferView = object
   method byteLength : int readonly_prop
 end
 
-class type ['a, 'b] typedArray = object
+class type ['a, 'b, 'c] typedArray = object
   inherit arrayBufferView
 
   method _BYTES_PER_ELEMENT : int readonly_prop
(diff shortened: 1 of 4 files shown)


OlivierNicole:
I rebased and marked the PR as ready for review.

hhugo:
Any idea how much breakage the change of API generates ?

lib/js_of_ocaml/typed_array.ml
```diff
@@ -49,47 +51,61 @@ class type ['a, 'b] typedArray = object
...
+type uint32Array = (uint32, Int32.t, Bigarray.int32_elt) typedArray
 
-type uint32Array = (int32, Bigarray.int32_elt) typedArray
+type float32Array = (number_t, float, Bigarray.float32_elt) typedArray
 
-type float32Array = (float, Bigarray.float32_elt) typedArray
+type float64Array = (number_t, float, Bigarray.float64_elt) typedArray
 
-type float64Array = (float, Bigarray.float64_elt) typedArray
+type ('bigarray, 'typed_array, 'elt) type' =
```
hhugo, line 85: Maybe this could be named kind ? similar to `Bigarray.kind`

lib/js_of_ocaml/typed_array.ml
```diff
@@ -49,47 +51,61 @@ class type ['a, 'b] typedArray = object
...
+     ('bigarray, 'elt, Bigarray.c_layout) Bigarray.Genarray.t
+  -> ('typed_array, 'bigarray, 'elt) typedArray t = "caml_ba_to_typed_array"
 
 external to_genarray :
-  ('a, 'b) typedArray t -> ('a, 'b, Bigarray.c_layout) Bigarray.Genarray.t
-  = "caml_ba_from_typed_array"
+     ('typed_array, 'bigarray, 'elt) typedArray t
+  -> ('bigarray, 'elt, Bigarray.c_layout) Bigarray.Genarray.t = "caml_ba_from_typed_array"
+
+let from_genarray _ a = from_genarray_impl a
```
hhugo, line 108: adding a type constraint to the `_` would help !

hhugo:
It seems we could now complete the api with `BigInt64Array` and `BigUint64Array`. One could do that in a subsequent PR.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array

lib/js_of_ocaml/typed_array.ml
OlivierNicole: I also like `kind` better (I updated the code), I’m just wondering whether it can be annoying in practice that both the `Bigarray` and `Typed_array` modules contain a type `kind`.

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Done

OlivierNicole:
> Any idea how much breakage the change of API generates ?

I’m not sure. I suppose we could run an opam-health-check on the PR.

lib/js_of_ocaml/typed_array.ml
```diff
@@ -20,7 +20,9 @@
 open! Import
 open Js
 
-type uint32 = float
+type int32 = number_t
```
hhugo, line 23: I wonder if we should have a different name to avoid confusion with the predefined `int32` type ? would `int32_` be enough ?

lib/js_of_ocaml/typed_array.mli
```diff
@@ -51,41 +53,58 @@ class type ['a, 'b] typedArray = object
...
 
-type uint32Array = (int32, Bigarray.int32_elt) typedArray
+type float32Array = (number_t, float, Bigarray.float32_elt) typedArray
 
-type float32Array = (float, Bigarray.float32_elt) typedArray
+type float64Array = (number_t, float, Bigarray.float64_elt) typedArray
 
-type float64Array = (float, Bigarray.float64_elt) typedArray
+type ('bigarray, 'typed_array, 'elt) kind =
+  | Char : (int, char, Bigarray.int8_unsigned_elt) kind
```
hhugo, line 88: Why is it an int here and not a char ?

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Why not. But in fact, I find the two type synonyms rather confusing and have removed them in the last version. What do you think?

lib/js_of_ocaml/typed_array.mli
OlivierNicole: Good question, I may be mistaken but I see no good reason why not (cc @vouillon). I have changed to `char, char` in the last version, and added a test for a `char` bigarray. One quirk is that converting such a bigarray to a typed array will yield a Uint8Array, but I think this is to be expected given that Javascript does not have CharArray.

lib/js_of_ocaml/typed_array.ml
```diff
@@ -221,13 +233,13 @@ class type dataView = object
 
   method setUint16_ : int -> int -> bool t -> unit meth
 
-  method setInt32 : int -> int -> unit meth
+  method setInt32 : int -> int32 -> unit meth
```
hhugo, line 236: shouldn't this be number_t ?

lib/js_of_ocaml/typed_array.ml
```diff
@@ -193,13 +205,13 @@ class type dataView = object
 
   method getUint16_ : int -> bool t -> int meth
 
-  method getInt32 : int -> int meth
+  method getInt32 : int -> int32 meth
```
hhugo, line 208: shouldn't this be number_t ?

lib/js_of_ocaml/typed_array.ml
```diff
@@ -193,13 +205,13 @@ class type dataView = object
...
 
-  method getInt32_ : int -> bool t -> int meth
+  method getInt32_ : int -> bool t -> int32 meth
```
hhugo, line 210: shouldn't this be number_t ?

lib/js_of_ocaml/typed_array.ml
```diff
@@ -221,13 +233,13 @@ class type dataView = object
 
   method setUint16_ : int -> int -> bool t -> unit meth
 
-  method setInt32 : int -> int -> unit meth
+  method setInt32 : int -> int32 -> unit meth
 
-  method setInt32_ : int -> int -> bool t -> unit meth
+  method setInt32_ : int -> int32 -> bool t -> unit meth
```
hhugo, line 238: shouldn't this be number_t ?

lib/js_of_ocaml/typed_array.ml
```diff
@@ -49,47 +47,61 @@ class type ['a, 'b] typedArray = object
...
+type uint32Array = (number_t, Int32.t, Bigarray.int32_elt) typedArray
 
-type uint32Array = (int32, Bigarray.int32_elt) typedArray
+type float32Array = (number_t, float, Bigarray.float32_elt) typedArray
 
-type float32Array = (float, Bigarray.float32_elt) typedArray
+type float64Array = (number_t, float, Bigarray.float64_elt) typedArray
 
-type float64Array = (float, Bigarray.float64_elt) typedArray
+type ('bigarray, 'typed_array, 'elt) kind =
```
vouillon, line 81:
```suggestion
type ('typed_array, 'bigarray, 'elt) kind =
```

lib/js_of_ocaml/typed_array.mli
OlivierNicole: Why is it unsound?

vouillon: This can be used to turn a float into an int, for instance:
```ocaml
type _ t =
  | A : float -> int t
  | B : int -> char t
  | C : float t
  | D : int32 t
  | E : int64 t
  | F : nativeint t
  | G : Complex.t t

let f (type a b) (x : (a, b) Bigarray.kind) : a t =
  match x with
  | Char -> B 10
  | Int8_unsigned -> A 10. | Int8_signed -> A 10.
  | Int16_unsigned -> A 10. | Int16_signed -> A 10. | Int -> A 10.
  | Float32 -> C | Float64 -> C
  | Int32 -> D
  | Int64 -> E
  | Nativeint -> F
  | Complex32 -> G | Complex64 -> G

let g (x : char t) = match x with B n -> n

let x = (Obj.magic Bigarray.Int8_unsigned : (char, Bigarray.int8_unsigned_elt) Bigarray.kind)

let () = Format.printf "%d@." (g (f x))
```

OlivierNicole: Got it. I just pushed a commit to disallow converting from `Char` bigarrays to typed arrays. It seems to me like the natural solution since there are no such TypedArrays, and the previous API also didn’t allow it.

hhugo:
@OlivierNicole, can you look at Jerome suggestion ? and rebase and squash commits ?

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Fixed (and documented)

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Done

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Done

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Dnoe

lib/js_of_ocaml/typed_array.ml
OlivierNicole: Done

OlivierNicole:
In any case, thank you for the review and @vouillon for the explanations!

hhugo:
And ended up squashing during the merge

hhugo:
This breaks ezjs_crypto. I am adding a constraint in opam. See https://github.com/ocaml/opam-repository/pull/26932/commits/e1c4e8e30ca0cd9014ea023f8f536367601e6986

PR #1657, opened 2024-07-30
Build_info/Unit_info: conversion to and from string maps

OlivierNicole:

### DIFF
--- a/compiler/lib/build_info.ml
+++ b/compiler/lib/build_info.ml
@@ -90,6 +90,10 @@ let parse s =
       in
       Some t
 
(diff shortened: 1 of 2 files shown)


hhugo:
Do we really need to introduce this in jsoo ? With a little bit of helper in jsoo, it seems that most of this could live in wasm_of_ocaml.
Is it enough to just introduce `Build_info.{to_map,of_map}` ?

vouillon:
That should be enough, indeed.

OlivierNicole:
This is ready for a new review, I think

OlivierNicole:
Thanks!

PR #1659, opened 2024-07-31
Document non-trivial function Code.constant_equal, and fix related bugs

OlivierNicole:
I think this will be useful to future us.

(@vouillon excavated the semantics of the function.)


OlivierNicole:
See

https://github.com/ocsigen/js_of_ocaml/blob/3204c33e5f7da4386755522083ac1e5d17e12fab/compiler/lib/code.ml#L404-L405

Is it definitely the case that in Javascript `a == b` is false when `a` is an `Int` (whatever that means) and `b` is an `Int64`?

vouillon:
An int64 is implemented as a JavaScript object, which is never equal to a JavaScript number.

compiler/lib/code.mli
```diff
@@ -185,6 +185,9 @@ type constant =
   | Tuple of int * constant array * array_or_not
   | Int of int32
 
+(** Guaranteed equality in terms of both OCaml [(=)] and Javascript [(==)]: if
+    [constant_equal a b = Some v], then [Poly.(=) a b = v] and
+    [Js_of_ocaml.Js.equals a b = v]. This is used for optimization purposes. *)
```
hhugo, line 190: I don't think the second part is correct. two strings, int64, tuple, or floatArray might be equal under constant_equal and different under Js_of_ocaml.Js.equals.
`caml_js_equals` in `eval_instr` looks incorrect.

vouillon:
Also `Float.equal 0. (-0.)` is false. So `(0., 0.) = (-0., 0.)` is miscompiled into `true`.

I think the fix in  #1048 is wrong and we should not use `Code.constant_equal` in `Flow.the_const_of`.

hhugo:
> Also `Float.equal 0. (-0.)` is false. So `(0., 0.) = (-0., 0.)` is miscompiled into `true`.

I'm guessing you mean
```
Also `Float.equal 0. (-0.)` is false. So `(0., 0.) = (-0., 0.)` is miscompiled into `false`
```

hhugo:
@vouillon, can you explain what's wrong with using `Code.constant_equal` in `Flow.the_const_of` ?

I think the way forward is to come up with regression tests showing how the current implementation is wrong. @OlivierNicole, do you want to look at this ?

compiler/lib/code.mli
OlivierNicole: Can you give an example of two strings that would be equal under constant_equal and not under `Js.equals`?

OlivierNicole:
I can look at this, but so far i’m confused:



> Also `Float.equal 0. (-0.)` is false.

Testing in ocaml 5.2.0 yields `true`.



> I'm guessing you mean
> 
> ```
> Also `Float.equal 0. (-0.)` is false. So `(0., 0.) = (-0., 0.)` is miscompiled into `false`
> ```

This equality is true in both ocaml 5.2.0 and Node (with `(==)`).

vouillon:
> I can look at this, but so far i’m confused:
> 
> > Also `Float.equal 0. (-0.)` is false.
> 
> Testing in ocaml 5.2.0 yields `true`.

We have this definition in `stdlib.ml`, which is really confusing:
```ocaml
  let equal (a : float) (b : float) =
    Int64.equal (Int64.bits_of_float a) (Int64.bits_of_float b)
```

compiler/lib/code.mli
vouillon: For strings, you need to set the option `use-js-string` to false.
You can try tuples: `Js.equals (0,0) (0, 0)` should be false.

OlivierNicole:
Yes, very!

vouillon:
> @vouillon, can you explain what's wrong with using `Code.constant_equal` in `Flow.the_const_of` ?

In `Flow.the_const_of`, we want to consider two constants the same if they cannot be distinguished. In particular, `0.` and `-0.` are different.

On the other hand, we want that if  `Code.constant_equal a b = Some v`, then `Poly.(=) a b = v`.

It is possible to define a function that satisfies both constraints by special-casing `0.` and `-0.`, but it seems simpler to use difference functions, especially since in `the_const_of` we only care about simple constants.

OlivierNicole:
> So `(0., 0.) = (-0., 0.)` is miscompiled into `false`

Regarding this, I suggest we stop shadowing the IEEE754-complying stdlib `Float.equal` with bitwise equality and call it `bitwise_equal` instead. This way we will use IEEE754-complying `Float.equal` in the rewritings and the problem should disappear.

vouillon:
> > So `(0., 0.) = (-0., 0.)` is miscompiled into `false`
> 
> Regarding this, I suggest we stop shadowing the IEEE754-complying stdlib `Float.equal` with bitwise equality and call it `bitwise_equal` instead. This way we will use IEEE754-complying `Float.equal` in the rewritings and the problem should disappear.

I agree that we should rename this function. But `Float.equal` is not quite IEEE754-complying since `Float.equal nan nan` is true (but `Poly.equal nan nan` is false). So we should also review where it is used.

OlivierNicole:
After checking, the only other place it is used is in `compiler/lib/javascript.ml`, in `Num.of_float`:

https://github.com/ocsigen/js_of_ocaml/blob/3204c33e5f7da4386755522083ac1e5d17e12fab/compiler/lib/javascript.ml#L106-L128

I’m not entirely sure what it does, but I expect we can adapt it if needed.

vouillon:
> After checking, the only other place it is used is in `compiler/lib/javascript.ml`, in `Num.of_float`:
> 
> [...]
> 
> I’m not entirely sure what it does, but I expect we can adapt it if needed.

`NaN` and zeroes have been excluded, so this is fine.

compiler/lib/code.mli
OlivierNicole: OK, it can be the case if the two values are structurally equal but are different objects.

OlivierNicole:
I have updated the comment, and attempted to fix the two compilation bugs that you have mentioned here.

OlivierNicole:
I will also add regression tests

compiler/lib/code.mli
OlivierNicole: > You can try tuples: Js.equals (0,0) (0, 0) should be false.

Mm, `(0,0) == (0,0)` is true in NodeJS.

compiler/lib/flow.ml
```diff
@@ -337,6 +337,21 @@ let the_def_of info x =
         x
   | Pc c -> Some (Constant c)
 
+(* If [constant_identical a b = Some true], then the two values cannot be
+   distinguished, i.e., they are not different objects (and [caml_js_equals a b
+   = true]) and if both are floats, they are bitwise equal. *)
+let constant_identical a b =
```
vouillon, line 343: This function could just return a boolean (meaning that we know that the two constants cannot be distinguished).

compiler/lib/code.mli
vouillon: `(0,0)` is not a syntax for tuples in JavaScript...
```
> [0,0,0]==[0,0,0]
false
```

OlivierNicole:
I have added a regression test.

compiler/tests-compiler/gh1659.ml
```diff
@@ -0,0 +1,9 @@
+let%expect_test _ =
+  let prog = {|
+let x = (0., 0.) = (-0., 0.);;
```
vouillon, line 3: I would also add a test for `Js.equals (0, 0) (0, 0)`.

hhugo:
> We have this definition in `stdlib.ml`, which is really confusing:
> 
> ```ocaml
>   let equal (a : float) (b : float) =
>     Int64.equal (Int64.bits_of_float a) (Int64.bits_of_float b)
> ```

Maybe we should rename it `Float.bits_equal` and have `Float.equal` the one from the stdlib. I'd prefer to limit the use of `Poly.equal` as much as possible.

compiler/lib/code.mli
```diff
@@ -185,7 +185,13 @@ type constant =
   | Tuple of int * constant array * array_or_not
   | Int of int32
 
-val constant_equal : constant -> constant -> bool option
+module Constant : sig
+  type t = constant
+
+  val ocaml_equal : t -> t -> bool option
+  (** Guaranteed equality in terms of both OCaml [(=)]: if [constant_equal a b =
```
hhugo, line 192:
```suggestion
  (** Guaranteed equality in terms of OCaml [(=)]: if [constant_equal a b =
```

compiler/lib/driver.ml
```diff
@@ -18,6 +18,7 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 open! Stdlib
+module Constant = Global_constant
```
hhugo, line 21: Just propage the name

compiler/lib/eval.ml
```diff
@@ -233,12 +233,41 @@ let the_cont_of info x (a : cont array) =
       | _ -> None)
     x
 
+(* If [constant_js_equal a b = Some v], then [caml_js_equals a b = Some v]). *)
```
hhugo, line 236:
```suggestion
(* If [constant_js_equal a b = Some v], then [caml_js_equals a b = v]). *)
```

compiler/lib/eval.ml
```diff
@@ -233,12 +233,41 @@ let the_cont_of info x (a : cont array) =
...
+let constant_js_equal a b =
+  match a, b with
+  | Int i, Int j -> Some (Int32.equal i j)
+  | Float a, Float b -> Some (Poly.equal a b)
+  | Int _, Float _ | Float _, Int _ -> None
+  (* All other values may be distinct objects and thus different by [caml_js_equals]. *)
+  | String _, _
+  | _, String _
+  | NativeString _, _
```
hhugo, line 245: native strings should be equal if their content is equal (at least in the jsoo case)
The same apply to String when `Config.Flag.use_js_string ()` is true

compiler/tests-compiler/gh1659.ml
hhugo: You can explicit add an external inside `prog`

```
external equals : 'a -> 'a -> bool = "caml_js_equals"
let x = equals (0, 0) (0, 0)
Printf.printf "%B\n" x;;
```

compiler/lib/flow.ml
```diff
@@ -337,6 +337,26 @@ let the_def_of info x =
...
+  match a, b with
+  | Int i, Int j -> Int32.equal i j
+  | Float a, Float b -> Float.bitwise_equal a b
+  | Int _, Float _ | Float _, Int _ -> false
+  (* All other values may be distinct objects and thus different by [caml_js_equals]. *)
+  | String _, _
+  | _, String _
+  | NativeString _, _
```
hhugo, line 351: two equal nativeString should be indistinguishable.
same for String when `use_js_string`

compiler/lib/code.mli
OlivierNicole: Done

compiler/lib/driver.ml
OlivierNicole: Done

compiler/lib/eval.ml
OlivierNicole: Thanks, fixed

compiler/lib/eval.ml
OlivierNicole: Done

compiler/tests-compiler/gh1659.ml
OlivierNicole: Done

OlivierNicole:
Thanks for your reviews, I have updated the changes.

Could you explain to me what are native strings exactly?

compiler/lib/flow.ml
```diff
@@ -337,6 +337,28 @@ let the_def_of info x =
...
 
+(* If [constant_identical a b = true], then the two values cannot be
+   distinguished, i.e., they are not different objects (and [caml_js_equals a b
+   = true]) and if both are floats, they are bitwise equal. *)
+let constant_identical a b =
+  match a, b with
+  | Int i, Int j -> Int32.equal i j
+  | Float a, Float b -> Float.bitwise_equal a b
+  | NativeString a, NativeString b -> Native_string.equal a b
+  | String a, String b -> Config.Flag.use_js_string () && String.equal a b
```
OlivierNicole, line 348: I expect that we may have to update this for Wasm.

vouillon: Indeed, this should be false for wasm_of_ocaml.
Maybe we can just drop this case?

OlivierNicole: Which one?

vouillon: Native strings are JavaScript strings. We have a hack to use JavaScript equality for them, so the first line is fine.

Also, since floats are boxed, one can distinguish two floats with the same value.

OlivierNicole: Maybe this suggests that the function should behave differently between the two backends (once there are two backends)?

compiler/lib/eval.ml
hhugo: I don't see the corresponding diff

compiler/lib/eval.ml
```diff
@@ -233,12 +233,41 @@ let the_cont_of info x (a : cont array) =
...
+      | Some e1, Some e2 -> (
+          match Code.Constant.ocaml_equal e1 e2 with
+          | None -> [ i ]
+          | Some c ->
+              let c = if c then 1l else 0l in
+              let c = Constant (Int c) in
+              Flow.Info.update_def info x c;
+              [ Let (x, c), loc ])
+      | _ -> [ i ])
+  | Let (x, Prim (Extern "caml_js_equals", [ y; z ])) -> (
```
hhugo, line 267: I think we could apply the same logic to `"caml_js_strict_equals"`

hhugo:
> > We have this definition in `stdlib.ml`, which is really confusing:
> > ```ocaml
> >   let equal (a : float) (b : float) =
> >     Int64.equal (Int64.bits_of_float a) (Int64.bits_of_float b)
> > ```
> 
> Maybe we should rename it `Float.bits_equal` and have `Float.equal` the one from the stdlib. I'd prefer to limit the use of `Poly.equal` as much as possible.

Could we keep a Float.equal in our stdlib and use it instead of Poly.equal ?

compiler/lib/eval.ml
OlivierNicole: Done

OlivierNicole:
> Could we keep a Float.equal in our stdlib and use it instead of Poly.equal ?

Sorry, I had misunderstood what you meant. Done.

compiler/lib/eval.ml
```diff
@@ -237,7 +237,7 @@ let the_cont_of info x (a : cont array) =
 let constant_js_equal a b =
   match a, b with
   | Int i, Int j -> Some (Int32.equal i j)
-  | Float a, Float b -> Some (Poly.equal a b)
+  | Float a, Float b -> Some (Float.equal a b)
```
vouillon, line 240: In JavaScript, `NaN == NaN` is false, but `Float.equal nan nan` is true.

compiler/lib/code.ml
```diff
@@ -387,9 +387,9 @@ module Constant = struct
           done;
           !same
     | Int64 a, Int64 b -> Some (Int64.equal a b)
-    | Float_array a, Float_array b -> Some (Array.equal Poly.equal a b)
+    | Float_array a, Float_array b -> Some (Array.equal Float.equal a b)
     | Int a, Int b -> Some (Int32.equal a b)
-    | Float a, Float b -> Some (Poly.equal a b)
+    | Float a, Float b -> Some (Float.equal a b)
```
vouillon, line 392: I think we need to keep `Poly.equal`. In Ocaml, `nan = nan` is false, but `Float.equal nan nan` is true.

dbuenzli:
> We have this definition in `stdlib.ml`, which is really confusing:

If memory serves this is to make `Float.equal` compatible with `Float.compare` which is implemented as `Stdlib.compare` which does not have the IEEE 754 semantics on `NaN`s so that you have a total order on `float` values. That in turn makes things sound when you pass `Float` to functors that expect an `OrderedType` signature (`{Set,Map}.Make`, etc.).

compiler/lib/eval.ml
hhugo: fixed. Now using `Float.ieee_equal` alias to `Poly.equal`

PR #1660, opened 2024-08-01
Runtime: implement Json output for Wasm

OlivierNicole:

In Wasm, some additional functions are required to output Json.

### DIFF
--- a/lib/js_of_ocaml/json.ml
+++ b/lib/js_of_ocaml/json.ml
@@ -173,8 +173,10 @@ let use_native_stringify () = !use_native_stringify_
 
 let set_use_native_stringify b = use_native_stringify_ := b
 
+let output_ x = to_json (Obj.repr x)
+
 let output obj =
   match Sys.backend_type with
   | Other "js_of_ocaml" when use_native_stringify () ->
       json##stringify_ obj (Js.wrap_callback output_reviver)
-  | _ -> Js.string (to_json (Obj.repr obj))
+  | _ -> Js.string (output_ obj)
--- a/lib/js_of_ocaml/json.ml
+++ b/lib/js_of_ocaml/json.ml
@@ -141,9 +141,10 @@ let input_reviver =
 let unsafe_input s =
   match Sys.backend_type with
   | Other "wasm_of_ocaml" ->
-      failwith
-        "Json.unsafe_input: not implemented (and not planned) in the Wasm backend as the \
-         inverse of `output` is not implementable"
+      (* https://github.com/ocsigen/js_of_ocaml/pull/1660#discussion_r1731099372
+         The encoding of OCaml values is ambiguous since both integers and floats
+         are mapped to numbers *)
+      failwith "Json.unsafe_input: not implemented in the Wasm backend"
   | _ -> json##parse_ s input_reviver
 
 class type obj = object
--- a/lib/js_of_ocaml/json.ml
+++ b/lib/js_of_ocaml/json.ml
@@ -20,6 +20,93 @@
 open Js
 open! Import
 
+(****)
+
+(* The writing logic for basic types is copied from [lib/deriving_json]. *)
+
+let write_string buffer s =
+  Buffer.add_char buffer '"';
+  for i = 0 to String.length s - 1 do
+    match s.[i] with
+    | '"' -> Buffer.add_string buffer {|\"|}
+    | '\\' -> Buffer.add_string buffer {|\\|}
+    | '\b' -> Buffer.add_string buffer {|\b|}
+    | '\x0C' -> Buffer.add_string buffer {|\f|}
+    | '\n' -> Buffer.add_string buffer {|\n|}
+    | '\r' -> Buffer.add_string buffer {|\r|}
+    | '\t' -> Buffer.add_string buffer {|\t|}
+    | c when Poly.(c <= '\x1F') ->
+        (* Other control characters are escaped. *)
+        Printf.bprintf buffer {|\u%04X|} (int_of_char c)
+    | c when Poly.(c < '\x80') -> Buffer.add_char buffer s.[i]
+    | _c (* >= '\x80' *) ->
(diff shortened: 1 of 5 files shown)


lib/js_of_ocaml/json.ml
```diff
@@ -20,6 +20,88 @@
 open Js
 open! Import
 
+(****)
+
+let write_string buffer s =
+  Buffer.add_char buffer '\"';
+  for i = 0 to String.length s - 1 do
+    match s.[i] with
+    | '\"' -> Buffer.add_string buffer "\\\""
```
hhugo, line 29: Maybe use quoted string on the RHS so get rid of the escaping.

lib/js_of_ocaml/json.ml
```diff
@@ -20,6 +20,88 @@
...
+        Buffer.add_char b ',';
+        write_float b (Obj.double_field v i)
+      done;
+      Buffer.add_char b ']')
+    else if t = Obj.custom_tag
+    then
+      match custom_identifier v with
+      | "_i" -> Printf.bprintf b "%ld" (Obj.obj v : int32)
+      | "_j" ->
+          let i : int64 = Obj.obj v in
```
hhugo, line 86: move the logic to a write_int64 helper

lib/js_of_ocaml/json.ml
```diff
@@ -20,6 +20,88 @@
 open Js
 open! Import
 
+(****)
```
hhugo, line 23: The write logic for basic types seems to be copied from lib/deriving_json, we could write a comment about this so that future us remembers that there are multiple copies of this logic.

runtime/obj.js
```diff
@@ -214,3 +214,9 @@ function caml_is_continuation_tag(t) {
 function caml_is_continuation_tag(t) {
   return (t == 245) ? 1 : 0;
 }
+
+//Provides: caml_custom_identifier
+//Requires: caml_string_of_jsstring
+function caml_custom_identifier (o) {
+  return caml_string_of_jsstring(o.custom_tag);
```
hhugo, line 221: This doesn't look correct, the field is called caml_custom

lib/js_of_ocaml/json.ml
```diff
@@ -71,4 +158,7 @@ let output_reviver _key (value : Unsafe.any) : Obj.t =
     Obj.repr (array [| 255; value##.lo; value##.mi; value##.hi |])
   else Obj.repr value
 
-let output obj = json##stringify_ obj (Js.wrap_callback output_reviver)
+let output obj =
+  match Sys.backend_type with
```
hhugo, line 162: We could use the following to control the implementation, it would allow to write tests checking the two implementations behave the same.

```ocaml
let use_native_stringify =
  ref (
    match Sys.backend_type with
      | Other "js_of_ocaml" -> true
      | Native | Bytecode | Other _ -> false)
```

lib/js_of_ocaml/json.ml
OlivierNicole: Done

lib/js_of_ocaml/json.ml
OlivierNicole: Done

lib/js_of_ocaml/json.ml
OlivierNicole: Done

lib/js_of_ocaml/json.ml
```diff
@@ -51,13 +138,18 @@ let input_reviver =
   in
   wrap_meth_callback reviver
 
-let unsafe_input s = json##parse_ s input_reviver
+let unsafe_input s =
+  match Sys.backend_type with
+  | Other "wasm_of_ocaml" -> failwith "Json.unsafe_input: not implemented"
```
hhugo, line 143: Can you add a comment explaining why it's not implemented ?

hhugo (changes requested):
I've changed the tests a little and added a request to add an extra comment.
Looks good otherwise.

lib/js_of_ocaml/json.ml
OlivierNicole: I changed to “not available in the Wasm backend”. What do you think?

hhugo: What about a comment answering these questions ?
Any reason it is not implemented ? Is an implementation possible ? Is it planned?

OlivierNicole: I would assume it could be implemented, but is not necessary to run existing packages and has not been a priority. @vouillon would know better.

vouillon: Function `output` is used by Eliom and in some Jane Street code, but `unsafe_input` does not seem to be used anywhere.
Also, we do not know whether the JSON value should be deserialized as a JS value or an OCaml value.

vouillon: And the encoding of OCaml values is ambiguous since both integers and floats are mapped to numbers.

OlivierNicole: I have updated the error message to try to summarize that.

OlivierNicole:
Thank you for the review!

lib/js_of_ocaml/json.ml
OlivierNicole: I got confused between. “comment” and “error message”, sorry. I see you fixed that.

lib/js_of_ocaml/json.mli
```diff
@@ -25,3 +25,16 @@ val output : 'a -> Js.js_string Js.t
 val unsafe_input : Js.js_string Js.t -> 'a
 (** Unmarshal a string in JSON format as an OCaml value (unsafe but
     fast !). *)
+
+(**/**)
+
+val output_ : 'a -> string
```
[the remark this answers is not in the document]
hhugo: I've added this while playing with tests. I could be removed indeed.

PR #1664, opened 2024-09-03
Decode sourcemap mappings only when necessary

OlivierNicole:
This was extracted from #1617. When linking, it was observed that parsing the mappings field of sourcemaps takes a big toll on the overall performance, and was unnecessary since the mappings are not modified.

### DIFF
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -1996,8 +1996,9 @@ let program ?(accept_unnamed_var = false) f ?source_map p =
               | None -> filename
               | Some _ -> Filename.concat "/builtin" filename)
         in
+        let sm_mappings = Source_map.Mappings.decode sm.mappings in
         let mappings =
-          List.rev_append_map !temp_mappings sm.mappings ~f:(fun (pos, m) ->
+          List.rev_append_map !temp_mappings sm_mappings ~f:(fun (pos, m) ->
               let gen_line = pos.PP.p_line + 1 in
               let gen_col = pos.PP.p_col in
               match m with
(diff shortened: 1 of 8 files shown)


compiler/lib/source_map.mli
```diff
@@ -44,7 +44,30 @@ type map =
...
+  type t
+
+  val empty : t
+  (** Represents the empty mapping. *)
+
+  val of_string : string -> t
+  (** By default, mappings are left uninterpreted, since many operations can be
+      performed efficiently directly on the encoded form. Therefore this
+      function is mostly a no-op and very cheap. It does not perform any
+      validation of its argument, unlike {!val:edit} or {!val:decode}. It is
```
hhugo, line 57: edit doesn't exists yet

compiler/lib/source_map.mli
```diff
@@ -44,7 +44,30 @@ type map =
...
+      guaranteed that {!val:of_string} and {!val:to_string} are inverse
+      functions. *)
+
+  val decode : t -> map list
+  (** Parse the mappings. Prefer using the more efficient {!val:edit} on the
```
hhugo, line 62: same here

hhugo (approved):
Happy to merge once the CI is happy.
I left two small comment.

OlivierNicole:
Thank you, done.

compiler/lib/js_output.ml
```diff
@@ -2012,6 +2013,7 @@ let program ?(accept_unnamed_var = false) f ?source_map p =
                   Source_map.Gen_Ori_Name
                     { gen_line; gen_col; ori_source; ori_line; ori_col; ori_name })
         in
+        let mappings = Source_map.Mappings.encode mappings in
```
vouillon, line 2016: @OlivierNicole Isn't this going to be quadratic? If we have `n` files, we are going to decode and encode a larger and larger sourcemap `n` times.

hhugo: This is gone in #1715

PR #1673, opened 2024-09-16
runtime: format code with biome

smorimoto:


hhugo:
I'd prefer to not reformat the ts file (zstd.ts) because it's not maintained here.

smorimoto:
@hhugo Updated!

biome.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "formatter": {
+    "enabled": true,
+    "useEditorconfig": true
+  },
+  "linter": {
+    "enabled": false
+  },
+  "organizeImports": {
+    "enabled": true
```
hhugo, line 15: What is this about ?

TyOverby:
Could we avoid making big formatting changes until `wasm_of_ocaml` is merged?

vouillon:
> Could we avoid making big formatting changes until `wasm_of_ocaml` is merged?

We are just going to copy the JavaScript runtime code without making any change, so this is not a problem.

hhugo:
> > Could we avoid making big formatting changes until `wasm_of_ocaml` is merged?
> 
> We are just going to copy the JavaScript runtime code without making any change, so this is not a problem.

@vouillon, why do you need such copy for wasm_of_ocaml?

biome.json
smorimoto: It wasn't really necessary, but the functionality to reorder imports.

vouillon:
What I mean is that we are going to directly use the version of the runtime files which are currently in the Js_of_ocaml repository rather than trying to merge them with those in the Wasm_of_ocaml repository.

PR #1674, opened 2024-09-17
JavaScript linker: add wasm flag

vouillon:


hhugo:
@vouillon, can you drop the two commits related to "target-specific code"

PR #1677, opened 2024-09-19
Runtime: small refactoring in parsing.js

hhugo:
spotted while working on https://github.com/ocsigen/js_of_ocaml/pull/1672

cc @smorimoto

### DIFF
--- a/runtime/parsing.js
+++ b/runtime/parsing.js
@@ -129,8 +129,8 @@ function caml_parse_engine(tables, env, cmd, arg) {
   var state = env[env_state];
   var errflag = env[env_errflag];
 
-  exit: for (;;) {
-    next: switch (cmd) {
+  the_loop: for (;;) {
+    switch (cmd) {
       case 0: //START:
         state = 0;
         errflag = 0;
@@ -147,7 +147,7 @@ function caml_parse_engine(tables, env, cmd, arg) {
           break;
         }
         res = READ_TOKEN;
-        break exit;
+        break the_loop;
       /* The ML code calls the lexer and updates */
       /* symb_start and symb_end */
       case 1: //TOKEN_READ:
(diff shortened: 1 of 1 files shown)


runtime/parsing.js
```diff
@@ -129,8 +129,8 @@ function caml_parse_engine(tables, env, cmd, arg) {
   var state = env[env_state];
   var errflag = env[env_errflag];
 
-  exit: for (;;) {
-    next: switch (cmd) {
+  the_loop: for (;;) {
```
smorimoto, line 132: Is that really needed?

```suggestion
  for (;;) {
```

runtime/parsing.js
```diff
@@ -147,7 +147,7 @@ function caml_parse_engine(tables, env, cmd, arg) {
           break;
         }
         res = READ_TOKEN;
-        break exit;
+        break the_loop;
```
smorimoto, line 150:
```suggestion
        break;
```

runtime/parsing.js
```diff
@@ -206,7 +206,7 @@ function caml_parse_engine(tables, env, cmd, arg) {
             ) {
               if (caml_parser_trace) log("Recovering in state " + state1);
               cmd = shift_recover;
-              break next;
+              continue the_loop;
```
smorimoto, line 209:
```suggestion
              continue;
```

runtime/parsing.js
smorimoto: nvm

smorimoto (approved):

runtime/parsing.js
vouillon: You want to exit the loop and not just terminate the current switch statement.

smorimoto: Indeed

PR #1678, opened 2024-09-19
Runtime: safe fixes

smorimoto:
All of these fixes are very general, independent of the JS runtime version.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -11,6 +11,7 @@
   "linter": {
     "enabled": true,
     "rules": {
+      "recommended": true,
       "complexity": {
         "noUselessSwitchCase": "off",
         "noUselessTernary": "off",
@@ -22,7 +23,6 @@
         "noInvalidUseBeforeDeclaration": "off",
         "noSwitchDeclarations": "off"
       },
-      "recommended": true,
       "security": {
         "noGlobalEval": "off"
       },
@@ -46,7 +46,6 @@
         "noDoubleEquals": "off",
         "noDuplicateCase": "off",
         "noFallthroughSwitchClause": "off",
-        "noGlobalIsFinite": "off",
         "noGlobalIsNan": "off",
         "noRedeclare": "off",
         "noRedundantUseStrict": "off",
--- a/biome.json
+++ b/biome.json
@@ -47,7 +47,6 @@
         "noDuplicateCase": "off",
         "noFallthroughSwitchClause": "off",
         "noRedeclare": "off",
-        "noRedundantUseStrict": "off",
         "noSelfCompare": "off",
         "useDefaultSwitchClauseLast": "off"
       }
--- a/biome.json
+++ b/biome.json
@@ -12,7 +12,6 @@
     "enabled": true,
     "rules": {
       "complexity": {
-        "noExtraBooleanCast": "off",
         "noUselessSwitchCase": "off",
         "noUselessTernary": "off",
         "useArrowFunction": "off",
--- a/biome.json
+++ b/biome.json
@@ -23,6 +23,9 @@
         "noInvalidUseBeforeDeclaration": "off",
         "noSwitchDeclarations": "off"
       },
+      "nursery": {
+        "noUselessEscapeInRegex": "error"
+      },
       "security": {
         "noGlobalEval": "off"
       },
--- a/biome.json
+++ b/biome.json
@@ -38,8 +38,7 @@
         "useExponentiationOperator": "off",
         "useNodejsImportProtocol": "off",
         "useSingleVarDeclarator": "off",
-        "useTemplate": "off",
-        "useWhile": "off"
+        "useTemplate": "off"
       },
       "suspicious": {
         "noAssignInExpressions": "off",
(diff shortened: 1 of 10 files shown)


smorimoto:
Ready for review

biome.json
```diff
@@ -35,7 +35,6 @@
         "noParameterAssign": "off",
         "noUselessElse": "off",
         "noVar": "off",
-        "useExponentiationOperator": "off",
```
hhugo: Can we leave this syntax for later. I was introduced with es6 or es7

runtime/fs.js
```diff
@@ -65,7 +65,7 @@ function make_path_is_absolute() {
     var isUnc = Boolean(device && device.charAt(1) !== ":");
 
     // UNC paths are always absolute
-    if (Boolean(result[2] || isUnc)) {
+    if (result[2] || isUnc) {
```
hhugo, line 68: The `Boolean` above might be unnecessary as well.

runtime/fs_fake.js
```diff
@@ -213,7 +213,7 @@ MlFakeDevice.prototype.is_dir = function (name) {
   return this.content[name_slash] ? 1 : 0;
 };
 MlFakeDevice.prototype.unlink = function (name) {
-  var ok = this.content[name] ? true : false;
+  var ok = !!this.content[name];
```
hhugo, line 216: I find this syntax too cryptic, maybe we should use `hasOwnProperty` ..

runtime/ieee_754.js
```diff
@@ -200,7 +200,7 @@ function caml_classify_float(x) {
     if (x != 0) return 1;
     return 2;
   }
-  return isNaN(x) ? 4 : 3;
+  return Number.isNaN(x) ? 4 : 3;
```
hhugo, line 205: FYI, the two are not strictly equivalent.
> Number.isNaN() doesn't attempt to convert the parameter to a number, so non-numbers always return false

runtime/mlBytes.js
```diff
@@ -89,8 +89,7 @@ function caml_utf8_of_utf16(s) {
     if (c < 0x80) {
       for (var j = i + 1; j < l && (c = s.charCodeAt(j)) < 0x80; j++);
       if (j - i > 512) {
-        t.substr(0, 1);
-        b += t;
+        b += t.slice(0, 1);
```
hhugo, line 92: This is just wrong. other changes about `substr` are wrong as well.

runtime/fs_fake.js
smorimoto: Indeed

runtime/fs.js
smorimoto: If you change it, the value will be `boolean | ""`

runtime/fs_fake.js
smorimoto: It's not really recommended and it's better to use `Object.hasOwn`, but that's a functionality added in ES2022, so how about revisiting it for subsequent PR to modernise?
https://biomejs.dev/linter/rules/no-prototype-builtins

runtime/ieee_754.js
smorimoto: Yeah, but I don't think it's a problem here. Am I missing something?

runtime/mlBytes.js
smorimoto: Oops... Since neither `substr` nor `slice` manipulate the original array, the operation itself was actually unnecessary.

biome.json
smorimoto: Done

runtime/mlBytes.js
hhugo: It was here on purpose, to force the js engine to fatten the string and improve performance. This is yet another example of things that are not as obvious as they seems. Can you leave this to another PR, we might need to benchmark

runtime/ieee_754.js
hhugo: It is correct here, but it's easy to miss the difference in behavior when mixed in the middle of other changes. It would simplify review if the PR was mentioning such difference.

runtime/mlBytes.js
smorimoto: Done

smorimoto:
@hhugo I think we can move this forward. I'm planning to introduce benchmarks before the next PR. Are you positive about that?

runtime/fs.js
hhugo: Can you be specific ?

runtime/fs_fake.js
hhugo: I'm still unconvinced by the change. Do we actually want to return a boolean..

hhugo:
Regarding the review process, can you try to close conversation on if there is a clear consensus

runtime/fs.js
smorimoto: In this case, an empty string should return an empty string without further evaluation beyond Logical AND.

This should work:

```js
var isUnc = device.length > 0 && device.charAt(1) !== ":";
```

runtime/fs_fake.js
smorimoto: At this point, just reverted.

runtime/fs.js
hhugo: Ok, can you make that change ?

smorimoto: @hhugo Done!

PR #1680, opened 2024-09-23
Fix global flow analysis

vouillon:
It needs to take into account values flowing out of the toplevel function (see #1556).


hhugo (approved):

vouillon:
This is a better fix than #1556. We can at least check that the test introduced in 1f29da6177682c5f3679a08b0cdc06f73fd29d5c still pass, but I'm not sure this solves any other issue.

Issue #1612, opened 2024-05-09
[BUG] js_of_ocaml is excessively memory hungry


JasonGross:
**Describe the bug**

`js_of_ocaml` is very cool!  I use it on CI to generate a webpage.  However, I cannot use it on the new GitHub Actions arm64 MacOS boxes, which have only 7 GB of RAM, because it sometimes eats 8--9 GB RAM to generate a single .js file.  For example [here is a table of build times and memory usages on linux](https://github.com/mit-plv/fiat-crypto/actions/runs/8790991439/job/24130114864):
```
     Time |   Peak Mem | File Name                                         
---------------------------------------------------------------------------
15m11.31s | 8904232 ko | Total Time / Peak Mem                             
---------------------------------------------------------------------------
 4m46.85s | 8808932 ko | ExtractionJsOfOCaml/bedrock2_fiat_crypto.js       
 4m37.20s | 7151532 ko | ExtractionJsOfOCaml/fiat_crypto.js                
 4m30.42s | 8904232 ko | ExtractionJsOfOCaml/with_bedrock2_fiat_crypto.js  
 0m25.80s | 3195952 ko | ExtractionJsOfOCaml/with_bedrock2_fiat_crypto.byte
 0m25.32s | 3193600 ko | ExtractionJsOfOCaml/bedrock2_fiat_crypto.byte     
 0m24.60s | 2814364 ko | ExtractionJsOfOCaml/fiat_crypto.byte              
 0m00.38s |  102780 ko | ExtractionJsOfOCaml/bedrock2_fiat_crypto.cmi      
 0m00.37s |   98468 ko | ExtractionJsOfOCaml/fiat_crypto.cmi               
 0m00.37s |  102360 ko | ExtractionJsOfOCaml/with_bedrock2_fiat_crypto.cmi 
```
It is similar on [mac](https://github.com/mit-plv/fiat-crypto/actions/runs/8790991430/job/24124235290#step:21:27), and a bit better on [debian sid](https://github.com/mit-plv/fiat-crypto/actions/runs/8790991433/job/24124235827#step:24:27).

I [invoke it with](https://github.com/mit-plv/fiat-crypto/blob/c5008e7450bb462f3ad9f0735e45ec751c9d266a/Makefile.standalone#L27-L30) `--source-map --no-inline --enable=effects` and invoke the compiler with ` -package js_of_ocaml -package unix -w -20 -g`

[...]

PR #1681, opened 2024-09-23
Improved global deadcode elimination

vouillon:
We now take into account of whether an instruction is in a live function or not when generating constraints, so that dead code can no longer result in making a variable live.

We also keep track of nested module.

The performance of the algorithm has been greatly improved as well.

On the `basic_website` example of Tyxml, compiled with `--opt 3`, I get the following reduction in code size:
|     | before | after | |
| ---  | ---: | ---: |  --- |
| uncompressed | 139304 | 125168 | -10% |
| gzip | 47980 | 41367 | -14% |
| bzip2 | 45298 | 38885 | -14% |

Here are the numbers regarding #1612, when compiling `with_bedrock2_fiat_crypto.byte`. For comparison, I have included an intermediate step which just optimizes the current algorithm.
| |before | optimized | after |
| -- | --: | --: | --: |
| duration (s) | 48 | 4.9 | 9.8 |
| peak memory (GiB) | 4 | 2.8 | 2.8 |

### DIFF
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -1031,10 +1031,32 @@ let throw_statement ctx cx k loc =
         , loc )
       ]
 
+let remove_unused_tail_args ctx exact trampolined args =
+  if exact && not trampolined
+  then
+    let has_unused_tail_args =
+      List.fold_left
+        ~f:(fun _ x -> Var.equal x ctx.Ctx.deadcode_sentinal)
+        ~init:false
+        args
+    in
+    if has_unused_tail_args
+    then
+      List.fold_right
+        ~f:(fun x args ->
+          match args with
+          | [] when Var.equal x ctx.Ctx.deadcode_sentinal -> []
+          | _ -> x :: args)
+        ~init:[]
+        args
+    else args
+  else args
+
 let rec translate_expr ctx queue loc x e level : _ * J.statement_list =
   match e with
   | Apply { f; args; exact } ->
       let trampolined = Var.Set.mem x ctx.Ctx.trampolined_calls in
+      let args = remove_unused_tail_args ctx exact trampolined args in
       let args, prop, queue =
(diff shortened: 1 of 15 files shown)


hhugo:
You need to check the effect on https://github.com/ocsigen/js_of_ocaml/issues/1612

hhugo:
> Regarding #1612, when compiling `with_bedrock2_fiat_crypto.byte` the analysis takes 40s with this PR. It takes 48s with `master`. But there is an intermediate step in this PR with an improved solver which reduces the duration of the initial analysis to 15s.

What about memory peak ?

vouillon:
I fixed a memory / speed issue. The numbers are pretty good now.

hhugo:
Cc @jasongross

compiler/lib/generate.ml
```diff
@@ -1124,8 +1124,7 @@ let rec translate_expr ctx queue loc x e level : _ * J.statement_list =
   | Special (Alias_prim name) ->
       let prim = Share.get_prim (runtime_fun ctx) name ctx.Ctx.share in
       (prim, const_p, queue), []
-  | Special Undefined ->
-      (J.(EVar (ident (Utf8_string.of_string_exn "undefined"))), const_p, queue), []
+  | Special Undefined -> (zero, const_p, queue), []
```
hhugo, line 1127: Maybe rename the Undefined constructor.

hhugo:
I didn't really reviewed https://github.com/ocsigen/js_of_ocaml/pull/1681/commits/df605821096b6f4a357b50c23834c3cc3edc2374 . The rest looks good. Happy to merge.

compiler/lib/generate.ml
vouillon: I have now just removed `Special Undefined` and use the integer constant `0` from the start.

PR #1682, opened 2024-09-25
runtime: fix useArrowFunction and noArguments

smorimoto:
Note: this partially introduces ES6!

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -13,7 +13,6 @@
     "rules": {
       "recommended": true,
       "complexity": {
-        "useArrowFunction": "off",
         "useOptionalChain": "off"
       },
       "correctness": {
@@ -26,7 +25,6 @@
         "noSubstr": "error"
       },
       "style": {
-        "noArguments": "off",
         "noCommaOperator": "off",
         "noParameterAssign": "off",
         "noUselessElse": "off",
--- a/runtime/js/backtrace.js
+++ b/runtime/js/backtrace.js
@@ -19,7 +19,7 @@
 //Requires: jsoo_sys_getenv
 var caml_record_backtrace_env_flag = FLAG("with-js-error");
 
-(function () {
+(() => {
   var r = jsoo_sys_getenv("OCAMLRUNPARAM");
   if (r !== undefined) {
     var l = r.split(",");
--- a/runtime/js/blake2.js
+++ b/runtime/js/blake2.js
@@ -1,6 +1,6 @@
 //Provides: blake2b
 //Version: >= 5.2
-var blake2b = (function () {
+var blake2b = (() => {
   // Blake2B in pure Javascript
   // Adapted from the reference implementation in RFC7693
   // Ported to Javascript by DC - https://github.com/dcposch
@@ -104,11 +104,7 @@ var blake2b = (function () {
   // These are offsets into a uint64 buffer.
   // Multiply them all by 2 to make them offsets into a uint32 buffer,
   // because this is Javascript and we don't have uint64s
-  const SIGMA82 = new Uint8Array(
-    SIGMA8.map(function (x) {
-      return x * 2;
-    }),
-  );
+  const SIGMA82 = new Uint8Array(SIGMA8.map((x) => x * 2));
 
   // Compression function. 'last' flag indicates last block.
   // Note we're representing 16 uint64s as 32 uint32s
--- a/runtime/js/effect.js
+++ b/runtime/js/effect.js
@@ -60,7 +60,7 @@ function caml_push_trap(handler) {
 //If: effects
 function caml_pop_trap() {
   if (!caml_exn_stack)
-    return function (x) {
+    return (x) => {
       throw x;
     };
   var h = caml_exn_stack[1];
--- a/runtime/js/fs_fake.js
+++ b/runtime/js/fs_fake.js
@@ -42,9 +42,7 @@ MlFakeDevice.prototype.create_dir_if_needed = function (name) {
     this.content[res] = Symbol("directory");
   }
 };
-MlFakeDevice.prototype.slash = function (name) {
-  return /\/$/.test(name) ? name : name + "/";
-};
+MlFakeDevice.prototype.slash = (name) => (/\/$/.test(name) ? name : name + "/");
 MlFakeDevice.prototype.lookup = function (name) {
   if (!this.content[name] && this.lookupFun) {
     var res = this.lookupFun(
@@ -350,18 +348,14 @@ MlFakeFile.prototype.read = function (offset, buf, pos, len) {
 //Requires: caml_raise_sys_error
 function MlFakeFd_out(fd, flags) {
   MlFakeFile.call(this, caml_create_bytes(0));
-  this.log = function (s) {
-    return 0;
-  };
+  this.log = (s) => 0;
   if (fd === 1 && typeof console.log === "function") this.log = console.log;
   else if (fd === 2 && typeof console.error === "function")
     this.log = console.error;
   else if (typeof console.log === "function") this.log = console.log;
   this.flags = flags;
 }
-MlFakeFd_out.prototype.length = function () {
-  return 0;
-};
+MlFakeFd_out.prototype.length = () => 0;
 MlFakeFd_out.prototype.write = function (offset, buf, pos, len) {
   if (this.log) {
     if (
--- a/runtime/js/fs_node.js
+++ b/runtime/js/fs_node.js
@@ -189,7 +189,7 @@ MlNodeDevice.prototype.opendir = function (name, raise_unix) {
     this.raise_nodejs_error(err, raise_unix);
   }
 };
-MlNodeDevice.prototype.raise_nodejs_error = function (err, raise_unix) {
+MlNodeDevice.prototype.raise_nodejs_error = (err, raise_unix) => {
   var unix_error = caml_named_value("Unix.Unix_error");
   if (raise_unix && unix_error) {
     var args = make_unix_err_args(err.code, err.syscall, err.path, err.errno);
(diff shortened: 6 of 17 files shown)


runtime/jslib.js
```diff
@@ -394,76 +388,55 @@ function caml_js_wrap_callback(f) {
...
-  return function () {
-    var len = arguments.length;
-    var args = new Array(len + 1);
-    args[0] = this;
-    for (var i = 0; i < len; i++) args[i + 1] = arguments[i];
-    var res = caml_callback(f, args);
+  return function (...args) {
+    const res = caml_callback(f, [this, ...args]);
     return res instanceof Function ? caml_js_wrap_callback(res) : res;
   };
```
smorimoto, line 416: In this case, the arrow function cannot be used for `this`.

runtime/jslib.js
```diff
@@ -394,76 +388,55 @@ function caml_js_wrap_callback(f) {
...
 //Provides: caml_js_wrap_callback_strict const
 //Requires: caml_callback
 function caml_js_wrap_callback_strict(arity, f) {
-  return function () {
-    var n = arguments.length;
-    var args = new Array(arity);
-    var len = Math.min(arguments.length, arity);
-    for (var i = 0; i < len; i++) args[i] = arguments[i];
+  return (...args) => {
+    args.length = arity;
```
smorimoto, line 397: You can use this trick to change the length of an array. I will comment the benchmark later.

runtime/jslib.js
```diff
@@ -394,76 +388,55 @@ function caml_js_wrap_callback(f) {
...
-  return function () {
-    var len = caml_js_function_arity(f) - 1;
-    var args = new Array(len + 1);
-    args[0] = this;
-    for (var i = 0; i < len; i++) args[i + 1] = arguments[i];
-    return caml_callback(f, args);
+  return function (...args) {
+    const len = caml_js_function_arity(f) - 1;
+    args.length = len;
+    return caml_callback(f, [this, ...args]);
```
smorimoto, line 439: For many arguments, using concat may have better performance than spreading.

TyOverby:
This PR looks like it contains some formatting changes (using arrow functions) and also rewrites some of the most foundational primitives in the jsoo runtime.  Could you split these up so that that latter can be thouroughly reviewed and benchmarked?

TyOverby:
Also, why are you interested in making the change to use arrow functions and avoid using `arguments`?

smorimoto:
### Context Differences

- The Arrow function inherits this from its parent. This may reduce the overhead of binding this by using the Arrow function.
- The function function binds this to its own scope, so you must explicitly bind this to prevent unintended behaviour in situations where this is treated differently.

The arrow function has no binding processing, so you might see some performance improvement in this case. That said, you should not see much difference in general.

TyOverby:
> The function function binds this to its own scope, so you must explicitly bind this to prevent unintended behaviour in situations where this is treated differently.

Does this ever come up in jsoo?

smorimoto:
@TyOverby See `runtime/fs_fake.js`

smorimoto:
> Also, why are you interested in using arrow functions and avoid using `arguments`?

`arguments` is not `Array`, it's `ArrayLike`, and converting it to `Array` has some cost. Also, newer JS engines have better optimisation for the rest parameter.

runtime/jslib.js
smorimoto: Oops, `splice` is better than `length`: https://jsben.ch/nRrGR

smorimoto: https://jsben.ch/dsvFX

runtime/jslib.js
smorimoto: https://jsben.ch/ZZfEd

smorimoto:
The rest parameter and arguments show almost no difference in the benchmark results for major browsers: https://jsben.ch/BQEVR

TyOverby:
> arguments is not Array, it's ArrayLike, and converting it to Array has some cost

But Chrome is converting it to an array under the hood too.  I benchmarked this a few years ago and found that using the spread operator had no impact on speed

smorimoto:
The rest parameter seems to be faster for many arguments: https://jsben.ch/Krmit

smorimoto:
> > arguments is not Array, it's ArrayLike, and converting it to Array has some cost
> 
> But Chrome is converting it to an array under the hood too. I benchmarked this a few years ago and found that using the spread operator had no impact on speed

It's true, but it has a certain advantage here because it simplifies the runtime code. If we are going to introduce const/let, ES6 is required, and if we don't see any performance regression, I think we can consider this kind of change positively.

TyOverby:
> The rest parameter and arguments show almost no difference in the benchmark results for major browsers: https://jsben.ch/BQEVR

On my machine at least, the new code is reliably 4% slower.

> The rest parameter seems to be faster for many arguments: https://jsben.ch/Krmit

I think we should prioritize calling functions that have a small number of arguments, as small-arity functions are more common in OCaml than functions with 100+ parameters.

> It's true, but it has a certain advantage here because it simplifies the runtime code. If we are going to introduce const/let, ES6 is required, and if we don't see any performance regression, I think we can consider this kind of change positively.

I modified your benchmark so that the arrays aren't empty, and to do some fast computation with them instead of logging (logging is slow, and mucks with benchmark results), and this new benchmark shows that even for 100-parameter calls, the old method is twice as fast (on Chrome at least) than using the new syntax: https://jsben.ch/sGBgk

smorimoto:
> I modified your benchmark so that the arrays aren't empty, and to do some fast computation with them instead of logging (logging is slow, and mucks with benchmark results), and this new benchmark shows that even for 100-parameter calls, the old method is twice as fast (on Chrome at least) than using the new syntax: https://jsben.ch/sGBgk

I'm seeing a strange result 🤔 (I'm also on the latest Chrome btw)

image

TyOverby:
Yeah, very strange.
Screenshot from 2024-09-26 12-54-30

smorimoto:
I ran it on the latest Chrome with Apple Silicon. What's your environment?

smorimoto:
Here is the result from Safari on the latest iOS:
IMG_0344

TyOverby:
> I ran it on the latest Chrome with Apple Silicon. What's your environment?

Latest Chrome x86-64

smorimoto:
I tested it again and it's still fast in all environments. I saw that it was very slow when ads loaded, so that might be the cause?

Chrome:
image


Firefox:
image

Safari:
image

hhugo:
> This PR looks like it contains some formatting changes (using arrow functions) and also rewrites some of the most foundational primitives in the jsoo runtime. Could you split these up so that that latter can be thouroughly reviewed and benchmarked?

@smorimoto, I think it would be good to split the two changes as suggested above so that they can be tested independently.

smorimoto:
To fix the noArguments, you need the arrow function, and to fix the useArrowFunction, you need to remove the arguments...

hhugo:
> To fix the noArguments, you need the arrow function, and to fix the useArrowFunction, you need to remove the arguments...

I don't see the relation, sorry. I would expect that only spread is needed to replace usage of `arguments`.

hhugo:
> After asking for the PR to be split up, I did the benchmarking which has me pretty confident that the change should instead be abandoned, so I don’t want anyone doing the extra work of splitting the feature up if it’s not going to be merged.

Which part is slower ?

TyOverby:
I suspect it’s the … args syntax, but I don’t know for sure.

hhugo:
@TyOverby, Have you run more that just micro benchmarks ? Were you able to see differences with real applications ?

TyOverby:
I haven’t run this on any real applications because we aren’t easily able to pull in jsoo branches until wasm is merged.  However, i think that microbenchmarks are sufficient to reject what is effectively a code formatting change

hhugo:
> I haven’t run this on any real applications because we aren’t easily able to pull in jsoo branches until wasm is merged. However, i think that microbenchmarks are sufficient to reject what is effectively a code formatting change

Except this PR is not only about formatting. Also, I've been tricked many times by javascript micro benchmarks because real program would sometime show opposite result. The last one was https://github.com/ocsigen/js_of_ocaml/pull/1647 that I ended up reverting.

TyOverby:
What is the purpose of the PR then?  Did some user of Js_of_ocaml need the runtime javascript to be more “modern” for some reason?

FWIW, I find the new style of code in  this PR to be harder to read than the original, so without a good reason to do it (like performance), i’m pretty opposed

hhugo:
> FWIW, I find the new style of code in this PR to be harder to read than the original, so without a good reason to do it (like performance), i’m pretty opposed

This PR will not be merge in its current state (You and I both suggested to split it up). I've extracted and reworked  part of it in https://github.com/ocsigen/js_of_ocaml/pull/1740 and I personally find the code much more digest in its new form.

smorimoto:
Benchmarking with CAMLBOY is probably useful: https://github.com/linoscope/CAMLBOY

smorimoto:
## 5.8.2

### First Attempt

```
ROM path: ./tobu.gb
  Frames: 5000
Duration: 3.127700
     FPS: 1598.618793
```

### Second attempt

```
ROM path: ./tobu.gb
  Frames: 5000
Duration: 2.946000
     FPS: 1697.216565
```

## This PR

### First Attempt

```
ROM path: ./tobu.gb
  Frames: 5000
Duration: 2.972200
     FPS: 1682.255568
```

### Second attempt

```
ROM path: ./tobu.gb
  Frames: 5000
Duration: 2.964000
     FPS: 1686.909582
```

smorimoto:
In fact, this PR scope is included, but neither good nor bad effects seem to happen, at least for CAMLBOY. We may need a benchmark that has a negative impact.

hhugo:
The more I look at it the more I dislike the systematic use of arrow functions. I think we should use them only when it improves readability significantly. If one show that it improves performance, we could perform the translation automatically on the fly. I believe we already have code for that.

hhugo:
I'm going to close this one in favor of #1740. I don't really know what do to with the systematic use of  arrow. Let reconsider later maybe

smorimoto:
OK

PR #1683, opened 2024-09-25
runtime: fix useNodejsImportProtocol

smorimoto:


hhugo:
@smorimoto, What's the mininum nodejs version for this change ? You should mention it in the readme.

smorimoto:
@hhugo See this doc: https://nodejs.org/api/esm.html#node-imports

hhugo:
That would be v16.0.0

PR #1684, opened 2024-09-25
!js-string implementation of caml_string_concat fix

rickyvetter:
If you write some code like (note this is not an ascii x!):
```ocaml
str ^ " ×"
```
And have compiled your code with `--disable use-js-string`, then the concat will not work correctly. I've changed to an implementation that seems correct, but likely has quite a few extra conversions. I'm not sure exactly what the original function was doing, but it felt complicated and so I was a bit nervous about fixing in a lower-level way.

I'd like to have tests in place but I'm not sure what the infrastructure looks like for testing with these flags?

### DIFF
--- a/runtime/mlBytes.js
+++ b/runtime/mlBytes.js
@@ -692,12 +692,12 @@ function caml_string_concat(a, b) {
 }
 
 //Provides: caml_string_concat
-//Requires: caml_convert_string_to_bytes, MlBytes
+//Requires: caml_string_of_jsbytes, caml_jsbytes_of_string
 //If: !js-string
 function caml_string_concat(s1, s2) {
-  s1.t & 6 && caml_convert_string_to_bytes(s1);
-  s2.t & 6 && caml_convert_string_to_bytes(s2);
-  return new MlBytes(s1.t, s1.c + s2.c, s1.l + s2.l);
+  return caml_string_of_jsbytes(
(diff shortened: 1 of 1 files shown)


hhugo:
> then the concat will not work correctly.

What does that mean ? What's the behavior you're seeing ?

> I'd like to have tests in place but I'm not sure what the infrastructure looks like for testing with these flags?

There are some test in `compiler/tests-compiler/test_string.ml`. They call `Util.compile_and_parse` but you might want to use  `Util.compile_and_run`.

hhugo:
Your implementation should just be equivalent to the existing one. If you inline a few function call. You get;
```js
  s1.t & 6 && caml_convert_string_to_bytes(s1);
  var a = s1.c;
  s2.t & 6 && caml_convert_string_to_bytes(s2);
  var b = s2.c;
  var x = a + b;
  return new MlBytes(0, x, x.lenght);
```

```js
  s1.t & 6 && caml_convert_string_to_bytes(s1);
  s2.t & 6 && caml_convert_string_to_bytes(s2);
  return new MlBytes(s1.t, s1.c + s2.c, s1.l + s2.l);
```

Also I wasn't able to reproduce your issue.

rickyvetter:
Sorry for the noise. You're right. There's a more complicated issue that isn't with this primitive.

PR #1687, opened 2024-09-26
Runtime: switch default case come last

hhugo:
It makes the code a bit easier to read
@vouillon, can you review this ?


hhugo:
cc @smorimoto

smorimoto (approved):

hhugo:
@smorimoto, I would have preferred a review from @vouillon before merging.

smorimoto:
Oops, sorry about that.

PR #1691, opened 2024-09-26
Runtime: Annotate all fall through switch case 

hhugo:
Based on #1687

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -40,7 +40,6 @@
       "suspicious": {
         "noAssignInExpressions": "off",
         "noDoubleEquals": "off",
-        "noFallthroughSwitchClause": "off",
         "noRedeclare": "off"
       }
     }
--- a/runtime/format.js
+++ b/runtime/format.js
@@ -76,9 +76,12 @@ function caml_parse_format(fmt) {
           i++;
         }
         i--;
+        break;
       case "d":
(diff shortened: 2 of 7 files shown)


runtime/format.js
```diff
@@ -76,9 +76,12 @@ function caml_parse_format(fmt) {
           i++;
         }
         i--;
+        break;
```
hhugo, line 79: @vouillon, I suspect the missing break was a mistake

runtime/parsing.js
```diff
@@ -225,12 +228,14 @@ function caml_parse_engine(tables, env, cmd, arg) {
           cmd = loop;
           continue the_loop;
         }
-      // Fall through
+      // Unreachable
```
hhugo, line 231: The condition above never "exit normally".

hhugo:
cc @smorimoto

biome.json
```diff
@@ -43,10 +43,9 @@
         "noAssignInExpressions": "off",
         "noControlCharactersInRegex": "off",
         "noDoubleEquals": "off",
-        "noFallthroughSwitchClause": "off",
+
```
smorimoto, line 46:
```suggestion
```

smorimoto (approved):

PR #1693, opened 2024-09-27
Compiler: introduce Targetint

hhugo:
This PR introduce a dedicated module and type of integers with the target/backend semantic.

The number of bits can be configured at runtime.


compiler/lib/parse_bytecode.ml
```diff
@@ -2305,7 +2303,10 @@ and compile infos pc state instrs =
         let x, _ = State.accu state in
         let y = Var.fresh () in
 
-        ( (Let (y, Prim (Eq, [ Pc (Int n); Pv x ])), loc) :: instrs
+        ( ( Let
+              (y, Prim (Eq, [ Pc (Int (Targetint.of_int32_warning_on_overflow n)); Pv x ]))
+          , loc )
+          :: instrs
```
OlivierNicole, line 2309: If the integer argument of the BEQ opcode does not fit in an int31, the bytecode is not correctly compilable to Wasm. Isn’t a mere warning a bit weak in this case? If the user decides to ignore it, they could get a garbage program.

compiler/lib/targetint.ml
```diff
@@ -0,0 +1,138 @@
+type t = Int32.t
+
+let num_bits_ = ref 0
+
+let set_num_bits x = num_bits_ := x
+
+let num_bits () =
+  match !num_bits_ with
+  | (31 | 32) as x -> x
+  | x -> failwith (Printf.sprintf "Targetint.num_bits %d unsupported" x)
```
OlivierNicole, line 10: My first reaction was: “Why hasn’t he used a module chosen according to the compilation target?” But then I understood that this information is not available at initialization phase. Interesting.

hhugo: We could do it at initialization phase by linking different libraries. But then, you cannot change your mind at runtime

compiler/lib/eval.ml
```diff
@@ -172,9 +119,7 @@ let eval_prim ~target x =
       | "caml_mul_float", _ -> float_binop l ( *. )
       | "caml_div_float", _ -> float_binop l ( /. )
       | "caml_fmod_float", _ -> float_binop l mod_float
-      | "caml_int_of_float", [ Float f ] ->
-          Some
-            (Int (Int32.of_float f |> Int.of_int32_warning_on_overflow |> Int.to_int32))
+      | "caml_int_of_float", [ Float f ] -> Some (Int (Targetint.of_float f))
```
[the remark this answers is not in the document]
hhugo: The result of int_of_float is unspecified if the argument doesn't fit in the range of representable integers. I think we should just not perform static evaluation in such case.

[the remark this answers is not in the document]
hhugo: fixed

compiler/lib/parse_bytecode.ml
hhugo: I've changed it to raise an exception

compiler/lib/targetint.ml
```diff
@@ -0,0 +1,144 @@
...
+  if offset = 0 then Int32.max_int else Int32.shift_right_logical (-1l) offset
+
+let max_int_i () = Int32.to_int (max_int ())
+
+let to_string x = Int32.to_string x
+
+let to_float x = Int32.to_float x
+
+let of_float_opt x =
+  if x < Int32.to_float (min_int ()) || x > Int32.to_float (max_int ())
```
hhugo, line 39: We need to swap branches to have this work correctly with nan

OlivierNicole (approved):
Looks good to me.

PR #1694, opened 2024-09-28
[compiler] [build-fs] Fix `target not set error`

ejgallego:
Regression from #1655 .

Not sure if `Config.target` should be better handled as a parameter?

### DIFF
--- a/compiler/bin-js_of_ocaml/build_fs.ml
+++ b/compiler/bin-js_of_ocaml/build_fs.ml
@@ -61,6 +61,7 @@ function jsoo_create_file_extern(name,content){
 }
 |}
   in
+  Config.set_target `JavaScript;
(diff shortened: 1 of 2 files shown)


hhugo:
> Not sure if `Config.target` should be better handled as a parameter?

This has been discussed in #1655. We already have similar situation with other configs. I'd rather solve the general issue all at once, instead of solving one specific case.

ejgallego:
IMO this doesn't need a changes entry, but happy to add one.

hhugo:
This PR means that we currently don't have any test running this command. Would you mind adding one ? Maybe in https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/tests-full/dune ?

ejgallego:
Done.

PR #1698, opened 2024-09-30
[ci] Add 32bit job for 4.14.2

ejgallego:
Let's see how far this gets, likely we will need some extra setup but depexts got much better in Opam 2.2.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,11 @@ jobs:
             skip-effects: true
             skip-test: false
             skip-doc: true
+          - os: ubuntu-latest
+            ocaml-compiler: "ocaml-variants.4.14.2+options,ocaml-option-32bit"
+            skip-effects: true
+            skip-test: false
+            skip-doc: true
           - os: macos-latest
             ocaml-compiler: "4.14"
             skip-effects: true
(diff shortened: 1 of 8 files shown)


ejgallego:
A first headache is that JSOO depends on the opam package `graphics`, but this package is broken on 32bit archs, so it seems.

Another build error that is easier to fix is:
```
File "benchmarks/sources/ml/splay.ml", line 192, characters 47-57:
192 |     let s = (s + 0x7ed55d16 + (s lsl 12)) land 0xffffffff in
                                                     ^^^^^^^^^^
Error: Integer literal exceeds the range of representable integers of type int
```

ejgallego:
Note that installing graphics locally with the correct multi-arch setup doesn't work either, so this is not due to missing `i386` packages in the Ubuntu version on CI.

hhugo:
You can rebase and should be able to drop the revert of targetint

hhugo:
> Note that installing graphics locally with the correct multi-arch setup doesn't work either, so this is not due to missing `i386` packages in the Ubuntu version on CI.

What is the error ?

hhugo:
Why not add a job for OCaml 5.2 32bit ?

ejgallego:
> Why not add a job for OCaml 5.2 32bit ?

I was actually thinking of that, it is a good idea IMO.

I kind of refrained last minute as I have no JSOO + OCaml 5.x experience, so I thought it could be wise to make the 4.12+32bit CI work first.

I am not in OCaml 5 myself due to lack of `memprof-limits` in 5.x series, which is critical to Coq coq-lsp users.

That doesn't matter for the JSOO version of Coq, however I still use 4.14 as my base dev branch. But I'm really looking forward to try JSOO 5.x with coq / coq-lsp / jsCoq

hhugo:
I'm merging this one, feel free to open a new one to add 5.2 32bit to the ci

ejgallego:
A problem with 5.2 + 32bit is that it is a byte only distribution, so JSOO will be a bytecode build, and thus quite slow.

Should we still go ahead with a byte-only CI?

I managed to compile coq-lsp with 5.2 + 32bit tho, but seems like this will become a dead end for us soon as coq-lsp requires `ppx` stuff, and for example, `ppx_inline_test` doesn't support 32bits anymore (`time_now.0.17.0` fails to build).

But at least it works now!

hhugo:
Testing byte only with 5.2 seems to be a good addition

PR #1700, opened 2024-10-01
[ci] Add 32bit 5.2.0 job

ejgallego:
Caveats:

- The OCaml switch in this case byte-only, so things are slow, as jsoo itself won't have a native mode

- JS ppx libs 0.17 don't really support 32bit builds, this will likely become a problem in the future (if not now actually, for example `ppx_inline_test.0.17.0` fails to build)

- jsCoq/coq-lsp works in this setup, which is IMO a great "data point" as to whether this setup is OK

p.s: Coq stops working in 5.2.0 if I pass `--enable=effects` tho, the error is `Reason: Uncaught RangeError: Maximum call stack size exceeded`, is that worth reporting?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,7 @@ jobs:
         # getting much better, but no luck yet, c.f:
         # https://github.com/ocaml/opam-repository/pull/26626
       - name: Install apt 32-bit dependencies
-        if: matrix.ocaml-compiler == 'ocaml-variants.4.14.2+options,ocaml-option-32bit' || matrix.ocaml-compiler == 'ocaml-variants.5.2.0+options,ocaml-option-32bit'
+        if: contains( matrix.ocaml-compiler, "ocaml-option-32bit")
         run: |
           sudo apt-get install aptitude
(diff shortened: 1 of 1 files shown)


ejgallego:
Indeed, even if I did check `git grep ppx_inline_test` in JSOO sources, the package got pulled transitively.

So we either:

- drop this PR
- fix `time_now.0.17.0` to build in 32bit (which may not be possible due to upstream policy)
- softer testing reqs for 5.2.0 + 32 bits

Any other idea?

hhugo:
No need to run the tests. Just build/install like we do for older ocaml version

ejgallego:
@hhugo should we still do `dune build @all --profile using-effects` in the 5.2.0+32bit job?

hhugo:
> @hhugo should we still do `dune build @all --profile using-effects` in the 5.2.0+32bit job?
> 
> As of now the `@runtest` part fails indeed.



> @hhugo should we still do `dune build @all --profile using-effects` in the 5.2.0+32bit job?
> 
> As of now the `@runtest` part fails indeed.

Yes, we could. You could change the ci to respect skip test when skip-effect is false

Issue #1704, opened 2024-10-04
[BUG] Fix implementation of ephemerons


vouillon:
Currently, we have a strong reference to the data. So, if the data references one of the keys, it will never be released.

The implementation in wasm_of_ocaml uses a chain of WeakMaps: to access the data, we iteratively perform a lookup using each key in turn.

The implementation is also probably racy. It uses a FinalizationRegistry which is asynchronous. So, a newly set data might get deleted if the previous data was considered as garbage. This is probably no longer an issue since ephemerons are immutable since OCaml 5.0.0.

PR #1707, opened 2024-10-08
Runtime: reimplement weak/ephemeron

hhugo:
First draft trying to address #1704.
I think we need more tests here.


hhugo:
@vouillon, this is ready for review.

runtime/weak.js
```diff
@@ -137,77 +156,100 @@ function caml_ephe_blit_key(a1, i1, a2, i2, len) {
...
+//Requires: caml_ephe_get_data
 //Requires: caml_obj_dup
 function caml_ephe_get_data_copy(x) {
-  if (x[caml_ephe_data_offset] === undefined) return 0;
-  else return [0, caml_obj_dup(x[caml_ephe_data_offset])];
+  var r = caml_ephe_get_data(x);
+  if (r === 0) return 0;
+  var z = r[1];
+  if (Array.isArray(z)) return [0, caml_obj_dup(z)];
+  return z;
```
vouillon, line 206:
```suggestion
  return r;
```

runtime/weak.js
```diff
@@ -137,77 +156,100 @@ function caml_ephe_blit_key(a1, i1, a2, i2, len) {
...
+    if (globalThis.WeakRef && k instanceof globalThis.WeakRef) {
+      var d = k.deref();
+      if (d === undefined) {
+        x[i] = caml_ephe_none;
+        continue;
+      }
+      if (globalThis.WeakMap) {
+        var data2 = new globalThis.WeakMap();
+        data2.set(k, data);
+        data = data2;
```
vouillon, line 224:
```suggestion
        data = new globalThis.WeakMap().set(k,data);
```

runtime/weak.js
```diff
@@ -114,21 +120,34 @@ function caml_ephe_get_key_copy(x, i) {
 }
 
 //Provides: caml_ephe_check_key mutable
-//Requires: caml_ephe_key_offset
+//Requires: caml_ephe_key_offset, caml_ephe_data_offset
+//Requires: caml_invalid_argument
+//Requires: caml_ephe_none
 //Alias: caml_weak_check
 function caml_ephe_check_key(x, i) {
+  if (i < 0 || caml_ephe_key_offset + i >= x.length)
```
vouillon, line 128: You can remove the checks. They have been moved to the OCaml code since OCaml 4.08 (https://github.com/ocaml/ocaml/pull/676).

vouillon (changes requested):
Except for a few points, this looks good to me. Thanks!

runtime/weak.js
hhugo: done

runtime/weak.js
hhugo: good catch

Issue #1622, opened 2024-06-10
[BUG] Using two Js_of_ocaml compiled libraries in the same runtime causes issues with js stubs


troederjs:
**Describe the bug**
If two separate ocaml libraries have JS stubs which define the same symbol ion `joo_global_object` (e.g. two different versions of [src-bindings/node/dune](https://github.com/ocamllabs/vscode-ocaml-platform/blob/ec65574c1df971a6632af3b6ffd2b051f0ebf649/src-bindings/node/dune#L6) in the OCaml VSCode extension) they will conflict when both libraries are loaded by different `Js_of_ocaml` executable.

This happens because `joo_global_object` is the `globalThis`, and so the two values override each other.

Normally, when running a single application this is not a problem since we can control all the dependencies and ensure they don't conflict. But specifically in the case of the VSCode extension, this causes issues when there are two `Js_of_ocaml` extensions that use incompatible versions of the nodejs binding stubs.

**Expected behavior**
Both `Js_of_ocaml` executable should be able to run independently in the same JS environment.

hhugo:
First note that `joo_global_object` should be avoided, support will be removed at some point. One should use `globalThis` instead.

It seems to me that vscode-ocaml-platform is part of the problem here. It should not update the global object,it should defined its own stubs.

hhugo:
I've worked on a bunch of changes to limit global side effects (#1708, #1696) 
Assuming bindings are written correctly (not by mutating the global object), two jsoo programs should be able to coexists without too much interference. I'll try to list and document the remaining side effects at some point.

hhugo:
I'm opening https://github.com/ocamllabs/vscode-ocaml-platform/issues/1617 and will close this one

PR #1708, opened 2024-10-09
Compiler: revisit static env handling

hhugo:


hhugo:
Should help with https://github.com/ocsigen/js_of_ocaml/issues/1622

Issue #1703, opened 2024-10-04
[BUG] Performance issue with Yojson


vouillon:
Parsing a large string (~ 1MiB) with Yojson is really slow. We should investigate why.

hhugo:
How slow is really slow ? Are you sure the slowness comes from yojson ? Can you describe the issue a bit more ?

hhugo:
I see. I was able to reproduce.
- `Yojson.Raw.from_string` is not (much less) affected because it processes the whole string at once.
- `--disable use-js-string` solves the issue
- We should probably fix theses patterns in the ocaml/ocaml codebase
- We should revisit the decision to change the underlying representation of bytes when converting to string (when use-js-string is enabled)

hhugo:
@vouillon, see https://github.com/ocaml/ocaml/pull/13543

vouillon:
Cool! Thanks!

hhugo:
- https://github.com/ocaml/ocaml/pull/13543 has been merged.
- #1713 alone fixes most of the issue reported here
- #1712 might need more testing to understand the impact

PR #1712, opened 2024-10-11
Runtime: don't change the shape of bytes when converting to string

hhugo:
should help with #1703


hhugo:
I'm no longer sure this is what we want in the short term. Turning into draft

Issue #1446, opened 2023-04-05
[FEATURE REQUEST] Use source-map sections in js_of_ocaml link command.


hhugo:
Support sourcemap sections fields so that we no longer need to merge sourcemaps mappings in `js_of_ocaml link`
```
Index map: supporting post processing
To support concatenating generated code and other common post processing, an alternate representation of a map is supported:
{
"version" : 3,
"file": “app.js”,
"sections": [
 { "offset": {"line":0, "column":0}, "url": “url_for_part1.map” }
 { "offset": {"line":100, "column":10}, "map": 
   {
     "version" : 3,
     "file": “section.js”,
     "sources": ["foo.js", "bar.js"],
     "names": ["src", "maps", "are", "fun"],
     "mappings": "AAAA,E;;ABCDE;"
   }
 }
],
}


The index map follow the form of the standard map
Line 1: The entire file is an JSON object.
Line 2: The version field. See the description of the standard map.
Line 3: The name field. See the description of the standard map.
Line 4: The sections field.

The “sections” field is an array of JSON objects that itself has two fields “offset” and a source map reference.  “offset” is an object with two fields, “line” and “column”, that represent the offset into generated code that the referenced source map represents.
The other field must be either “url” or “map”. A “url” entry must be a URL where a source map can be found for this section and the url is resolved in the same way as the “sources” fields in the standard map. A “map” entry must be an embedded complete source map object.  An embedded map does not inherit any values from the containing index map.

The sections must be sorted by starting position and the represented sections may not overlap.

```

PR #1714, opened 2024-10-14
Handle composite sourcemaps

OlivierNicole:
This has been extracted from #1617 upon @hhugo's suggestion. “Index” sourcemaps are useful to merge multiple sourcemaps at a low cost and is part of the optimization in #1617.

### DIFF
--- a/compiler/lib/source_map.ml
+++ b/compiler/lib/source_map.ml
@@ -373,25 +373,31 @@ module Standard = struct
   let json t =
     let stringlit s = `Stringlit (Yojson.Safe.to_string (`String s)) in
     `Assoc
-      [ "version", `Intlit (string_of_int t.version)
-      ; "file", stringlit (rewrite_path t.file)
-      ; ( "sourceRoot"
-        , stringlit
-            (match t.sourceroot with
-            | None -> ""
-            | Some s -> rewrite_path s) )
-      ; "names", `List (List.map t.names ~f:(fun s -> stringlit s))
-      ; "sources", `List (List.map t.sources ~f:(fun s -> stringlit (rewrite_path s)))
-      ; "mappings", stringlit (Mappings.to_string t.mappings)
-      ; ( "sourcesContent"
-        , `List
-            (match t.sources_content with
-            | None -> []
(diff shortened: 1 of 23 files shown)


compiler/lib/source_map.ml
```diff
@@ -438,8 +438,116 @@ let of_json (json : Yojson.Raw.t) =
...
+                     section"
+              in
+              { gen_line; gen_column }
+          | _ ->
+              invalid_arg "Source_map_io.Index.of_json: 'offset' field of unexpected type"
+        in
+        (match List.assoc "url" json with
+        | _ ->
+            invalid_arg
+              "Source_map_io.Index.of_json: URLs in index maps are not currently \
```
hhugo, line 514: Is this a limitation of the spec or of this implementation ( why dont we support it ?)

hhugo: I found two sources that disagree on the presence of the url form in the spec.

vouillon: It's part of [the spec](https://sourcemaps.info/spec.html), but since the Chrome DevTools does not support it, I'm not sure many tools generate such files.
https://github.com/ChromeDevTools/devtools-frontend/blob/d8095064b341d27f160550b05904a86d05b7ac82/front_end/core/sdk/SourceMap.ts#L211-L216

OlivierNicole: And we happen not to need it.

hhugo: https://tc39.es/source-map/#section-object doesn't mention it

vouillon: Apparently, no one supports this, so it has been dropped from the spec. See https://github.com/tc39/source-map/issues/29.

hhugo:
@OlivierNicole, I took the liberty to refactor the source_map module a bit and fix the CI.
I've also added a small separate commit to use the index sourcemap instead of merge inside Link_js.

hhugo:
partially fix https://github.com/ocsigen/js_of_ocaml/issues/1446

PR #1715, opened 2024-10-17
Compiler: refactor sourcemap generation, use Index sourcemap

### LINKED ISSUE #1446: [FEATURE REQUEST] Use source-map sections in js_of_ocaml link command.

hhugo:
This PR leverages the recent support for index sourcemap to make sourcemap generation faster.
We no longer decode or edit mappings.

When compiling cma to js. We emit one sourcemap section per compilation unit. This allows to copy one compilation unit with its mappings without having to decode/change its mapping.

Perf
===
Compiling js_of_ocaml with itself (`dune build compiler/bin-js_of_ocaml/js_of_ocaml.bc.js`).
The final link time goes from 660ms on master to 440ms with this PR

Building the toplevel in toplevel/examples/lwt_toplevel/,
The final link time goes from 1350ms (620 ms processing sourcemap) on master to 730ms (90ms processing sourcemap) with this PR
  

X-link
====
build on #1714 
replace #1617 
fix #1446


hhugo:
@OlivierNicole, do you want to take a look at this ?

compiler/lib/link_js.ml
```diff
@@ -447,22 +449,48 @@ let link ~output ~linkall ~mklib ~toplevel ~files ~resolve_sourcemap_url ~source
...
+                    sm.Source_map.Index.sections
+                    ~f:(fun ({ gen_line; gen_column }, `Map sm) ->
+                      ( gen_line + Source_map.Mappings.first_line sm.mappings
+                      , gen_line + Source_map.Mappings.number_of_lines sm.mappings
+                      , gen_line
+                      , gen_column
+                      , sm ))
+            in
+            List.concat_map reloc ~f:(function
+                | `Drop _ -> []
```
hhugo, line 473: This part needs to be reworked

compiler/lib/link_js.ml
```diff
@@ -447,22 +449,48 @@ let link ~output ~linkall ~mklib ~toplevel ~files ~resolve_sourcemap_url ~source
...
+                | `Copy (src, dst, len) ->
+                    List.filter_map sm ~f:(fun (first, last, gen_line, gen_column, sm) ->
+                        if first > src + len || last < src
```
hhugo, line 476: This needs to be adjusted

compiler/tests-sourcemap/dump_sourcemap.ml
```diff
@@ -74,12 +74,12 @@ let print_mapping lines ?(line_offset = 0) (sm : Source_map.Standard.t) =
                 ori_line
                 ori_col
                 gen_col
-                (mark gen_col lines.(gen_line - 1 + line_offset))
```
hhugo, line 77: This diff is suspicious

hhugo: fixed

compiler/lib/link_js.ml
hhugo: added a comment

compiler/lib/link_js.ml
hhugo: done

hhugo:
@rickyvetter, would you be able to test this PR ?

OlivierNicole:
What happens to lines that can be added, such as here? Don’t they require editing the mappings?

https://github.com/ocsigen/js_of_ocaml/blob/81aec6da0250660783a4e3ca8ab1208f8aace610/compiler/lib/link_js.ml#L123

OlivierNicole:
Ok I understand, in the end, it’s fine because it never results in adding lines in the middle of a file, but this mechanism (that sometimes add a `//# <file> <line>` comment when changing files in `Line_writer.write`) seems brittle and makes the code unclear. Could we do this outside of `Line_writer.write`?

hhugo:
> Ok I understand, in the end, it’s fine because it never results in adding lines in the middle of a file, but this mechanism (that sometimes add a `//# <file> <line>` comment when changing files in `Line_writer.write`) seems brittle and makes the code unclear. Could we do this outside of `Line_writer.write`?

I've just simplified the logic a bit.

rickyvetter:
I'm seeing similar improvements to as #1617. I ran across 6 of our bigger apps and they all averaged 45% reduction in linking time.

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -65,6 +74,52 @@ let output_gen ~standalone ~custom_header ~build_info ~source_map output_file f
...
+  match Builtins.find file with
+  | Some f -> Some (Source_map.Source_content.create (Builtins.File.content f))
+  | None ->
+      if Sys.file_exists file && not (Sys.is_directory file)
+      then
+        let content = Fs.read_file file in
+        Some (Source_map.Source_content.create content)
+      else None
+
+let build_source_map_with_offset
```
OlivierNicole, line 87: It’s particularly obscure what this function does (apparently, turning a `Source_map.info` into a `Source_map.Standard.t` while ignoring the initial unmapped lines and using filenames from an existing `Source_map.Standard.t`…). A comment would save many minutes of code reading.

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -65,6 +74,52 @@ let output_gen ~standalone ~custom_header ~build_info ~source_map output_file f
...
+    List.map sources ~f:(fun filename ->
+        match Builtins.find filename with
+        | None -> filename
+        | Some _ -> Filename.concat "/builtin" filename)
+  in
+  let line_offset, mappings = Source_map.Mappings.encode_with_offset mappings in
+  ( line_offset
+  , { (sm : Source_map.Standard.t) with sources; sources_content; names; mappings } )
+
+let build_source_map_many base l =
```
OlivierNicole, line 105: It’s not much clearer what this one does. In fact, all the code here seems to rest on what `Driver.f` does with sourcemaps, which is undocumented and takes time to figure out as it is spread across so many files. Couldn’t we document `Driver.f`?

compiler/lib/source_map.mli
```diff
@@ -57,10 +59,16 @@ module Mappings : sig
...
       {!val:of_string} and {!val:to_string} are inverse functions. *)
 
-  val decode : t -> map list
+  val decode : t -> decoded
   (** Parse the mappings. *)
 
-  val encode : map list -> t
+  val encode : decoded -> t
+
+  val encode_with_offset : decoded -> int * t
```
OlivierNicole, line 67: We should really document how this function differs from `encode` (IIUC it discards the first unmapped lines and returns their number).

OlivierNicole (approved):
This looks good to me, except for the fact that the sourcemap handling in `Compile` is hard to understand and review, and I think it would really help to add three documentation comments at the right places.

compiler/lib/source_map.mli
hhugo: Done some more refactoring/renaming/documenting

compiler/bin-js_of_ocaml/compile.ml
hhugo: What make you think it depends on what `Driver.f` does ? `Driver.f` just return a `Source_map.info`.  There is no much magic there.

compiler/bin-js_of_ocaml/compile.ml
hhugo: I've done some refactoring and renaming. I did not add comment yet but it should be much easier to read.

compiler/lib/link_js.ml
```diff
@@ -447,22 +451,62 @@ let link ~output ~linkall ~mklib ~toplevel ~files ~resolve_sourcemap_url ~source
...
+                               sourcemap, but we could in theory filter out part of it. *)
+                            assert (src <= first && last <= src + len);
+                            Some (first, last, gen_line + dst - src, gen_column, sm))))
+            in
+            (* Make sure dropped sections are not overlapping selected sourcemap. *)
+            List.iter reloc ~f:(function
+                | `Copy _ -> ()
+                | `Drop (src, len) ->
+                    List.iter maps ~f:(fun (first, last, _, _, _) ->
+                        if first > src + len || last < src then () else assert false));
```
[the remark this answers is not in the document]
hhugo: no lines are added or removed in the middle of lines covered by a single `Sourcemap.Standard.t`

hhugo:
@OlivierNicole, may I request another round of review ?

compiler/bin-js_of_ocaml/compile.ml
OlivierNicole: My bad, I thought there was some accumulation going on in the returned source map but I misread.

OlivierNicole (commented):
Looks good to me

PR #1716, opened 2024-10-18
Source map improvements

vouillon:
I started working on improving the source map output of Wasm_of_ocaml, and got a bit side-tracked...

Many changes to preserve debug information from the bytecode as well as possible and put this information in the source map at appropriate places for the JavaScript debuggers.


compiler/lib/js_output.ml
```diff
@@ -39,6 +39,78 @@
...
+       ^
+
+Chrome uses the name associated to the location of each bound variable
+to determine its name.
+
+   function f(x) { var y = ... }
+            ^ ^        ^
+
+Chrome uses the location of the opening parenthesis of a function
+declaration to determine the function name in the stack.
```
hhugo, line 108: Where do these infos come from ? Are there any link you can join to this doc ?

compiler/lib/js_output.ml
```diff
@@ -456,6 +456,33 @@ struct
     in
     traverse l e
 
+  let stop_on_statement st =
```
hhugo, line 492: Maybe add a comment here 
```
The debuggers do not stop on some statements, like function
declarations. So there is no point in outputting some debug information
there.
```

Where does this info come from ?

compiler/lib/js_traverse.ml
```diff
@@ -1750,8 +1750,13 @@ class simpl =
           | If_statement (ENum n, _, iffalse) when Num.is_zero n -> opt_cons iffalse rem
           (* if (e1) return e2 else return e3 --> return e1 ? e2 : e3 *)
           | If_statement
-              (cond, (Return_statement (Some e1), _), Some (Return_statement (Some e2), _))
-            -> (Return_statement (Some (ECond (cond, e1, e2))), loc) :: rem
+              ( cond
+              , (Return_statement (Some e1, _), _)
+              , Some (Return_statement (Some e2, _), _) ) ->
+              ( Return_statement
+                  (Some (ECond (cond, e1, e2)), U (*ZZZ Use end of function? *))
```
hhugo, line 1757: Is this a todo ?

compiler/lib/ocaml_compiler.ml
```diff
@@ -50,6 +50,10 @@ let rec find_loc_in_summary ident' = function
   | Env.Env_empty -> None
   | Env.Env_value (_summary, ident, description) when Poly.(ident = ident') ->
       Some description.Types.val_loc
+  | Env.Env_module (_summary, ident, _, description) when Ident.same ident ident' ->
+      Some description.Types.md_loc
+  | Env.Env_extension (_summary, ident, description) when Ident.same ident ident' ->
+      Some description.Types.ext_loc
```
hhugo, line 56: Why doesn't this affect tests-full ?

compiler/lib/specialize_js.ml
```diff
@@ -185,6 +185,30 @@ let specialize_instrs ~target info l =
...
+      ; (Event _, _)
+      ; (Let (bytes, Prim (Extern "caml_create_bytes", [ Pv len' ])), _)
+      ; (Event _, _)
+      ; ( Let
+            ( u1
+            , Prim
+                ( Extern "caml_blit_string"
+                , [ Pv a'; Pc (Int zero1); Pv bytes'; Pc (Int zero2); Pv alen'' ] ) )
+        , _ )
+      ; (Event _, _)
```
hhugo, line 202: This pattern seems fragile, can we do better ?

compiler/tests-compiler/effects_continuations.ml
```diff
@@ -103,79 +103,86 @@ let%expect_test "test-compiler/lib-effects/test1.ml" =
...
     }
     //end
     function cond1(b, cont){
-     function _o_(ic){return cont([0, ic, 7]);}
+     function _p_(ic){return cont([0, ic, 7]);}
      return b
-             ? caml_cps_call2(Stdlib[79], cst_toto$0, _o_)
-             : caml_cps_call2(Stdlib[79], cst_titi, _o_);
+             ? caml_cps_call2
+               (Stdlib[79], cst_toto$0, function(_q_){return _p_(_q_);})
```
hhugo, line 147: Why do we have this diff ?

hhugo: Maybe some inlining optim needs to be updated ?

toplevel/examples/lwt_toplevel/dune
```diff
@@ -39,7 +39,10 @@
   (:standard -linkall))
  (modes byte js)
  (js_of_ocaml
-  (link_flags (:standard))
+  (link_flags
+   :standard
+   ; The source map would be too large on 32-bit architectures otherwise
+   --no-source-map)
```
hhugo, line 45: Wouldn't it be enough to disable the inline sourcemap ?
```
(link_flags (:standard --source-map \ --source-map-inline)) 
```

Or maybe just disable sourcemap on 32bit arch only

compiler/lib/generate.ml
```diff
@@ -501,55 +501,165 @@ let constant ~ctx x level =
 type queue_elt =
   { prop : int
   ; ce : J.expression
-  ; loc : J.location
```
hhugo, line 507: The `let*` rewrite make it hard to review the rest of the diff. could this commit be split, maybe ?

compiler/lib/parse_bytecode.ml
```diff
@@ -975,6 +967,42 @@ and compile infos pc state instrs =
...
+            push_event Before source event instrs
+        | { ev_kind = Event_after _ | Event_pseudo; _ }, _ ->
+            if debug_parser () then Format.eprintf "Ignored useless event@.";
+            (* Not interesting:
+               - before a throw instruction, but we already have an event
+                 for the exception
+               - omitted else clause
+            *)
+            instrs
+        | _, _ ->
```
hhugo, line 971: Can we have an exhaustive pattern for event (at least on ev_kind) ?

compiler/lib/js_output.ml
```diff
@@ -1996,25 +2149,52 @@ let program ?(accept_unnamed_var = false) f ?source_map p =
               | None -> filename
               | Some _ -> Filename.concat "/builtin" filename)
         in
+        let ignore_list =
+          List.filter sources ~f:(fun filename ->
+              String.length filename >= 9
+              && String.equal (String.sub filename ~pos:0 ~len:9) "/builtin/")
```
hhugo, line 2155: String.starts_with

compiler/lib/js_output.ml
vouillon: I did some tests.

compiler/tests-compiler/effects_continuations.ml
vouillon: For the CPS transformation, we are splitting blocks if a function call is followed by some instructions.
Since I added an event instruction, this resulted in more splitting. But it seems fine to just skip events in this case.

compiler/lib/generate.ml
vouillon: It is splitted.

toplevel/examples/lwt_toplevel/dune
vouillon: Yes, this is enough.

I'm not sure how to disable sourcemap on 32bit arch only.

compiler/lib/js_output.ml
vouillon: This is only available since OCaml 3.13. But maybe we should add it to `stdlib.ml`?

hhugo: We have String.is_prefix in our stdlib

toplevel/examples/lwt_toplevel/dune
hhugo: > I'm not sure how to disable sourcemap on 32bit arch only.

Emit flags in a file and use `(:include filename)` in the dune stanza.

vouillon: Do you think it is worth it?

hhugo: Looking more at it. Having sourcemap enabled here was probably a mistake in the first place. We commit the generated file in the wikidoc branch. Better to not  bloat it with sourcemap.

hhugo:
Can you rebase (and squash relevant commits) into master. I've just merged #1715

compiler/lib/js_output.ml
```diff
@@ -1371,7 +1371,13 @@ struct
 
   and statement ?(last = false) f (s, loc) =
     let can_omit_semi = PP.compact f && last in
-    let last_semi () = if can_omit_semi then () else PP.string f ";" in
+    let last_semi ?(ret = false) () =
+      if can_omit_semi
+      then ()
+      else if ret && source_map_enabled
+      then PP.string f "; "
```
hhugo, line 1411: This can result in a trailing white space. Do we need a space if the next char is a newline ?

compiler/lib/ocaml_compiler.ml
hhugo: Should the commit message be "indicate the origin location of a modules ..." ?

compiler/lib/js_output.ml
vouillon: Actually we don't need to add a space when pretty-printing. And a newline make more sense in compact mode. I have updated this.

compiler/lib/js_output.ml
```diff
@@ -1732,19 +1731,19 @@ struct
         | SideEffect -> ()
         | Default i ->
             PP.space f;
-            ident f i
+            ident f ~kind:`Binding i
         | Namespace (def, i) ->
             Option.iter def ~f:(fun def ->
                 PP.space f;
-                ident f def;
+                ident f ~kind:`Reference def;
```
hhugo, line 1778: Shouldn't this be binding?

compiler/lib/js_output.ml
```diff
@@ -853,9 +852,9 @@ struct
     | EAssignTarget t -> (
         let property f p =
           match p with
-          | TargetPropertyId (Prop_and_ident id, None) -> ident f id
+          | TargetPropertyId (Prop_and_ident id, None) -> ident f ~kind:`Reference id
```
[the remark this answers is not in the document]
vouillon: This could be an implicit variable declaration, indeed. But this is bad practice and the identifier should have been declared somewhere else. The expression `EVar v` can also declare a variable, by the way, when it is on the right-hand side of an assignment operator.
I'm not sure we want to have larger source maps just to support this.

compiler/lib/source_map.ml
```diff
@@ -520,7 +565,8 @@ module Standard = struct
   let to_file m file = Yojson.Raw.to_file file (json m)
 
   let invariant
-      { version; file = _; sourceroot = _; names; sources_content; sources; mappings } =
+      { version; file = _; sourceroot = _; names; sources_content; sources; mappings; _ }
```
hhugo, line 568: You could check that ignore list is a subset of sources, or add a comment if we should not do that

compiler/lib/effects.ml
```diff
@@ -139,6 +139,28 @@ let dominance_frontier g idom =
     g.preds;
   frontiers
 
+(* Last instruction of a block, ignoring events *)
+let rec last_instr l =
+  match l with
+  | [] -> None
+  | [ i ] | [ i; (Event _, _) ] -> Some i
```
hhugo, line 146: This works because we avoid consecutive events. Maybe this invariant should be added to code.ml

compiler/lib/js_output.ml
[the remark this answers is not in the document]
hhugo: Ok.

compiler/lib/ocaml_compiler.ml
vouillon: Indeed. But actually, we no longer use this information.

compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -497,7 +497,7 @@
      i =  /*<<stdlib.ml:285:2>>*/ 0;
     for(;;){
       /*<<stdlib.ml:280:4>>*/ if(l <= i)
-       /*<<stdlib.ml:280:19>>*/ return s1 + ".";
+       /*<<stdlib.ml:213:2>>*/ return s1 + ".";
```
hhugo: Why do we now see the location of the body of string concat ?

compiler/lib/js_output.ml
```diff
@@ -886,9 +885,9 @@ struct
         let element f p =
           match p with
           | TargetElementHole -> ()
-          | TargetElementId (id, None) -> ident f id
+          | TargetElementId (id, None) -> ident f ~kind:`Binding id
           | TargetElementId (id, Some (e, _)) ->
-              ident f id;
+              ident f ~kind:`Binding id;
```
hhugo, line 916: Why is this a `Binding` while it is a `Reference` in TargetPropertyId ?

compiler/lib/js_output.ml
```diff
@@ -1941,21 +2102,35 @@ let program ?(accept_unnamed_var = false) ?(source_map = false) f p =
...
+              Source_map.Gen { gen_col; gen_line }
+          | Source_map.Gen_Ori m -> Source_map.Gen_Ori { m with gen_line; gen_col }
+          | Source_map.Gen_Ori_Name m ->
+              Source_map.Gen_Ori_Name { m with gen_line; gen_col }
+        in
+        let rec build_mappings pos mapping prev_mappings =
+          match mapping with
+          | [] -> prev_mappings
+          | (pos', m) :: rem ->
+              if pos'.PP.p_line = pos.PP.p_line
```
hhugo, line 2119: You should add a comment explaining the mapping duplication here.

compiler/lib/inline.ml
```diff
@@ -303,7 +309,9 @@ let inline ~first_class_primitives live_vars closures name pc (outer, p) =
         | Let (x, Closure (l, (pc, []))), loc when first_class_primitives -> (
             let block = Addr.Map.find pc p.blocks in
             match block with
-            | { body = [ (Let (y, Prim (Extern prim, args)), _loc) ]
+            | { body =
+                  ( [ (Let (y, Prim (Extern prim, args)), _loc) ]
+                  | [ (Event _, _); (Let (y, Prim (Extern prim, args)), _loc) ] )
```
hhugo, line 314: I think there can be an event after the `Let` as well

hhugo:
I see diffs such as 
```diff
-|    /*<<stdlib.ml:32:0>>*/  /*<<stdlib.ml:32:0>>*/ var
-|    Exit = [248, "Stdlib.Exit", runtime.caml_fresh_oo_id(0)];
+|   var
+|    Exit =
+|       /*<<stdlib.ml:24:44>>*/ [248,
+|       "Stdlib.Exit",
+|       runtime.caml_fresh_oo_id(0)];
```

The new location seems to points to the end of the previous function call in the file. It could be what we want for sourcemap but it degrades the debug output quite a bit. What do you think ?

hhugo:
running #1717 on `bedrock2_fiat_crypto.byte`

master gives:
> 87.50user 1.11system 1:28.66elapsed 99%CPU (0avgtext+0avgdata 4179372maxresident)k
24inputs+187296outputs (0major+1076275minor)pagefaults 0swaps

 #1717 gives:
> 71.06user 1.26system 1:12.37elapsed 99%CPU (0avgtext+0avgdata 4491440maxresident)k
0inputs+189256outputs (0major+1154258minor)pagefaults 0swaps

The speedup is nice ! Any intuition about the source of the memory peak increase ?

cc @JasonGross

hhugo:
without sourcemap I get a smaller memory footprint with #1717

master:
> 58.62user 1.07system 1:00.04elapsed 99%CPU (0avgtext+0avgdata 3468748maxresident)k
0inputs+19688outputs (0major+1020857minor)pagefaults 0swaps

1717 
> 59.62user 1.01system 1:00.84elapsed 99%CPU (0avgtext+0avgdata 3025052maxresident)k
0inputs+19688outputs (0major+947949minor)pagefaults 0swaps

compiler/tests-full/stdlib.cma.expected.js
vouillon: I was wondering whether to keep the event at the beginning of the function. But we don't really need it if the function is inlined. And if the function is not inlined, the Chrome debugger is going to skip over it if there is not event, which is probably what we want as well.

compiler/lib/js_output.ml
vouillon: Oh, this is wrong.

vouillon:
Indeed, the locations have been optimized for source maps. The issue is that allocations can be reordered with other instructions. So, if you annotated them with their actual location, you could move back and forth when single stepping in the debugger, which is not intuitive. Also, the debugger will not stop on allocations, only on variable declaration. So, if the allocation is nested within an expression, there is no point to indicate each location in source maps.

If we want an accurate debug output, we would need a different strategy.

vouillon:
The memory increase might be because events hold a `Parse_info.t` value.
Maybe we don't need to keep the debug information until code generation. We still use it to get the end location of functions. But we could put this information into the closure constructor:
```diff
-  | Closure of Var.t list * cont
+  | Closure { params : Var.t list; cont : cont; end_loc : Parse_info.t option }
```

hhugo:
I think we can merge the current diff.

I have done superficial review of individual commits as well as a superficial review of the entire diff.

I'm not very happy with the state of the debug output but because it reflects the behavior needed for sourcemap it is useful in this form as well.

@vouillon, Can you squash the fixup commits again ? (before we merge)

Note that I've pushed 2 small cleanups.

hhugo:
> The memory increase might be because events hold a `Parse_info.t` value. Maybe we don't need to keep the debug information until code generation. We still use it to get the end location of functions. But we could put this information into the closure constructor:
> 
> ```diff
> -  | Closure of Var.t list * cont
> +  | Closure { params : Var.t list; cont : cont; end_loc : Parse_info.t option }
> ```
> 
> This is a bit invasive, so maybe this could be done later.

We can leave such optimisation for another time. Feel free to get side-tracked !

PR #1717, opened 2024-10-18
Remove location from instructions

vouillon:
To save some rebasing work, this should probably be merged after #1461.


hhugo:
I've rebased the branch

PR #1718, opened 2024-10-18
Global dead code: add regression test for bug fixed in #1681

vouillon:


compiler/tests-compiler/global_deadcode.ml
```diff
@@ -129,3 +129,30 @@ let%expect_test "Omit unused return expressions" =
...
+          x.a <- 1; (* This has to be handled after [x] is returned *)
+          {a = 3; b = 4}
+        )
+      let g = ref (fun _ -> assert false)
+      let _ =
+        (* We should not track that [f] is used below *)
+        g := f; prerr_int ((!g true).b + (!g false).b)
+      |}
+  in
+  (* Expect return value of f to be omitted. *)
```
hhugo, line 152: I don't understand this comment ?

vouillon: I copy-pasted the code and forgot to update the comment...

PR #1721, opened 2024-10-26
Consume Hints produced by ocaml#13438

hhugo:
See https://github.com/ocaml/ocaml/pull/13438


OlivierNicole:
Do we have an idea of the scale of the performance gains that could be made with this?

hhugo:
> Do we have an idea of the scale of the performance gains that could be made with this?

I don't know exactly, I've rebased the PR and its ocaml companion https://github.com/hhugo/ocaml/tree/optimization-hints but we can't currently run the benchmarks with OCaml 5.3 (see #2035)

hhugo:
@vouillon do you mind looking at https://github.com/ocsigen/js_of_ocaml/actions/runs/28660395878/job/84999338096?pr=1721 ?

vouillon:
> @vouillon do you mind looking at https://github.com/ocsigen/js_of_ocaml/actions/runs/28660395878/job/84999338096?pr=1721 ?

Your fix looks good. The issue is already with OCaml 5.5.

hhugo:
@vouillon, have you reviewed this already ? Do you think it's mergeable in its current state ?

vouillon:
> @vouillon, have you reviewed this already ? Do you think it's mergeable in its current state ?

I have just updated `effect-native.wat` for ocaml 5.6. If the CI is all green, I think this is mergeable.

hhugo:
The ocaml 5.6 job take 50min for jsoo.. Maybe we should only test it on master (while the issue is beeing fixed)

PR #1631, opened 2024-06-26
Use dune test stanza 

vouillon:
This requires some fixes in dune...

PR #1724, opened 2024-10-31
Merge wasm_of_ocaml

OlivierNicole:
This aims to merge wasm_of_ocaml, currently hosted at https://github.com/ocaml-wasm/wasm_of_ocaml, back into this repo.

~~This is pending on Dune support ocaml/dune#11029. For now, Dune is pinned for the CI jobs.~~


vouillon:
@hhugo You can have a first look.

Regarding the directory structure, the wasm-specific code of wasm_of_ocaml is in `compiler/lib/wasm` (and  `compiler/lib/bin-wasm_of_ocaml` for the binary itself).  Initially, these files were included in the `js_of_ocaml-compiler` library, so I added a `wa_` prefix to avoid confusing them with JavaScript related files. Now that it's a separate library, I should probably remove the prefix. The runtime files are in `runtime/wasm`. Maybe we should move the JavaScript runtime to `runtime/js` to get something more symmetric?

compiler/lib/wasm/wa_closure_conversion.ml
```diff
@@ -0,0 +1,173 @@
...
+                    let arities =
+                      Var.Map.fold
+                        (fun f (params, _) m -> Var.Map.add f (List.length params) m)
+                        l
+                        Var.Map.empty
+                    in
+                    List.map ~f:(fun f -> f, Var.Map.find f arities) fun_lst
+                  in
+                  (*
+                  Format.eprintf "AAA";
```
hhugo, line 146: do you wan't to keep this comment ?

compiler/lib/wasm/sexp.ml
```diff
@@ -0,0 +1,161 @@
...
+    | Atom "false" -> false
+    | _ -> assert false
+
+  let mandatory f v =
+    match v with
+    | Some v -> f v
+    | None -> assert false
+end
+(*
+parse
```
hhugo, line 155: this could be dropped or moved to a test directory

compiler/lib/wasm/sexp.ml
```diff
@@ -0,0 +1,161 @@
...
+let from_string s =
+  let v, pos = parse (Buffer.create 16) s 0 in
+  if pos < String.length s then parse_error ();
+  v
+
+module Util = struct
+  let single f v =
+    match v with
+    | [ v ] -> f v
+    | _ -> assert false
```
hhugo, line 114: should this assert be `invalid_arg` ? same below ..

README.md
```diff
@@ -1,144 +1,69 @@
...
-[opam](https://github.com/ocsigen/js_of_ocaml/blob/master/js_of_ocaml-compiler.opam)
-file for version constraints.
-
-### Toplevel requirements
-
-- tyxml, reactiveData
-- ocp-indent: needed to support indentation in the toplevel
-- higlo: needed to support Syntax highlighting in the toplevel
-- cohttp: needed to build the toplevel webserver
+# Wasm_of_ocaml
```
hhugo, line 1: Should we merge the two readmes ?

biome.json
```diff
@@ -37,7 +37,8 @@
       },
       "suspicious": {
         "noAssignInExpressions": "off",
-        "noRedeclare": "off"
+        "noRedeclare": "off",
+        "noRedundantUseStrict": "off"
```
hhugo, line 41: Why is this ?

compiler/lib/wasm/dune
```diff
@@ -0,0 +1,7 @@
+(library
+ (name wasm_of_ocaml_compiler)
```
hhugo, line 2: how do you feel about moving this lib to compiler/lib-wasm ?

.github/workflows/merge.yml
```diff
@@ -0,0 +1,39 @@
+name: Automatic Merge with Js_of_ocaml
```
hhugo, line 1: I presume this file is not meant for this repo

README.md
vouillon: Yes, certainly.

hhugo:
> The runtime files are in `runtime/wasm`. Maybe we should move the JavaScript runtime to `runtime/js` to get something more symmetric?

Sounds good

compiler/lib/wasm/wa_source_map.ml
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+type t = Yojson.Raw.t
+
+type u =
+  { mappings : string
+  ; mutable pos : int
+  }
+
+module Vlq = struct
```
hhugo, line 16: This should probably be merged with the vlq module from lib

compiler/tests-compiler/gen-rules/gen.ml
```diff
@@ -60,6 +60,13 @@ let test_enabled_if = function
   | "gh1051" -> B64
   | _ -> Any
 
+let enabled_if cond =
+  match cond with
+  | Any -> "(and (<> %{profile} wasm) (<> %{profile} wasm-effects))"
+  | GE5 ->
+      "(and (<> %{profile} wasm) (<> %{profile} wasm-effects) (>= %{ocaml_version} 5))"
+  | B64 -> "(and (<> %{profile} wasm) (<> %{profile} wasm-effects) %{arch_sixtyfour})"
+
```
hhugo, line 69: Can't we do something for the entire directory instead of duplicating `(and (<> %{profile} wasm) (<> %{profile} wasm-effects))`

vouillon: I have simplified a bit the generated rules. But I don't see how to do something for the entire directory at once.

Maybe we could use environment variables to test whether we are generating JS or Wasm code, and for effects rather than matching on the profile's name?

hhugo:
Wasm support was merged in dune (https://github.com/ocaml/dune/pull/11093). This PR or #1726 can be cleaned and undrafted

.github/workflows/build.yml
```diff
@@ -112,13 +112,31 @@ jobs:
...
+          opam pin add dune.3.17 https://github.com/ocaml-wasm/dune.git#wasm_of_ocaml
+
       - run: opam install conf-pkg-config
         if: runner.os == 'Windows'
 
       - run: opam install . --best-effort
         if: ${{ matrix.skip-test }}
 
-      - run: opam install . --with-test
+      - run: cat VERSION | xargs opam pin wasm_of_ocaml-compiler . -n --with-version
```
hhugo, line 125: Do we really need to ping wasm_of_ocaml explicitly ?

vouillon: The issue is that the other packages take their version from the opam repository while `wasm_of_ocaml-compiler`, which is not in the repository, is given version `dev` by default.

vouillon:
I'm not sure how we should proceed now. I need to write some documentation. And we still have the issue that the Jane Street packages have not been updated recently and some still assume that OCaml floats are JavaScript numbers.

hhugo:
> I'm not sure how we should proceed now. I need to write some documentation. And we still have the issue that the Jane Street packages have not been updated recently and some still assume that OCaml floats are JavaScript numbers.

I think we could make a release soon without this PR so that we have something ready for the ocaml 5.3 release. In parallel, we can finish the review for the wasm merge (we need to wait for a dune release anyway). 

There are few points to clarify. 
- who's going to maintain the wasm backend ?
- should wasm_of_ocaml be released at the same time as other jsoo packages.
- is this repo becoming the main development repo ? Or should I expect other synchronisation PRs? 

After merging this PR I'll probably bump the major version number and make another release

compiler/tests-jsoo/dune
```diff
@@ -16,22 +16,32 @@
...
+  (modes js wasm best))
+ (preprocess
+  (pps ppx_expect)))
+
+(library
+ (name jsoo_testsuite_parsing)
+ (modules test_parsing calc_parser calc_lexer)
+ (libraries unix compiler-libs.common js_of_ocaml-compiler)
+ (enabled_if
+  (>= %{ocaml_version} 5.1.1))
```
hhugo, line 28: Why is this conditioned on ocaml >= 5.1.1 ?

compiler/tests-jsoo/lib-effects/dune
```diff
@@ -1,12 +1,22 @@
+(env
+ (using-effects)
+ (wasm)
+ (wasm-effects)
+ (_
+  (js_of_ocaml
+   (flags
+    (:standard --enable effects))
+   ;; separate compilation doesn't work when using
```
hhugo, line 9: I would expect this comment to no longer be true

compiler/tests-num/dune
```diff
@@ -1,31 +1,18 @@
...
- (target main.referencejs)
- (deps main.bc.js)
- (action
-  (with-stdout-to
-   %{target}
-   (run node ./main.bc.js))))
-
-(rule
- (target main.reference)
+ (target main.expected)
```
hhugo, line 13: Do we need to keep this rule ?

compiler/tests-jsoo/dune
vouillon: Because this PR landed on OCaml 5.1.1: https://github.com/ocaml/ocaml/pull/12046

compiler/tests-num/dune
vouillon: This produces the file `main.expected`, which is then used as a reference when testing the generated JavaScript code. But one could commit this file instead.

compiler/tests-jsoo/dune
hhugo: Ok, so this is do that we can drop the [flush_stubs.c](https://github.com/ocsigen/js_of_ocaml/pull/1724/files/c5d3694429833fd137748ee8230f7efd18962fc7#diff-1d121ef25538fe6d01a2ad0a42e4b26cf7f92b3af30a594882ebeebf679c0096)

compiler/tests-jsoo/lib-effects/dune
hhugo: I've removed it on master

compiler/tests-num/dune
hhugo: Let's keep it then

compiler/lib-wasm/wasm_source_map.ml
```diff
@@ -0,0 +1,174 @@
...
+
+let concatenate l =
+  Source_map.Index
+    { version = 3
+    ; file = None
+    ; sections =
+        List.map
+          ~f:(fun (ofs, sm) ->
+            match sm with
+            | Source_map.Index _ -> assert false
```
hhugo, line 107: Can't we implement this ?

compiler/lib-wasm/wasm_source_map.ml
```diff
@@ -0,0 +1,174 @@
...
+      0
+      { Vlq64.string = mappings; pos = 0; len = String.length mappings }
+      resize_data
+      1
+      col0
+      delta0
+      0;
+    Buffer.contents buf
+
+let resize resize_data sm =
```
hhugo, line 85: What's this logic about ? Can we use Index sourcemap instead ?

hhugo:
There are files without headers and license. I'm sure there are some missings in the repo already. Maybe it's a good time to instrument adding/checking this.

compiler/bin-wasm_of_ocaml/build_runtime.ml
```diff
@@ -0,0 +1,32 @@
...
+ *)
+
+open! Js_of_ocaml_compiler.Stdlib
+
+let info =
+  Info.make
+    ~name:"build-runtime"
+    ~doc:"Build standalone runtime. Used for separate compilation."
+    ~description:
+      "Js_of_ocaml is a compiler from OCaml bytecode to Javascript. It makes it possible \
```
hhugo, line 26: Update this

compiler/bin-wasm_of_ocaml/info.ml
```diff
@@ -0,0 +1,48 @@
...
+    ; `S "BUGS"
+    ; `P
+        "Bugs are tracked on github at \
+         $(i,https://github.com/ocsigen/js_of_ocaml/issues)."
+    ; `S "SEE ALSO"
+    ; `P "ocaml(1)"
+    ; `S "AUTHORS"
+    ; `P "Jerome Vouillon, Hugo Heuzard."
+    ; `S "LICENSE"
+    ; `P "Copyright (C) 2010-2020."
```
hhugo, line 35: bump date

wasm_of_ocaml-compiler.opam
```diff
@@ -0,0 +1,50 @@
...
+authors: ["Ocsigen team <dev@ocsigen.org>"]
+license: [
+  "GPL-2.0-or-later" "LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception"
+]
+homepage: "https://ocsigen.org/js_of_ocaml/latest/manual/overview"
+doc: "https://ocsigen.org/js_of_ocaml/latest/manual/overview"
+bug-reports: "https://github.com/ocsigen/js_of_ocaml/issues"
+depends: [
+  "dune" {>= "3.17"}
+  "ocaml" {>= "4.14" & < "5.3"}
```
hhugo, line 16: Why is it not compatible with older version ?

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -0,0 +1,228 @@
...
+        uses: actions/cache/save@v4
+        with:
+          path: binaryen
+          key: ${{ runner.os }}-binaryen-version_118
+
+      - name: Set binaryen's path
+        run: |
+          echo "$GITHUB_WORKSPACE/binaryen/bin" >> $GITHUB_PATH
+
+      - name: Restore cached OCaml
```
hhugo, line 83: Is this still needed with setup-ocaml v3 ?

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -0,0 +1,228 @@
...
+        # for graphics
+        run: brew install xquartz
+
+      - name: Set opam path
+        if: steps.cache-ocaml.outputs.cache-hit
+        run: |
+          echo /opt/hostedtoolcache/opam/*/x86_64 >> $GITHUB_PATH
+          echo /Users/runner/hostedtoolcache/opam/*/arm64 >> $GITHUB_PATH
+
+      - name: Cache OCaml
```
hhugo, line 118: same question here

tools/ci_setup.ml
```diff
@@ -0,0 +1,296 @@
...
++          1073741823 -> ( 6) \001\253\255\255\255\063
++         -1073741824 -> ( 6) \001\253\000\000\000\192
++          1073741824 -> (12) \011\0101073741824
++         -1073741825 -> (13) \011\011\0451073741825 |}]
++    ;;
++
+     let%expect_test "bin_io de-serialization V2" =
+       (* Some bignums will have two bin_io representation depending on where their
+          were serialized.  Make sure we're able to parse things back regardless of the
+|bignum}
```
hhugo, line 146: https://github.com/orgs/community/discussions/42534

hhugo: https://discuss.ocaml.org/t/syntax-highlighting-on-github/11656/5

tools/ci_setup.ml
```diff
@@ -0,0 +1,296 @@
...
++  then raise_s [%message "mismatch" (x : float) (x' : float) (x |> of_float : Bignum.t)]
+ ;;
+ 
+ let%expect_test "roundtrip: f |> Bignum.of_float_decimal |> Bignum.to_float" =
+@@ -783,7 +788,7 @@ let%test_module _ =
+          -1073741825 -> ( 6) \001\253\255\255\255\191 |}]
+     ;;
+ 
+-    let%expect_test ("bin_io serialization V2 (javascript)" [@tags "js-only"]) =
++    let%expect_test ("bin_io serialization V2 (javascript)" [@tags "js-only", "no-wasm"]) =
```
hhugo, line 107: Why is the expect test different in js and wasm ?

tools/node_wrapper.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+export PATH=$(echo $PATH | cut -d : -f 2-) # Do not call oneself recursively
```
hhugo, line 2: This is a bit fragile. There could be other directories added by dune. I'll commit an alternative.

tools/ci_setup.ml
vouillon: The integer size differs.

wasm_of_ocaml-compiler.opam
vouillon: It is difficult to test older versions since we uses a version of packages like `ppx_inline_test`, `ppx_expect`, `base` which depend on `ocaml >= 4.14`. I'm not really willing to modify older versions of these packages to add Wasm runtime files...

compiler/lib-wasm/wasm_source_map.ml
hhugo: https://github.com/ocaml-wasm/wasm_of_ocaml/pull/129

compiler/lib-wasm/wasm_source_map.ml
vouillon: It's a bit as if we were reformatting some JavaScript code. We could implement the corresponding source map update by splitting the input source map into an index containing a large number of output source maps, for each unchanged piece of code. But it makes a lot more sense to keep a single standard source map and adjust the position of the mappings.

.github/workflows/build-wasm_of_ocaml.yml
vouillon: Removing this cache adds about 30s to the github workflow. I guess it's ok to remove it since it adds some complexity.

tools/ci_setup.ml
vouillon: I have removed this piece of comment.

compiler/bin-wasm_of_ocaml/cmd_arg.ml
```diff
@@ -0,0 +1,247 @@
...
+    let doc = "Link JavaScript and WebAssembly files [$(docv)]. " in
+    Arg.(value & pos_all string [] & info [] ~docv:"RUNTIME_FILES" ~doc)
+  in
+  let output_file =
+    let doc = "Set output file name to [$(docv)]." in
+    Arg.(required & opt (some string) None & info [ "o" ] ~docv:"FILE" ~doc)
+  in
+  let no_sourcemap =
+    let doc =
+      "Don't generate source map. All other source map related flags will be be ignored."
```
hhugo, line 179:
```suggestion
      "Don't generate source map. All other source map related flags will be ignored."
```

compiler/lib-wasm/wasm_source_map.ml
hhugo: > For this, one needs to renumber the module components

This is the part I don't really understand, can you point me to where the renumbering of components happens ?

vouillon: We are building a single Wasm module by linking together several Wasm modules.
There is a function `scanner` in `wasm_link.ml` which scan the Wasm binary code and renumber the components.
https://github.com/ocsigen/js_of_ocaml/blob/dcd52fc075934d9f7b800b60df09f2b1d58c8bbd/compiler/lib-wasm/wasm_link.ml#L894
The renumbering is performed by functions `rewrite` and `rewrite_signed`, which read the index and write the updated index, keeping track of code size changes.
https://github.com/ocsigen/js_of_ocaml/blob/dcd52fc075934d9f7b800b60df09f2b1d58c8bbd/compiler/lib-wasm/wasm_link.ml#L950

vouillon:
I think it's going to be much simpler to use this repo as the main development repo and to release wasm_of_ocaml at the same time as other jsoo packages, given how tightly it depends on the js_of_ocaml compiler library.

hhugo:
> Ideally, big packages that haven’t yet done a release with explicit float conversions should do so first, before wasm_of_ocaml is released. This is because after the release, the type `Js.number_t` will no longer be a synonym of `float` and so these packages would break. A number of Jane Street packages, in particular, are impacted.

I don't really understand why we would need to wait here ? Isn't it better to have wasmoo out first so that other packages can add support for their next releases. It would be better to get the correct constraint in the opam-repository anyway.

I want to release the current-ish state of master soon (with limited breakage). And then a major release with wasmoo. People not able to upgrade to the latest version of jsoo won't miss much features/changes in that case.

OlivierNicole:
I suppose this works, if all these packages have upper bounds on their version of js_of_ocaml.

hhugo:
One thing that would help A LOT would be identify all packages broken by this PR and add an upper bound on the js_of_ocaml version (`< 6.0`) in the opam-repo.
That would make releasing the next version much much easier.

hhugo:
Can you rebase this PR ?

Do you plan to migrate issues and PRs to this repo ?

vouillon:
> Do you plan to migrate issues and PRs to this repo ?

I'm not sure what to do regarding the current issues. I don't think we can migrate issues across organizations.
There are not many opened PR. I think we can just create new PRs in this repo.

vouillon:
> One thing that would help A LOT would be identify all packages broken by this PR and add an upper bound on the js_of_ocaml version (`< 6.0`) in the opam-repo. That would make releasing the next version much much easier.

See https://github.com/ocaml/opam-repository/pull/26980

hhugo:
#1631 has been merged. Can you rebase/update this PR one more (final?) time

compiler/tests-ocaml/lib-effects/evenodd.ml
```diff
@@ -18,5 +18,5 @@ and odd n =
   else even (n-1)
 
 let _ =
-  let n = 100_000 in
+  let n = 10_000 in
```
hhugo, line 21: Is this change necessary ?

hhugo: or is it just to speedup tests ?

vouillon: This is just to speed up this test.

dune-project
```diff
@@ -132,3 +132,29 @@
...
   (re (and (>= 1.9.0) :with-test))
 ))
+
+(package
+ (name wasm_of_ocaml-compiler)
+ (synopsis "Compiler from OCaml bytecode to WebAssembly")
+ (description
+  "Wasm_of_ocaml is a compiler from OCaml bytecode to WebAssembly. It makes it possible to run pure OCaml programs in JavaScript environment like browsers and Node.js")
+ (depends
+  (ocaml (and (>= 4.14) (< 5.3)))
```
hhugo, line 142: jsoo is compatible with 5.3. What's missing for wasmoo ?

.github/workflows/build.yml
```diff
@@ -118,7 +118,21 @@ jobs:
...
         if: ${{ matrix.skip-test }}
 
-      - run: opam install . --with-test
+      - run: cat VERSION | xargs opam pin . -n --with-version 
+        if: ${{ !matrix.skip-test }}
+        shell: bash
+
+      - run: opam install conf-c++
+        # Otherwise, the next step fails reinstalling gcc while compiling
+        # other packages
```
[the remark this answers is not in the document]
vouillon: I think the issue is that `dune build @all` will invoke `wasm-merge` and `wasm-opt` to build `runtime/wasm/runtime.wasm`.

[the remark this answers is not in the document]
vouillon: We could cache binaryen, as is done in build-wasm_of_ocaml.yml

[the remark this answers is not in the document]
hhugo: Let's try that

hhugo:
@vouillon, I'm testing https://github.com/ocsigen/js_of_ocaml/pull/1757/commits/9b964d91ac54301d6957bf91eb9e81ed28c22b0e in #1757

hhugo:
> @vouillon, I'm testing [9b964d9](https://github.com/ocsigen/js_of_ocaml/commit/9b964d91ac54301d6957bf91eb9e81ed28c22b0e) in #1757

I think you should cherry-pick https://github.com/ocsigen/js_of_ocaml/pull/1757/commits/cf96d62dd332b827823c9d538b39ce277b20f8df

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -0,0 +1,184 @@
+name: Build wasm_of_ocaml
+
+on:
+  pull_request:
+  push:
```
hhugo, line 5:
```suggestion
  pull_request:
  push:
    branches:
      - master
```

runtime/wasm/runtime.js
```diff
@@ -0,0 +1,564 @@
...
+    },
+    weak_map_new: () => new WeakMap(),
+    map_new: () => new Map(),
+    map_get: (m, x) => {
+      var v = m.get(x);
+      return v === undefined ? null : v;
+    },
+    map_set: (m, x, v) => m.set(x, v),
+    map_delete: (m, x) => m.delete(x),
+    log: (x) => console.log("ZZZZZ", x),
```
hhugo, line 460: Update this

tools/node_wrapper.sh
hhugo: Let's keep it this way for now.

hhugo:
Thanks for all the hard work

dbuenzli:
Bravo! Is there some high level docs on how to use the stuff ?

vouillon:
> Bravo! Is there some high level docs on how to use the stuff ?

There is a short [`README_wasm_of_ocaml.md`](https://github.com/ocsigen/js_of_ocaml/blob/master/README_wasm_of_ocaml.md). But I still need to write some documentation...

rickyvetter:
Incredible! This is awesome.

dbuenzli:
> There is a short [`README_wasm_of_ocaml.md`](https://github.com/ocsigen/js_of_ocaml/blob/master/README_wasm_of_ocaml.md).

This looks to good to be true :–) The browsers APIs can be used as is ? 

However it is possible to make it work over the `file://` protocol ? And what's the story with webworkers ? 

I have a number crunching app (computing thousands of polygon intersections) that I distribute as a single double-clickable `.html` file. It would be interesting to see the gains there. However it uses [extremely dirty tricks](https://gist.github.com/dbuenzli/0cbad35b9d0a828b17d15aa026e230f2#file-work_queue-ml-L31-L44)   to make webworkers work out of a single js_of_ocaml compiled file with code paths differentiated in the main via [`Worker.ami`](https://erratique.ch/software/brr/doc/Brr_webworkers/Worker/index.html#val-ami).

vouillon:
> The browsers APIs can be used as is ?

Right.

> However it is possible to make it work over the `file://` protocol ?

Not at the moment. [Source Phase Imports](https://github.com/tc39/proposal-source-phase-imports) might make it possible at some point. Or maybe we could have an option to somehow embed the Wasm code in the JavaScript code.

> And what's the story with webworkers ?

I'm not sure. What problem could there be?

> I have a number crunching app (computing thousands of polygon intersections) that I distribute as a single double-clickable `.html` file. It would be interesting to see the gains there. However it uses [extremely dirty tricks](https://gist.github.com/dbuenzli/0cbad35b9d0a828b17d15aa026e230f2#file-work_queue-ml-L31-L44) to make webworkers work out of a single js_of_ocaml compiled file with code paths differentiated in the main via [`Worker.ami`](https://erratique.ch/software/brr/doc/Brr_webworkers/Worker/index.html#val-ami).

This might just work if we embed the Wasm code.

dbuenzli:
> I'm not sure. What problem could there be?

I'm concerned about data transfers. With webworkers you are supposed to [transfer](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) JavaScript values that that go through the [structured cloned algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). 

As [we once discussed](https://discuss.ocaml.org/t/js-of-ocaml-ocamls-values-and-the-structured-clone-algorithm/10615) that worked relatively flawlessly with js_of_ocaml's encoding of OCaml values encoded as JavaScript values. 

Now I don't know how OCaml values are represented in `wasm_of_ocaml`, but in the [webworker work queue](https://gist.github.com/dbuenzli/0cbad35b9d0a828b17d15aa026e230f2) I linked to I'm basically [using `Obj.magic`](https://gist.github.com/dbuenzli/0cbad35b9d0a828b17d15aa026e230f2#file-work_queue-ml-L56) on OCaml values to transfer work items but if those are no longer represented as JavaScript values then I expect kaboom.

vouillon:
OCaml values are represented using Wasm GC values, so indeed the structured clone algorithm will not work on them. You have to serialize them somehow. You can try `Marshal` which works much better than with Js_of_ocaml (integers and floats do not share the same representation).

I have [some experimental code to embed the Wasm code within the JavaScript code](https://github.com/ocsigen/js_of_ocaml/pull/1758). Maybe you could give it a try.

micahcantor:
Congrats @vouillon @OlivierNicole @hhugo 🎉 !

PR #1726, opened 2024-11-04
Wasm_of_ocaml with alternative dune support

vouillon:
See https://github.com/ocaml/dune/pull/11093#issuecomment-2454803413


lib/tests/gen-rules/gen.ml
```diff
@@ -79,6 +81,12 @@ let () =
            (Hashtbl.hash prefix mod 100)
            (match enabled_if basename with
            | Any -> "true"
-           | GE5 -> "(>= %{ocaml_version} 5)"
-           | No_effects -> "(<> %{profile} using-effects)")
+           | GE5 ->
+               (* ZZZ /static not yet implemented *)
+               "(and (>= %{ocaml_version} 5) (<> %{profile} wasm) (<> %{profile} \
+                wasm-effects))"
+           | Not_wasm -> "(and (<> %{profile} wasm) (<> %{profile} wasm-effects))"
```
hhugo, line 88: Could we change `modes` instead of encoding this by listing wasm profiles   ?

PR #1729, opened 2024-11-06
Fix compare function

vouillon:


hhugo:
Thanks. A regression test would be nice if you can

PR #1734, opened 2024-11-15
[draft] Implement the `popover` API.

SylvainBoilard:
Hello,

This PR adds support for the `popover` API.

See:
- https://html.spec.whatwg.org/multipage/popover.html ;
- https://developer.mozilla.org/en-US/docs/Web/API/Popover_API .

Please let me know of any issue; in fact, I would appreciate your input regarding a couple things (see below).

Cheers,

### DIFF
--- a/lib/js_of_ocaml/dom_html.ml
+++ b/lib/js_of_ocaml/dom_html.ml
@@ -462,6 +462,8 @@ and toggleEvent = object
   method newState : js_string t readonly_prop
 
   method oldState : js_string t readonly_prop
+
+  method source : element t opt readonly_prop
 end
 
 and mediaQueryListEvent = object
@@ -568,6 +570,10 @@ and eventTarget = object ('self)
   method onpointerover : ('self t, pointerEvent t) event_listener writeonly_prop
 
   method onpointerup : ('self t, pointerEvent t) event_listener writeonly_prop
+
+  method onbeforetoggle : ('self t, toggleEvent t) event_listener writeonly_prop
+
+  method ontoggle : ('self t, toggleEvent t) event_listener writeonly_prop
 
   method dispatchEvent : event t -> bool t meth
 end
@@ -692,6 +698,16 @@ and tokenList = object
   method stringifier : js_string t prop
 end
 
+and showPopover_options = object
+  method source : element t writeonly_prop
+end
+
+and togglePopover_options = object
+  method force : bool t writeonly_prop
+
+  method source : element t writeonly_prop
+end
+
 and element = object
   inherit Dom.element
 
@@ -747,6 +763,8 @@ and element = object
 
   method scrollHeight : int prop
 
+  method popover : js_string t opt prop
+
   method getClientRects : clientRectList t meth
 
   method getBoundingClientRect : clientRect t meth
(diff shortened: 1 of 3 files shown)


lib/js_of_ocaml/dom_html.ml
```diff
@@ -747,6 +753,18 @@ and element = object
...
+  method showPopover : unit meth
+
+  method showPopover_options : _ -> unit meth
+
+  method togglePopover : bool t meth
+
+  method togglePopover_force : bool t -> bool t meth
+
+  method togglePopover_options : _ -> bool t meth
+
```
SylvainBoilard, line 767: [According to the specs](https://html.spec.whatwg.org/multipage/dom.html#showpopoveroptions), these methods accept a dictionary as a parameter for passing options. How would that be reflected in Js_of_ocaml?

lib/js_of_ocaml/dom_html.ml
```diff
@@ -557,6 +557,10 @@ and eventTarget = object ('self)
 
   method onpointerup : ('self t, pointerEvent t) event_listener writeonly_prop
 
+  method onbeforetoggle : ('self t, toggleEvent t) event_listener writeonly_prop
+
+  method ontoggle : ('self t, toggleEvent t) event_listener writeonly_prop
+
```
SylvainBoilard, line 563: There are similar methods related to event handlers in this file that have the `prop` type instead of `writeonly_prop`. In fact, this PR initially used the latter type. Is this new version correct?

hhugo: writeonly_prop is often used for situation where the property might be undefined.

lib/js_of_ocaml/dom_html.ml
hhugo: I think there is an example in lib/js_of_ocaml/eventSource.ml

hhugo:
I apologies for the extra long delay reviewing your PR.

lib/js_of_ocaml/dom_html.ml
```diff
@@ -1284,6 +1308,10 @@ class type buttonElement = object
   method _type : js_string t readonly_prop
 
   method value : js_string t prop
+
+  method popovertarget : element t opt prop
+
+  method popovertargetaction : js_string t prop
```
hhugo: Shouldn't this be `popoverTargetAction` ?

lib/js_of_ocaml/dom_html.ml
```diff
@@ -1284,6 +1308,10 @@ class type buttonElement = object
   method _type : js_string t readonly_prop
 
   method value : js_string t prop
+
+  method popovertarget : element t opt prop
```
hhugo: Shouldn't this be `popoverTargetElement` ?

hhugo (changes requested):
First batch of review.
Maybe we should have a small demo in examples/  so that we can tests such feature.

lib/js_of_ocaml/dom_html.ml
SylvainBoilard: Alright I think `writeonly_prop` is the right one then.

lib/js_of_ocaml/dom_html.ml
SylvainBoilard: I had a look at a few places in the code for inspiration and came up with something, please let me know what you think.

hhugo:
replaced by #2252, thanks for the contribution and sorry for delays

SylvainBoilard:
Thank you for merging this, and sorry for not getting around to making the demo you had requested.

Cheers!

PR #1740, opened 2024-11-21
Runtime: no longer use arguments

hhugo:
subset of #1682

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -26,7 +26,6 @@
         "noSubstr": "error"
       },
       "style": {
-        "noArguments": "off",
         "noCommaOperator": "off",
         "noParameterAssign": "off",
         "noUselessElse": "off",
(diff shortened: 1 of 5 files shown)


hhugo:
@TyOverby, could you test this ?

hhugo:
@TyOverby, if we end up merging this, we should probably also merge stdlib_modern.js and stdlib.js and get rid of stdlib_modern.js

hhugo:
@TyOverby, Wasm support has been merged. I hoping that it means you can now easily test this PR ?

smorimoto (approved):

smorimoto:
This should help simplify and increase runtime performance (a bit)

hhugo:
@vouillon, any opinion here ?

vouillon (approved):
I'm not sure about the performance. This looks good otherwise.

PR #1743, opened 2024-11-25
Lib: small refactoring around js exception printing

hhugo:


TyOverby:
Does this change any semantics?

hhugo:
Previously, We would call toString to print any value that doesn't look like an OCaml exception.
With this PR, the printer only looks for js `Error`s.

So yes, there is a semantic difference. It was extracted from the wasm backend branch. Maybe @vouillon can tell if/why it is necessary for the wasm backend.

vouillon:
If the value was an JavaScript array, we would consider it as an OCaml exception. Otherwise, we would use `toString` to turn get an error message, assuming that it was a JavaScript exception. Since does not work with wasm_of_ocaml, since OCaml exceptions are Wasm arrays, which are not JavaScript array and does not have a `toString` method. With this change, we instead test whether the value is a JavaScript exception.

I'm not sure we still need this code, actually. We now call `caml_wrap_exception` systematically to turn JavaScript exceptions into OCaml exceptions, so this code should never see any JavaScript exception.

PR #1745, opened 2024-11-25
Hints phys equal

### LINKED ISSUE #1372: Use `Object.is` instead of `===` to implement phys_equal

hhugo:
Fix https://github.com/ocsigen/js_of_ocaml/issues/1372


TyOverby:
There's a lot of unexpected changes in this PR, I think you may need to rebase it?

smorimoto:
Now, when you merge the hints branch, this PR will automatically be directed to master branch.

PR #1749, opened 2024-11-28
biome: enable all performance rules

smorimoto:


smorimoto:
Found some regular expression warnings: https://github.com/ocsigen/js_of_ocaml/actions/runs/12069505071/job/33656910919?pr=1749#step:4:1

hhugo:
Why is the CI not failing, can we make it fail ?

smorimoto:
The `all` option will be removed in the future...
- https://github.com/biomejs/biome/pull/4664

smorimoto:
> Why is the CI not failing, can we make it fail ?

I think it's because the rule is treated as a warning by default.

Of course we can make it an error.

hhugo:
Did you close the PR on purpose ? Can you explain why ?

PR #1752, opened 2024-12-02
Compiler: introduce mechanism to deprecated runtime primitives, resto…

hhugo:
…re caml_new_string


hhugo:
https://github.com/ocaml/dune/pull/10777

PR #1753, opened 2024-12-02
Remove *.opam.template

smorimoto:


hhugo:
The CI doesn't like this change. 
Could you try to motivate changes in the description whenever you open PRs ?

.github/workflows/build.yml
```diff
@@ -23,62 +23,29 @@ jobs:
...
+            ocaml-compiler: "ocaml-variants.4.14.2+options,ocaml-option-32bit"
+          - os: ubuntu-latest
+            ocaml-compiler: "ocaml-variants.5.2.1+options,ocaml-option-32bit" # Note this OCaml compiler is bytecode only
 
     runs-on: ${{ matrix.os }}
 
     steps:
       - name: Set git to use LF
-        if: matrix.ocaml-compiler < 5.2
+        if: matrix.os == 'windows-latest'
```
hhugo, line 48: I don't understand this change. `ocaml-compiler < 5.2` was on purpose.

.github/workflows/build.yml
```diff
@@ -109,41 +76,33 @@ jobs:
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          dune-cache: true
-          opam-pin: false
-
-      - run: opam install conf-pkg-config
```
hhugo, line 115: We need this, this is a temporary fix for https://github.com/dra27/mingw-w64-shims/issues/1

.github/workflows/build.yml
```diff
@@ -109,41 +76,33 @@ jobs:
...
+        if: matrix.ocaml-compiler >= 5.0
 
       - run: opam exec -- git diff --exit-code
-        if: ${{ !matrix.skip-test }}
+        if: matrix.ocaml-compiler >= 4.14
 
-      - name: build doc
-        if: ${{ !matrix.skip-doc && github.event_name == 'push' && github.ref_name == 'master'}}
+      - name: Build doc
+        if: matrix.os == 'ubuntu-latest' && matrix.ocaml-compiler == '5.2' && github.event_name == 'push' && github.ref_name == 'master'
```
hhugo, line 96: I would prefer to only bump version once when it come to change the version we generate the doc with.

.github/workflows/build.yml
```diff
@@ -23,62 +23,29 @@ jobs:
           - "4.11"
           - "4.12"
           - "4.13"
+          - "4.14"
           - "5.0"
           - "5.1"
-        skip-test:
```
hhugo, line 28: I find the previous version much easier to understand.

hhugo: This PR kind of inline the logic previously encoded by skip-* parameters. We know have to go over all step understand what's done for what version. I think this is going in the wrong direction.

hhugo:
Can you please explain why you want to remove opam.template and try to keep the diff minimal in the CI

smorimoto:
I was going to remove the excess as much as possible, but as you said, it was not as elegant as I expected.

PR #1758, opened 2024-12-05
Embed the Wasm code inside the JavaScript script

vouillon:
- [ ] Enable through a flag
- [ ] Make it work with separate compilation


hhugo:
@vouillon, what's you ETA for his PR ?

PR #1764, opened 2024-12-10
Modernize stdlib

hhugo:
and remove `modern_stdlib.js`


smorimoto (approved):

hhugo:
@TyOverby, any opinion ?

TyOverby:
I think we may have moved away from stdlib_modern for performance reasons; I can check when I get back to work in January

TyOverby:
Looks like we're using `modern` anyway

TyOverby (approved):

PR #1772, opened 2024-12-17
[wasm] Option to implement OCaml strings with JavaScript strings

vouillon:
- [x] Conditional compilation of Wasm runtime
- [x] Import string constants (define JS strings in Wasm runtime)
- [x] default should  be not to use js strings
- [x] Compatibilty for when JS string builtins proposal is not available
- [ ] Performance improvements
  - [x] skip conversions for ASCII strings
  - [x] lower level version of caml_blit_string
  - [ ] Use I16 array for conversions 
  - [x] Remove unnecessary conversion
- [x] clean-up / check for missing tests
- [ ] adapt Jane street libraries
- [ ] performance measurements

### DIFF
--- a/benchmarks/benchmark-partial-render-table/Makefile
+++ b/benchmarks/benchmark-partial-render-table/Makefile
@@ -3,7 +3,6 @@
 export NAME=Partial Render Table
 
 SHELL=/usr/bin/env bash -o pipefail
-TIME=$(shell whereis time | cut -d ' ' -f 2)
 
 bench:
 	@date -u +"%FT%TZ - $(NAME): starting"
@@ -15,7 +14,7 @@ bench:
 	@date -u +"%FT%TZ - $(NAME): done"
 
 perform:
-	$(TIME) -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty --enable use-js-string main.bc-for-jsoo -o out.js 2>&1 | \
+	env time -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty --enable use-js-string main.bc-for-jsoo -o out.js 2>&1 | \
 	tee /dev/stderr | \
 	ocaml -I +str str.cma ../utils/compilation_metrics.ml $(COMPILER) "$(NAME)" out.js | \
 	sh ../utils/aggregate.sh $(KIND)
--- a/benchmarks/benchmark-partial-render-table/Makefile
+++ b/benchmarks/benchmark-partial-render-table/Makefile
@@ -2,7 +2,8 @@
 
 export NAME=Partial Render Table
 
-SHELL=/bin/bash -o pipefail
+SHELL=/usr/bin/env bash -o pipefail
+TIME=$(shell whereis time | cut -d ' ' -f 2)
 
 bench:
 	@date -u +"%FT%TZ - $(NAME): starting"
@@ -14,7 +15,7 @@ bench:
 	@date -u +"%FT%TZ - $(NAME): done"
 
 perform:
-	/usr/bin/time -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty --enable use-js-string main.bc-for-jsoo -o out.js 2>&1 | \
+	$(TIME) -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty --enable use-js-string main.bc-for-jsoo -o out.js 2>&1 | \
 	tee /dev/stderr | \
 	ocaml -I +str str.cma ../utils/compilation_metrics.ml $(COMPILER) "$(NAME)" out.js | \
 	sh ../utils/aggregate.sh $(KIND)
--- a/benchmarks/benchmark-partial-render-table/Makefile
+++ b/benchmarks/benchmark-partial-render-table/Makefile
@@ -14,7 +14,7 @@ bench:
 	@date -u +"%FT%TZ - $(NAME): done"
 
 perform:
-	/usr/bin/time -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty main.bc-for-jsoo -o out.js 2>&1 | \
+	/usr/bin/time -f "%E %R" $(COMPILER) --debug times --opt 2 --pretty --enable use-js-string main.bc-for-jsoo -o out.js 2>&1 | \
 	tee /dev/stderr | \
 	ocaml -I +str str.cma ../utils/compilation_metrics.ml $(COMPILER) "$(NAME)" out.js | \
 	sh ../utils/aggregate.sh $(KIND)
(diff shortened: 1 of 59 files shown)


compiler/lib-wasm/gc_target.ml
```diff
@@ -788,13 +805,48 @@ module Memory = struct
...
+          ~import_module:"wasm:js-string"
+          ~name:"charCodeAt"
+          (Fun
+             { W.params = [ Ref { nullable = true; typ = Extern }; I32 ]
+             ; result = [ I32 ]
+             })
+      in
+      let* e = string_value e in
+      let* e' = Value.int_val e' in
+      Value.val_int (return (W.Call (f, [ e; e' ])))
```
vouillon:
```suggestion
      let* e' = e' in
      return (W.Call (f, [ e; e' ]))
```

compiler/lib-wasm/generate.ml
```diff
@@ -300,29 +300,30 @@ module Generate (Target : Target_sig.S) = struct
...
-      ~tz:(Int Unnormalized)
-      (fun x y z -> seq (Memory.bytes_set x y z) Value.unit);
     register_tern_prim
       "caml_bytes_unsafe_set"
       ~ty:(Int Normalized)
       ~tz:(Int Unnormalized)
       (fun x y z -> seq (Memory.bytes_set x y z) Value.unit);
+    let string_get context x y =
+      seq
+        (let* cond = Arith.uge (Value.int_val y) (Memory.string_length x) in
```
vouillon:
```suggestion
        (let* cond = Arith.uge y (Memory.string_length x) in
```

compiler/lib-wasm/runtime.ml
```diff
@@ -24,7 +24,9 @@ let build ~allowed_imports ~link_options ~opt_options ~variables ~inputs ~output
       let missing_imports =
         List.filter
           ~f:(fun { Link.Wasm_binary.module_; _ } ->
-            not (List.mem ~eq:String.equal module_ allowed_imports))
+            not
+              (String.equal module_ ""
+              || List.mem ~eq:String.equal module_ allowed_imports))
```
vouillon:
```suggestion
            not (List.mem ~eq:String.equal module_ allowed_imports))
```

compiler/lib-wasm/wat_preprocess.ml
```diff
@@ -508,35 +516,116 @@ and rewrite st elt =
       then raise (Error (position_of_loc loc_value, "Expecting a string"));
       let s = parse_string loc_value value in
       write st pos;
+      if variable_is_set st "use-js-string"
+      then (
+        Printf.bprintf
+          st.head_buf
+          "(import \"\" %s (global %s$string externref)) "
```
vouillon: We are now using module name `str` for JavaScript strings, rather than an empty module name.
```suggestion
          "(import \"str\" %s (global %s$string externref)) "
```

compiler/lib-wasm/wat_preprocess.ml
```diff
@@ -508,35 +516,116 @@ and rewrite st elt =
...
+      if not (is_string value)
+      then raise (Error (position_of_loc loc_value, "Expecting a string"));
+      let s = parse_string loc_value value in
+      let name = generate_id st s in
+      write st pos;
+      if variable_is_set st "use-js-string"
+      then (
+        Printf.bprintf
+          st.head_buf
+          "(import \"\" %s (global %s$string externref)) "
```
vouillon:
```suggestion
          "(import \"str\" %s (global %s$string externref)) "
```

compiler/lib-wasm/runtime.ml
OlivierNicole: Shouldn’t it be:

```suggestion
            not
              (String.equal module_ "str"
              || List.mem ~eq:String.equal module_ allowed_imports))
```

But even then I’m getting errors…

compiler/lib-wasm/wat_preprocess.ml
```diff
@@ -508,35 +516,116 @@ and rewrite st elt =
...
+          [ { desc = Atom "@jsstring"; _ }
+          ; { desc = Atom name; _ }
+          ; { desc = Atom value; _ }
+          ]
+    ; loc = pos, pos'
+    } ->
+      write st pos;
+      Printf.bprintf
+        st.head_buf
+        "(import \"\" %s (global %s$string externref)) "
```
vouillon:
```suggestion
        "(import \"str\" %s (global %s$string externref)) "
```

compiler/lib-wasm/wat_preprocess.ml
```diff
@@ -508,35 +516,116 @@ and rewrite st elt =
...
     ; loc = pos, pos'
     } ->
       if not (is_string value)
       then raise (Error (position_of_loc loc_value, "Expecting a string"));
       let s = parse_string loc_value value in
+      let name = generate_id st s in
       write st pos;
+      Printf.bprintf
+        st.head_buf
+        "(import \"\" %s (global %s$string externref)) "
```
vouillon:
```suggestion
        "(import \"str\" %s (global %s$string externref)) "
```

compiler/lib-wasm/runtime.ml
vouillon: You should update the list of allowed imports in `compile.ml` instead.

benchmarks/benchmark-partial-render-table/Makefile
```diff
@@ -2,7 +2,8 @@
 
 export NAME=Partial Render Table
 
-SHELL=/bin/bash -o pipefail
+SHELL=/usr/bin/env bash -o pipefail
+TIME=$(shell whereis time | cut -d ' ' -f 2)
```
vouillon:
```suggestion
TIME=$(shell which time)
```
Or you can use `env time` or simply `\time` instead of `/usr/bin/time` below.

OlivierNicole: Thanks!

OlivierNicole:
The benchmark results are… surprising

runtime/js/mlBytes.js
```diff
@@ -79,6 +79,115 @@ function caml_sub_uint8_array_to_jsbytes(a, i, len) {
   return s;
 }
 
+//Provides: caml_utf8_of_utf16
+function caml_utf8_of_utf16(s) {
```
hhugo, line 83: Why do you need to bring this back ?

OlivierNicole: It’s temporary to have the benchmark CI run again, I don’t intend to keep it.

OlivierNicole:
Rebased on master.

hhugo:
@vouillon, What should we do about this PR ?

PR #1781, opened 2024-12-30
Remove unused files

smorimoto:
Check the CI does not complain...

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,5 @@
+# Default behaviour, for if core.autocrlf isn't set
+* text=auto
+
 *.ml linguist-language=OCaml
 *.mli linguist-language=OCaml
-
-# We are pinning wasm_of_ocaml using this file in the CI. This would
-# fail on Windows otherwise.
-VERSION -text


.github/workflows/build.yml
```diff
@@ -108,13 +107,12 @@ jobs:
       - name: Set-up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: lts/*
+          node-version: latest
```
hhugo, line 110: @smorimoto, diff to the yml files have nothing to do with this PR.
Can you keep the scope of PRs to the minimum, please, explaining in the description the reason for the CHANGES.
For example, why should we remove the dune cache ? I tink I know the reason but it forces me to dig information too much.

.gitattributes
```diff
@@ -1,6 +0,0 @@
-*.ml linguist-language=OCaml
-*.mli linguist-language=OCaml
```
hhugo, line 2: This needs to stay

.github/workflows/build.yml
```diff
@@ -83,7 +83,6 @@ jobs:
 
     steps:
       - name: Set git to use LF
-        if: matrix.ocaml-compiler < 5.2
```
hhugo, line 86: I want to keep this.

.github/workflows/build.yml
smorimoto: Sure

.github/workflows/build.yml
smorimoto: OK

.gitattributes
smorimoto: OK

.github/workflows/build.yml
```diff
@@ -83,7 +83,7 @@ jobs:
...
+        if: matrix.os == 'windows-latest' || matrix.ocaml-compiler < 5.2
```
hhugo, line 86: You're missing the point, I don't want to change LF setting in git if not required. Please keep the condition as it was before

.gitattributes
```diff
@@ -1,6 +1,5 @@
+# Default behaviour, for if core.autocrlf isn't set
+* text=auto
+
 *.ml linguist-language=OCaml
 *.mli linguist-language=OCaml
-
-# We are pinning wasm_of_ocaml using this file in the CI. This would
-# fail on Windows otherwise.
-VERSION -text
```
hhugo, line 6: This needs to stay for now

smorimoto: The latest master includes a fix that makes this no longer needed.

smorimoto: In fact, CI is not failing!

.github/workflows/build.yml
hhugo: @smorimoto, can you open a PR to restore the previous version

.gitattributes
hhugo: Changes in the file don't fit with the commit message. I'll merge and amend the message

hhugo: FYI, I'm reverting this change in https://github.com/ocsigen/js_of_ocaml/pull/1789

PR #1782, opened 2024-12-31
Update GitHub Actions workflows

smorimoto:
Check the CI does not complain...

### DIFF
--- a/.github/workflows/build-wasm_of_ocaml.yml
+++ b/.github/workflows/build-wasm_of_ocaml.yml
@@ -9,8 +9,6 @@ on:
 jobs:
   build:
     env:
-      OPAMJOBS: 2
-      OPAMYES: true
       WASM_OF_OCAML: true
 
     strategy:
@@ -19,44 +17,43 @@ jobs:
         os:
           - ubuntu-latest
         ocaml-compiler:
-          - 4.14.x
-          - 5.0.x
-          - 5.1.x
-          - 5.2.x
+          - "4.14"
+          - "5.0"
+          - "5.1"
+          - "5.2"
           - ocaml-compiler.5.3.0~beta2
         separate_compilation:
           - true
         include:
           - os: macos-latest
-            ocaml-compiler: 5.2.x
+            ocaml-compiler: "5.2"
             separate_compilation: true
           - os: ubuntu-latest
-            ocaml-compiler: 4.14.x
+            ocaml-compiler: "4.14"
             separate_compilation: false
 
     runs-on: ${{ matrix.os }}
 
     steps:
-      - name: Set git to use LF
-        run: |
-          git config --global core.autocrlf false
-          git config --global core.eol lf
-          git config --global core.ignorecase false
+      - name: Checkout tree
+        uses: actions/checkout@v4
 
-      - name: Install node
+      - name: Checkout Jane Street opam repository
+        uses: actions/checkout@v4
+        with:
+          repository: janestreet/opam-repository
+          ref: feaf8f831051fd5f316963b28efd728cf0b0eca1
+          path: janestreet/opam-repository
+
+      - name: Set-up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 23
+          node-version: latest
 
-      - name: Install OCaml ${{ matrix.ocaml-compiler }}
+      - name: Set-up OCaml ${{ matrix.ocaml-compiler }}
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-
-      - name: Checkout code
-        uses: actions/checkout@v4
-        with:
-          path: wasm_of_ocaml
 
       - name: Set-up Binaryen
         uses: Aandreba/setup-binaryen@v1.0.0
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -102,23 +105,15 @@ jobs:
...
+          node-version: latest
 
       - name: Set-up OCaml ${{ matrix.ocaml-compiler }}
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          dune-cache: true
-          opam-pin: false
-
-      - run: opam install conf-pkg-config
```
hhugo, line 120: I think this is the second time you try to remove that line. It was here on purpose to work-around an opam packaging issue. Are you sure it is no longer needed

.github/workflows/lint.yml
```diff
@@ -28,7 +27,6 @@ jobs:
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: "5.2"
-          dune-cache: true
```
hhugo, line 31: You still haven't explained this change

hhugo:
@smorimoto, I'm not very happy with how you push changes without explanation/description and review.

.github/workflows/build.yml
smorimoto: Yes, it's no longer needed.

.github/workflows/lint.yml
smorimoto: From dune 3.17, dune cache no longer works with opam workflow. There is a work going on upstream about this, but there is no timeline yet...

smorimoto:
I foolishly thought it was clear enough

.github/workflows/build.yml
smorimoto: To be more clear, zarith now explicitly requires it, and conf-pkg-config is work properly in the Cygwin environment.

hhugo: I don't remember this being related to zarith

smorimoto: At least for now, `zarith` is the only library which requires `conf-pkg-config` (on Windows)

hhugo: I though the issue was with mingw-64-shims and its optional dep on conf-pkg-config. Here is an extract of a private discussion with @dra27 

> Oh, I see what's happening here - it's a race
> There's an instant during the recompile where the old mingw-64-shims package has been removed (so there's no x86_64-w64-mingw32-gcc shim installed) and those packages then try to call the C compiler.

@dra27 do you remember this issue, can you confirm it was fixed ?

smorimoto: If I remember correctly, it was fixed in `mingw-w64-shims.0.2.0`.

hhugo: No, I have failing logs involving mingw-w64-shims.0.2.0

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -19,45 +17,44 @@ jobs:
...
             separate_compilation: true
           - os: ubuntu-latest
-            ocaml-compiler: 4.14.x
+            ocaml-compiler: "4.14"
             separate_compilation: false
 
     runs-on: ${{ matrix.os }}
 
     steps:
-      - name: Set git to use LF
```
hhugo, line 40: This will need to come back once we start testing windows

smorimoto: In the upcoming PR, I'm going to make changes to separate the jobs for testing and installation testing so that we can run all tests on all platforms.

hhugo:
> I foolishly thought it was clear enough

At least the following deserves an explanation and is clearly not clear enough if you push it through without further comment.
  
```
From dune 3.17, dune cache no longer works with opam workflow. There is a work going on upstream about this, but there is no timeline yet...
```

Can you point to GH issue tracking this ? I haven't seen any CI faliure related to this.

tools/ci_setup.ml
```diff
@@ -212,7 +212,7 @@ let pin delay nm =
     (Printf.sprintf
        "opam pin add -n %s https://github.com/ocaml-wasm/%s.git#wasm"
        (try List.assoc nm aliases
-        with Not_found -> if List.mem_assoc nm packages then nm ^ ".v0.16.0" else nm)
+        with Not_found -> if List.mem_assoc nm packages then nm ^ ".v0.16.1" else nm)
```
hhugo, line 215: Yet another change hidden in the noise that is not explained.

.github/workflows/build.yml
hhugo: https://github.com/ocsigen/js_of_ocaml/actions/runs/12629900105/job/35188670458?pr=1785

hhugo:
> > I foolishly thought it was clear enough
> 
> At least the following deserves an explanation and is clearly not clear enough if you push it through without further comment.
> 
> ```
> From dune 3.17, dune cache no longer works with opam workflow. There is a work going on upstream about this, but there is no timeline yet...
> ```
> 
> Can you point to GH issue tracking this ? I haven't seen any CI faliure related to this.

Ping, can you explain the issue ?

.github/workflows/build.yml
hhugo: reverted in https://github.com/ocsigen/js_of_ocaml/pull/1786

.github/workflows/build-wasm_of_ocaml.yml
hhugo: > separate the jobs for testing and installation testing

I don't understand what you mean by that.

vouillon:
It seems some tests are no longer run:

hhugo:
> It seems some tests are no longer run

Attempted fix: https://github.com/ocsigen/js_of_ocaml/pull/1789

PR #1783, opened 2025-01-01
Use dune-build-info to manage version information

smorimoto:
The purpose of using `dune-build-info` is to simplify managing the version of packages, but I still need a pointer because I don't know where to ideally place the version retrieval file.

Ref: https://dune.readthedocs.io/en/stable/dune-libs.html#dune-build-info-library


compiler/bin-js_of_ocaml/dune
```diff
@@ -4,10 +4,9 @@
  (package js_of_ocaml-compiler)
  (libraries
   jsoo_cmdline
-  js_of_ocaml-compiler
```
hhugo, line 7: js_of_ocaml-compiler is a (direct) dependency

compiler/bin-jsoo_minify/dune
```diff
@@ -2,7 +2,7 @@
  (name jsoo_minify)
  (public_name jsoo_minify)
  (package js_of_ocaml-compiler)
- (libraries jsoo_cmdline js_of_ocaml-compiler cmdliner compiler-libs.common)
```
hhugo, line 5: js_of_ocaml-compiler  is a direct dependency

hhugo:
- The version is no longer reported in the toplevel
- The version is only available after installation, I suppose this doesn't play well with mono-repos. 
- The version reported is incorrect because the tag of latest version `5.9.1` is not an ancestor of master. With this PR, I get `5.9.0-617-g3b9f656` but I would expect a `5.9.1` prefix.

PR #1784, opened 2025-01-01
Add other textMetrics property

illusory0x0:
add other TextMetrics property https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics and comment with links

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
   and direct style, resulting in better performance when
   no effect handler is installed
 * Compiler/Runtime: Make resuming a continuation more efficient in js (#1765)
+* Lib: Add other textMetrics property (#1784)
 
 ## Bug fixes
 * Fix small bug in global data flow analysis (#1768)
--- a/lib/js_of_ocaml/dom_html.ml
+++ b/lib/js_of_ocaml/dom_html.ml
@@ -1998,7 +1998,35 @@ and canvasGradient = object
 end
 
 and textMetrics = object
+  method actualBoundingBoxAscent : number_t readonly_prop
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/actualBoundingBoxAscent *)
+
+  method actualBoundingBoxDescent : number_t readonly_prop
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/actualBoundingBoxDescent *)
+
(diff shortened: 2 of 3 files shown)


illusory0x0:
add  changelog

CHANGES.md
```diff
@@ -15,6 +15,11 @@
 * Fix small bug in global data flow analysis (#1768)
 * Runtime: no longer leak channels
 
+# 5.9.2 (2025-01-02) - Lille
+
+## Features/Changes
+* Lib: add other TextMetrics property
```
hhugo, line 21: This should be moved to the dev section above

lib/js_of_ocaml/dom_html.ml
```diff
@@ -1998,7 +1998,41 @@ and canvasGradient = object
...
+  method actualBoundingBoxLeft : number_t readonly_prop
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/actualBoundingBoxLeft *)
+
+  method actualBoundingBoxRight : number_t readonly_prop
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/actualBoundingBoxRight *)
+
+  method alphabeticBaseline : number_t readonly_prop
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/alphabeticBaseline *)
+
+  method emHeightAscent : number_t readonly_prop
```
hhugo, line 2016: This doesn't seem to be well supported

lib/js_of_ocaml/dom_html.ml
```diff
@@ -1998,7 +1998,41 @@ and canvasGradient = object
...
+  (** https://developer.mozilla.org/docs/Web/API/TextMetrics/emHeightAscent *)
+
+  method emHeightDescent : number_t readonly_prop
```
hhugo, line 2019: This doesn't seem to be well supported

hhugo: Do you really need it ? If no, we could drop it. If yes, we could use `number_t optdef readonly_prop`, forcing the user to handle the case where it's not available.

illusory0x0: no need.

CHANGES.md
illusory0x0: Thank you for your useful tips.

hhugo:
thanks

PR #1790, opened 2025-01-07
[wasm] Optimize calls to a statically known function

vouillon:
When the function to be called is statically known, we can call it directly instead of loading the code pointer from the closure.

### DIFF
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -186,10 +186,16 @@ module Generate (Target : Target_sig.S) = struct
 
   let zero_divide_pc = -2
 
+  let exact_call kind =
+    match kind with
+    | Generic -> false
+    | Exact | Known _ -> true
+
   let rec translate_expr ctx context x e =
     match e with
-    | Apply { f; args; exact }
-      when exact || List.length args = if Var.Set.mem x ctx.in_cps then 2 else 1 ->
+    | Apply { f; args; kind }
+      when exact_call kind || List.length args = if Var.Set.mem x ctx.in_cps then 2 else 1
+      ->
         let rec loop acc l =
           match l with
           | [] -> (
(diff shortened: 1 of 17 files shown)


compiler/lib/driver.ml
```diff
@@ -115,9 +115,9 @@ let effects ~deadcode_sentinal p =
       p
       |> Effects.f ~flow_info:info ~live_vars
       |> map_fst
-           (match effects with
-           | `Double_translation -> Fun.id
-           | `Cps -> Lambda_lifting.f)
+           (match Config.target (), effects with
+           | `Wasm, _ | _, `Double_translation -> Fun.id
+           | `JavaScript, `Cps -> Lambda_lifting.f)
```
hhugo, line 120: Is this a bug fix ?

compiler/lib-wasm/generate.ml
```diff
@@ -204,13 +210,14 @@ module Generate (Target : Target_sig.S) = struct
...
-                | W.RefFunc g ->
+                match funct, kind with
+                | W.RefFunc g, _ ->
                     (* Functions with constant closures ignore their
                        environment. In case of partial application, we
                        still need the closure. *)
-                    let* cl = if exact then Value.unit else return closure in
+                    let* cl = if exact_call kind then Value.unit else return closure in
                     return (W.Call (g, List.rev (cl :: acc)))
+                | _, Known g -> return (W.Call (g, List.rev (closure :: acc)))
```
hhugo, line 220: can we have `f <> g` here ?

vouillon: Sure. For instance, if you have a call `StringMap.iter h m`, `f` will be bound to the expression which accesses the relevant field of module `StringMap` while `g` will be bound to the actual closure (assuming that the functor application has not been optimized away).

hhugo:
Is this ready for review ?

PR #1791, opened 2025-01-07
Effects / double translation: keep more functions in direct style

vouillon:
We don't require escaping functions nor functions called from a CPS call site to have a CPS variant. This makes it possible to reduce the amount of code duplication. 
This implies that CPS call sites have to be able do deal with direct style functions. This makes CPS calls a little bit slower, but we don't care too much about performance in this case.

### DIFF
--- a/compiler/lib/effects.ml
+++ b/compiler/lib/effects.ml
@@ -542,15 +542,22 @@ let rewrite_instr ~st (instr : instr) : instr =
                 (Extern "caml_alloc_dummy_function", [ size; Pc (Int (Targetint.succ a)) ])
             )
       | _ -> assert false)
-  | Let (x, Apply { f; args; _ }) when not (Var.Set.mem x st.cps_needed) ->
-      (* At the moment, we turn into CPS any function not called with
-         the right number of parameter *)
-      assert (
-        (* If this function is unknown to the global flow analysis, then it was
+  | Let (x, Apply { f; args; exact }) when not (Var.Set.mem x st.cps_needed) ->
+      if double_translate ()
+      then
+        let exact =
+          (* If this function is unknown to the global flow analysis, then it was
            introduced by the lambda lifting and we don't have exactness info any more. *)
-        Var.idx f >= Var.Tbl.length st.flow_info.info_approximation
-        || Global_flow.exact_call st.flow_info f (List.length args));
-      Let (x, Apply { f; args; exact = true })
+          exact
+          || Var.idx f < Var.Tbl.length st.flow_info.info_approximation
+             && Global_flow.exact_call st.flow_info f (List.length args)
+        in
+        Let (x, Apply { f; args; exact })
+      else (
+        (* At the moment, we turn into CPS any function not called with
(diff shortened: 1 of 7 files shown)


hhugo:
@vouillon, do you want to merge this before a release ?

OlivierNicole:
I’m looking at this, but I think I need to reflect a bit more to understand this statement:


> We don't require escaping functions nor functions called from a CPS call site to have a CPS variant.

compiler/lib/effects.ml
```diff
@@ -542,15 +542,22 @@ let rewrite_instr ~st (instr : instr) : instr =
...
-      Let (x, Apply { f; args; exact = true })
+          exact
+          || Var.idx f < Var.Tbl.length st.flow_info.info_approximation
+             && Global_flow.exact_call st.flow_info f (List.length args)
+        in
+        Let (x, Apply { f; args; exact })
+      else (
+        (* At the moment, we turn into CPS any function not called with
+         the right number of parameter *)
+        assert (Global_flow.exact_call st.flow_info f (List.length args));
```
OlivierNicole, line 559: I can’t find where in the code this happens—making sure that all non-exact calls use the CPS calling convention. Would you know?

OlivierNicole (approved):
This looks good to me! I only have a question on an unrelated point.

compiler/lib/effects.ml
vouillon: The global flow analysis does not track partial applications and over applications: when a call is not exact, it considers that the function might escape. Then, the partial CPS analysis consider that this function has to be in CPS, and thus that any call point that may call this function has to be in CPS. With this PR, we no longer force the function to be in CPS, so this is no longer true.

OlivierNicole:
Here’s how this PR affects the (bzip2-ed) code size blowup induced by `--effects=double-translation` compared to `--effects=cps`:


| Revision |  `master` (3d645b9) | this PR |
--------------|----------------------------|------------|
| `ocamlc` (doesn't use effect handlers) | 161 % | 155 % |
| [`generator.ml`](https://github.com/ocaml-multicore/effects-examples/blob/8a7e80dd29e68e2d05fe4b0399c8535a51f36395/generator.ml) | 111.3 % | 111.0 % |

On `generator.ml` which is a small program using effect handlers, the effect is very small.

compiler/lib/effects.ml
OlivierNicole: I see, so it goes through escaping.

OlivierNicole:
If the calling convention can be chosen dynamically, doesn’t it open the way to being able to link together files compiled with `--effects=cps` and files compiled with `--effects=double-translation`?

vouillon:
> If the calling convention can be chosen dynamically, doesn’t it open the way to being able to link together files compiled with `--effects=cps` and files compiled with `--effects=double-translation`?

With `--effects=double-translation` , we expect that there will always be a direct style version of all functions, which we can call directly. This is not the case with `--effects=cps`.

PR #1794, opened 2025-01-09
CI: test using version 0.18 of Jane Street packages

vouillon:


runtime/wasm/bigstring.wat
```diff
@@ -201,7 +201,29 @@
                (local.set $len (i32.sub (local.get $len) (i32.const 1)))
                (local.set $pos (i32.add (local.get $pos) (i32.const 1)))
                (br $loop))))
-     (ref.i31 (i32.const -1)))
+      (ref.i31 (i32.const -1)))
+
+   (func (export "caml_bigstring_memrchr")
```
hhugo, line 206: Does this belong here ? Or in another repo ?

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -98,51 +121,81 @@ jobs:
...
         run: opam exec -- dune runtest
 
       - name: Run Bonsai tests
-        if: ${{ matrix.ocaml-compiler < '5.' && matrix.separate_compilation }}
-        working-directory: ./janestreet/lib/bonsai
+        if: matrix.all_jane_street_tests
+        working-directory: ./janestreet/lib/bonsai_test
+        run: opam exec -- dune runtest
+
+      - name: Write node wrapper
```
hhugo, line 168: Can we do something inside `ci_setup.ml` instead ? we could update janestreet/dune-workspace  by adding a `binaries` section and dumping the node wrapper somewhere in the janestreet directory

runtime/wasm/bigstring.wat
vouillon: Since I may change the implementation of big arrays at some point, I think it's simpler to put this here for now.

PR #1796, opened 2025-01-11
CI: test wasm on windows

### LINKED ISSUE #1797: Make sure wasmoo works on windows and update the CI
See https://github.com/ocsigen/js_of_ocaml/pull/1796 and https://github.com/ocsigen/js_of_ocaml/actions/runs/12747474600/job/35525613956?pr=1796
```[tasklist]
### Tasks
- [x] Sys.os_type is not set properly
- [ ] The asset directory is not copied into the inline test sandbox
- [x] There are sometimes errors when filling the asset directory
```

hhugo:
fix #1797


runtime/wasm/dune
```diff
@@ -55,10 +55,6 @@
   (glob_files *.wat))
  (action
   (progn
-   (bash
-    "which wasm-merge > /dev/null || (echo 'Error: Binaryen tools not found in the PATH'; false)")
-   (bash
-    "wasm-merge --version | grep -q 'version \\(11[89]\\|1[2-9][0-9]\\)' || (echo 'Error: Binaryen version 118 or greater is currently required'; false)")
```
hhugo, line 61: This doesn't work when running locally. I'm not sure why. Maybe we can replace it with a small ocaml program. or just drop it.

vouillon: I think we can just drop this. This is a lot less useful now that binaryen is installed by opam.

hhugo:
The CI is now happy.

hhugo:
Replaced by #1799 so we can avoid all renamings

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -48,7 +53,7 @@ jobs:
 
     steps:
       - name: Set git to use LF
-        if: ${{ matrix.os == 'windows-latest' && matrix.ocaml-compiler < 5.2 }}
+        if: ${{ matrix.os == 'windows-latest' }}
```
hhugo, line 56: I'd like to have the tests pass in presence of CRLF

.github/workflows/build-wasm_of_ocaml.yml
```diff
@@ -33,6 +33,11 @@ jobs:
             separate_compilation: true
             jane_street_tests: false
             all_jane_street_tests: false
+          - os: windows-latest
+            ocaml-compiler: "5.2"
+            separate_compilation: true
+            jane_street_tests: true
```
hhugo, line 39: @vouillon, `jane_street_tests` currently doesn't change whether we perform more and less tests. Is that expected ?

PR #1798, opened 2025-01-15
Runtime: use Uint8array method subarray rather than slice when possible to avoid a copy

vouillon:


runtime/js/marshal.js
```diff
@@ -880,10 +880,10 @@ function caml_output_value_to_bytes(v, flags) {
 }
 
 //Provides: caml_output_value_to_buffer
-//Requires: caml_output_val, caml_failwith, caml_blit_bytes
+//Requires: caml_output_val, caml_failwith, caml_blit_bytes, caml_bytes_of_array
 function caml_output_value_to_buffer(s, ofs, len, v, flags) {
   var t = caml_output_val(v, flags);
   if (t.length > len) caml_failwith("Marshal.to_buffer: buffer overflow");
-  caml_blit_bytes(t, 0, s, ofs, t.length);
```
hhugo, line 887: Can you confirm this is a bugfix ? Can we have a regression tests ?

runtime/js/mlBytes.js
```diff
@@ -78,7 +78,7 @@ function caml_subarray_to_jsbytes(a, i, len) {
   if (i === 0 && len <= 4096 && len === a.length) return f.apply(null, a);
   var s = "";
   for (; 0 < len; i += 1024, len -= 1024)
-    s += f.apply(null, a.slice(i, i + Math.min(len, 1024)));
+    s += f.apply(null, a.subarray(i, i + Math.min(len, 1024)));
```
hhugo, line 81: It is not clear that `caml_subarray_to_jsbytes`  is not used on `Array`.  How do you feel about introducing a new function and updating the callers ? `caml_sub_int8_array_to_jsbytes` ? We would need another function for the `caml_string_of_array` counterpart. 

Edit: It seems that BigStringReader.readstr is indirectly calling caml_subarray_to_jsbytes with an `Array`, so is `caml_ml_output_int`

runtime/js/marshal.js
vouillon: Indeed.

runtime/js/mlBytes.js
```diff
@@ -511,6 +511,12 @@ function caml_string_of_array(a) {
   return caml_string_of_jsbytes(caml_subarray_to_jsbytes(a, 0, a.length));
 }
 
+//Provides: caml_string_of_uint8_array
+//Requires: caml_sub_uint8_array_to_jsbytes, caml_string_of_jsbytes
+function caml_string_of_uint8_array(a) {
+  return caml_string_of_jsbytes(caml_sub_uint8_array_to_jsbytes(a, 0, a.length));
```
hhugo, line 517: Should we consider this version with js-string is not used ?
```js
//Provides: caml_string_of_uint8_array
//Requires: caml_bytes_of_uint8_array
//If: !js-string
function caml_string_of_uint8_array(a) {
  return caml_bytes_of_uint8_array(a.slice());
}
```

runtime/js/mlBytes.js
```diff
@@ -526,6 +526,12 @@ function caml_bytes_of_array(a) {
   return new MlBytes(4, a, a.length);
 }
```
hhugo, line 527: I think we should mark caml_bytes_of_array as deprecated because its semantic is confusing (using / copying its argument)

hhugo: It's still used in the runtime

hhugo:
This looks good, thanks.
Please add a changelog entry for the bugfix at least

PR #1799, opened 2025-01-16
CI: test wasm on windows (shorter dune path)

hhugo:


hhugo:
It seems we are hitting https://github.com/libuv/libuv/issues/3622

vouillon (approved):

PR #1803, opened 2025-01-17
Runtime: Add float16 support to bigarray

hhugo:
float16array is not widely supported.


hhugo:
@vouillon, I think the wasm runtime needs to be updated somehow. Can you take a look ?

hhugo:
@vouillon, I think there is a bug in the wasm runtime of bigarrays plus we need to add support for float16.
I'd like to include this in the 6.0 release. Can you take a look ?

hhugo:
> @hhugo Maybe I should switch to using `Uint16array` typed arrays, actually. Eventually, we should use `Float16Array`, but it's still under a flag in Chrome, and cannot be used with WebGL in Safari: https://github.com/WebKit/WebKit/blob/9fca11bf1d0cc85c923d86c2a8e71045c30dfc98/Source/WebCore/html/canvas/WebGLRenderingContextBase.cpp#L3767-L3769

Should I wait for this change ?

vouillon:
> Should I wait for this change ?

I'll try to do that tomorrow

PR #1806, opened 2025-01-20
CI: workaround opam pkg bug

hhugo:
Work-around a race between reinstalling mingw-w64-shims
(e.g. because of conf-pkg-config optional dep) and installing other
packages that implicitly depend on mingw-w64-shims.


hhugo:
@dra27 do you already have an issue for this ? if no, should we create one ?

dra27:
Yes, it's tracked in https://github.com/ocaml-windows/papercuts/issues/2.

PR #1810, opened 2025-01-23
Runtime: fresh obj_id when un-marshaling

hhugo:
This brings the behavior closer to the upstream runtime


hhugo:
@vouillon, this needs to be updated in the wasm runtime. Can you take a look ?

vouillon:
So, you can marshal an object if it has no method. And people are actually doing it!

PR #1811, opened 2025-01-23
Tests: Fix some more tests

hhugo:


hhugo:
@vouillon, yet another test failing in wasm. We probably just need to check more bounds when creating arrays

runtime/wasm/array.wat
```diff
@@ -55,7 +57,9 @@
       (param $n (ref eq)) (param $v (ref eq)) (result (ref eq))
       (local $sz i32) (local $f f64)
       (local.set $sz (i31.get_s (ref.cast (ref i31) (local.get $n))))
-      (if (i32.lt_s (local.get $sz) (i32.const 0))
+      (if (i32.or
+             (i32.lt_s (local.get $sz) (i32.const 0))
+             (i32.ge_s (local.get $sz) (i32.const 0x7ffffff)))
```
vouillon, line 62:
```suggestion
      (if  (i32.ge_u (local.get $sz) (i32.const 0x7ffffff))
```

PR #1813, opened 2025-01-25
Add OCaml sys tests

vouillon:


hhugo:
@vouillon, I've updated the js runtime to fix the tests. Can you make a similar fix on the wasm side ?

PR #1815, opened 2025-01-27
Many more tests

hhugo:


compiler/tests-ocaml/misc/pr7168.ml
```diff
@@ -0,0 +1,82 @@
...
+  let x = x+x in  let x = x+x in  let x = x+x in  let x = x+x in
+  let x = x+x in  let x = x+x in  let x = x+x in  let x = x+x in
+  let x = x+x in  let x = x+x in  let x = x+x in  let x = x+x in
+  let x = x+x in
+  let _ = f x in
+  ()
+
+let _ =
+  try f 1
+  with Stack_overflow -> Printf.printf "OK\n"
```
vouillon, line 82: Stack overflows cannot be caught in Wasm. So, there is no way we can make this example pass.

hhugo: disabled in wasm

hhugo:
@vouillon, can you look at the failing test `compiler/tests-ocaml/runtime-objects/` in wasm ?

PR #1820, opened 2025-01-28
Runtime fixes

vouillon:


runtime/js/fs_node.js
```diff
@@ -360,7 +360,7 @@ function MlNodeFd() {}
 //Provides: caml_sys_open_for_node
 //Requires: MlNodeFd
 function caml_sys_open_for_node(fd, flags) {
-  if (flags.name) {
+  if (flags.altname) {
```
vouillon, line 364: @hhugo Do you remember what this was for?

hhugo: I think it was related to https://github.com/grain-lang/grain/pull/614

hhugo: trying to read from fd 0 in the node interpreter gives me.
```
fs.readSync(0,b)
Uncaught Error: EAGAIN: resource temporarily unavailable, read
    at Object.readSync (node:fs:742:18) {
  errno: -11,
  code: 'EAGAIN',
  syscall: 'read'
}
````
while reading from '/dev/stdin' works.

File: runtime/js/fs_node.js [hunk 2]
```diff
@@ -360,7 +361,7 @@ function MlNodeFd() {}
...
     try {
       var fs = require("node:fs");
       var fd2 = fs.openSync(flags.name, "rs");
```
hhugo, line 367: This need to be updated as well ..

PR #1821, opened 2025-01-28
Run all tests with double translation too

OlivierNicole:
This requires some care to avoid Dune setting the flag `--effects=` twice.


compiler/tests-jsoo/lib-effects-2/dune
```diff
@@ -1,4 +1,13 @@
 (env
+ (with-effects-double-translation
+  (flags
+   (:standard -w -38))
+  (js_of_ocaml
+   ;; separate compilation doesn't yet work when using
+   ;; '--effect=double-translation' since Dune doesn't know it should compile a
+   ;; different version of the dependencies.
+   ;; TODO: remove once support in ocaml/dune#11222 is released.
+   (compilation_mode whole_program)))
```
vouillon, line 10:
```suggestion
 (with-effects-double-translation
  (flags
   (:standard -w -38)))
```

vouillon: Actually, I think we can just remove the directories `compiler/tests-jsoo/lib-effects-2/` and `compiler/tests-ocaml/effects-2`.

OlivierNicole:
@vouillon It looks like you have erased some of my commits when force-pushing. I’ll re-add them on top when the CI has finished.

.github/workflows/build.yml
```diff
@@ -148,7 +153,9 @@ jobs:
       - run: opam exec -- make tests
         if: ${{ !matrix.skip-test }}
 
-      - run: opam exec -- dune build @all @runtest @runtest-js --profile with-effects
+      - run: |
+          opam exec -- dune build @all @runtest @runtest-js --profile with-effects
```
vouillon, line 157: You should split this into two actions

vouillon (approved):

dune
```diff
@@ -25,15 +21,12 @@
   (js_of_ocaml
    (compilation_mode separate)
    (flags
-    (:standard --enable effects --effects double-translation))
+    (:standard --effects double-translation))
    (build_runtime_flags
-    (:standard --enable effects --effects double-translation)))
+    (:standard --effects double-translation)))
```
vouillon, line 26: You should grep for `build_runtime_flags` and remove this option wherever it is not needed.

tools/node_wrapper.ml
```diff
@@ -4,6 +4,8 @@ let extra_args_for_wasoo =
   ; "--stack-size=10000"
   ]
 
+let extra_args_for_jsoo = [ "--stack-size=3000" ]
```
hhugo, line 7: Can we have a comment explaining why we need this ?

hhugo:
I've pushed a commit to have tests-dynlink-js and lwt_toplevel work with double-translation.

hhugo:
manual/effects.wiki is outdated regarding dune support, can someone update it either in this PR or in a separate one ? We need this before the 6.0 release.

tools/node_wrapper.ml
vouillon: I have added that the double translation needed a larger stack. But I don't have a good explanation for that.

hhugo: I've pushed a change to only set stack-size with the with-effects-double-translation profile

PR #1822, opened 2025-01-30
Wasm text file preprocessor

vouillon:
For a start, it is only used to manage runtime differences between OCaml versions, and to provide some syntactic sugar for strings. But I plan to use it for a WASI version of the runtime and [to implement the use-js-string flag](https://github.com/ocsigen/js_of_ocaml/pull/1772).


compiler/lib-wasm/wat_preprocess.ml
```diff
@@ -0,0 +1,483 @@
...
+      then
+        raise
+          (Error
+             ( Sedlexing.lexing_positions lexbuf
+             , Printf.sprintf "Unexpected end of file.\n" ));
+      List.rev acc, (start_loc, loc')
+  | ATOM s ->
+      parse_list lexbuf toplevel start_loc ({ loc = loc, loc'; desc = Atom s } :: acc)
+
+let parse lexbuf =
```
hhugo, line 183: Isn't this the second sexp parser  in this lib ?

hhugo: Sexplib / [Parsexp ](https://github.com/janestreet/parsexp) are probably fine to add. The major downside is that new releases require ocaml >= 5.1.

> The first one is used internally to read library metadata. This is used for linking, so it's performance critical. And the metadata can be quite large (for core, for instance). Atoms are arbitrary strings, so it is convenient to serialize / deserialize data. I would be happy to use sexplib instead but that would add an external dependency.

Couldn't we use marshal for this ?

We can probably keep the 2 sexp implementation for now if it's easier but I'd like to clarify one aspect, the syntax for comments. In addition to line comments, wa_preprocess has block comments but it seems to use a syntax different from parsexp/sexplib. Does the syntax exist elsewhere ?

hhugo:
Have you considered using this to add sugars to help write control flows (e.g having a `switch` instruction )  ?

compiler/lib-wasm/wat_preprocess.ml
vouillon: > Sexplib / [Parsexp ](https://github.com/janestreet/parsexp) are probably fine to add. The major downside is that new releases require ocaml >= 5.1.

I think its fine to use an older version for OCaml < 5.1

> Couldn't we use marshal for this ?

A text format is much more practical for debugging

> We can probably keep the 2 sexp implementation for now if it's easier but I'd like to clarify one aspect, the syntax for comments. In addition to line comments, wa_preprocess has block comments but it seems to use a syntax different from parsexp/sexplib. Does the syntax exist elsewhere ?

This is not standard indeed. It seems to be only used by the [Wasm text format](https://webassembly.github.io/spec/core/text/lexical.html#comments).

vouillon:
> Have you considered using this to add sugars to help write control flows (e.g having a `switch` instruction ) ?

That seems a good idea, indeed!

compiler/lib-wasm/wat_preprocess.ml
hhugo: > This is not standard indeed. It seems to be only used by the [Wasm text format](https://webassembly.github.io/spec/core/text/lexical.html#comments).

Can you add a reference to the spec somewhere ?

compiler/bin-wasmoo_util/dune
```diff
@@ -0,0 +1,17 @@
+(executable
+ (name wasmoo_util)
```
hhugo, line 2: Have you consider having subcommands inside wasm_of_ocaml instead ?

hhugo (approved):

compiler/bin-wasmoo_util/dune
vouillon: We have the issue that `wasmoo_util link` is used to build the runtime which is then included in `wasm_of_ocaml`.
But maybe I can have an internal binary to do that and also expose the same functionality as a wasm_of_ocaml subcommand.

hhugo: Shouldn't the runtime be linked later ? If we want to have it conditional on flags such as use-js-strings ?

vouillon: My current plan is to have a set of precompiled runtimes for the most common options, and to compile runtimes on the fly otherwise.
It's a bit slow to compile and link the runtime. That's especially noticeable when we have a large number of executables, like with the Js_of_ocaml test suite.
But maybe we could share the runtimes produced by `wasm_of_ocaml build-runtime` between executables, using a similar trick in dune as for ppxs?

compiler/lib-wasm/binaryen.mli
```diff
@@ -16,12 +16,18 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
+type link_input =
+  { module_name : string
```
hhugo, line 20: Can you document the fields ?

compiler/lib-wasm/link.mli
```diff
@@ -24,6 +24,10 @@ module Wasm_binary : sig
     ; name : string
     }
 
+  val check : contents:string -> bool
```
hhugo, line 27: doc comments would be welcome ?

compiler/lib-wasm/wat_preprocess.mli
```diff
@@ -0,0 +1,22 @@
...
+  | String of string
+  | Version of int * int * int
+
+val f : variables:(string * value) list -> filename:string -> contents:string -> string
+
+type source =
+  | File (* Binary file (skipped) *)
+  | Contents of string (* Contents to preprocess *)
+
+type input =
```
hhugo, line 12: some documentation would be welcome

compiler/bin-wasmoo_util/dune
hhugo: Sounds good

PR #1823, opened 2025-01-31
Add some OCaml Unix tests and implement the missing primitives

vouillon:


runtime/wasm/runtime.js
```diff
@@ -135,7 +139,42 @@
...
+      large,
+      s.dev,
+      s.ino | 0,
+      kind,
+      s.mode,
+      s.nlink,
+      s.uid,
+      s.gid,
+      s.rdev,
+      BigInt(s.size),
```
hhugo, line 169: You should update ECMASCRIPT.md to mention bigint

compiler/tests-ocaml/lib-unix/isatty/dune
```diff
@@ -0,0 +1,7 @@
+(tests
+ (names isatty_std)
+ ; isatty_tty does not work on the CI since we are not running in a tty there
```
hhugo, line 3: Can we keep it when running tests locally ? Is there any magic `(enabled_if <XXX>)` we can use ?

compiler/tests-ocaml/lib-unix/common/dune
```diff
@@ -0,0 +1,23 @@
...
+  channel_of
+  cmdline_prog
+  reflector
+  rename
+  truncate
+  uexit
+  utimes)
+ (libraries ocaml_testing unix)
+ (flags
+  (:standard -no-strict-formats \ -strict-formats))
```
hhugo, line 13: Maybe we should change the default in an env stanza in compiler/tests-ocaml/dune

compiler/tests-ocaml/lib-unix/isatty/dune
hhugo: Pushed a fix

hhugo:
Can you add some changelog entry ?

compiler/tests-ocaml/lib-unix/common/dune
hhugo: done in #1824

PR #1825, opened 2025-02-03
Various runtime fixes

vouillon:
@hhugo I'm a bit worried of making large changes right before a release. So, if you are unsure of some of the changes, I can just omit them.

### DIFF
--- a/runtime/js/sys.js
+++ b/runtime/js/sys.js
@@ -18,9 +18,9 @@
 ///////////// Sys
 
 //Provides: caml_raise_sys_error (const)
-//Requires: caml_raise_with_string, caml_global_data
+//Requires: caml_raise_with_arg, caml_global_data, caml_string_of_jsstring
 function caml_raise_sys_error(msg) {
-  caml_raise_with_string(caml_global_data.Sys_error, msg);
+  caml_raise_with_arg(caml_global_data.Sys_error, caml_string_of_jsstring(msg));
 }
 
 //Provides: caml_sys_exit
--- a/runtime/js/unix.js
+++ b/runtime/js/unix.js
@@ -220,7 +220,7 @@ function caml_unix_stat(name) {
 function caml_unix_stat_64(name) {
   var root = resolve_fs_device(name);
   if (!root.device.stat) {
-    caml_failwith("caml_unix_stat: not implemented");
+    caml_failwith("caml_unix_stat_64: not implemented");
   }
   return root.device.stat(
     root.rest,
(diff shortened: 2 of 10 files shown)


runtime/js/unix.js
```diff
@@ -188,7 +188,8 @@ function make_unix_err_args(code, syscall, path, errno) {
       errno = -9999;
     }
     // If none of the above variants, fallback to EUNKNOWNERR(int)
-    variant = BLOCK(0, errno);
+    // errno is expected to be positive
+    variant = BLOCK(0, -errno);
```
hhugo, line 192: `Math.abs` Maybe ?

vouillon: If you look at `util.getSystemErrorMap()`, the `errno` value is always negative in nodejs. See https://ant.readthedocs.io/en/latest/errors.html:
> Implementation detail: on Unix error codes are the negated errno (or -errno), while on Windows they are defined by libuv to arbitrary negative numbers. 

So we can just take the opposite.

runtime/js/unix.js
```diff
@@ -294,19 +295,16 @@ function caml_unix_rmdir(name) {
...
 
 //Provides: caml_unix_symlink
-//Requires: resolve_fs_device, caml_failwith
+//Requires: resolve_fs_device, caml_failwith, caml_jsstring_of_string
 //Alias: unix_symlink
 function caml_unix_symlink(to_dir, src, dst) {
-  var src_root = resolve_fs_device(src);
   var dst_root = resolve_fs_device(dst);
-  if (src_root.device !== dst_root.device)
-    caml_failwith("caml_unix_symlink: cannot symlink between two filesystems");
```
hhugo, line 303: I'm unsure about his one change, because the current implem will not be able to follow links across device.

runtime/js/sys.js
```diff
@@ -18,9 +18,9 @@
 ///////////// Sys
 
 //Provides: caml_raise_sys_error (const)
-//Requires: caml_raise_with_string, caml_global_data
+//Requires: caml_raise_with_arg, caml_global_data, caml_string_of_jsstring
 function caml_raise_sys_error(msg) {
-  caml_raise_with_string(caml_global_data.Sys_error, msg);
+  caml_raise_with_arg(caml_global_data.Sys_error, caml_string_of_jsstring(msg));
```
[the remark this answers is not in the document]
hhugo: @smorimoto, copilot seems to only emits noises, Can you stop assigning it on PR you don't work on / review  ?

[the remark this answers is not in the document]
smorimoto: Sure

hhugo:
> I'm a bit worried of making large changes right before a release

I don't mind minor releases if something break.

Can you update CHANGES.md ?

runtime/js/unix.js
vouillon: This test is wrong anyway, since `src` should be interpreted relatively to `dst`, not relatively to the current working directory.

PR #1827, opened 2025-02-04
Misc: prepare the release

hhugo:


vouillon:
> @hhugo There are some concerns from Jane Street that #1769 might be too aggressive. In particular, they are relying on the fact that the method `localStorage` of `window` has type `storage t optdef readonly_prop` to access safely the local storage when it is available, without crashing in other contexts (node / service workers). Also, Chrome is sending [fake `mousedown` events](https://github.com/chromium/chromium/commit/55518e17850cac0bdc6fca7b24092bea479e34db) without a `key` nor a `code` property.

I think it makes sense to revert the changes for the `localStorage` and `sessionStorage`. I'm not so sure regarding key events. @rickyvetter @TyOverby  What do you think?

TyOverby:
I think that something needs to be done about the events handler types.  

The “correct” change would be to remove the subtypes from all event handlers, requiring the user to try to upcast an Event to Keyboard event.

I think it would be easier to keep all of the fields as optdef, allowing people to continue “duck typing” with the fields.

But the current state is not great for us, we have a lot of applications that handle these fields not being present, and if we were to remove the optdef checks, they’d just crash instead

PR #1828, opened 2025-02-04
Revert some of #1769 for compatibility

vouillon:


lib/js_of_ocaml/dom_html.mli
```diff
@@ -2201,9 +2204,13 @@ class type window = object
 
   method scrollBy : number_t -> number_t -> unit meth
 
-  method sessionStorage : storage t readonly_prop
+  (* These two properties are not available on non-Web environments
+     (for instance, Web workers, node). So we keep the [optdef]
+     annotation for now. *)
```
hhugo, line 2209: Isn't this true for most method inside `window` ?

vouillon: Yes, indeed. The point is that existing programs have been written (and tested) with this in mind but, if one is not careful, one might introduce a bug when removing the test when accessing the local/session storage.

tools/ci_setup.ml
```diff
@@ -34,7 +34,6 @@ let forked_packages =
     ; "base_bigstring"
     ; "bin_prot"
     ; "bonsai_test"
-    ; "bonsai_web"
```
hhugo, line 37: Why don't you fix bonsai_web instead ?

vouillon: It no longer needs to be fixed, so I have removed it from the list of packages which I have modified. The Jane Street version will be downloaded instead.

hhugo: ok

PR #1829, opened 2025-02-04
Implement more Unix primitives

vouillon:


runtime/js/unix.js
```diff
@@ -77,6 +91,22 @@ function caml_unix_startup() {}
 //Alias: win_cleanup
 function caml_unix_cleanup() {}
 
+//Provides: caml_sys_putenv (const, const)
+//Requires: caml_jsstring_of_string, jsoo_static_env
+function caml_sys_putenv(name, value) {
+  var name = caml_jsstring_of_string(name);
+  var value = caml_jsstring_of_string(value);
+  delete jsoo_static_env[name];
```
hhugo, line 99: getenv can be optimized at compile time based on the value at static env variable. I'm not sure we should alter it  at runtime.

vouillon: I have removed `putenv`. It is not very useful anyway.

hhugo:
Aren't there any test in the ocaml testsuite for theses ?

vouillon:
> Aren't there any test in the ocaml testsuite for theses ?

Many of them are not tested. I have added a test but it is far from complete. I'll write more.

runtime/js/unix.js
```diff
@@ -294,6 +349,26 @@ function caml_unix_rmdir(name) {
...
+//Provides: caml_unix_link
+//Requires: resolve_fs_device, caml_failwith, caml_raise_system_error
+//Alias: unix_link
+function caml_unix_link(follow, src, dst) {
+  var src_root = resolve_fs_device(src);
+  var dst_root = resolve_fs_device(dst);
+  if (!src_root.device.link) {
+    caml_failwith("caml_unix_link: not implemented");
+  }
+  if (typeof follow !== "number")
```
hhugo, line 361: Can you add a comment explain this ?

runtime/js/unix.js
```diff
@@ -589,10 +730,43 @@ function caml_unix_getuid(unit) {
   return 1;
 }
 
-//Provides: caml_unix_getpwuid
+//Provides: caml_unix_geteuid
+//Alias: unix_geteuid
+function caml_unix_geteuid(unit) {
+  if (globalThis.process && globalThis.process.geteuid) {
```
hhugo, line 736: We use `?.` elsewhere

runtime/wasm/io.wat
```diff
@@ -605,6 +617,17 @@
                   (struct.get $channel $fd (local.get $ch))))
             (i64.extend_i32_s (struct.get $channel $curr (local.get $ch))))))
 
+(;ZZZ
```
hhugo, line 620: Do you want to keep this ?

PR #1831, opened 2025-02-06
WASI runtime

vouillon:
You can produce a WASI binary by running `wasm_of_ocaml` with the `--enable wasi` flag:
```
wasm_of_ocaml --enable wasi foo.byte -o foo.js
```
This produces some Wasm code that can be directly executed by the [Wizard engine](https://github.com/titzer/wizard-engine):
```
wizeng.x86-64-linux -ext:stack-switching -ext:legacy-eh foo.assets/code.wasm
```
and by [wasmtime](https://github.com/bytecodealliance/wasmtime):
```
wasmtime -W=all-proposals=y foo.assets/code.wasm
```

Note that we keep the same output convention, at least for now: a JavaScript file `foo.js` and a directory `foo.assets` containing the Wasm code. The JavaScript file can be use to run the WASI binary with `node`.

Blocked by:
- [x] #2177 
- [x] https://github.com/ocaml/dune/pull/13613
- [x] and a new release of dune
- [x] #2189


compiler/lib/build_info.ml
```diff
@@ -156,7 +157,7 @@ let configure t =
   StringMap.iter
     (fun k v ->
       match k with
-      | "use-js-string" -> Config.Flag.set k (bool_of_string v)
+      | "use-js-string" | "wasi" -> Config.Flag.set k (bool_of_string v)
```
hhugo: does that mean we need to change dune rules as well ?

vouillon: Yes, indeed. And for #1832 as well...

Maybe we should try to move some of the logic which is currently in dune to js_of_ocaml.

hhugo: Do you have anything in mind ?
I was hoping we could do some kind of dynamic rule generation similar to what's done for ppxes.

vouillon: Something like adding to js_of_ocaml an option `--output-config` which would just output the path that dune should use, and an option `--with-config` that would decode the path and override other command line parameters.

compiler/bin-wasm_of_ocaml/link_wasm.ml
```diff
@@ -99,6 +99,15 @@ let link
...
     } =
+  Js_of_ocaml_compiler.Config.set_effects_backend
+    (match
+       List.find_map
+         ~f:(fun (k, v) -> if String.equal k "effects" then Some v else None)
+         variables.set
+     with
+    | Some "native" -> `Native
+    | Some "cps" -> `Cps
+    | _ -> `Jspi);
```
hhugo: Waybe we should not silently ignore unknown  values

hhugo: There is similar diff in #2189, is there a lot of intersection between the two ?

vouillon: The PR is on top of #2189.

hhugo:
Do you want this in the upcoming release ?

vouillon:
> Do you want this in the upcoming release ?

That would be cool.

.github/workflows/wasm_of_ocaml.yml
```diff
@@ -103,6 +116,25 @@ jobs:
         with:
           node-version: ${{ matrix.os == 'windows-latest' && 'latest' || 'v26.0.0-v8-canary20260216631fb6e5ef' }}
 
+      - name: Set-up Rust toolchain
+        if: matrix.wasi
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Checkout Wasmtime
```
hhugo: Could we use https://github.com/bytecodealliance/actions/#install-wasmtime

vouillon: Done

runtime/wasm/unix.wat
```diff
@@ -790,6 +1857,217 @@
...
+               (if (local.get $res)
+                  (then
+                     (call $caml_unix_error (local.get $res) (@string "write")
+                        (global.get $no_arg))))
+               (local.set $n (i32.load (local.get $nwritten)))
+               (local.set $written (i32.add (local.get $written) (local.get $n)))
+               (local.set $pos (i32.add (local.get $pos) (local.get $n)))
+               (local.set $len (i32.sub (local.get $len) (local.get $n)))
+               (br $loop))))
+      (ref.i31 (local.get $n)))
```
hhugo: Is the the total number of bytes written ?

runtime/wasm/unix.wat
```diff
@@ -149,6 +217,102 @@
...
+                  (else
+                     (local.set $errcode (i32.const -1))))))
+         (else
+            (local.set $errcode
+               (i31.get_u
+                  (ref.cast (ref i31)
+                     (array.get $block
+                        (ref.cast (ref $block) (local.get $err))
+                        (i32.const 1)))))))
+      (if (i32.gt_u (local.get $errcode)
```
hhugo: should this be ge_u ?

runtime/wasm/sys.wat
```diff
@@ -241,6 +506,28 @@
 
    (@string $toString "toString")
 
+(@if wasi
+(@then
+   (func $caml_handle_sys_error (export "caml_handle_sys_error")
+      (param $arg (ref eq)) (param $errno i32)
+      (local $msg (ref eq))
+      (local.set $msg
+         (if (result (ref eq)) (i32.gt_u (local.get $errno)
```
hhugo: Should this be ge_u ?

runtime/wasm/unix.wat
```diff
@@ -471,7 +1074,147 @@
...
+               (call $free (struct.get $directory $buffer (local.get $dir)))
+               (struct.set $directory $buffer (local.get $dir) (local.get $buf))
+               (struct.set $directory $size (local.get $dir)
+                  (local.get $entry_size))))
+         (block $done
+            (br_if $done
+               (i32.and
+                  (i32.ne (i32.const 0)
+                     (struct.get $directory $available (local.get $dir))
+                  (i32.lt_u (struct.get $directory $available (local.get $dir))
```
hhugo, line 1172: Is this meant to be a an argument of `i32.and`  ? I think the parents are wrong

runtime/wasm/marshal.wat
```diff
@@ -45,11 +45,76 @@
...
+      (struct
+         (field $size (mut i32))
+         (field $keys (mut (ref $block)))
+         (field $values (mut (ref $block)))))
+   (func $map_new (result (ref any))
+      (struct.new $map
+         (i32.const 0)
+         (array.new $block (ref.i31 (i32.const 0)) (i32.const 2))
+         (array.new $block (ref.i31 (i32.const 0)) (i32.const 2))))
+   (func $map_get (param $map (ref any)) (param $k (ref eq)) (result i31ref)
```
hhugo, line 61: Is this function doing a linear scan ? Can we do better ?

runtime/wasm/unix.wat
```diff
@@ -324,6 +709,48 @@
...
+         (ref.i31 (i32.wrap_i64 (i64.load (local.get $p))))
+         (ref.i31 (i32.wrap_i64 (i64.load offset=8 (local.get $p))))
+         (ref.i31
+            (array.get $constr_table
+               (global.get $file_kinds) (i32.load8_u offset=16 (local.get $p))))
+         (ref.i31 (i32.const 384 (;0600;)))
+         (ref.i31 (i32.wrap_i64 (i64.load offset=24 (local.get $p))))
+         (ref.i31 (i32.const 1))
+         (ref.i31 (i32.const 1))
+         (ref.i31 (i32.wrap_i64 (i64.load (local.get $p))))
```
hhugo: Should this be zero ?

runtime/wasm/fs.wat
```diff
@@ -132,11 +683,34 @@
...
 
+(@if wasi
+(@then
+   (func (export "caml_sys_file_exists")
+      (param $name (ref eq)) (result (ref eq))
+      (local $p (tuple i32 i32 i32))
+      (local $res i32) (local $buffer i32)
+      (local.set $p (call $wasi_resolve_path (local.get $name)))
+      (if (i32.lt_s (tuple.extract 3 0 (local.get $p)) (i32.const 0))
+         (then (return (ref.i31 (i32.const 0)))))
```
hhugo: Aren't we missing a ` (call $free (tuple.extract 3 1 (local.get $p)))` ?

runtime/wasm/fs.wat
```diff
@@ -52,13 +92,291 @@
...
+            (tuple.extract 2 0 (local.get $res))
+            (local.get $p)
+            (array.len (tuple.extract 2 1 (local.get $res))))))
+
+   (func $caml_sys_resolve_path (export "caml_sys_resolve_path")
+      (param $path (ref eq)) (result i32 i32 i32)
+      (local $res (tuple i32 i32 i32))
+      (local.set $res (call $wasi_resolve_path (local.get $path)))
+      (if (i32.lt_s (tuple.extract 3 0 (local.get $res)) (i32.const 0))
+         (then ;; ENOENT
```
hhugo, line 368: Should we `(call $free (tuple.extract 3 1 (local.get $p)))` here ?

runtime/wasm/fs.wat
```diff
@@ -88,7 +453,130 @@
...
+               (local.set $new_buf
+                  (call $checked_malloc (local.get $entry_size)))
+               (call $free (local.get $buf))
+               (local.set $buf (local.get $new_buf))
+               (local.set $size (local.get $entry_size))))
+         (block $done
+            (br_if $done
+               (i32.and
+                  (i32.ne (i32.const 0) (local.get $available))
+                  (i32.lt_u (local.get $available) (local.get $size))))
```
hhugo, line 557: Don't we need to reset $available if the buffer was refill ?

runtime/wasm/unix.wat
vouillon: Yes, `Unix.write` repeats the writing operation until all bytes have been written or an error occurs (in which case it raises).

runtime/wasm/marshal.wat
vouillon: No, we can't. We can only check for pointer equality in Wasm.

runtime/wasm/unix.wat
hhugo: Shouldn't we return `$written` ?

runtime/wasm/fs.wat
vouillon: No, the path is not allocated in case of error.

runtime/wasm/unix.wat
vouillon: Oh, right

tools/node_wrapper.ml
```diff
@@ -19,16 +32,34 @@ let env =
       else e)
     env
 
-let args =
+let environment_args () =
+  List.filter
+    (fun e -> not (String.contains e ','))
```
hhugo, line 40: Why/What do we filter here ?

vouillon: Some tests need to have access to some environment variables. But environment variable whose value contains a comma would be misparsed (`--env=A=1,B=2` is interpreted as `--env=A=1 --env=B=2`).

An alternative would be to just include a fixed list of environment variables.

hhugo: Maybe worth a small comment

tools/ci_setup-mainstream.ml
```diff
@@ -52,6 +52,7 @@ let node_wrapper =
  (name node_wrapper)
        (libraries unix))|} )
   ; "node_wrapper/node_wrapper_per_profile.ml", {|let args = []|}
+  ; "node_wrapper/node_wrapper_per_engine.ml", {|let engine = "node"|}
```
hhugo, line 54: Do we still use the _per_profile.ml file ?

runtime/wasm/prng.wat
```diff
@@ -16,13 +16,30 @@
...
       (func $dv_set_i64 (param externref i32 i64 i32)))
+   (import "bindings" "littleEndian" (global $littleEndian i32))
+))
+   (import "bigarray" "caml_ba_get_data"
+      (func $caml_ba_get_data (param (ref eq)) (result (ref extern))))
    (import "bigarray" "caml_ba_get_view"
       (func $caml_ba_get_view (param (ref eq)) (result (ref extern))))
-   (import "bindings" "littleEndian" (global $littleEndian i32))
+
+(@if wasi
```
hhugo: Can you move this inside the `(@if wasi ..)` above

tools/ci_setup-mainstream.ml
vouillon: No, indeed.

hhugo:
I didn't review the code myself, it's a lot of wasm added, I just went through some parts highlighted by Claude code.
That makes we wonder how to maintain all this.. Are there any opportunities for simplification ? Do we have enough coverage in the testsuites ?

This PR target wasi preview1, What about preview2 ?

I see https://github.com/vouillon/wax is moving a lot. How long before we can review wax files instead of wast ?

runtime/wasm/prng.wat
```diff
@@ -16,13 +16,26 @@
 ;; Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
 (module
+
+   (import "bigarray" "caml_ba_get_data"
```
hhugo: This doesn't seem to be used

compiler/lib-wasm/generate.ml
```diff
@@ -2593,6 +2626,10 @@ let add_start_function = G.add_start_function
 
 let add_init_function = G.add_init_function
 
+let add_missing_primitives =
+  let module G = Generate (Gc_target) in
```
hhugo:
```suggestion
```

G is already bound above at 2604

tools/node_wrapper.ml
```diff
@@ -1,3 +1,16 @@
+let wizard_args = [ "--ext:stack-switching"; "--stack-size=2M"; "--dir=."; "--dir=/tmp" ]
```
hhugo, line 1: The wasm_overview.wiki says
```
and the Wizard engine (with
the {{{--ext:gc --ext:exception-handling}}} flags)
```

manual/wasm_overview.wiki
```diff
@@ -85,6 +85,27 @@ runtime with support for the WasmFX extension (currently available, behind the
...
+Wasmtime does not support the legacy Wasm exception-handling
+instructions, so when {{{--enable wasi}}} is used the compiler emits
+the new {{{exnref}}}-based instructions instead.
+
+For now, the output remains the same as without the {{{--enable
+wasi}}} flag: a JavaScript file {{{foo.js}}} and a directory
+{{{foo.assets}}} containing the Wasm code {{{code.wasm}}}. The
+JavaScript file can be used to run the WASI binary with {{{node}}},
+while the Wasm code can be run directly by other Wasm engines.
+
```
hhugo, line 112: What are the differences when target wasi ? non-weak weakref ? Marshall sharing is O(n2) ? Dynlink ? any other limitation ? Maybe we can have a few words about it.

hhugo:
Other raw review straight from AI

  - readdir after closedir reads a dangling buffer — runtime/wasm/unix.wat:1201-1219.
  closedir frees the entry buffer but leaves $pos/$available, so a subsequent Unix.readdir
  parses dirents out of low linear memory and returns garbage strings. Resetting
  $pos/$available makes it fall through to fd_readdir on the closed fd → proper EBADF.
  - Channel refill/output hooks silently ignored under WASI —
  runtime/wasm/io.wat:752-760,1254-1279. The setters still store closures but the wasi
  read/flush paths never consult them (Sys_js.set_channel_flusher etc. silently does fd
  I/O). Similarly caml_create_file (fs.wat:763-765) discards content and returns success.
  Failing loudly would be less surprising.

vouillon:
> I didn't review the code myself, it's a lot of wasm added, I just went through some parts highlighted by Claude code. That makes we wonder how to maintain all this.. Are there any opportunities for simplification ? Do we have enough coverage in the testsuites ?

I have made some simplifications.

I think we have a pretty good coverage with the OCaml test suite and a few additional tests.

> This PR target wasi preview1, What about preview2 ?

preview2 is completely different, unfortunately.
 
> I see https://github.com/vouillon/wax is moving a lot. How long before we can review wax files instead of wast ?

This is getting close.

tools/node_wrapper.ml
vouillon: I have updated the documentation.

manual/wasm_overview.wiki
vouillon: I have added a section in the documentation.

runtime/wasm/dune
```diff
@@ -57,3 +94,34 @@
...
+     -w
+     /src
+     ghcr.io/webassembly/wasi-sdk
+     /opt/wasi-sdk/bin/clang
+     -O2
+     libc.c
+     -flto
+     -o
+     -)
+    (run wasm-opt -Oz --strip-debug --strip-dwarf - -o -)))))
```
hhugo: I think this target is part of the default alias, and will cause failures if one tries to compile without docker available. Can you verify that ? Can we think of a better workflow ? This question is also relevant for #2249

vouillon: Indeed. I have pushed a fix.

hhugo (approved):

PR #1832, opened 2025-02-06
Implementation of effects based on the Stack Switching proposal

vouillon:
To generate Wasm code that makes use of the Stack Switching proposal, you first need to install the development version of Binaryen (or version 124 once it is released).

Then, you need to run `wasm_of_ocaml` with the `--enable wasi` flag:
```
wasm_of_ocaml --enable wasi foo.byte -o foo.js
```
This produces some Wasm code that can be directly executed by the [Wizard engine](https://github.com/titzer/wizard-engine):
```
wizeng.x86-64-linux -ext:stack-switching -ext:legacy-eh foo.assets/code.wasm
```
Note that we keep the same output convention, at least for now: a JavaScript file `foo.js` and a directory `foo.assets` containing the Wasm code. The JavaScript file should be useful to run the tests with `node` once it supports this extension. It is used at the moment in https://github.com/ocsigen/js_of_ocaml/pull/1831 to test the WASI runtime.


linxuanm:
Hi! We are working on Wizard's stack switching and would like to generate benchmarks with `wasm_of_ocaml`. I tried to build `wasm_of_ocaml` but it seems that it depends on `binaryen-bin-1.19` while an external installation of `binaryen-1.24` is needed. I was able to get around this by manually editing path but was wondering if there are any plans for removing the outdated `binaryen-bin` from dependency so that building `wasm_of_ocaml` wouldn't incorrectly use the outdated executables from `opam` path, or if there's a better way to streamline the installation for CI? Thank you!

hhugo:
Is this replaced by #2189

vouillon:
> Is this replaced by #2189

This PR performs some tests which are not performed by #2189. I need to rebase it on top of #1831.

vouillon:
Now implemented in #1831.

PR #1842, opened 2025-02-24
Add wasm_of_ocaml benchmarks with current-bench output

OlivierNicole:
@vouillon and I (so far, mostly @vouillon) have started to look at ways of improving performance of Wasm code. To keep track of our progress, we need to set up a benchmarking suite that gets run regularly.

This PR makes the repository compatible for monitoring by https://github.com/ocurrent/current-bench/: the benchmark suite is extended with the current-bench output format and I modified the scripts so that Wasm benchmarks are first-class citizens. The existing benchmarking utilities should still work, I ran a few checks to verify that.

This is a first step, we will soon want to add macrobenchmarks as well — presumably by having a `benchmarks/sources/macro` directory where each subdirectory is a macrobenchmark with its own dune file.

### DIFF
--- a/manual/wasm_overview.wiki
+++ b/manual/wasm_overview.wiki
@@ -7,10 +7,6 @@ It provides an alternative way to run pure OCaml programs in JavaScript environm
 
 The compiler is provided by the wasm_of_ocaml-compiler package. The <<a_manual chapter="overview" |Js_of_ocaml libraries>> are compatible with this compiler.
 
-== Dependencies
-
-Wasm_of_ocaml depends on [[https://github.com/WebAssembly/binaryen|the Binaryen toolchain]] version 119 or later.
-
 == Installation
 
 The easiest way to install wasm_of_ocaml is to use opam.
--- a/wasm_of_ocaml-compiler.opam
+++ b/wasm_of_ocaml-compiler.opam
@@ -27,6 +27,7 @@ depends: [
   "menhirLib"
   "menhirSdk"
   "yojson" {>= "2.1"}
+  "binaryen-bin"
   "odoc" {with-doc}
 ]
 depopts: ["ocamlfind"]


OlivierNicole:
Note: this is ready for review but I still need to verify details—namely, how current-bench deals with multiple opam files, and what should be the name of the results file—so we should hold off from merging yet.

OlivierNicole:
Thanks to @punchagan’s advice, this should now hopefully work with current-bench.

vouillon:
I think `make bench` needs to build `wasm_of_ocaml`. The opam file is here only to install external dependencies.

vouillon:
I would put all the results together and [grouping them](https://github.com/ocurrent/current-bench/blob/main/doc/json_spec.md#groupping-metrics)? 
```Json
{
  "name": "Wasm_of_ocaml",
  "results": [
    {
      "name": "Microbenchmarks",
      "metrics": [
        {
          "name": "microbenchmark/almabench",
          "value": 1.6087265,
          "units": "s"
        },
        {
          "name": "microbenchmark/bdd",
          "value": 0.31883700000000004,
          "units": "s"
        },
```

OlivierNicole:
This will cause all the run times to appear on one unique graph. I’m not sure we want that?

OlivierNicole:
> I think `make bench` needs to build `wasm_of_ocaml`. The opam file is here only to install external dependencies.

Actually, I disagree. When you look at the last line of https://github.com/ocurrent/current-bench/blob/b0528d005b34b8ea1ff4cd7f893165ce782ef5db/pipeline/lib/custom_dockerfile.ml#L78-L96, it also installs regular dependencies.

I have rearranged the metrics list as requested.

> Maybe we could report the geometric mean of all the running times as well?

A geometric mean of absolute values? Isn’t the geometric mean usually used on relative evolution numbers?

OlivierNicole:
This repo has been [added](https://bench.ci.dev/ocsigen/js_of_ocaml/) to the current-bench infra and the results for this PR can be checked [here](https://bench.ci.dev/ocsigen/js_of_ocaml/pull/1842/base/master).

bench.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ocaml/opam:debian-ocaml-5.3
+WORKDIR /bench-dir
+RUN sudo apt-get update && \
+    sudo apt-get install -qq -yy --no-install-recommends pkg-config libgmp-dev \
+      wget
+RUN sudo ln -sf /usr/bin/opam-2.1 /usr/bin/opam
```
hhugo, line 6: Why do we use an old version of opam ?

bench.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN wget https://nodejs.org/dist/v22.14.0/node-v22.14.0-linux-x64.tar.xz && \
+    tar xJvf node-v22.14.0-linux-x64.tar.xz && \
+    sudo cp node-v22.14.0-linux-x64/bin/node /usr/bin/node
+COPY --chown=opam:opam ./*.opam ./VERSION ./
+RUN opam exec -- git init && \
+    opam exec -- git add . && \
+    opam exec -- git commit -m dummy && \
+    opam exec -- git tag -a $(cat VERSION)-dev -m dummy
+RUN opam pin -yn --with-version=dev .
+RUN opam install -y --depext-only wasm_of_ocaml-bench && \
```
hhugo, line 18: I don't understand the purpose of wasm_of_ocaml-bench, why don't we install wasm_of_ocaml-compiler instead ?

hhugo:
I don't understand why we need all the logic in the docker file.
Other open PR seem to fail with
```
make: *** No rule to make target 'bench'.  Stop.
docker-run failed with exit-code 2
```
Why isn't it just a matter of added a makefile target ?

bench.Dockerfile
OlivierNicole: Current-bench’s default Dockerfile needed a dedicated opam file for the benchmarks. But now that we have a custom Dockerfile, I can probably drop it.

bench.Dockerfile
OlivierNicole: Good question. I took inspiration from https://github.com/ocaml-multicore/picos/blob/main/bench.Dockerfile which does this. I’ll try without this line.

OlivierNicole:
In the absence of a `bench.Dockerfile` file, by default current-bench will:

- Pin all the `./*.opam` files
- Install their dependencies
- Run `make bench`.

This does not work for us for several reasons:

- The benchmarks require a more recent version of Node than what is currently available in the default Debian image;
- the build of js_of_ocaml and wasm_of_ocaml fails at the `dune subst` step because for some reason, there is no Git version information in the container
- Everything is built in one step, which is not great for caching: `binaryen` is rebuilt at every run.

bench.Dockerfile
OlivierNicole: By default opam is even older (2.0). I made it use 2.3 instead.

bench.Dockerfile
```diff
@@ -15,12 +15,12 @@ RUN opam exec -- git init && \
     opam exec -- git commit -m dummy && \
     opam exec -- git tag -a $(cat VERSION)-dev -m dummy
 RUN opam pin -yn --with-version=dev .
-RUN opam install -y --depext-only wasm_of_ocaml-bench && \
+RUN opam install -y --depext-only wasm_of_ocaml-compiler && \
```
hhugo, line 18: Isn't this line unnecessary?

bench.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    tar xJvf node-v22.14.0-linux-x64.tar.xz && \
+    sudo cp node-v22.14.0-linux-x64/bin/node /usr/bin/node
+COPY --chown=opam:opam ./*.opam ./VERSION ./
+RUN opam exec -- git init && \
+    opam exec -- git add . && \
+    opam exec -- git commit -m dummy && \
+    opam exec -- git tag -a $(cat VERSION)-dev -m dummy
+RUN opam pin -yn --with-version=dev .
+RUN opam install -y --deps-only js_of_ocaml-compiler && \
+    opam install -y binaryen-bin
```
hhugo, line 19:
```suggestion
```
I don't think this is useful, we're already installing wasm_of_ocaml-compiler below

bench.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+COPY --chown=opam:opam . ./
+RUN opam exec -- git add . && \
+    opam exec -- git commit -m dummy && \
+    opam exec -- git tag -a $(cat VERSION)-dev-2 -m dummy
```
hhugo, line 23: We already have the same logic above, can we drop it ?
```suggestion
```

bench.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    sudo cp node-v22.14.0-linux-x64/bin/node /usr/bin/node
+COPY --chown=opam:opam ./*.opam ./VERSION ./
+RUN opam exec -- git init && \
+    opam exec -- git add . && \
+    opam exec -- git commit -m dummy && \
+    opam exec -- git tag -a $(cat VERSION)-dev -m dummy
+RUN opam pin -yn --with-version=dev .
+RUN opam install -y --deps-only js_of_ocaml-compiler && \
+    opam install -y binaryen-bin
+COPY --chown=opam:opam . ./
```
hhugo, line 20: What's the purpose of this ?

bench.Dockerfile
OlivierNicole: Yes, but having the installation of binaryen-bin in a different RUN command allows it to be cached separately. As a consequence, when wasm_of_ocaml-compiler is modified, the long build of binaryen will not be run again.

bench.Dockerfile
OlivierNicole: I need to commit the changes in order for them to be taken into account in the pinned packages. The tag may be optional. See https://github.com/ocsigen/js_of_ocaml/issues/1852 for the reason of the dirty hack.

bench.Dockerfile
OlivierNicole: This is also in order to optimize cache behaviour:

- First only the `*.opam` files are copied. As long as those files do not change, the subsequent steps of installing binaryen and the dependencies of JSOO do not need to be re-run.
- Then we copy the rest of the files and install the rest.

At first I was copying all the files straight away, but then binaryen was rebuilt for every small change, even unrelated.

bench.Dockerfile
hhugo: Should we use `RUN opam install -y --deps-only wasm_of_ocaml-compiler` then ?

hhugo:
I've just merged #1853 , you should be able to remove the git logic

bench.Dockerfile
hhugo: Also, what do you think about using a released binary https://github.com/WebAssembly/binaryen/releases/tag/version_122 instead of recompiling it ?

OlivierNicole: > Should we use `RUN opam install -y --deps-only wasm_of_ocaml-compiler` then ?

That would install js_of_ocaml-compiler, which requires the files to be copied first, which I wanted to avoid for the cache.

Maybe I should add comments to document.



> Also, what do you think about using a released binary https://github.com/WebAssembly/binaryen/releases/tag/version_122 instead of recompiling it ?

I thought about it, but it’s been a bit difficult due to the choice of installing binaryen as an opam package (binaryen-bin). I think we may want to come back on that choice.

OlivierNicole:
I have opened https://github.com/ocsigen/js_of_ocaml/pull/1854 which should unblock the installation of Binaryen as a binary here.

benchmarks/Makefile
```diff
@@ -10,6 +10,14 @@ GRAPHSEFF = time-effects.pdf size-effects.pdf size-bzip2-effects.pdf
 # For full benchs:
 all: _perf graphsnopr graphspr graphseff _noperf
 
+# This target is the one run by `current-bench`,
+# see https://github.com/ocurrent/current-bench
+.PHONY: bench
+bench: __run_wasm
```
hhugo, line 16: Shouldn't this be called bench-wasm ?

benchmarks/run.ml
```diff
@@ -239,107 +258,221 @@ let _ =
...
-  compile_jsoo "--disable compact" code Spec.byte code Spec.js_of_ocaml_compact;
-  compile_jsoo "--disable optcall" code Spec.byte code Spec.js_of_ocaml_call;
-  compile_jsoo ~effects:true "" code Spec.byte code Spec.js_of_ocaml_effects;
   compile "ocamlc -unsafe" src Spec.ml code Spec.byte_unsafe;
   compile "ocamlopt" src Spec.ml code Spec.opt_unsafe;
-  compile_jsoo "" code Spec.byte_unsafe code Spec.js_of_ocaml_unsafe;
+  if wasm then compile_jsoo ~wasm "" code Spec.byte code Spec.wasm_of_ocaml;
+  if js
+  then (
+    compile_jsoo ~wasm "" code Spec.byte code Spec.js_of_ocaml;
```
hhugo, line 294: This doesn't look correct. if both js and wasm are set, this will use `wasm_of_ocaml` instead of `js_of_ocaml` to compile the file.
I think it would be better to rename `compile_jsoo` with something like
```ocaml
val compile_js : backend:[`Jsoo | `Wasmoo] -> ...
```

benchmarks/run.ml
```diff
@@ -239,107 +258,221 @@ let _ =
...
-  compile_jsoo "--enable=use-js-string" code Spec.byte code Spec.js_of_ocaml_js_string;
-  compile_jsoo "--disable inline" code Spec.byte code Spec.js_of_ocaml_inline;
-  compile_jsoo "--disable deadcode" code Spec.byte code Spec.js_of_ocaml_deadcode;
-  compile_jsoo "--disable compact" code Spec.byte code Spec.js_of_ocaml_compact;
-  compile_jsoo "--disable optcall" code Spec.byte code Spec.js_of_ocaml_call;
-  compile_jsoo ~effects:true "" code Spec.byte code Spec.js_of_ocaml_effects;
   compile "ocamlc -unsafe" src Spec.ml code Spec.byte_unsafe;
   compile "ocamlopt" src Spec.ml code Spec.opt_unsafe;
-  compile_jsoo "" code Spec.byte_unsafe code Spec.js_of_ocaml_unsafe;
+  if wasm then compile_jsoo ~wasm "" code Spec.byte code Spec.wasm_of_ocaml;
```
hhugo, line 291: Why don't we take effects into account ?

benchmarks/run.ml
```diff
@@ -239,107 +258,221 @@ let _ =
...
-        ; Some Spec.js_of_ocaml_o3
-        ; Some Spec.js_of_ocaml_js_string
-        ; Some Spec.js_of_ocaml_unsafe
-        ; Some Spec.js_of_ocaml_inline
-        ; Some Spec.js_of_ocaml_deadcode
-        ; Some Spec.js_of_ocaml_compact
-        ; Some Spec.js_of_ocaml_call
-        ; Some Spec.js_of_ocaml_effects
-        ] )
-    else if effects
```
hhugo, line 329: You've dropped this logic. The documentation of `effects` was previously: 
```
only run with and without effect handler support
``` 
which I think meant that we wanted to run a small subset of the runs.

hhugo (commented):
The PR does many things at once
- Setup docker LGTM
- Some refactoring of the benchmarks (e.g introducing Measure). Can we move that to a separate PR ? 
- Add a new output format. Can we move that to a separate PR (or together with the refactoring above) 
- Add support for effects backend. I don't think it is correct. We probably want to be able to bench different backends / bench all backend or none ? Can this be moved to a different PR
- Add support for wasm in benchmarks. Can we have this in a dedicated PR ?

Maybe we could have the following stack of PR
- refactoring and new output format
- wasm support and docker setup
- effects backends

The benchmark scripts have received very little attention so far. Maybe it's a good time for larger refactoring.

run.ml is responsible for producing all data for reports. It has a few flags to produce a subset of data, presumably to speedup the process.
We could image a way to only generate the data we need for a set of reports.

hhugo:
See #1855

hhugo:
I think it would be useful to extend the scope of this to track things like
- run time for jsoo files
- build time for large executable (ocamlc ? / jsoo-toplevel ?) with jsoo and wasmoo
- memory consumption when building theses larges executables

vouillon:
> > Maybe we could report the geometric mean of all the running times as well?
> 
> A geometric mean of absolute values? Isn’t the geometric mean usually used on relative evolution numbers?

The relative evolution of the geometric mean is the geometric mean of the relative evolution numbers. So this provides a way to evaluate the global performance evolution.

OlivierNicole:
I see, it makes sense to me then.

PR #1849, opened 2025-03-03
Fix initialization of standard streams under Windows

### LINKED ISSUE #1845: Getting fstat error since latest release when running JS code on Windows
I recently started getting test failures on a Github Actions job using Windows and JSOO-generated code. I don't have a Windows machine directly available, so I apologize for being unable to provide a minimal reproducible example. 

The exact error is this ([link](https://github.com/johnridesabike/acutis/actions/runs/13456966660/job/37748174480)):

```
node:fs:1644
  handleErrorFromBinding(ctx);
  ^

Error: EINVAL: invalid argument, fstat
    at Object.fstatSync (node:fs:1644:3)
    at new AB (D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:661:12)
    at XO (D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:852:29)
    at y9 (D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:880:21)
    at D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:881:14
    at D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:881:202
    at Object.<anonymous> (D:\a\acutis\acutis\_build\.sandbox\e437599ddd7afc5fce4ce1a65f7b45e0\default\main.cjs:4927:119)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32) {
  fd: 0,
  errno: -4071,
  syscall: 'fstat',
  code: 'EINVAL'
 }

Node.js v18.20.6
```

The line number in the error message seems to point to this code from the JSOO runtime source:

https://github.com/ocsigen/js_of_ocaml/blob/937e84920335b58a5598f222a613f62f781912f7/runtime/js/fs_node.js#L332

Some notes:
- I don't believe the JSOO program in my tests calls any IO functions directly. 
- I didn't modify anything in the test before the test started breaking (the only commits were non-code).
- This does not happen on MacOS or Linux tests.

vouillon:
We are using fstat to test whether the file is seekable. This function  fails on standard streams under Windows with node 18 (and lower). See  https://github.com/libuv/libuv/pull/3811. So, we just assume that the   file is not seekable in case an exception is risen.

Fixes #1845

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 
 ## Bug fixes
 * Runtime: fix path normalization (#1848)
+* Runtime: fix initialization of standard streams under Windows (#1849)
 
 # 6.0.1 (2025-02-07) - Lille
 
--- a/compiler/tests-jsoo/gh1845.t
+++ b/compiler/tests-jsoo/gh1845.t
@@ -1,0 +1,9 @@
+
+  $ echo 'prerr_endline "test"' > test.ml
+
+  $ ocamlc test.ml -o test.bc
+
+  $ dune exec -- js_of_ocaml test.bc -o test.js
+
+  $ node test.js
+  test
--- a/runtime/js/fs_node.js
+++ b/runtime/js/fs_node.js
@@ -392,9 +392,15 @@ class MlNodeFd extends MlFile {
     this.fs = require("node:fs");
     this.fd = fd;
     this.flags = flags;
-    var stats = this.fs.fstatSync(fd);
-    flags.noSeek =
-      stats.isCharacterDevice() || stats.isFIFO() || stats.isSocket();
+    try {
+      var stats = this.fs.fstatSync(fd);
+      flags.noSeek =
+        stats.isCharacterDevice() || stats.isFIFO() || stats.isSocket();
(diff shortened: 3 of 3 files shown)


hhugo:
@vouillon, I was able to reproduce this issue with node 18 but not with node 20

vouillon:
@hhugo How do you reproduce this issue ? Is it just a matter of using an older version of node?

vouillon:
I think this was fixed by this PR: https://github.com/libuv/libuv/pull/3811

hhugo (approved):
Can you add a changelog entry

PR #1854, opened 2025-03-05
Remove dependency of wasm_of_ocaml on binaryen-bin

OlivierNicole:
Depending on an external tool as an opam package is non-standard and makes it harder to set up a CI (#1842), or when Binaryen is installed in some other way. I think we should just remove it.

### DIFF
--- a/manual/wasm_overview.wiki
+++ b/manual/wasm_overview.wiki
@@ -7,6 +7,10 @@ It provides an alternative way to run pure OCaml programs in JavaScript environm
 
 The compiler is provided by the wasm_of_ocaml-compiler package. The <<a_manual chapter="overview" |Js_of_ocaml libraries>> are compatible with this compiler.
 
+== Dependencies
+
+Wasm_of_ocaml depends on [[https://github.com/WebAssembly/binaryen|the Binaryen toolchain]] version 119 or later.
+
 == Installation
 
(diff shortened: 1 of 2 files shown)


vouillon:
If we want to remove this dependency, we need to provide a clear error message when the binaryen tools are not installed, or an old version is installed. People don't read the documentation...

In a CI, we can just do:
```
opam install --fake binaryen-bin
```

vouillon:
Also, making it harder to install `wasm_of_ocaml` may discourage people to try it.

OlivierNicole:
Is there a reason why we don't specify it as a depext?

vouillon:
> Is there a reason why we don't specify it as a depext?

We could do that for some OSs / distributions, indeed, now that Binaryen 119 is not so recent anymore. But Ubuntu 24.10 still only provides version 108. And Github Actions are still using Ubuntu 24.04.

hhugo:
> If we want to remove this dependency, we need to provide a clear error message when the binaryen tools are not installed, or an old version is installed. People don't read the documentation...
> 
> In a CI, we can just do:
> 
> ```
> opam install --fake binaryen-bin
> ```

@OlivierNicole, see https://github.com/OlivierNicole/js_of_ocaml/pull/3/files

OlivierNicole:
Thank you for your input, #1842 has been updated to rather use `opam install --fake`. Closing.

PR #1861, opened 2025-03-09
runtime(js): fix useOptionalChain

smorimoto:
It's already widely used in the runtime and can be unlocked since it's already included in the ECMASCRIPT documentation.


runtime/js/sys.js
```diff
@@ -146,7 +146,7 @@ var caml_argv = (function () {
   var main = "a.out";
   var args = [];
 
-  if (process && process.argv && process.argv.length > 1) {
+  if (process?.argv && process.argv.length > 1) {
```
hhugo, line 149: This can probably be simplified further

smorimoto: Ah, indeed

smorimoto: Fixed

PR #1862, opened 2025-03-09
runtime(js): fix useTemplate

smorimoto:
Template is another one that is already widely used and documented.


hhugo (commented):
I don't like complex expressions inside templates. They makes review harder  in my experience.
I think I would prefer if templates only embed simple identifiers.

hhugo:
@smorimoto, do you plan to do anything with this PR ?

PR #1866, opened 2025-03-10
ppx: upgrade to ppxlib 0.36

anmonteiro:
- Upgrades the JSOO ppxes to ppxlib 0.36 with https://github.com/ocaml-ppx/ppxlib/pull/514
- I realize `ppx_expect` and its transitive deps need to be upgraded first before this change can be considered, but I'm opening this up to unblock some of my PRs in other repos (this builds fine if you don't run the tests)

### DIFF
--- a/ppx/ppx_js/lib_internal/ppx_js_internal.ml
+++ b/ppx/ppx_js/lib_internal/ppx_js_internal.ml
@@ -236,7 +236,6 @@ let invoker ?(extra_types = []) uplift downlift body arguments =
     let args = List.map ~f:ident arguments in
     body args
   in
-  let annotated_ebody = Exp.constraint_ ebody tfunc_res in
   (* Build the function.
      The last arguments is just used to tie all types together.
      It's unused in the implementation.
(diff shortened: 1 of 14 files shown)


hhugo (commented):
Blocked until the CI goes green

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -585,7 +585,14 @@ let preprocess_literal_object mappper fields :
...
           match exp.pexp_desc with
-          | Pexp_fun (label, _, _, body) -> Arg.make ~label () :: create_meth_ty body
+          | Pexp_function (params, _, Pfunction_body body) ->
+              let params =
+                List.filter_map params ~f:(function
+                  | { pparam_desc = Pparam_newtype _; _ } -> None
+                  | { pparam_desc = Pparam_val (label, _, _arg); _ } ->
+                      Some (Arg.make ~label ()))
+              in
+              params @ create_meth_ty body
```
hhugo, line 595: I'm not sure we need to recurse here.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -585,7 +585,14 @@ let preprocess_literal_object mappper fields :
         let body = format_meth (mappper body) in
         let rec create_meth_ty exp =
           match exp.pexp_desc with
-          | Pexp_fun (label, _, _, body) -> Arg.make ~label () :: create_meth_ty body
+          | Pexp_function (params, _, Pfunction_body body) ->
+              let params =
+                List.filter_map params ~f:(function
+                  | { pparam_desc = Pparam_newtype _; _ } -> None
```
hhugo, line 591: I'm not sure that this is what we want. I'll have to look futher

ppx/ppx_js/lib_internal/ppx_js_internal.ml
anmonteiro: @hhugo the reason why I added recursion was for cases like this:

```ocaml
class x =
  object
    method foo a = fun b -> a + b
  end
```

due to https://github.com/ocaml/ocaml/pull/12236 preserving both `Pexp_function` nodes in the AST. I'm not very familiar with how JSOO handles this case, so I trust you to know whether it's unnecessary.

hhugo:
jsoo can now build with ppxlib.0.36. Running tests still requires ppxlib < 0.36

hhugo:
Tests now run with ppxlib.0.36

PR #1867, opened 2025-03-11
Fix function declaration scopes in Js_traverse.rename_variable

### LINKED ISSUE #1865: Minifier crash
The minifier crashes on the following piece of code:
```js
function f (p) {
    return e
    if (p) {
        function e () {
        }
        e ();
    }
}
```
We get the following error:
```
Some variables escaped (#1). Use [--debug js_assign] for more info.
jsoo_minify: You found a bug. Please report it at https://github.com/ocsigen/js_of_ocaml/issues :
Error: File "compiler/lib/js_assign.ml", line 442, characters 5-11: Assertion failed
```

vouillon:
Function declarations are block-scoped, except at toplevel where they are treated like var declarations.

This fixes #1865.

### DIFF
--- a/compiler/lib/js_traverse.mli
+++ b/compiler/lib/js_traverse.mli
@@ -165,14 +165,14 @@ class free : freevar
 
 type scope =
   | Module
+  | Script
   | Lexical_block
   | Fun_block of ident option
 
 class rename_variable : esm:bool -> object ('a)
   inherit mapper
 
-  method update_state :
-    ?toplevel:bool -> scope -> Javascript.ident list -> Javascript.statement_list -> 'a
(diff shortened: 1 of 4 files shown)


compiler/lib/js_traverse.mli
```diff
@@ -171,7 +171,8 @@ type scope =
 class rename_variable : esm:bool -> object ('a)
   inherit mapper
 
-  method update_state : scope -> Javascript.ident list -> Javascript.statement_list -> 'a
+  method update_state :
+    ?toplevel:bool -> scope -> Javascript.ident list -> Javascript.statement_list -> 'a
```
hhugo, line 175: I think we could merge toplevel with the scope data type. We already have `Module` which is somewhat similar.

PR #1870, opened 2025-03-11
Fix assert failure in double translation

### LINKED ISSUE #1868: Double translation: assertion failure
Consider the following piece of code:
```ocaml
exception Nested of exn

let wrap f =
  try f () with exn ->
  let rec unwrap exn =
    match exn with
    | Nested e -> unwrap e
    | _ -> raise exn
  in
  unwrap exn
```
Trying to compile it with double translation results in an assertion failure:
```
$ js_of_ocaml --pretty --effect=double-translation  exn.cmo
js_of_ocaml: You found a bug. Please report it at https://github.com/ocsigen/js_of_ocaml/issues :
Error: File "compiler/lib/generate.ml", line 1950, characters 10-16: Assertion failed
```

vouillon:
We have a pass that removes empty blocks. This can trigger an assertion failure in code generation. It is safe to remove this assertion.

This fixes #1868.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 * Runtime: fix initialization of standard streams under Windows (#1849)
 * Compiler: fix stack overflow issues with double translation (#1869)
 * Compiler: minifier fix (#1867)
+* Compiler: fix assert failure with double translation (#1870)
 
 # 6.0.1 (2025-02-07) - Lille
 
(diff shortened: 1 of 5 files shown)


vouillon:
> @vouillon, can you review the changes I'm made ?

This seems like a lot of work for something that hardly ever happen. Otherwise, this looks correct.

PR #1871, opened 2025-03-12
Make Obj.dup work on floats and boxed integers

vouillon:


hhugo:
Should we add a slice method to `MlNat` and `Ml_Bigarray` ?

vouillon:
> Should we add a slice method to `MlNat` and `Ml_Bigarray` ?

It seems `Obj.dup` is not really supported for bigarrays in the native runtime (we get two bigarrays that share the same buffer, so the buffer gets freed when either is freed). `nat` is pretty much deprecated.

hhugo (approved):

PR #1875, opened 2025-03-12
Improve Link_js performance

OlivierNicole:
This is a pair of patches by @Enoumy, in use in the internal version of JSOO at Jane Street.

### DIFF
--- a/compiler/lib/link_js.ml
+++ b/compiler/lib/link_js.ml
@@ -58,7 +58,19 @@ end = struct
 
   let open_ fname =
     let lines =
-      In_channel.input_all (open_in_bin fname) |> String.split_on_char ~sep:'\n'
+      (* If we are not on 32-bit hardware (where the max string length is too
+         small), read the entire file and split it in lines. This is faster
+         than reading it line by line. On 32 bits, we fall back to a
+         line-by-line read. *)
+      let ic = open_in_bin fname in
+      let x =
+        if Sys.word_size >= 64
+        then
+          really_input_string ic (in_channel_length ic) |> String.split_on_char ~sep:'\n'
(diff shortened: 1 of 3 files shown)


hhugo:
What improvement does that give in practice?

OlivierNicole:
See the first commit message. In addition, when building js_of_ocaml with itself in separate compilation mode, the final linking step falls down from ~550ms on master to ~120ms here.

OlivierNicole:
Erratum: I can’t reproduce the numbers above. After disabling Intel turbo boost on my laptop CPU, I get 2.24 ± 0.022 s on master (559fbce) compared to 2.15 ± 0.024 s on this PR, computed on 20 runs.

hhugo:
I've make small cleanup to the commits. Please make sure you're still happy with the end result.

hhugo (approved):

compiler/lib/link_js.ml
```diff
@@ -37,57 +39,61 @@ module Line_reader : sig
...
+    let lines =
+      (* If we are not on 32-bit hardware (where the max string length is too
+         small), read the entire file and split it in lines. This is faster
+         than reading it line by line. On 32 bits, we fall back to a
+         line-by-line read. *)
+      let ic = open_in_bin fname in
+      let x =
+        if Sys.word_size >= 64
+        then
+          really_input_string ic (in_channel_length ic) |> String.split_on_char ~sep:'\n'
```
hhugo, line 69: For the record, the two branches are not equivalent.
for empty content, `String.split_on_char` returns `[""]`, input_lines return `[]`.
for content with a trailing newline, `String.split_on_char` will return an extra empty line compared to `input_lines`.

PR #1877, opened 2025-03-13
Fix jsoo_mkcmis

vouillon:
The target and the effect mode was not initialized.


hhugo:
We need to add a test for jsoo_mkcmis in the repo

vouillon:
> We need to add a test for jsoo_mkcmis in the repo

I'm not sure what it should look like.

hhugo:
> > We need to add a test for jsoo_mkcmis in the repo
> 
> I'm not sure what it should look like.

Just having the tool run during runtest would be a start.

hhugo:
I've added tests

hhugo (approved):

PR #1881, opened 2025-03-13
Wasm runtime: keep track of backtrace status

vouillon:
Wasm does not support backtraces. But some applications reuse this flag to decide whether to output code locations or not.


runtime/wasm/runtime.js
```diff
@@ -143,6 +143,18 @@
...
+  function getenv(n) {
+    if (isNode && process.env[n] !== undefined) return process.env[n];
+    return globalThis.jsoo_env?.[n];
+  }
+
+  let record_backtrace_flag = 0;
+
+  for (const l of getenv("OCAMLRUNPARAM")?.split(",") || []) {
+    if (l === "b") record_backtrace_flag = 1;
+    if (l.startsWith("b=")) record_backtrace_flag = +!!+l[i].slice(2);
```
hhugo, line 155: Why do we need the additional `+!!` ?

vouillon: To make sure that we value of the flag is a boolean (0 or 1), not an arbitrary number.

hhugo: I think I would prefer a more explicit
`+l[i].slice(2) ? 1 : 0`

hhugo (approved):
Let's squash and merge noce CI is happy.

PR #1882, opened 2025-03-13
Wasm_of_ocaml: rewrite file paths in source maps

vouillon:


compiler/lib/source_map.mli
```diff
@@ -147,6 +147,8 @@ val of_string : ?tmp_buf:Buffer.t -> string -> t
 
 val of_file : ?tmp_buf:Buffer.t -> string -> t
 
+val rewrite_paths : t -> t
```
OlivierNicole, line 150:
```suggestion
val rewrite_paths : t -> t
(** Rewrite paths in source map in accordance with the [BUILD_PATH_PREFIX_MAP] environment
    variable. *)
```

OlivierNicole:
So there is no need to do rewriting at link time, because we assume all source maps have been generated by us?

OlivierNicole (approved):

hhugo:
Maybe we should use the new `rewrite_paths` function in the `json` functions that currently call `rewrite_path` multiple times.

vouillon:
So, I think that the issue was actually that we were just copying the source map without rewriting paths in one case: 
https://github.com/ocsigen/js_of_ocaml/blob/e347ad98e3ae49d3c5c7ec6fac160ca1c2dc00a2/compiler/bin-wasm_of_ocaml/compile.ml#L270-L272

We were also rewriting paths twice during linking.

PR #1883, opened 2025-03-13
Add flag `--empty-sourcemap`

OlivierNicole:
This patch implements the `--empty-sourcemap` flag. It is used to generate an empty sourcemap, which is cheap. Sometimes application authors would like the cheapness of not building sourcemap files to speed up development iterations, but their build pipelines do not handle the case where `.map` files are not produced at all (think, e.g., custom Dune rules depending on `.map` files and passing them to a bundler). In such cases, they would like to still generate the empty `.map` files.

cc @rickyvetter

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@
 * Ppx: explicitly disallow polymorphic method (#1897)
 * Ppx: allow "function" in object literals (#1897)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
+* Compiler: add the `--empty-sourcemap` flag
 * Compiler: static evaluation of more primitives (#1912)
 * Compiler: faster compilation by stopping sooner when optimizations become unproductive (#1939)
 * Runtime: use Dataview to convert between floats and bit representation
--- a/compiler/bin-js_of_ocaml/cmd_arg.ml
+++ b/compiler/bin-js_of_ocaml/cmd_arg.ml
@@ -54,7 +54,7 @@ type t =
   { common : Jsoo_cmdline.Arg.t
   ; (* compile option *)
     profile : Driver.profile option
-  ; source_map : (string option * Source_map.Standard.t) option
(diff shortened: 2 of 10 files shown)


rickyvetter:
Thanks for adding this @OlivierNicole! To add a little context for the motivation here:

Say you are building (foo.bc.js+map) and then you want to pass those artifacts to a new rule, potentially in another directory entirely. This rule tars up all your artifacts for easier deployment. That rule might have deps that look like `../client/foo.bc.js ../client/foo.bc.map (dir ../assets)`. If, in dev, you'd like to disable sourcemaps entirely to be able to iterate faster, you can't do so without changing the rules in both your `client` directory and your deployment directory.

I suspect that this oddity is partially why the dune rules default to inline in `dev` and none in `release` as opposed to any separate sourcemap usage, which would provide better performance characteristics for the actual web page in `dev`. I could imagine a future change which might make the dune defaults here `dev->separate` and `release->empty`.

hhugo (changes requested):
As far as I see, the PR only affects separate compilation link phase. Is this on purpose ?
What about compilation of intermediate cma/cmo to js ? What about whole program compilation ?

rickyvetter:
Intermediate artifact support could not be easily achieved without breaking sourcemaps of shared dependencies in dune. It also has a much smaller effect on incremental rebuild times. If the intent isn't performance, but rather to actually hide a specific library's source maps in all cases, then I think no-sourcemaps is sufficient since intermediate artifacts don't support separate sourcemaps either.

I think this flag should be supported in whole program compilation.

hhugo:
@OlivierNicole, can you rebase and add some tests

OlivierNicole:
I did that, as well as making the flag work for whole program compilation.

All CI jobs seem broken for some independent reason.

OlivierNicole:
Wasm_of_ocaml puts generated sourcemaps in a directory, and IIUC that makes the presence or absence of sourcemaps much easier to handle with build systems such as Dune, so this is not needed.

PR #1884, opened 2025-03-14
Wasm_of_ocaml: do not minify twice the always-included JavaScript

vouillon:


hhugo:
Where is the other minification coming from ?

vouillon:
> Where is the other minification coming from ?

From `Driver.link_and_pack`. But maybe I should move `coloring` to `Driver.full` instead?

hhugo:
> instead

Maybe.

I don't really understand why we do `ouput_js prelude ^ output_js launcher` instead of `output_js (prelude @ launcher)`. Any important reason ?

vouillon:
> I don't really understand why we do `ouput_js prelude ^ output_js launcher` instead of `output_js (prelude @ launcher)`. Any important reason ?

I don't think so. I guess it made some sense at some point, but not any longer.

hhugo (approved):

Issue #1792, opened 2025-01-07
Double translation: lambda lifting improvements


vouillon:
- We don't need to lift all CPS functions. We just need that the code which is going to be duplicated contain no closures.
- Lifted functions which have no free variable don't need to be wrapped.

PR #1885, opened 2025-03-14
Lambda lifting: only lift functions that have free variables

OlivierNicole:
This addresses the second remark in #1792: lifted functions that have no free variables don't need to be wrapped.


hhugo:
What's the status of this ?

OlivierNicole:
It’s a small improvement, currently only used for `--effects=double-translation`, and is ready for review.

OlivierNicole:
We could, it would be a further improvement on the gain that the present state of this PR brings.

PR #1886, opened 2025-03-14
Do more aggressive lambda lifting

OlivierNicole:
With @vouillon we realized that the `Lambda_lifting_simple` pass that is performed by double translation makes some programs significantly faster. We measured roughly a 1.45 speedup on a large (proprietary) Bonsai benchmark. Presumably V8 is much faster with more toplevel functions and less nested closures.

Making this a draft PR for now because:

- It should probably be behind a flag: it makes the generated Javascript further from the source code structure;
- It theoretically results in a quadratic code size.

We probably need to do some measurements before merging. (And maybe with other engines.)

### DIFF
--- a/compiler/bin-js_of_ocaml/cmd_arg.ml
+++ b/compiler/bin-js_of_ocaml/cmd_arg.ml
@@ -79,7 +79,6 @@ type t =
   ; effects : Config.effects_backend
   ; build_config : bool
   ; apply_build_config : string option
-  ; lambda_lift_all : bool
   }
 
 let wrap_with_fun_conv =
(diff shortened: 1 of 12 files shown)


OlivierNicole:
This seems to result in a modest speedup on average, including on `partial_render_table`. Is this the version of the benchmark that has a memory leak?

OlivierNicole:
I re-ran the CI a few times to double-check whether this new pass increases the compilation time. The compilation time seems to oscillate around ±2 %, so I would say it’s in the noise.

hhugo:
You should probably rebase and rerun the bench now that #1935  is merged to check that this PR still improves perfs

hhugo:
@OlivierNicole, there might be some improvement and the size increase seems small. Do you want to rebase the PR one more time and have the feature under a flag ? That way, we can merge instead of letting the PR rot.

compiler/bin-js_of_ocaml/cmd_arg.ml
```diff
@@ -306,6 +307,14 @@ let options =
           None
       & info [ "effects" ] ~docv:"KIND" ~doc)
   in
+  let lambda_lift_all =
```
hhugo: Most of optional optimizations are controlled with a `--enable X` and `--disable X`  argument. Should we use the same mechanism for this one ? See `config.ml`

OlivierNicole: Done.

OlivierNicole:
Done. The CI failure seems unrelated.

PR #1887, opened 2025-03-15
Fix: marshal.js BigStringReader use offset method on bigstring

TyOverby:
`BigStringReader.readstr` and `BigStringReader.readuint8array` were calling an `offset` method on `self`, when it should have been going through `self.s`.

The tests that caught this bug are proprietary and not trivial to public-release, so I'd appreciate some advice on how best to add a regression test for this code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 * Runtime: make Obj.dup work with floats and boxed numbers (#1871)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
 * Runtime: implement BLAKE2b primitives for Wasm (#1873)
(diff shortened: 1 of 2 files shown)


hhugo:
I suspect janestreet to be the sole user of this code. We could consider removing it from the jsoo codebase and let you maintain it yourself. Or you can move some code to this repo with some tests.

TyOverby:
@hhugo I'm happy to move it into [base_bigstring](https://github.com/janestreet/base_bigstring/blob/master/src/runtime.js), but can we coordinate releases so that someone using base_bigstring from opam, but a nightly js_of_ocaml won't have to deal with a long period of breakage where neither package has the function?

hhugo:
Are there any publicly released package that depend on BigStringReader ? I didn't find any

Issue #1779, opened 2024-12-30
[FEATURE REQUEST] Make resuming a continuation more efficient in wasmoo


hhugo:
Following #1765 and #1658

PR #1765, opened 2024-12-10
Make resuming a continuation more efficient

hhugo:
Fix #1658
See corresponding OCaml PR https://github.com/ocaml/ocaml/pull/12735

- [x] some comments needs to be updated.
- [ ] The wasm part has not been optimized

PR #1892, opened 2025-03-17
Wasm runtime: make resuming continuations more efficient

vouillon:
This was done in #1765 for the JavaScript runtime.
See corresponding OCaml PR ocaml/ocaml#12735

fix https://github.com/ocsigen/js_of_ocaml/issues/1779

### DIFF
--- a/runtime/wasm/effect.wat
+++ b/runtime/wasm/effect.wat
@@ -24,6 +24,7 @@
    (import "obj" "caml_fresh_oo_id"
      (func $caml_fresh_oo_id (param (ref eq)) (result (ref eq))))
    (import "obj" "cont_tag" (global $cont_tag i32))
+   (import "obj" "object_tag" (global $object_tag i32))
    (import "stdlib" "caml_named_value"
       (func $caml_named_value (param (ref eq)) (result (ref null eq))))
    (import "fail" "ocaml_exception" (tag $ocaml_exception (param (ref eq))))
@@ -115,20 +116,21 @@
 
    ;; Stack of fibers
 
-   (type $handlers
-      (struct
-         (field $value (ref eq))
-         (field $exn (ref eq))
-         (field $effect (ref eq))))
-
-   (type $generic_fiber (sub (struct (field $handlers (mut (ref $handlers))))))
+   (type $generic_fiber
+      (sub
+         (struct
+            (field $value (mut (ref eq)))
+            (field $exn (mut (ref eq)))
+            (field $effect (mut (ref eq))))))
 
    (type $fiber
       (sub final $generic_fiber
          (struct
-            (field $handlers (mut (ref $handlers)))
-            (field $cont (ref $cont))
-            (field $next (ref null $fiber)))))
+            (field $value (mut (ref eq)))
+            (field $exn (mut (ref eq)))
+            (field $effect (mut (ref eq)))
+            (field $cont (mut (ref $cont)))
+            (field $next (mut (ref null $fiber))))))
 
    (@string $effect_unhandled "Effect.Unhandled")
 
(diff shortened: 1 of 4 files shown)


hhugo:
@OlivierNicole, Would you be able to review this PR ?

hhugo:
I used the deep_state benchmarks and show the following improvement

compile the bench with:
```
WASM_OF_OCAML=true dune build  compiler/tests-jsoo/lib-effects/deep_state.bc.wasm.js --profile with-effects
```

before:
```
$ time node _build/default/compiler/tests-jsoo/lib-effects/deep_state.bc.wasm.js 128 500_000
Running deepstate: depth=128 ops=500000

real	0m6,003s
user	0m5,980s
sys	0m0,033s
```
after:
```
$ time node _build/default/compiler/tests-jsoo/lib-effects/deep_state.bc.wasm.js 128 500_000
Running deepstate: depth=128 ops=500000

real	0m2,359s
user	0m2,342s
sys	0m0,025s
```

stack switching is however very slow
```
$ time node --experimental-wasm-stack-switching _build/default/compiler/tests-jsoo/lib-effects/deep_state.bc.wasm.js 128 500_000
Running deepstate: depth=128 ops=500000

real	1m11,467s
user	1m11,388s
sys	0m0,122s
```

vouillon:
> stack switching is however very slow

Yes, that's expected. It goes through the JavaScript event loop for each stack switch.

hhugo:
> > stack switching is however very slow
> 
> Yes, that's expected. It goes through the JavaScript event loop for each stack switch.

Should the doc/readme say something about this ?

vouillon:
> Should the doc/readme say something about this ?

We have this at the moment:
https://github.com/ocsigen/js_of_ocaml/blob/55c622862c506a2d403d3f1d5d5c6bfda6bf45d7/manual/wasm_overview.wiki?plain=1#L56

hhugo (approved):
I didn't review the wat diff but the CI is happy. @vouillon, I'll let you decide whether to merge now or wait for more review

OlivierNicole:
> @OlivierNicole, Would you be able to review this PR ?

I am swamped for now, but maybe I can start on Friday.

runtime/wasm/effect.wat
```diff
@@ -115,20 +116,21 @@
 
    ;; Stack of fibers
 
-   (type $handlers
-      (struct
-         (field $value (ref eq))
-         (field $exn (ref eq))
-         (field $effect (ref eq))))
-
```
OlivierNicole, line 123: Is there a particular reason for removing this type?

OlivierNicole (commented):
From a high level point of view, I’m not sure to understand why this change affects `%perform` and `%reperform`, while ocaml/ocaml#12735 only affected `%resume`. There was also code shared between perform and reperform (in line with the native and bytecode backends) that this PR seems to remove, and I’m not sure to understand why.

runtime/wasm/effect.wat
vouillon: Since we now update these fields instead of allocation another structure, there is no reason to keep one level of indirection.

vouillon:
You can have a look at #1765 and #1891.

We have some changes to make in perform/reperform since we were storing the fibers in reversed order in the continuation. They were [already stored in the right order in the OCaml runtime](https://github.com/ocaml/ocaml/commit/fc6cacea67bd965bf3765975e7b860d551d824b9). Also, we need to [store the tail of the stack in the continuation](https://github.com/ocaml/ocaml/pull/12309/files#diff-463184d7ba0c7e61cae25a864aaf737689642847d21a94dbbf4c1abe6cfdbc54).

Separating perform and reperform also allows some simplifications regarding unhandled effects: for perform, we can just raise the exception right away, and for reperform, we can resume the continuation with an exception. One no longer needs to put a default handler at the top of the current stack.

OlivierNicole (approved):
Thanks, this looks good to me.

Do we have an effects benchmark to assess the impact of this PR?

vouillon:
> Do we have an effects benchmark to assess the impact of this PR?

@hhugo [did some measurements](https://github.com/ocsigen/js_of_ocaml/pull/1892#issuecomment-2732853198).

OlivierNicole:
Oh, sorry, I had scanned through this message but for some reason thought it was on the other PR (the Javascript runtime one).

PR #1894, opened 2025-03-21
Benchmarks: updates

vouillon:


OlivierNicole (approved):
See also #1895.

benchmarks/report.ml
```diff
@@ -264,6 +264,18 @@ let output_table =
     stats t;
     f t
 
+let rec transpose (l : (string * 'a list) list) : (string * 'a) list list =
+  match l with
+  | (_, _ :: _) :: _ ->
+      List.map ~f:(fun (label, data) -> label, List.hd data) l
+      :: transpose (List.map ~f:(fun (label, data) -> label, List.tl data) l)
+  | (_, []) :: _ | [] -> []
```
OlivierNicole, line 272:
```suggestion
let rec transpose (l : (string * 'a list) list) : (string * 'a) list list =
  match l with
  | (_, _ :: _) :: _ ->
      List.map ~f:(fun (label, data) -> label, List.hd data) l
      :: (transpose [@tailcall]) (List.map ~f:(fun (label, data) -> label, List.tl data) l)
  | (_, []) :: _ | [] -> []
  [@@tail_mod_cons]
```
This will be more efficient, in case we care.

OlivierNicole (approved):

benchmarks/report.ml
vouillon: The function is calling itself recursively once, so we don't really care.

benchmarks/report.ml
```diff
@@ -273,21 +285,44 @@ let current_bench_output
...
+  assert (List.length measures = List.length measure_descs);
+  let summary =
+    let metrics =
+      List.map2 measure_descs measures ~f:(fun desc measures ->
+          let description =
+            Option.value desc.Measure.description ~default:desc.Measure.name
+          in
+          let mean = geometric_mean measures in
+          `Assoc
+            [ "name", `String description
```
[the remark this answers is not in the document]
vouillon: I have added it to the name of the result below: `Microbenchmarks - summary (geometric means)`

PR #1898, opened 2025-03-24
Wasm runtime: support unmarhalling compressed data

vouillon:


hhugo (approved):
I didn't really review the wat changes, but the CI is happy so I'm happy

PR #1899, opened 2025-03-24
More benchmarks

vouillon:


OlivierNicole:
For some reasons the benchmarks results don’t show in the logs.

vouillon:
> For some reasons the benchmarks results don’t show in the logs.

Are you looking at the right place?
https://bench.ci.dev/ocsigen/js_of_ocaml/pull/1899/base/master/benchmark/Wasm_of_ocaml

OlivierNicole:
Apparently not. I was tripped on the fact that clicking on a benchmark set resets the PR selection.

OlivierNicole:
There seems to still be ppxlib-related issues in the benchmarking CI…

OlivierNicole (approved):

vouillon:
Let's wait. I have not seen the result of my latest changes.

OlivierNicole:
(The benchmarking server looks like it’s in a better state now.)

vouillon:
> Some benchmarks results have “?” as a result still.

I think this is because the result or the metric has been renamed.

OlivierNicole:
In that case, feel free to merge.

PR #1900, opened 2025-03-24
Global flow analysis: performance improvements

vouillon:


hhugo:
Do you have numbers to share regarding the improvement?

vouillon:
This makes a large difference for `bedrock2_fiat_crypto.ml`: from 25 seconds down to 3.7 seconds. And the peak memory usage goes down from 1.5 GiB to 1.2 GiB. This does not seem to make much of a difference for other benchmarks.

PR #1904, opened 2025-03-27
Fix bound check for empty float array

vouillon:
A value of type `float array` is an unboxed float array, unless it is empty. The bound checking code was wrongly assuming that it was always an unboxed float array.

### DIFF
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -343,10 +343,17 @@ module Generate (Target : Target_sig.S) = struct
                   x
             | Extern "caml_check_bound_float", [ x; y ] ->
                 seq
-                  (let* cond =
-                     Arith.uge (Value.int_val y) (Memory.float_array_length x)
+                  (let a = Code.Var.fresh () in
+                   let* () = store a x in
+                   let label = label_index context bound_error_pc in
+                   (* If this is not a float array, it must be the
+                      empty array, and the bound check should fail. *)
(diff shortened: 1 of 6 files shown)


OlivierNicole:
Oh, funny!

compiler/lib-wasm/generate.ml
```diff
@@ -343,10 +343,17 @@ module Generate (Target : Target_sig.S) = struct
                   x
             | Extern "caml_check_bound_float", [ x; y ] ->
                 seq
-                  (let* cond =
-                     Arith.uge (Value.int_val y) (Memory.float_array_length x)
+                  (let a = Code.Var.fresh_n "a" in
```
hhugo, line 346:
```suggestion
                  (let a = Code.Var.fresh () in
```

OlivierNicole:
This adds a type test to all float arrays bound checks. Couldn’t this be considered a bug in OCaml?

vouillon:
> This adds a type test to all float arrays bound checks. Couldn’t this be considered a bug in OCaml?

There is no way to know whether an empty array is going to be used with type `float array` or not. So, the only representation that makes sense is of a generic array.

The OCaml runtime does not have to perform a test since the array length is stored the same way for both boxed and unboxed arrays. For us, there is a difference since we omit the tag of unboxed float arrays.

I don't expect this test to costs much. If the Wasm engine (or binaryen) is able to statically determine that it has float array, the test will be optimized away. Otherwise, the cast performed before accessing the value can be optimized away because of this test.

hhugo (approved):

PR #1905, opened 2025-03-27
Small fixes

vouillon:


runtime/wasm/runtime.js
```diff
@@ -152,7 +152,7 @@
 
   for (const l of getenv("OCAMLRUNPARAM")?.split(",") || []) {
     if (l === "b") record_backtrace_flag = 1;
-    if (l.startsWith("b=")) record_backtrace_flag = +l[i].slice(2) ? 1 : 0;
+    if (l.startsWith("b=")) record_backtrace_flag = +l.slice(2) ? 1 : 0;
```
hhugo, line 155: Maybe you should check from free variable a build time.

vouillon: I thought I did. But it seems not...

PR #1907, opened 2025-03-28
Use more precise Wasm types

vouillon:


hhugo:
@vouillon, What should we do with this PR ?

PR #1908, opened 2025-03-28
More precise function type

vouillon:
Casting the environment to its exact type should be slightly faster since we avoid the test for i31.


hhugo:
What should we do about this PR

PR #1912, opened 2025-04-01
Optimization of primitives: some improvements

vouillon:
- evaluate statically more primitives
- add some missing `const` annotations in the JavaScript runtime
- provide mutability information for primitives handled internally when generating Wasm code

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 * Ppx: explicitly disallow polymorphic method (#1897)
 * Ppx: allow "function" in object literals (#1897)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
+* Compiler: static evaluation of more primitives (#1912)
 
 ## Bug fixes
 * Compiler: fix stack overflow issues with double translation (#1869)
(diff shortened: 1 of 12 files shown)


compiler/lib/eval.ml
```diff
@@ -119,22 +218,131 @@ let eval_prim x =
...
+      | "caml_int32_sub", _ -> int32_binop l Int32.sub
+      | "caml_int32_mul", _ -> int32_binop l Int32.mul
+      | "caml_int32_and", _ -> int32_binop l Int32.logand
+      | "caml_int32_or", _ -> int32_binop l Int32.logor
+      | "caml_int32_xor", _ -> int32_binop l Int32.logxor
+      | "caml_int32_div", [ _; Int32 i ] when not (Int32.equal i 0l) ->
+          int32_binop l Int32.div
+      | "caml_int32_mod", [ _; Int32 i ] when not (Int32.equal i 0l) ->
+          int32_binop l Int32.rem
+      | "caml_int32_shift_left", _ -> int32_shiftop l Int32.shift_left
```
hhugo, line 280: Many of theses are currently only meaningful to the wasm backend. There reason is that the js backend alias some int32/nativeint primitives to their int counterpart.
I find this state quite confusing so I'm experimenting with #1915.

hhugo:
I've rebased this on top of master.

hhugo (approved):

PR #1915, opened 2025-04-03
Don't noop conversion between ints/float with the same runtime repr.

hhugo:
Previously, primitives taking float arguments
could receive Int/Int32/NativeInt/Float arg


hhugo:
Experimentation based on this comment https://github.com/ocsigen/js_of_ocaml/pull/1649#discussion_r1699175384

hhugo:
With the current PR, compiling ocamlc result in the following diff. Two `4` become `4.` 

```diff
$ diff master.js pr.js 
35587c35587
<          Math.floor(v14535 * 4 / v14534 * 1000. + 0.5) / 1000.),
---
>          Math.floor(v14535 * 4. / v14534 * 1000. + 0.5) / 1000.),
35595c35595
<     var v14543 = [0, v14542.length - 2 | 0], v14545 = v14544 * 4;
---
>     var v14543 = [0, v14542.length - 2 | 0], v14545 = v14544 * 4.;
```

compiler/lib/code.ml
```diff
@@ -859,11 +858,7 @@ let invariant { blocks; start; _ } =
         Var.ISet.add defs x)
     in
     let check_constant = function
-      | NativeInt _ | Int32 _ ->
-          assert (
-            match target with
-            | `Wasm -> true
-            | _ -> false)
+      | NativeInt _ | Int32 _ -> ()
```
vouillon, line 861: You can remove `check_constant`.

compiler/lib/generate.ml
```diff
@@ -1457,15 +1456,21 @@ let rec translate_expr ctx loc x e level : (_ * J.statement_list) Expr_builder.t
...
-            | Some f -> f l ctx loc
-            | None ->
-                if String.is_prefix name ~prefix:"%"
-                then failwith (Printf.sprintf "Unresolved internal primitive: %s" name);
-                let prim = Share.get_prim (runtime_fun ctx) name ctx.Ctx.share in
-                let* () = info ~need_loc:true (kind (Primitive.kind name)) in
-                let* args = list_map (fun x -> access' ~ctx x) l in
-                return (J.call prim args loc))
+            match name, l with
+            | "%identity-ints-repr", [ x ] ->
```
vouillon, line 1460: I would use ``register_un_prim "%identity-ints-repr" `Pure (fun cx _ -> cx)``.

compiler/lib/eval.ml
```diff
@@ -86,9 +82,21 @@ let eval_prim x =
   | Eq, [ Int i; Int j ] -> bool Targetint.(i = j)
   | Neq, [ Int i; Int j ] -> bool Targetint.(i <> j)
   | Ult, [ Int i; Int j ] -> bool (Targetint.(j < zero) || Targetint.(i < j))
-  | Extern name, l -> (
-      let name = Primitive.resolve name in
+  | Extern name', l -> (
+      let name = Primitive.resolve name' in
       match name, l with
```
vouillon, line 87: Since these operations make sense in Wasm as well, maybe you should do:
```ocaml
match if String.equal name "%identity-ints-repr" then name' else name with
```

compiler/lib/primitive.ml
```diff
@@ -76,7 +76,11 @@ let has_arity nm a = try Hashtbl.find arities (resolve nm) = a with Not_found ->
 
 let is_pure nm =
   match nm with
-  | "%identity" | "%direct_int_div" | "%direct_int_mod" | "%direct_int_mul" -> true
+  | "%identity"
+  | "%identity-ints-repr"
```
vouillon, line 80: Not needed if you use `register_un_prim`.

vouillon:
> The issue now is that wasmoo uses information from the js runtime but some `//Alias:` annotations are not wasmoo compatible. @vouillon, do you have any plan for another mechanism to keep "purity" information for the wasm runtime ?

I did not plan to do that. One issue is that if we start duplicating this information, we will have to make sure to keep things in sync. The other issue is that I don't know where else to put this information.

runtime/wasm/jslib.wat
```diff
@@ -414,47 +414,6 @@
          (call $wrap_fun_arguments
             (call $wrap_callback_strict (i32.const 1) (local.get 0)))))
 
-   (func (export "caml_callback")
```
hhugo, line 417: I think this was deadcode, because generate.ml was aliasing `caml_callback` to `caml_trampoline`. @vouillon, can you confirm this is correct ?

compiler/lib-wasm/generate.ml
```diff
@@ -661,7 +661,11 @@ module Generate (Target : Target_sig.S) = struct
                 in
                 Memory.allocate ~tag:0 ~deadcode_sentinal:ctx.deadcode_sentinal l
             | Extern name, l -> (
-                let name = Primitive.resolve name in
+                let name =
+                  match name with
+                  | "caml_alloc_stack" when effects_cps () -> "caml_cps_alloc_stack"
```
hhugo, line 666: @vouillon, could this be moved to the wasm runtime somehow using conditional ? `(@if ..`

runtime/wasm/jslib.wat
vouillon: This function is called from `runtime/wasm/runtime.js`.

It might be clearer to use `caml_trampoline` in `compiler/lib/effects.ml`, and alias it to `caml_callback` when generating JavaScript code.

compiler/lib-wasm/generate.ml
vouillon: See #1920

runtime/wasm/jslib.wat
vouillon: See #1922

hhugo:
This PR is now includes
- part of #1912 (for more static evaluation)
- #1922

With this PR, wasmoo no longer resolves aliases during codegen. Aliases are still used to compute purity though.

I think there is still an issue with #1922, can you take a look ?

compiler/lib/eval.ml
```diff
@@ -66,18 +63,111 @@ let float_binop (l : constant list) (f : float -> float -> float) : constant opt
...
+
+  let atanh l = None
+
+  let cbrt l = None
+
+  let erf l = None
+
+  let erfc l = None
+end
+[@@if ocaml_version < (4, 13, 0)]
```
vouillon, line 169: We no longer need this

compiler/lib-wasm/generate.ml
```diff
@@ -1195,14 +1194,7 @@ module Generate (Target : Target_sig.S) = struct
     List.rev_append context.other_fields (imports @ constant_data)
 end
 
-let init () =
-  let l =
-    [ "caml_callback", "caml_trampoline"
-    ; "caml_make_array", "caml_array_of_uniform_array"
-    ]
-  in
-  Primitive.register "caml_array_of_uniform_array" `Mutable None None;
```
vouillon, line 1204: We no longer specify anywhere that `caml_array_of_uniform_array` / `caml_make_array` is `Mutable`.

hhugo: reintroduced

compiler/lib/eval.ml
hhugo: removed

vouillon (approved):

PR #1916, opened 2025-04-04
More extensive Wasm AST

vouillon:
Plus small type-related fix in the Wasm linker

### DIFF
--- a/compiler/lib-wasm/wasm_link.ml
+++ b/compiler/lib-wasm/wasm_link.ml
@@ -697,6 +697,12 @@ module Read = struct
         let pos = st.type_index_count in
         let pos' = add_rectype types ty in
         let count = Array.length ty in
+        let len = Array.length st.type_mapping in
+        if pos + count > len
+        then (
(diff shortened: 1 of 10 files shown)


compiler/lib-wasm/wasm_link.ml
```diff
@@ -697,6 +697,12 @@ module Read = struct
         let pos = st.type_index_count in
         let pos' = add_rectype types ty in
         let count = Array.length ty in
+        while pos + count > Array.length st.type_mapping do
+          let l = Array.length st.type_mapping in
+          let m = Array.make (l + (l / 20) + 1) 0 in
```
hhugo, line 702: We only need to re-allocate the array once
```
if not_enough_space
then
  let new_size = ref .. in
  while .. do ... done;
  let m = Array.make !new_size 0
  Array.blit ~src:st.type_mapping ~src_pos:0 ~dst:m ~dst_pos:0 ~len:l;
  st.type_mapping <- m
```it

hhugo (approved):
LGTM

Issue #1788, opened 2025-01-06
Split JSPI and Cps effect runtime to improve readability 


hhugo:
They currently both live in effect.wat with similar names.
- caml_alloc_stack vs caml_cps_alloc_stack
- %perform vs caml_perform_effect
- ...

We could move jspi and cps part in their own file together with their helpers.
@vouillon, what do you think ?

I also wonder if making names more symmetric could help

PR #1920, opened 2025-04-07
Use a specific runtime for CPS-based effects

vouillon:


hhugo:
Probably enough to close https://github.com/ocsigen/js_of_ocaml/issues/1788

PR #1921, opened 2025-04-07
Compiler: improved compilation of switches

vouillon:


compiler/lib-wasm/generate.ml
```diff
@@ -1225,9 +1258,36 @@ let fix_switch_branches p =
       l
   in
   Addr.Map.iter
-    (fun _ block ->
+    (fun pc block ->
       match block.branch with
-      | Switch (_, l) -> fix_branches l
+      | Switch (x, l) -> (
+          match find_outlier_index l with
+          | `All_equals ->
```
hhugo, line 1265: This optimization should already be handled by `Eval.eval_branch`. Maybe the `Distinguished` case should be performed there as well ..

vouillon: `Eval.eval_branch` only deals with the case where `x` is known.

vouillon: I moved it earlier. jsoo already optimizes switches. So, this does not make much of a difference.  The code is simpler, though.

compiler/lib/specialize.ml
```diff
@@ -117,3 +117,60 @@ let specialize_instrs ~function_arity p =
...
+              | `All_equals -> Addr.Map.add pc { block with branch = Branch l.(0) } blocks
+              | `Distinguished i ->
+                  let block =
+                    let v = Var.fresh () in
+                    let c = Var.fresh () in
+                    { block with
+                      body =
+                        block.body
+                        @ [ Let (v, Constant (Int (Targetint.of_int_exn i)))
+                          ; Let (c, Prim (Eq, [ Pv x; Pv v ]))
```
hhugo, line 166: You could use `Pc (Int ..)`

compiler/lib/specialize.ml
```diff
@@ -117,3 +117,60 @@ let specialize_instrs ~function_arity p =
   { p with blocks; free_pc }
 
 let f = specialize_instrs
+
+(***)
+
+(* For switches, at this point, we know that this it is sufficient to
+   check the [pc]. *)
```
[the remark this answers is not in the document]
vouillon: For a given pc, they are all the same:
https://github.com/ocsigen/js_of_ocaml/blob/967b503335915cb226ecd45aaff09eeeeeaa1776/compiler/lib/parse_bytecode.ml#L866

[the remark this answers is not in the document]
hhugo: We should probably add this to the invariant function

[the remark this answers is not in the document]
vouillon: `Effects.remove_empty_blocks` can break this invariant, though.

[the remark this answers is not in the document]
hhugo: Which invariant ? that for a given pc, we always have the same args/params ?

compiler/lib/specialize.ml
```diff
@@ -117,3 +117,101 @@ let specialize_instrs ~function_arity p =
...
+          then `Distinguished i
+          else `Splitted i
+      | `Distinguished j ->
+          if equal a0 arr.(j)
+          then
+            match find a0 j with
+            | `All_equals ->
+                if j = i + 1 then `Distinguished i else `Splitted_shifted (i, j)
+            | `Distinguished _ -> `Many_cases
+          else `Many_cases)
```
vouillon, line 154: I think you can skip the test `equal a0 arr.(j)` since `find a0 j` will return `Distinguished j` if the test is false:
```suggestion
            match find a0 j with
            | `All_equals ->
                if j = i + 1 then `Distinguished i else `Splitted_shifted (i, j)
            | `Distinguished _ -> `Many_cases
          )
```

hhugo (approved):

hhugo:
@vouillon, I'll let you merge if you're happy with the current state

PR #1922, opened 2025-04-07
Effects: introduce primitive caml_cps_trampoline

vouillon:
This is a better name for the primitive that sets up a trampoline to switch to CPS.

### DIFF
--- a/runtime/js/jslib.js
+++ b/runtime/js/jslib.js
@@ -86,6 +86,7 @@ var caml_callback = caml_call_gen;
 //If: !doubletranslate
 //Requires: caml_stack_depth, caml_call_gen, caml_wrap_exception
 //Requires: caml_current_stack
+//Alias: caml_cps_trampoline
 function caml_callback(f, args) {
   var saved_stack_depth = caml_stack_depth;
   var saved_current_stack = caml_current_stack;
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -1186,7 +1186,6 @@ let init () =
   let l =
     [ "caml_ensure_stack_capacity", "%identity"
     ; "caml_process_pending_actions_with_root", "%identity"
-    ; "caml_callback", "caml_trampoline"
     ; "caml_make_array", "caml_array_of_uniform_array"
     ]
   in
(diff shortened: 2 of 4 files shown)


runtime/js/jslib.js
```diff
@@ -86,6 +86,7 @@ var caml_callback = caml_call_gen;
 //If: !doubletranslate
 //Requires: caml_stack_depth, caml_call_gen, caml_wrap_exception
 //Requires: caml_current_stack
+//Alias: caml_cps_trampoline
```
hhugo, line 89: I think this make the CI fail because wasmoo will "resolve" `caml_cps_trampoline` to `caml_callback`

hhugo:
The following diff is missing from the PR. It would be nice if we could check that import/export are coherent instead of relying on bad error reporting such as
```
File "compiler/tests-wasm_of_ocaml/dune", line 2, characters 8-12:
2 |  (names gh38 gh46 gh107 gh112 gh1904)
            ^^^^
(cd _build/default/compiler/tests-wasm_of_ocaml && ../../.bin/node gh38.bc.wasm.js)
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

[TypeError: WebAssembly.instantiate(): Import #110 "effect": module is not an object or function]
``` 
```diff
 (func $caml_is_continuation (param (ref eq)) (result i32)))
(@if (= effects "cps")
(@then
-   (import "effect" "caml_trampoline"
-      (func $caml_trampoline (param (ref eq) (ref eq)) (result (ref eq))))
+   (import "effect" "caml_cps_trampoline"
+      (func $caml_cps_trampoline (param (ref eq) (ref eq)) (result (ref eq))))
))


@@ -460,14 +460,14 @@
(@then
   (func $caml_callback_1 (export "caml_callback_1")
      (param $f (ref eq)) (param $x (ref eq)) (result (ref eq))
-      (return_call $caml_trampoline
+      (return_call $caml_cps_trampoline
         (local.get $f)
         (array.new_fixed $block 2 (ref.i31 (i32.const 0)) (local.get $x))))

   (func (export "caml_callback_2")
      (param $f (ref eq)) (param $x (ref eq)) (param $y (ref eq))
      (result (ref eq))
-      (return_call $caml_trampoline
+      (return_call $caml_cps_trampoline
         (local.get $f)
         (array.new_fixed $block 3 (ref.i31 (i32.const 0))
           (local.get $x) (local.get $y))))
```

vouillon:
> The following diff is missing from the PR.

That's because this code was added in master since the PR was made.

> It would be nice if we could check that import/export are coherent instead of relying on bad error reporting such as
[...]

See #1931

hhugo:
included in #1915

PR #1927, opened 2025-04-09
Benchmarks: move version-specific set-up from docker file to makefiles

vouillon:


OlivierNicole:
Note to myself: from @vouillon's explanations, my understanding is that we hope that building a number of executables will not be put in the cache, since the invocation of `make` in not in the `bench.Dockerfile`. The CI should not be slowed down by this because Docker is unable to cache builds that depend on external source code.

Ideally, we would have a way for Docker to know that these benchmarks’ code hasn’t changed, and to reuse their build output.

OlivierNicole (approved):

PR #1931, opened 2025-04-10
Runtime compilation: fail if importing from unexpected modules

vouillon:


hhugo:
> With this PR, you get the following error message:
> 
> ```
> The runtime contains unknown imports:
>   effect caml_trampoline23
> ```

Cool, I was testing with an older commit which didn't `exit 2` in case of errors (1c298823006bccc89c84448b5a697d8e9c8cea18)

PR #1934, opened 2025-04-12
 Compiler: remove empty blocks

hhugo:
Removing empty blocks seems to improve codegen a bit.


vouillon:
I think you can just change this line: 
https://github.com/ocsigen/js_of_ocaml/blob/46f55cfe045cf817340b1e4d2ab1fc0c1e676388/compiler/lib-wasm/generate.ml#L1241

hhugo:
@vouillon, could you approve the PR if you think it should be merged in the current state or comment if you have any concerns.

PR #1935, opened 2025-04-14
Rewrite inlining pass

vouillon:


compiler/lib/inline.ml
```diff
@@ -18,131 +18,247 @@
...
+  cache
+    info.cont
+    info.init_code
+    (sum
+    @@ fun { body; _ } ->
+    List.fold_left
+      ~f:(fun n i ->
+        match i with
+        | Let (_, (Closure _ | Field _ | Constant _ | Block _)) -> n + 1
+        | _ -> n)
```
hhugo, line 147: It would be easier to review if ignored instructions were listed explicitly

compiler/lib/code.ml
```diff
@@ -780,6 +784,31 @@ let fold_closures_outermost_first { start; blocks; _ } f accu =
...
+      pc
+      blocks
+      accu
+  in
+  let accu = visit None blocks start f accu in
+  f None None [] (start, []) accu
+
+let rec last_instr l =
+  match l with
+  | [] -> None
```
hhugo, line 808:
```suggestion
  | [] | [ Event _ ] -> None
```

compiler/lib/specialize.ml
```diff
@@ -54,7 +54,18 @@ let add_event loc instrs =
   | Some loc -> Event loc :: instrs
   | None -> instrs
 
-let specialize_instr function_arity ((acc, free_pc, extra), loc) i =
+let rec args_equal xs ys =
+  match xs, ys with
+  | [], [] -> true
+  | x :: xs, Pv y :: ys -> Code.Var.compare x y = 0 && args_equal xs ys
```
hhugo, line 60: `Code.Var.equal`

compiler/lib/inline.ml
```diff
@@ -21,129 +21,457 @@
...
+  { f : Var.t
+  ; params : Var.t list
+  ; cont : Code.cont
+  ; enclosing_function : Var.t option
+  ; recursive : bool
+  ; loops : bool cache
+  ; body_size : int cache
+  ; full_size : int cache
+  ; closure_count : int cache
+  ; init_code : int cache
```
hhugo, line 175: Could you add a comment to some of these fields explain what there are for ? 
At least for `init_code`, `interesting_params`, `full_size`

compiler/lib/inline.ml
```diff
@@ -21,129 +21,457 @@
...
+   allow other optimizations: direct function calls, more precise dead
+   code elimination, ...
+*)
+let functor_like ~context info =
+  (context.aggressive || body_size ~context info <= 15)
+  && (not info.recursive)
+  && (not (contains_loop ~context info))
+  && returns_a_block ~context info
+  && count_init_code ~context info * 2 > body_size ~context info
+  && full_size ~context info - body_size ~context info <= 20 * closure_count ~context info
```
hhugo, line 332: maybe add a small comment explaining the two heuristics above , `init_code * 2 > body_size` and `full_size <= 20 * closure_count`

hhugo:
I've pushed a fixup to the testsuite.
We should check how this PR affects functor heavy programs (something using core maybe).
@TyOverby could you tests this PR on your side ?

hhugo:
We need a changelog entry

hhugo:
I'm not certain I read the benchmark correctly.
It seems that partial render table sees a code size increase of 10%, memory increase of ~50%, compilation time increase of 30% for not runtime improvement.

hhugo:
Maybe we can wait for https://github.com/ocsigen/js_of_ocaml/pull/1962 to get better measurements.

compiler/lib/inline.ml
hhugo: @vouillon, just making sure you saw this comment

hhugo:
We don't have the latest benchmark.
The last result we have shows a runtime regression for ocamlc (maybe some noise ?).

With this PR, we seem to double the time spent in inline. We can probably live with that.

hhugo:
@vouillon, we need to investigate the macos failure.
Can we do anything to reduce compilation time with `--opt 1`  ?

@TyOverby, gentle ping on testing this PR.

This PR definitely need a changelog entry

hhugo:
> @vouillon, we need to investigate the macos failure.

It could be an issue with node 24.

vouillon:
> @vouillon, we need to investigate the macos failure.

The test pass a lazy list to a function. This value is unfolded into something pretty large. The function is no longer inlined (we avoid inlining functions containing loops at toplevel since we expect that the toplevel code will not get optimized). Since node retains function parameters, the value is not garbage-collected and node eventually runs out of memory.

vouillon:
I need to go through this code again and add some comments.
I'll see whether I can reduce compilation time with --opt 1.

hhugo:
> > @vouillon, we need to investigate the macos failure.
> 
> The test pass a lazy list to a function. This value is unfolded into something pretty large. The function is no longer inlined (we avoid inlining functions containing loops at toplevel since we expect that the toplevel code will not get optimized). Since node retains function parameters, the value is not garbage-collected and node eventually runs out of memory.

Downgrading to node 23 fixes the CI

hhugo:
There is a large increase for maxresident bewteen node 23 and node 24
node 23
```
$ /bin/time node _build/default/compiler/tests-ocaml/match-exception/streams.bc.js
iter_stream with handler case (match) is tail recursive
3.79user 0.52system 0:04.00elapsed 107%CPU (0avgtext+0avgdata 2077648maxresident)k
0inputs+0outputs (0major+510186minor)pagefaults 0swaps
```
node 24
```
$ /bin/time node _build/default/compiler/tests-ocaml/match-exception/streams.bc.js
iter_stream with handler case (match) is tail recursive
3.77user 0.92system 0:02.89elapsed 162%CPU (0avgtext+0avgdata 3259072maxresident)k
0inputs+0outputs (0major+825840minor)pagefaults 0swaps
```

TyOverby:
Apologies for the delay; I didn't see this thread for a while.  We should have some test and benchmark results ready for you next week.

TyOverby:
out of curiosity, what was the osx / node-24 issue?  consuming too much memory?

hhugo:
> Apologies for the delay; I didn't see this thread for a while. We should have some test and benchmark results ready for you next week.

Many improvements landed on master in the past few days. It would be nice to test both base and tip of the PR so we can understand the impact of the PR alone

vouillon:
> out of curiosity, what was the osx / node-24 issue? consuming too much memory?

@TyOverby See [my comment above](https://github.com/ocsigen/js_of_ocaml/pull/1935#issuecomment-2865840364).

vouillon:
> I'm not certain I read the benchmark correctly. It seems that partial render table sees a code size increase of 10%, memory increase of ~50%, compilation time increase of 30% for not runtime improvement.

Right, the aggressive inlining of functors does not really seem to result into any runtime improvement with js_of_ocaml. So it is not enabled only with wasm_of_ocaml.

compiler/lib/inline.ml
vouillon: I have added a comment to the functions that fill these fields.

hhugo:
I've pushed commits to only inline (small) functors in o3 with jsoo. Let's wait for the benchmarks

hhugo:
fannkuch_redux and fft seem to take longer now. Can you take a look ? Compilation time increase everywhere but I guess we could live with that given recent improvement everywhere else

vouillon:
For `fft`, it's because a function no longer gets inline because I have reduced the inlining limit from 200  down to 150.
For `fankuch_redux`, the function `fannkuch` is no longer inlined at toplevel, so it not optimized with the assumption that `n = 10`.
```
  let n = 10 in
  let _maxflips, _checksum = fannkuch n in
```
Inlining small functions make a significant difference for `raytrace`.

hhugo:
> For `fft`, it's because a function no longer gets inline because I have reduced the inlining limit from 200 down to 150. For `fankuch_redux`, the function `fannkuch` is no longer inlined at toplevel, so it not optimized with the assumption that `n = 10`.
> 
> ```
>   let n = 10 in
>   let _maxflips, _checksum = fannkuch n in
> ```
> 
> Inlining small functions make a significant difference for `raytrace`.

Are you ok to merge in the current state ?

hhugo:
> Apologies for the delay; I didn't see this thread for a while. We should have some test and benchmark results ready for you next week.

@TyOverby, any update on this ?

vouillon:
> Are you ok to merge in the current state ?

I would prefer to wait for some feedbacks from Ty.

rickyvetter:
We were able to run pull this in internally and performance looks very good! Substantially faster and more consistent on PRT and our other internal benchmarks. For Bonsai benchmarks we are seeing 50%-80% reduction in benchmarking times. Binary size looks like <1% increase in separate compilation and 0-2% in whole program. There are a couple outlier programs that increase in the 10-16% range.

We've reached out about a miscompilation issue on Slack and initially we believe it's unrelated to this PR directly, but it looks like applying this patch actually causes a very similar miscompilation in a program that didn't have it before. This one new case is the only test we have failing, and I suspect that if we resolve the minimal repro for the original issue, then we might also see how to resolve this new instance in this PR?

TyOverby:
For the bonsai benchmarks, I suspect that the large improvements are due to the inlining-related memory leak being resolved by this PR

hhugo:
> We were able to run pull this in internally and performance looks very good! Substantially faster and more consistent on PRT and our other internal benchmarks. For Bonsai benchmarks we are seeing 50%-80% reduction in benchmarking times. Binary size looks like <1% increase in separate compilation and 0-2% in whole program. There are a couple outlier programs that increase in the 10-16% range.
> 
> We've reached out about a miscompilation issue on Slack and initially we believe it's unrelated to this PR directly, but it looks like applying this patch actually causes a very similar miscompilation in a program that didn't have it before. This one new case is the only test we have failing, and I suspect that if we resolve the minimal repro for the original issue, then we might also see how to resolve this new instance in this PR?

Anything to say on compilation times?

hhugo:
Let's merge and move from there.

PR #1937, opened 2025-04-16
Modernize runtime

hhugo:


runtime/js/ieee_754.js
```diff
@@ -17,74 +17,35 @@
...
-  var k = Math.pow(2, 24);
-  var r3 = x | 0;
-  x = (x - r3) * k;
-  var r2 = x | 0;
-  x = (x - r2) * k;
-  var r1 = x | 0;
-  r3 = (r3 & 0xf) | sign | (exp << 4);
+  caml_bufferf64[0] = x;
+  var lo32 = caml_bufferi32[0];
+  var hi32 = caml_bufferi32[1];
```
vouillon, line 36: This assumes a little endian architecture. It would be more portable to use a DataView:
```js
  var buffer = new ArrayBuffer(8);
  var view = new DataView(buffer);
  view.setFloat64(0, x, true); // little endian
  var lo32 = view.getUint32(0, true);
  var hi32 = view.getUint32(4, true);
```

hhugo:
following https://github.com/ocsigen/js_of_ocaml/pull/1912/commits/06e456937ae8960d5a49cbe156ea0aa324161b4a, this PR also preserves the payload of nan

vouillon:
I'm not sure it is worth it preserving the payload of NaN. Safari and Firefox use NaN boxing, so they have a single NaN values, the others being used internally to encode pointers.

hhugo:
> I'm not sure it is worth it preserving the payload of NaN. Safari and Firefox use NaN boxing, so they have a single NaN values, the others being used internally to encode pointers.

Yeah, this was mostly an experiment, I've removed this change.

PR #1938, opened 2025-04-16
Fix performance issue with Structure.is_merge_node

vouillon:
`Structure.is_merge_node` is called for many successors of a node and was linear in the number of predecessor, which could yield a quadratic behavior during Wasm code generation, especially after #1934 which removes some intermediate blocks (see [Fiat-Crypto benchmark](https://bench.ci.dev/ocsigen/js_of_ocaml/pull/1934/base/master/benchmark/Wasm_of_ocaml?worker=autumn&image=bench.Dockerfile)).


hhugo:
How does this affect perf compared to pre-1934 ?

vouillon:
This does not seem to make any difference.

PR #1939, opened 2025-04-17
Make optimized compilation terminate sooner

vouillon:


compiler/lib/code.ml
```diff
@@ -787,12 +787,32 @@ let eq p1 p2 =
...
-         with Not_found -> false)
+         match Addr.Map.find pc p2.blocks with
+         | exception Not_found -> false
+         | block2 ->
+             List.equal ~eq:Var.equal block1.params block2.params
+             && Poly.(block1.branch = block2.branch)
+             && List.equal
+                  ~eq:(fun i i' ->
+                    match i, i' with
+                    | Let (x, Constant (Float f)), Let (x', Constant (Float f')) ->
```
hhugo, line 798: - `constant` is recursive and may contain floats under `Tuple`.
- I was wondering whether we should use the bit representation  of float in the IR. to avoid such thing to happen again in the future.

vouillon: - Oh, indeed!
- I tried that. What to you think of it?

hhugo: I like the new state

compiler/lib/code.ml
```diff
@@ -783,18 +789,13 @@ let fold_closures_outermost_first { start; blocks; _ } f accu =
 let eq p1 p2 =
   p1.start = p2.start
   && Addr.Map.cardinal p1.blocks = Addr.Map.cardinal p2.blocks
```
hhugo, line 791: Is it useful to check  Addr.Map.cardinal first ? (now that the code uses `Map.equal`)

vouillon: Probably not

hhugo: Can you do the change and merge the PR ?

PR #1942, opened 2025-04-22
Update and clarify wasm_of_ocaml documentation

OlivierNicole:
This proposes to update wasm_of_ocaml documentation in several ways:

- Usage and installation instructions are removed from `README_wasm_of_ocaml.md` and a pointer to the relevant manual page is added instead. Several of these instructions were out of date. Similarly, the “Implementation status” section is removed in favor of the clearer “Supported features” section in the linked manual page.
- In the menu, the link “Binding JS libraries” is renamed to “Writing Wasm primitives”, which is the actual title of the page, and seems clearer to me. Also, a duplicate link is removed.
- In the manual, some superfluous indentations are removed (unless I’m missing something about the wikidoc format) and a subsection is added with a few explanations about primitives, and a link to the relevant page.


vouillon:
Thanks!

PR #1943, opened 2025-04-22
Compiler: make tailcall optim more robust

hhugo:
We introduce an intermediate block at the beginning of the closure (if necessary) and jump back to it when we see a tailcall.

With this, we can:
- remove the "skip_param" argument from "Flow.f"
- simplify the O3 profile. Well, O3 is now identical to O2

### DIFF
--- a/compiler/lib/tailcall.ml
+++ b/compiler/lib/tailcall.ml
@@ -1,3 +1,4 @@
+1
 (* Js_of_ocaml compiler
  * http://www.ocsigen.org/js_of_ocaml/
  * Copyright (C) 2010 Jérôme Vouillon
@@ -17,6 +18,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
+
 open! Stdlib
 
 let times = Debug.find "times"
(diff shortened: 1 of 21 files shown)


hhugo:
rebased on top of #1947

hhugo:
Rebased. @vouillon, can you take a look ?

compiler/tests-compiler/gh1007.ml
```diff
@@ -158,8 +158,9 @@ let ()  = M.myfun M.x
      var x$0 = x;
      for(;;){
       if(! x$0) return 0;
+      var next = x$0[1], cmp = runtime.caml_int_compare;
+      let cmp$0 = cmp;
```
hhugo, line 162: I don't understand this change.

vouillon: Initially, we have:
```
 1 v645 = fun(v647, v646){5 (v646, v647)}

==== 5 (v648, v649) ====
 1 v650 = "caml_int_compare"(v649, v648)
   return v650
```
On master, the function is inlined.
With this PR, the parameters of the initial continuation are eliminated right away (`fun(v647, v646){5 ()}`, so it is turned into an alias:
```
 1 v645 = alias caml_int_compare
```
Then, since the variable is defined in a loop and used in a function, a let binding is introduced.

compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -17233,385 +17218,426 @@
...
+         f:
+         {
+          g:
+          {
+           h:
+           {
+            i:
+            {
+             j:
+             {
```
vouillon, line 17241: Do you understand this?

compiler/tests-compiler/gh1007.ml
hhugo: Maybe we should only introduce aliases once after all optimizations passes

compiler/tests-full/stdlib.cma.expected.js
hhugo: I did not look at it yet

compiler/lib/tailcall.ml
```diff
@@ -39,53 +39,107 @@ let rec tail_call x f l =
...
-        blocks
-        pc
-        (fun pc (visited, blocks) ->
-          let visited, blocks = traverse f pc visited blocks in
-          visited, blocks)
-        (visited, blocks)
-    in
-    visited, blocks
+    let blocks' = rewrite_block f pc blocks in
+    if not (phys_equal blocks blocks')
```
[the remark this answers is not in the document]
hhugo: Now returning an option

compiler/tests-compiler/gh1007.ml
hhugo: I've pushed a commit to only emit aliases at the end

compiler/tests-full/stdlib.cma.expected.js
hhugo: This is because we have new tailcalls introducing loops and  `shrink_loops` tries to exit the loop for non-small code-path. Here, I believe the presence of closure is responsible for exiting the loop, we have the following in `Structure.measure`
```
          (* A closure is never small *)
          | Let (_, Closure _) -> -1
```

hhugo: I'm not sure what to do about this

vouillon: So, that's the right thing to do, then.

vouillon (approved):

PR #1947, opened 2025-04-23
Compiler: disconnect closure cont from closure location. 

hhugo:
This gives us more freedom to do code manipulation without losing debug info / sourcemap.

This refactoring also means that we no longer need debug information after parsing bytecode, which simplifies  code a bit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
 * Compiler: static evaluation of more primitives (#1912)
 * Compiler: faster compilation by stopping sooner when optimizations become unproductive (#1939)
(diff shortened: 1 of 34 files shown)


vouillon:
Maybe we can just pass the source location instead.
```ocaml
Parse_bytecode.Debug.find_loc debug ~position:after pc
```
This would let us drop the debugging information after parsing the bytecode.

hhugo:
> Maybe we can just pass the source location instead.
> 
> ```ocaml
> Parse_bytecode.Debug.find_loc debug ~position:after pc
> ```
> 
> This would let us drop the debugging information after parsing the bytecode.

I've pushed the change.
The PR now need some more attention around  `lamda_lift*.ml`, `effects.ml` and  `generate_closure.ml`

vouillon (approved):

PR #1949, opened 2025-04-24
Some small improvement to the optimization passes

hhugo:
on top of #1943

### DIFF
--- a/compiler/lib/driver.ml
+++ b/compiler/lib/driver.ml
@@ -78,12 +78,10 @@ let specialize_js_once_after p =
   if debug () then Format.eprintf "Specialize js once...@.";
   Specialize_js.f_once_after p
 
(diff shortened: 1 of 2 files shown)


compiler/lib/driver.ml
```diff
@@ -147,65 +141,52 @@ let print p =
...
   +> deadcode
   +> tailcall
   +> phi
   +> flow
-  +> specialize'
+  +> specialize
   +> eval
   +> inline
   +> deadcode
   +> print
```
vouillon, line 173: I'm wondering why there is a `print` here.

hhugo: history -> f28eb2b6fdf6395ec620e60ee077cb7b99bd35eb

PR #1958, opened 2025-04-30
Reference unboxing

vouillon:


compiler/lib/phisimpl.ml
```diff
@@ -67,6 +67,10 @@ let program_deps { blocks; _ } =
     (fun _pc block ->
       List.iter block.body ~f:(fun i ->
           match i with
+          | Let (x, Prim (Extern "%identity", [ Pv y ])) ->
```
hhugo, line 70: "%identity" is eliminated by parse_bytecode

hhugo:
rebased

compiler/lib/phisimpl.ml
hhugo: I didn't realize the PR introduce such prim back

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,173 @@
...
+              if check_field_access depth x then Var.Hashtbl.replace updated x start_pc
+          | Set_field (x, _, Non_float, y) ->
+              discard y;
+              if check_field_access depth x then Var.Hashtbl.replace updated x start_pc
+          | _ -> Freevars.iter_instr_free_vars discard i)
+        block.body;
+      Freevars.iter_last_free_var discard block.branch;
+      match block.branch with
+      | Pushtrap ((pc', _), _, (pc'', _)) ->
+          traverse (depth + 1) start_pc pc';
```
hhugo: Can you add a comment explaining why we can't optimize across try-catch ?

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,173 @@
...
+               | _ -> false -> m, rem
+        | _ -> m, i :: rem)
+      block.body
+      ~init:(m, [])
+  in
+  m, List.rev l
+
+let rewrite_cont relevant_vars vars (pc', args) =
+  let refs, _ = Int.Hashtbl.find relevant_vars pc' in
+  let vars = Var.Map.filter (fun x _ -> Var.Set.mem x refs) vars in
```
hhugo: Is the filtering necessary ? isn't vars always a subset of refs ?

vouillon: Not when you jump outside the scope of a reference. `refs` is the set of references which are in scope a the beginning of block `pc'`; `vars` maps references in scope in the current block to the variables which contain their current value.

compiler/lib/ref_unboxing.ml
vouillon: I have updated the comment at the beginning of the file

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,176 @@
...
+        ~f:(fun i ->
+          match i with
+          | Let (x, Block (0, [| _ |], (NotArray | Unknown), Maybe_mutable)) ->
+              Freevars.iter_instr_free_vars discard i;
+              Var.Hashtbl.replace candidates x depth
+          | Let (_, Closure (_, (pc', _), _)) -> traverse (depth + 1) pc' pc'
+          | Let (_, Field (x, 0, Non_float)) -> ignore (check_field_access depth x)
+          | Offset_ref (x, _) ->
+              if check_field_access depth x then Var.Hashtbl.replace updated x start_pc
+          | Set_field (x, _, Non_float, y) ->
```
hhugo, line 180: Shouldn't this be ?
```suggestion
          | Set_field (x, 0, Non_float, y) ->
```

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,176 @@
...
+  let t = Timer.make () in
+  let candidates = Var.Hashtbl.create 128 in
+  let updated = Var.Hashtbl.create 128 in
+  let visited = BitSet.create' p.free_pc in
+  let discard x = Var.Hashtbl.remove candidates x in
+  let check_field_access depth x =
+    match Var.Hashtbl.find candidates x with
+    | exception Not_found -> false
+    | depth' ->
+        if depth' = depth
```
hhugo, line 150: This check is not enough, we could be at the same level but inside a different try-catch. For example, 
```ocaml
   let f () =
     let r =
       try 
         assert (Random.int 3 < 10);
         ref 1
       with _ -> assert false
     in
     try 
       if (Random.int 3 < 0)
       then incr r
       else decr r;
       !r
     with _ -> assert false
```

compiler/lib/ref_unboxing.ml
vouillon: It does not really matter since we know there is only one field. But this is more coherent with the case `Let (_, Field (x, 0, Non_float))`.

compiler/lib/ref_unboxing.ml
vouillon: Oh, indeed!

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,191 @@
...
+            , Let
+                ( y
+                , Prim
+                    ( Extern "%int_add"
+                    , [ Pv (Var.Map.find x m); Pc (Int (Targetint.of_int_exn n)) ] ) )
+              :: rem )
+        | Set_field (x, 0, Non_float, y) when Var.Map.mem x m -> Var.Map.add x y m, rem
+        | Event _
+          when match rem with
+               | Event _ :: _ -> true
```
hhugo: Can we have a comment explain why it's ok to keep the first event, and drop the second one

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,191 @@
...
+        ~init:vars
+        block.body
+    in
+    Int.Hashtbl.add relevant_vars pc (vars, vars');
+    Addr.Set.iter (fun pc' -> traverse_tree g pc' vars') (Structure.get_edges g pc)
+  in
+  traverse_tree g pc Var.Set.empty;
+  let rec traverse_tree' g pc blocks =
+    let block = Addr.Map.find pc p.blocks in
+    let vars, refs = Int.Hashtbl.find relevant_vars pc in
```
hhugo: We have line 54 
```ocaml
let refs, _ = Int.Hashtbl.find relevant_vars pc' in
```
and line 78
```ocaml
 let vars, refs = Int.Hashtbl.find relevant_vars pc in
```

Using `refs` in both position when matching the tuple is a bit confusing. I think changing variable names with something more explicit would help.

compiler/lib/ref_unboxing.ml
vouillon: Oh, we should keep the last one.

compiler/lib/driver.ml
```diff
@@ -217,6 +217,7 @@ let round profile : 'a -> 'a =
   print
   +> tailcall
   +> (flow +> specialize +> eval +> fst)
+  +> Ref_unboxing.f
```
hhugo, line 219: What's the rational for placing the pass here ? Have you tried other arrangements ?

vouillon: Maybe it makes more sense to put it right before phi simplification, since it relies on it to rename variables.

vouillon: I have changed the pass so that it performs a substitution directly rather that relying on the phi simplification pass followed by a deadcode elimination pass (which would happen in another round).
Then, it is not clear which placement is best.

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,225 @@
...
+
+let rewrite refs block m m' =
+  let m, m', l =
+    List.fold_left
+      ~f:(fun (m, m', acc) i ->
+        match i with
+        | Let (x, Block (0, [| y |], (NotArray | Unknown), Maybe_mutable))
+          when Var.Set.mem x refs -> Var.Map.add x y m, m', acc
+        | Let (y, Field (x, 0, Non_float)) when Var.Map.mem x m ->
+            (* Optimized away by Phisimpl *)
```
hhugo: stale comment

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,225 @@
...
+            m, Var.Map.add y (Var.Map.find x m) m', acc
+        | Offset_ref (x, n) when Var.Map.mem x m ->
+            let y = Var.fresh () in
```
hhugo: Should we use `Var.fork x` ?

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,225 @@
...
+up to date whenever a reference is updated.
+*)
+
+let debug = Debug.find "unbox-refs"
+
+let times = Debug.find "times"
+
+let stats = Debug.find "stats"
+
+let rewrite refs block m m' =
```
hhugo: More explicit naming, maybe.
```suggestion
let rewrite unboxed_refs block refs_content subst =
```

hhugo: Actually, I think it would be better to name it `rewrite_body` and have it take a `body` instead of a `block`.

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,230 @@
...
+      Switch (x, Array.map ~f:(fun cont -> rewrite_cont relevant_vars ref_contents cont) a)
+  | Pushtrap (cont, x, cont') ->
+      Pushtrap
+        ( rewrite_cont relevant_vars ref_contents cont
+        , x
+        , rewrite_cont relevant_vars ref_contents cont' )
+  | Poptrap cont -> Poptrap (rewrite_cont relevant_vars ref_contents cont)
+
+let rewrite_function p ~unboxed_refs pc subst =
+  let g = Structure.(dominator_tree (build_graph p.blocks pc)) in
```
hhugo, line 101: This is ok because we don't have unboxed refs in `cont` / `body.params`.  Maybe we could add a comment.

compiler/lib/ref_unboxing.ml
```diff
@@ -0,0 +1,230 @@
...
+ *)
+
+open! Stdlib
+open Code
+
+(*
+ocamlc does not perform reference unboxing when emitting debugging
+information. Inlining can also enable additional reference unboxing.
+
+We currently does not unbox references which are used within the scope
```
hhugo:
```suggestion
We currently do not unbox references which are used within the scope
```

compiler/lib/ref_unboxing.ml
vouillon: I don't understand what you mean.

hhugo: Ignore my comment. I must have got confused.

PR #1961, opened 2025-05-05
Compiler: faster specialize

hhugo:


compiler/lib/specialize.ml
```diff
@@ -97,39 +101,44 @@ let specialize_instr opt_count function_arity ((acc, free_pc, extra), loc) i =
...
+        then
+          let (body, free_pc, extra), _ =
+            List.fold_left
+              block.body
+              ~init:(([], free_pc, []), None)
+              ~f:(fun acc i ->
+                match i with
+                | Event loc ->
+                    let (body, free_pc, extra), _ = acc in
+                    (i :: body, free_pc, extra), Some loc
```
vouillon, line 121: While you are at it, it seems this location is not used.

PR #1962, opened 2025-05-05
Improved optimization pass scheduling

vouillon:


compiler/lib/driver.ml
```diff
@@ -157,38 +155,20 @@ let rec loop max name round i (p : 'a) : 'a =
...
-  +> inline
-  +> deadcode
-  +> print
-  +> flow
-  +> specialize
-  +> eval
-  +> inline
-  +> deadcode
-  +> phi
+let o1 = loop 2 "round" round 1 +> phi +> flow +> specialize +> eval +> print
```
hhugo, line 163: How did you decide to add the `phi +> flow +> specialize +> eval` ?

hhugo:
I was playing with the same idea and got the same diffs in the tests, do you understand where they come from ?

vouillon:
> I was playing with the same idea and got the same diffs in the tests, do you understand where they come from ?

It's because we add a `phi` pass before the first `deadcode` pass. So the name of some variables which are actually unused get propagated to previous variables.

compiler/lib/driver.ml
vouillon: These are just the first passes in `round`. It does not really make sense to end with an inlining pass, since the principal interest of this pass is to allow more simplifications afterward. I have omitted the tailcall pass since there is not much opportunity for more tail call optimizations after two rounds of optimizations.

compiler/lib/deadcode.ml
```diff
@@ -183,6 +183,26 @@ let annot st pc xi =
 
 (****)
 
+let remove_unused_blocks p =
```
hhugo, line 186: How do you feel about adding to the invariant that there are no unused blocks ?

compiler/lib/deadcode.ml
```diff
@@ -319,6 +344,8 @@ let f ({ blocks; _ } as p : Code.program) =
             })
       blocks
   in
+  let p = { p with blocks } in
+  let p = remove_empty_blocks ~live_vars:st.live p in
   if times () then Format.eprintf "  dead code elim.: %a@." Timer.print t;
```
hhugo, line 349: at line 323, we have 
```
if not (Addr.Set.mem pc st.reachable_blocks)
        then (
          st.deleted_blocks <- st.deleted_blocks + 1;
          None)
```
Maybe we could remove this logic (and turn the `filter_map` into a `map`) and let remove_unused_blocks do its job.

compiler/lib/deadcode.ml
```diff
@@ -242,15 +262,20 @@ let remove_empty_blocks ~live_vars (p : Code.program) : Code.program =
             (let branch = block.branch in
              match branch with
              | Branch cont -> Branch (resolve cont)
-             | Cond (x, cont1, cont2) -> Cond (x, resolve cont1, resolve cont2)
+             | Cond (x, cont1, cont2) ->
+                 let cont1' = resolve cont1 in
+                 let cont2' = resolve cont2 in
+                 if Code.cont_equal cont1' cont2'
```
hhugo, line 268: Maybe add a comment saying that this logic also exists in Eval.f or just remove it  from eval.
Note that his can introduce more deadcode if x is no longer used which means that Deadcode might not reach a fixpoint. This might be undesirable.

vouillon: I actually added this to reach a fixpoint faster,. If you have a complex pattern matching whose result is unused, it can take several iterations to reach a fixpoint: simplifying the conditional allows to empty the block, which allows to simplify a previous conditional. This happens when compiling `ocamlc`.

Maybe we can remove this logic from Eval.f, indeed.

compiler/lib/code.ml
```diff
@@ -801,11 +801,35 @@ let eq p1 p2 =
...
+let cont_compare (pc, args) (pc', args') =
+  let c = compare pc pc' in
+  if c <> 0 then c else List.compare ~cmp:Var.compare args args'
+
 let with_invariant = Debug.find "invariant"
 
 let check_defs = false
 
-let invariant { blocks; start; _ } =
+let mark_unused_blocks p =
```
hhugo, line 814: used_blocks

hhugo:
@vouillon, I've rebased your branch and made some changes to accommodate  the new check

compiler/lib/deadcode.ml
vouillon: It seems better to remove this dead code early

hhugo:
It's a bit difficult to analyze the benchmarks for this PR because the number of optimization passes is now different.

The only noticeable regression is with Fiat-Crypto, `phi + flow` went from 12s to 17s. I don't know if we can do anything about it.

hhugo:
Can you also add a Changelog entry

vouillon:
> The only noticeable regression is with Fiat-Crypto, `phi + flow` went from 12s to 17s. I don't know if we can do anything about it.

 I have pushed a commit to skip the phi simplification pass in the first round. It is expensive (since blocks still have a lot of parameters) and not very helpful.

hhugo (approved):

PR #1963, opened 2025-05-05
Speed-up dead code elimination and make Pure_fun.f more effective

vouillon:


compiler/lib/pure_fun.ml
```diff
@@ -40,39 +45,47 @@ let pure_instr pure_funs i =
...
+    let pure =
       fold_children
         blocks
         pc
-        (fun pc (pure, visited, funs) ->
-          let pure', visited, funs = traverse blocks pc visited funs in
-          pure && pure', visited, funs)
-        (true, visited, funs)
+        (fun pc pure ->
+          let pure' = traverse blocks pc visited pure_blocks funs in
```
hhugo, line 59: Do we really need to traverse if  `pure = true` ?

compiler/lib/deadcode.ml
```diff
@@ -42,7 +42,7 @@ type t =
   { blocks : block Addr.Map.t
   ; live : variable_uses
   ; defs : def list array
-  ; mutable reachable_blocks : Addr.Set.t
+  ; reachable_blocks : BitSet.t
   ; pure_funs : Var.Set.t
```
hhugo, line 46: The PR doesn't change this code, but 
- Have you considered making pure_funs a bitset ?
- Maybe the result of `Pure_fun.f`  could be made abstract.

hhugo:
let's wait for #1959

compiler/lib/deadcode.ml
vouillon: I don't think it is worth it to use a bitset.

PR #1965, opened 2025-05-05
Flow.the_const_of: allow to use different comparison functions

vouillon:
Whether two constants should be considered equal depends on the context in which they are used.

### DIFF
--- a/compiler/lib/flow.ml
+++ b/compiler/lib/flow.ml
@@ -360,53 +360,25 @@ let the_def_of info x =
         x
   | Pc c -> Some (Constant c)
 
-(* If [constant_identical a b = true], then the two values cannot be
-   distinguished, i.e., they are not different objects (and [caml_js_equals a b
-   = true]) and if both are floats, they are bitwise equal. *)
-let constant_identical ~(target : [ `JavaScript | `Wasm ]) a b =
-  match a, b, target with
-  | Int i, Int j, _ -> Targetint.equal i j
-  | Float a, Float b, `JavaScript -> Int64.equal a b
-  | Float _, Float _, `Wasm -> false
-  | NativeString a, NativeString b, `JavaScript -> Native_string.equal a b
-  | NativeString _, NativeString _, `Wasm ->
-      false
-      (* Native strings are boxed (JavaScript objects) in Wasm and are
-          possibly different objects *)
-  | String a, String b, `JavaScript -> Config.Flag.use_js_string () && String.equal a b
-  | String _, String _, `Wasm ->
-      false (* Strings are boxed in Wasm and are possibly different objects *)
-  | Int32 _, Int32 _, `Wasm ->
-      false (* [Int32]s are boxed in Wasm and are possibly different objects *)
-  | Int32 a, Int32 b, `JavaScript -> Int32.equal a b
-  | NativeInt _, NativeInt _, `Wasm ->
-      false (* [NativeInt]s are boxed in Wasm and are possibly different objects *)
-  | NativeInt a, NativeInt b, `JavaScript -> Int32.equal a b
-  (* All other values may be distinct objects and thus different by [caml_js_equals]. *)
-  | Int64 _, Int64 _, _ -> false
-  | Tuple _, Tuple _, _ -> false
-  | Float_array _, Float_array _, _ -> false
-  | (Int _ | Float _ | Int64 _ | Int32 _ | NativeInt _), _, _ -> false
-  | (String _ | NativeString _), _, _ -> false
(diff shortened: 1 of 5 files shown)


compiler/lib/flow.ml
```diff
@@ -360,53 +360,25 @@ let the_def_of info x =
...
-          | Expr (Constant ((Int32 _ | NativeInt _) as c)), `JavaScript -> Some c
-          | Expr (Constant (String _ as c)), _ when Config.Flag.safe_string () -> Some c
-          | Expr (Constant c), _ ->
-              if Var.ISet.mem info.info_possibly_mutable x then None else Some c
+          match info.info_defs.(Var.idx x) with
+          | Expr
+              (Constant
+                 ((Float _ | Int _ | Int32 _ | Int64 _ | NativeInt _ | NativeString _) as
+                  c)) -> Some c
+          | Expr (Constant c)
```
hhugo, line 374:
```suggestion
          | Expr (Constant (String _ as c))
```

compiler/lib/flow.ml
```diff
@@ -360,53 +360,25 @@ let the_def_of info x =
...
+            when Config.Flag.safe_string ()
+                 || not (Var.ISet.mem info.info_possibly_mutable x) -> Some c
```
hhugo, line 376:
```suggestion
          | Expr (Constant (String _ as c)), _ when Config.Flag.safe_string () -> Some c
          | Expr (Constant c) -> 
             if  Var.ISet.mem info.info_possibly_mutable x then None else Some c
```

compiler/lib/flow.ml
```diff
@@ -429,13 +401,13 @@ let the_int info x =
   | Pc (Int c) -> Some c
   | Pc _ -> None
 
-let the_string_of ~target info x =
-  match the_const_of info ~target x with
+let the_string_of info x =
+  match the_const_of ~eq:(fun _ _ -> false) info x with
```
hhugo, line 405: Can you explain the `~eq:(fun _ _ -> false)` here and below ?

compiler/lib/flow.ml
vouillon: I think neither code is correct.
When `Config.Flag.safe_string ()` is true, all constants are really constant.
If `Var.ISet.mem info.info_possibly_mutable x` if false, the root of the constant is not possibly modified. So, if it is a string, it is a constant string. If it's a tuple, we are not sure.

compiler/lib/flow.ml
vouillon: This is used to specialize JavaScript primitives. We expect that a single string is directly passed to the primitive. So, I don't think we need to handle the case when there are several possible strings. But I can use `constant_equal` if you prefer.

compiler/lib/flow.ml
```diff
@@ -352,61 +352,38 @@ let the_def_of info x =
...
+                 (( Float _
+                  | Int _
+                  | Int32 _
+                  | Int64 _
+                  | NativeInt _
+                  | NativeString _
+                  | Float_array _ ) as c)) -> Some c
+          | Expr (Constant (String _ as c))
+            when not (Var.ISet.mem info.info_possibly_mutable x) -> Some c
+          | Expr (Constant c) when Config.Flag.safe_string () -> Some c
```
hhugo, line 381: I didn't follow your previous explanation, why would this be true for tuple ?

compiler/lib/flow.ml
```diff
@@ -352,61 +352,38 @@ let the_def_of info x =
...
+          match info.info_defs.(Var.idx x) with
+          | Expr
+              (Constant
+                 (( Float _
+                  | Int _
+                  | Int32 _
+                  | Int64 _
+                  | NativeInt _
+                  | NativeString _
+                  | Float_array _ ) as c)) -> Some c
```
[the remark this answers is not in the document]
vouillon: Yes, this is for a record of float or an immutable array. `caml_obj_dup` is called when the array is not actually constant:
```
ocaml -dinstr
# [|1.;2.;3.;4.;5.|];;
	const [|1. 2. 3. 4. 5.|]
	ccall caml_obj_dup, 1
```

compiler/lib/flow.ml
vouillon: Constants are not ever modified, except possibly strings (and therefore constants containing strings) when `Config.Flag.safe_string ()` is false. This is enforced by the OCaml compiler.

hhugo (approved):

hhugo:
Can you add a changelog entry ?

PR #1968, opened 2025-05-06
Compiler: compact program Addr.t

hhugo:
Blocks `Addr.t` are sparse in a program. This can be wasteful when using bitset to encode visited blocks.
The PR propose to "compact" `Addr.t` in a program if needed.


hhugo:
This reduces memory consumption by ~3 to ~5% on large programs in the benchmarks

PR #1971, opened 2025-05-07
Compiler: speedup coloring

hhugo:
This PR removes a bottleneck in the coloring algorithm and significantly speedup the pass.


hhugo:
@OlivierNicole, the benchmark CI job disappeared. Who should we ping ?

vouillon (approved):

OlivierNicole:
Sorry for the delay. Jérôme pinged the relevant people internally and it seems solved. The external point of contact is https://github.com/ocurrent/current-bench/issues, we can relay if necessary.

PR #1973, opened 2025-05-09
Adjust some microbenmarks

hhugo:
Some benchmarks terminate very fast and are very noisy due to startup overhead. The PR make soli, binary_trees and fft do more work so that take in the order of a second to finish.


vouillon (approved):

vouillon:
Maybe `raytrace` as well?

PR #1977, opened 2025-05-12
Fix bigarray hashing

vouillon:


hhugo (approved):
LGTM. Just needs a change log entry

PR #1979, opened 2025-05-12
Wasm runtime: use DataView primitives to implement bigarrays

vouillon:


hhugo:
should we extend the scope of this PR to the js runtime ?

vouillon:
From Wasm, we have to call a JavaScript function to access typed arrays, which is slow compared to the array access itself, while DataView primitives get optimized by V8. We don't have this issue with the JavaScript runtime.

hhugo:
I'm going to just trust the CI for this one. Not doing a full review.

PR #1981, opened 2025-05-12
Deprecate List function hiding polymorphic comparison

hhugo:
and avoid using polymorphic hashtbl


hhugo:
> The bench reveal a 18% increase for the "check" pass when building Fiat-Crypto. I don't understand where this is coming from. @vouillon, maybe another pair of eyes could help here.

I was apparently just some noise

vouillon:
> I was apparently just some noise

I suspect you were not up to date with master. So, this PR was missing some performance improvements that were on master.

PR #1986, opened 2025-05-15
Improve Js_assign logic

hhugo:
The logic now handles named variable and prevents collisions. One no longer need to globally "reserve"  named free variables.

### DIFF
--- a/compiler/tests-full/stdlib.cma.expected.js
+++ b/compiler/tests-full/stdlib.cma.expected.js
@@ -437,8 +437,8 @@
     cst_true = cst_true$0,
     cst_false = cst_false$0,
     cst_bool_of_string = "bool_of_string",
-    _a_ = [0, 1],
-    _b_ = [0, 0];
+    a = [0, 1],
+    b = [0, 0];
    function char_of_int(n){
      /*<<stdlib.ml:224:2>>*/ if(0 <= n && 255 >= n)
       /*<<stdlib.ml:224:58>>*/ return n;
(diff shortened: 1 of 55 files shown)


compiler/tests-full/stdlib.cma.expected.js
```diff
@@ -1823,8 +1815,8 @@
      var
       xs$0 =  /*<<seq.ml:133:12>>*/ match[2],
       x = match[1],
-      _an_ =  /*<<seq.ml:137:6>>*/ caml_call1(p, x);
-      /*<<seq.ml:137:9>>*/ if(! _an_) return _an_;
+      a =  /*<<seq.ml:137:6>>*/ caml_call1(p, x);
+      /*<<seq.ml:137:9>>*/ if(! a) return a;
```
vouillon, line 1819: The rationale for this style of variable name `_an_` is that we could search for all occurrences of a variable. This becomes harder when all variable names are shorten.

hhugo: For debugging, you can always use `--debug stable_var` or we could implement another variation.

hhugo: Do you have a use case in mind where neither the new behavior nor the stable_var one work for you ?

hhugo:
This PR reduces the size of fiat-crypo by 10%. This is surprising, because I would expect smaller size with `--pretty` only and the fiat-crypto benchmark doesn't use pretty.

hhugo:
> This PR reduces the size of fiat-crypo by 10%. This is surprising, because I would expect smaller size with `--pretty` only and the fiat-crypto benchmark doesn't use pretty.

The reason is that we're now using a fresh `Var_printer.t` in js_assign and can properly assign short variable starting from "index zero".

 `master` is using `Var.to_string` and rely on it's internal state. `effects.ml` uses debug such as `debug_print "@[<v>cname = %s@,@]" @@ Var.to_string v` which was forcing the evaluation of `Var.to_string` even when debug was not enabled.

hhugo:
fiat-crypto compiles  5% faster.

compiler/tests-full/stdlib.cma.expected.js
hhugo: I've added a new debug `--debug var` that emits variables in the `_an_` style and used it in `stdlib.cma.expected.js`

vouillon: > Do you have a use case in mind where neither the new behavior nor the stable_var one work for you ?

I think this is mostly useful for constants. You want to easily find where a constant is used (or find the definition of a constant), but still get names from the source code when available.

hhugo:
I think this is ready to be merged. @vouillon, do you want to review this before I merge ?

vouillon (approved):

PR #1987, opened 2025-05-15
Fewer free variable traversal

hhugo:
~~On top of #1986,~~ This PR removes two `Js_traverse.free` traversal and replace them by a cheaper over-approximation.


hhugo:
probably replaced by https://github.com/ocsigen/js_of_ocaml/pull/1989

PR #1988, opened 2025-05-15
Compiler: faster constant sharing

hhugo:


hhugo:
> Are there any benchmark results for this change?

You can look at https://bench.ci.dev/ocsigen/js_of_ocaml/branch/master/benchmark/Js_of_ocaml?worker=autumn&image=bench.Dockerfile. There is a 1.5s drop in compilation time for fiat-crypto

PR #1991, opened 2025-05-15
Try to run the tests with jsc

vouillon:


hhugo:
Don't you think we should aim at being able to test multiple environments ? I wanted to test deno for instance

vouillon:
> Don't you think we should aim at being able to test multiple environments ? I wanted to test deno for instance

I have some code to test other environments in #1831.
Deno is supposed to be Node-compatible, so it should not be too hard to use it.
The issue with shells like `jsc`, `sm` or `v8` is that they are too limited to run the tests unchanged.

PR #1992, opened 2025-05-15
Faster wat output

hhugo:


hhugo:
I've added timing for binaryen invocations but that makes the benchmarks graph pretty useless unless you untick the binaryen passes.

vouillon:
> I've added timing for binaryen invocations but that makes the benchmarks graph pretty useless unless you untick the binaryen passes.

Right. That's why I did not add them.

vouillon:
`Wasm_output` can output a Wasm binary module, which should be much faster to write than a text file, and faster to parse by binaryen. I'm not really using it at the moment since it does not support source maps yet. But maybe we could use it when source maps are disabled (and `--debug wat` is not set).

PR #2000, opened 2025-05-16
Output a Wasm binary module rather than a text file when possible

vouillon:
This is faster to write a binary file, and then for binaryen to parse a binary file.

We still output a Wasm text file when source maps are enabled (we don't yet supported writing a source map file together with a Wasm binary module) or when `--debug wat` is set.

### DIFF
--- a/benchmarks/benchmark-fiat-crypto/Makefile
+++ b/benchmarks/benchmark-fiat-crypto/Makefile
@@ -10,8 +10,9 @@ bench:
 	$(MAKE) perform COMPILER=wasm_of_ocaml EXTRA_ARGS="" KIND=wasm
 	@date -u +"%FT%TZ - $(NAME): done"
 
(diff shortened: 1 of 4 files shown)


hhugo:
wasm-merge is WAAAAAYYYY faster. What would it take to support sourcemap with the binary format ?

vouillon:
> wasm-merge is WAAAAAYYYY faster. What would it take to support sourcemap with the binary format ?

I guess they have a slow parser...

Not that much, I guess. We need to replicate what is done in `Js_output`. So, push a mapping for each `Event`, plus one at the end of each function (after the byte `0x0B`) to avoid that an events bleeds over the next function. There is a single line; the column is the position in the binary file.
Plus some plumbing to write the source map to a temporary file and pass the file name to binaryen.

hhugo:
Maybe we could remove sourcemap from the fiat crypto bench for now

PR #2001, opened 2025-05-16
Compiler: reorder optimization passes

hhugo:


hhugo:
This makes jsoo find the fixpoint of ocamlc one round earlier (from ~8.6s to ~7.8s on my machine)
I don't know why it doesn't show clearly on the benchmark.

vouillon:
> This makes jsoo find the fixpoint of ocamlc one round earlier (from ~8.6s to ~7.8s on my machine) I don't know why it doesn't show clearly on the benchmark.

We are using OCaml 5.2 for the benchmarks, which already only needs 5 rounds.

vouillon (approved):

PR #1645, opened 2024-07-26
Text encode decode

hhugo:

PR #2008, opened 2025-05-19
Runtime: fix caml_utf8_of_utf16 bug in high surrogate case

### LINKED ISSUE #2006: [BUG] `caml_utf8_of_utf16` has a bug in high surrogate case
**Describe the bug**

`caml_utf8_of_utf16(String.fromCharCode(0xdbff, 0xdfff))` generate wrong utf8 encoding string

https://github.com/ocsigen/js_of_ocaml/blob/246df644334aa97103d1051aec128337ea1868f9/runtime/js/mlBytes.js#L106-L113

line 107 should be `c > 0xdbff` since `0xdbff` is a valid surrogate

**Expected behavior**

expected result: `b'\xf4\x8f\xbf\xbf'` (get from python `bytes.fromhex('dbffdfff').decode('utf-16be').encode('utf-8')`)

actual result: `ef bf bd ed bf bf`, get from

```js
Array.from(caml_utf8_of_utf16(String.fromCharCode(0xdbff, 0xdfff)))
    .map((c) => c.charCodeAt(0).toString(16).padStart(2, "0"))
    .join(" ")
```

**Versions**

latest version of jsoo contains this bug.

hhugo:
fix #2006

### DIFF
--- a/lib/tests/dune.inc
+++ b/lib/tests/dune.inc
@@ -100,6 +100,16 @@
   (pps ppx_js_internal ppx_expect)))
 
 (library
+ ;; lib/tests/test_string.ml
+ (name test_string_75)
+ (enabled_if true)
+ (modules test_string)
+ (libraries js_of_ocaml unix)
+ (inline_tests (modes js wasm))
+ (preprocess
+  (pps ppx_js_internal ppx_expect)))
+
+(library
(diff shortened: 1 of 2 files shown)


hhugo:
Maybe a good time to reconsider #1645

PR #2010, opened 2025-05-21
Compiler: do not prevent deadcode to preserve tailcall

hhugo:


OlivierNicole:
I’m still confused. In `Let (x, e); Return x`, if `e` is not pure, IIUC `x` should not be dead. So it seems to me that this patch is the equivalent of reverting to the previous state of unconditionally emitting `Return (zero_var x)`. Indeed, the only case where `live_tc` is true is when this is a tail call to an impure function, but then `x` should be live, so `zero_var x` is `x`.

hhugo:
> I’m still confused. In `Let (x, e); Return x`, if `e` is not pure, IIUC `x` should not be dead. So it seems to me that this patch is the equivalent of reverting to the previous state of unconditionally emitting `Return (zero_var x)`. Indeed, the only case where `live_tc` is true is when this is a tail call to an impure function, but then `x` should be live, so `zero_var x` is `x`.

you can have `e` is not pure and `x` is dead. `x` is dead just means `x` is not used but we still need to keep e if it has side-effects.

OlivierNicole:
I see, thanks for the explanation!

PR #2013, opened 2025-05-22
Fix invariant

hhugo:
## Effects
The effects pass was previously  leaving unused blocks behind.

### The issue
I do not have a good grasp on what's is going on but it seems that a parent closure can sometimes be optimized before the closures it contains (recursive cycles).

```
function A () 
   fun B() {  A () }
   fun C() { B () }
   C();
}
```
- We first optimize A and decide nothing happens
- Then C and decide to inline B and B becomes dead
- We need to cleanup A but we already visited it

### The fix
- We maintain a table for a closure to the block it is declared into.
- We collect the names of all inlined closures, filter out the one that are still live
- We map the dead closures to their block and rewrite theses blocks only.

### DIFF
--- a/compiler/lib/lambda_lifting.ml
+++ b/compiler/lib/lambda_lifting.ml
@@ -225,15 +225,16 @@ let rec traverse var_depth (program, functions) pc depth limit =
     program.blocks
     (program, functions)
 
-let f program =
+let f p =
   let t = Timer.make () in
   let nv = Var.count () in
   let var_depth = Array.make nv (-1) in
-  let program, functions =
+  let p, functions =
     let threshold = Config.Param.lambda_lifting_threshold () in
     let baseline = Config.Param.lambda_lifting_baseline () in
-    traverse var_depth (program, []) program.start 0 (baseline + threshold)
+    traverse var_depth (p, []) p.start 0 (baseline + threshold)
   in
   assert (List.is_empty functions);
   if Debug.find "times" () then Format.eprintf "  lambda lifting: %a@." Timer.print t;
-  program
+  Code.invariant p;
+  p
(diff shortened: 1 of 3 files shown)


compiler/lib/effects.ml
```diff
@@ -1140,6 +1140,7 @@ let f ~flow_info ~live_vars p =
   let p = split_blocks ~cps_needed p in
   let p, trampolined_calls, in_cps = cps_transform ~live_vars ~flow_info ~cps_needed p in
   if Debug.find "times" () then Format.eprintf "  effects: %a@." Timer.print t;
+  let p = Deadcode.remove_unused_blocks p in
```
hhugo, line 1143: @OlivierNicole , @vouillon , is it expected that the effects pass produces unused  blocks ?

compiler/lib/inline.ml
```diff
@@ -409,7 +420,8 @@ and should_inline ~context info args =
       || Option.equal Var.equal info.enclosing_function context.current_function
       || (not !(context.has_closures))
          && Option.equal Var.equal info.enclosing_function context.enclosing_function
-  | `Wasm, _ | `JavaScript, `Double_translation -> true
+  | `JavaScript, `Double_translation -> true
```
hhugo, line 423: @vouillon, I think we need a comment explaining this line

compiler/lib/effects.ml
OlivierNicole: I would expect so, yes. It creates a bunch of new blocks when creating all the continuation closures.

vouillon: @hhugo How do you get it to produce unused blocks? I removed this line and `dune build @runtest-js --profile with-effects` still runs without error.

hhugo: OCAMLRUNPARAM=b=1 dune exec js_of_ocaml `which ocamlc.byte` --  -o ocamlc.js --opt 1  --debug stats  --debug invariant --effects double-translation

hhugo:
This invariant has been fixed in other PRs

PR #2022, opened 2025-05-27
Wasm_of_ocaml: more efficient definition of constant Javascript strings

vouillon:
Instead of storing JavaScript strings in the JavaScript launcher script, we use the ['imported string constants'](https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md#string-constants) mechanism from the JS String Builtins proposal.

### DIFF
--- a/compiler/lib-wasm/gc_target.ml
+++ b/compiler/lib-wasm/gc_target.ml
@@ -987,7 +987,7 @@ module Constant = struct
         in
         let* x =
           register_import
-            ~import_module:""
+            ~import_module:"str"
             ~name:s
             (Global { mut = false; typ = Ref { nullable = false; typ = Extern } })
         in
--- a/runtime/wasm/runtime.js
+++ b/runtime/wasm/runtime.js
@@ -572,7 +572,7 @@
       "wasm:js-string": string_ops,
       "wasm:text-decoder": string_ops,
       "wasm:text-encoder": string_ops,
-      "": new globalThis.Proxy(
+      str: new globalThis.Proxy(
         {},
         {
           get(_, prop) {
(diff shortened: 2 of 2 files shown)


runtime/wasm/runtime.js
```diff
@@ -570,11 +570,22 @@
       "wasm:js-string": string_ops,
       "wasm:text-decoder": string_ops,
       "wasm:text-encoder": string_ops,
+      "": new globalThis.Proxy(
```
hhugo: Do we really need this ? Is it a fallback when the extension is not implemented ?

compiler/lib-wasm/gc_target.ml
```diff
@@ -989,22 +982,18 @@ module Constant = struct
...
         let* x =
-          let* name = unit_name in
           register_import
-            ~import_module:
-              (match name with
-              | None -> "strings"
-              | Some name -> name ^ ".strings")
-            ~name:(string_of_int i)
-            (Global { mut = false; typ = Ref { nullable = false; typ = Any } })
+            ~import_module:""
```
hhugo: It would look less magic if we use an import_module that is not the empty string. What do you think ?

runtime/wasm/runtime.js
vouillon: Right. And I think it is still necessary for the version of `node` that we are using on the CI.

compiler/lib-wasm/gc_target.ml
vouillon: The module name is repeated for each string, so it is better to keep it short. On the other hand, we probably do not have many JavaScript strings in practice.

PR #2026, opened 2025-05-30
Wasm_of_ocaml: make it possible to disable effects with `--effects disabled`

vouillon:
There can be a significant cost to call a JavaScript function from Wasm due to stack switching when JSPI is enabled (which will happen with node 25).

### DIFF
--- a/compiler/bin-js_of_ocaml/cmd_arg.ml
+++ b/compiler/bin-js_of_ocaml/cmd_arg.ml
@@ -39,8 +39,8 @@ let trim_trailing_dir_sep s =
 
 let normalize_include_dirs dirs = List.map dirs ~f:trim_trailing_dir_sep
 
-let normalize_effects (effects : [ `Cps | `Double_translation ] option) common :
-    Config.effects_backend =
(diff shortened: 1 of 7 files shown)


hhugo:
We could be explicit about disabling effects in the following logics:

```
compiler/tests-compiler/util/util.ml
280:        | `Double_translation -> [ "--effects=double-translation" ]
281:        | `Cps -> [ "--effects=cps" ]

compiler/tests-dynlink-js/effects_flags.ml
5:    | [ "with-effects-double-translation" ], true -> [ "--effects"; "double-translation" ]
6:    | [ "with-effects" ], true -> [ "--effects"; "cps" ]

toplevel/examples/lwt_toplevel/effects_flags.ml
5:    | [ "with-effects-double-translation" ], true -> [ "--effects"; "double-translation" ]
6:    | [ "with-effects" ], true -> [ "--effects"; "cps" ]
7:    | _, true -> [ "--effects"; "cps" ]

benchmarks/run.ml
252:         | `Cps -> "--effects=cps"
253:         | `Double_translation -> "--effects=double-translation"))
262:         | `Cps -> "--effects=cps"))
```

PR #2030, opened 2025-06-04
Prepare 5.4

hhugo:


hhugo:
@vouillon, can I let you update the wasm runtime ?

runtime/js/domain.js
```diff
@@ -31,6 +31,12 @@ function caml_atomic_load(ref) {
   return ref[1];
 }
 
+//Provides: caml_atomic_load_field
+//Version: >= 5.4
+function caml_atomic_load_field(b, i) {
```
hhugo, line 36: Maybe the we should transform this into a `Field` access in parse_bytecode.

runtime/js/sys.js
```diff
@@ -350,6 +361,41 @@ function caml_sys_is_regular_file(name) {
...
+function caml_sys_io_buffer_size(_unit) {
+  return caml_io_buffer_size;
+}
+
+//Provides: caml_sys_temp_dir_name
+//Requires: os_type
+//Requires: caml_string_of_jsstring
+//Version: >= 5.4
+function caml_sys_temp_dir_name(_unit) {
+  if (os_type === "Win32" && globalThis.os.tmpdir) {
```
vouillon:
```suggestion
  if (os_type === "Win32" && require('node:os').tmpdir) {
```

runtime/js/sys.js
```diff
@@ -350,6 +361,41 @@ function caml_sys_is_regular_file(name) {
...
+  return caml_io_buffer_size;
+}
+
+//Provides: caml_sys_temp_dir_name
+//Requires: os_type
+//Requires: caml_string_of_jsstring
+//Version: >= 5.4
+function caml_sys_temp_dir_name(_unit) {
+  if (os_type === "Win32" && require("node:os").tmpdir) {
+    return caml_string_of_jsstring(globalThis.os.tmpdir());
```
vouillon: If think the test implies that the program is run with node, and thus that `require("node:os").tmpdir)()` will work.
```suggestion
  if (os_type === "Win32") {
    return caml_string_of_jsstring(require("node:os").tmpdir)());
```

hhugo: applied

runtime/js/domain.js
hhugo: see https://github.com/ocsigen/js_of_ocaml/pull/2030/commits/176fae6281766b648f4e4133da956fff2af34496

PR #2032, opened 2025-06-04
Wasm: integer optimizations

vouillon:
Try to avoid `ref.i31`/`i31.get_s`, which are not well optimized by binaryen and V8.


hhugo:
Looking at the benchmarks, I don't see improvements. Is this expected?

hhugo:
> Looking at the benchmarks, I don't see improvements. Is this expected?

There are in fact nice improvement in the micro benchmarks

hhugo:
I dont have time to understand the entire change, the CI is happy, you can merge if you want or wait for someone else to do some review (@OlivierNicole ?)

PR #2036, opened 2025-06-11
Microbenchmarks: disable Wasm baseline compiler

vouillon:
We are interested in the performance of optimized code. But V8 is currently not able to switch to optimized code while executing a loop in Wasm. So, we can get stuck running unoptimized code.

### DIFF
--- a/benchmarks/run.config
+++ b/benchmarks/run.config
@@ -1,6 +1,8 @@
 #
 #interpreter v8 / node: apt-get install nodejs
-interpreter node /usr/bin/env node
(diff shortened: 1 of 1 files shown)


benchmarks/run.config
```diff
@@ -1,6 +1,6 @@
 #
 #interpreter v8 / node: apt-get install nodejs
-interpreter node /usr/bin/env node
+interpreter node /usr/bin/env node --no-liftoff
```
hhugo, line 5: I think you should leave a comment in the file explaining this.

PR #2039, opened 2025-06-12
Fix method lookup cache with separate compilation

hhugo:
Cache for method lookup was slightly broken in separate compilation because separate units/libs would use the same cache slot.

Also, for GETDYNMET, we were using a unique cache slot (zero). This PR no longer uses the cache for it.

### DIFF
--- a/runtime/wasm/obj.wat
+++ b/runtime/wasm/obj.wat
@@ -457,14 +457,8 @@
                (br $loop))))
       (array.set $int_array (global.get $method_cache) (local.get $cacheid)
          (i32.add (local.get $li) (i32.const 1)))
-      (if (result (ref eq))
-          (ref.eq (local.get $vtag)
-             (array.get $block (local.get $meths)
-                (i32.add (local.get $li) (i32.const 1))))
-         (then
-            (array.get $block (local.get $meths) (local.get $li)))
-         (else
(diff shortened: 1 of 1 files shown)


hhugo:
@vouillon, can you review the wasm change ?

runtime/js/obj.js
```diff
@@ -142,19 +142,42 @@ function caml_lazy_make_forward(v) {
...
+    hi = meths[1] * 2 + 1,
+    mi;
+  while (li < hi) {
+    mi = ((li + hi) >> 1) | 1;
+    if (tag < meths[mi + 1]) hi = mi - 2;
+    else li = mi;
   }
+  caml_method_cache[cacheid] = li - 3;
+  /* return 0 if tag is not there */
+  return tag === meths[li + 1] ? meths[li] : 0;
```
vouillon: In this case, we know that the method exists.
```suggestion
  return meths[li];
```

vouillon (changes requested):

runtime/js/obj.js
vouillon: You should do the same in the Wasm code

vouillon: Oh, I missed it!

PR #2040, opened 2025-06-12
Use the type analysis to avoid using JavasScript strict equality

vouillon:
If we know that one of the compared values is not a JavaScript value, we can directly use physical equality.

This is the number of use of JS physical equality when compiling `ocamlc` 5.3. Some uses were already eliminated in PR #2032 (when comparing two integers).

|PR| Number of use of JS physical equality|
|---|---|
|Initially | 748|
|PR #2032  | 658
|This PR | 190

### DIFF
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -226,16 +226,18 @@ module Generate (Target : Target_sig.S) = struct
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) x lsl const 1l)
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) y lsl const 1l)
     | Top, Top ->
-        (if negate then Value.phys_neq else Value.phys_eq)
-          ~relaxed:true
+        Value.js_eqeqeq
+          ~negate
           (transl_prim_arg ctx ~typ:Top x)
           (transl_prim_arg ctx ~typ:Top y)
-    | Int (Normalized | Unnormalized), (Int Ref | Top | Bot | Number _ | Tuple _)
-    | (Int Ref | Top | Bot | Number _ | Tuple _), Int (Normalized | Unnormalized)
-    | ( (Int Ref | Top | Bot | Number _ | Tuple _)
-      , (Int Ref | Top | Bot | Number _ | Tuple _) ) ->
+    | Bot, _ | _, Bot ->
+        (* this is deadcode *)
         (if negate then Value.phys_neq else Value.phys_eq)
-          ~relaxed:false
+          (transl_prim_arg ctx ~typ:Top x)
+          (transl_prim_arg ctx ~typ:Top y)
+    | (Int _ | Number _ | Tuple _), _ | _, (Int _ | Number _ | Tuple _) ->
+        (* Only Top may contain JavaScript values *)
+        (if negate then Value.phys_neq else Value.phys_eq)
           (transl_prim_arg ctx ~typ:Top x)
           (transl_prim_arg ctx ~typ:Top y)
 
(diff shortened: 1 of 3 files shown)


compiler/lib-wasm/generate.ml
```diff
@@ -215,15 +215,26 @@ module Generate (Target : Target_sig.S) = struct
...
+        (if negate then Arith.( = ) else Arith.( <> ))
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) x lsl const 1l)
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) y lsl const 1l)
-    | _ -> op' (transl_prim_arg ctx ~typ:Top x) (transl_prim_arg ctx ~typ:Top y)
+    | Top, Top ->
+        (if negate then Value.phys_eq else Value.phys_neq)
+          ~relaxed:true
+          (transl_prim_arg ctx ~typ:Top x)
+          (transl_prim_arg ctx ~typ:Top y)
+    | _ ->
```
hhugo: Exhaustive pattern would be better here to reason about correctness

compiler/lib-wasm/generate.ml
```diff
@@ -215,15 +215,29 @@ module Generate (Target : Target_sig.S) = struct
...
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) y lsl const 1l)
-    | _ -> op' (transl_prim_arg ctx ~typ:Top x) (transl_prim_arg ctx ~typ:Top y)
+    | Top, Top ->
+        (if negate then Value.phys_neq else Value.phys_eq)
+          ~relaxed:true
+          (transl_prim_arg ctx ~typ:Top x)
+          (transl_prim_arg ctx ~typ:Top y)
+    | Int (Normalized | Unnormalized), (Int Ref | Top | Bot | Number _ | Tuple _)
+    | (Int Ref | Top | Bot | Number _ | Tuple _), Int (Normalized | Unnormalized)
+    | ( (Int Ref | Top | Bot | Number _ | Tuple _)
```
hhugo: Should this apply with `Bot, Bot`, `Bot, Top` and `Top, Bot` ? if so, can you explain why ?

vouillon: Only `Top` may contain JavaScript values, and we only need to use JavaScript strict  equality if both arguments are JavaScript values.

If one argument is `Bot`, this is some dead code, and we can emit whatever we want, as long as it typechecks.

hhugo: Thanks, I've pushed a small refactoring to reflect your comment

compiler/lib-wasm/generate.ml
```diff
@@ -226,16 +226,18 @@ module Generate (Target : Target_sig.S) = struct
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) x lsl const 1l)
           Arith.(transl_prim_arg ctx ~typ:(Int Unnormalized) y lsl const 1l)
     | Top, Top ->
+        Value.js_eqeqeq
+          ~negate
+          (transl_prim_arg ctx ~typ:Top x)
+          (transl_prim_arg ctx ~typ:Top y)
+    | Bot, _ | _, Bot ->
+        (* this is deadcode, deadcode is off *)
```
vouillon: This can also be dead code which is missed by our dead code passes. For instance, if a function with an optional parameter is always passed `None`, we don't eliminate the code that corresponds to the `Some` case.

hhugo: updated

hhugo (approved):

hhugo:
LGTM, @vouillon, you can merge if you think it's good.

PR #2041, opened 2025-06-12
Use more precise types in function environments

vouillon:
This refines the type of locals based on the expression they are initialized with. The main motivation for this is that the environment types get refined as well, which binaryen is not able to do. In turn, some casts become unnecessary. On `ocamlc.byte`, we get the following number of casts:

| Before | After
|-|-
| 37949 | 35553


hhugo:
Do you plan to remove unnecessary cast ?

PR #2055, opened 2025-06-23
CI: additional deps for 32bits

hhugo:


hhugo:
X-link https://github.com/ocaml/opam/issues/6413 which should allow to specify such deps upstream

PR #2056, opened 2025-06-23
Audit unused js variable in the js runtime 

hhugo:


runtime/js/str.js
```diff
@@ -293,29 +292,62 @@ var re_match = (function () {
...
+      if (res) return res;
+      pos--;
+    } while (pos >= 0);
+  } else {
+    do {
+      var res = re_match(re, s, pos, 0);
+      if (res) return res;
+      pos--;
+    } while (pos >= 0);
   }
```
hhugo, line 350: @vouillon, should we implement that logic in the wasm runtime ? I see that you've left `ZZZ startchars` comments about it ?

PR #2057, opened 2025-06-23
Compiler: simplify branch

hhugo:


hhugo:
The effect of this PR on fiat-crypto is big. ~22% on codesize and compilation time.
It improves code size a little bit (-0.5%) on `ocamlc` and `Partial Render Table` but make compilation longer by 1-2%

hhugo:
> Maybe you should drop events when merging branches. This could be really confusing otherwise if it looks like we are reaching a wrong case.

I've done that and moved the optimization to Specialize.switches that runs only once.

PR #2058, opened 2025-06-23
Allow js_of_ocaml-compiler to be installed on OCaml 5.4

kit-ty-kate:
The master branch seems to work but this constraint is forbidding opam users to install it


hhugo:
The constraint needs to be bumped in the dune-project file.
I would also add a CI job for 5.4

PR #2059, opened 2025-06-24
Closure optimization: omit code pointer when not used

vouillon:
This reduces the amount of memory allocated. But I think this also allows Binaryen to make more optimizations, since it becomes clear that the function is only called from known places.

### DIFF
--- a/compiler/lib-wasm/call_graph_analysis.ml
+++ b/compiler/lib-wasm/call_graph_analysis.ml
@@ -1,0 +1,61 @@
+open! Stdlib
+open Code
+
+let debug = Debug.find "call-graph"
+
+let get_approx info x =
+  (* Specialization can add some variables *)
+  if Var.idx x < Var.Tbl.length info.Global_flow.info_approximation
+  then Var.Tbl.get info.Global_flow.info_approximation x
+  else Top
+
+let block_deps ~info ~non_escaping ~ambiguous ~blocks pc =
+  let block = Addr.Map.find pc blocks in
+  List.iter block.body ~f:(fun i ->
+      match i with
+      | Let (_, Apply { f; _ }) -> (
+          match get_approx info f with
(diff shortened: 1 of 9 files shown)


compiler/lib-wasm/generate.ml
```diff
@@ -1520,6 +1513,7 @@ let start () = make_context ~value_type:Gc_target.Type.value
 
 let f ~context ~unit_name p ~live_vars ~in_cps ~deadcode_sentinal ~global_flow_data =
   let state, info = global_flow_data in
+  let fun_info = Call_graph_analysis.f p info in
```
hhugo, line 1516: We're missing debug timing for Call_graph_analysis and Typing

hhugo:
compilation time for fiat-crypto increases from 128s to 135s. Do you know where it's spent ?

vouillon:
> compilation time for fiat-crypto increases from 128s to 135s. Do you know where it's spent ?

I think `wasm-opt` takes longer to execute.

compiler/lib-wasm/call_graph_analysis.ml
```diff
@@ -0,0 +1,63 @@
...
+            | Values { known; others } ->
+                if others || Var.Set.cardinal known > 1
+                then Var.Set.iter (fun x -> Var.Hashtbl.replace ambiguous x ()) known;
+                if debug ()
+                then
+                  Format.eprintf
+                    "CALL others:%b known:%d@."
+                    others
+                    (Var.Set.cardinal known)
+          with Invalid_argument _ -> ())
```
hhugo: This deserves a comment

vouillon: Specialization can add some variables. I'm now using an helper function which checks that `Var.Tbl.get` will not fail.

PR #2070, opened 2025-07-04
Add Intl.RelativeTimeFormat

toastal:
https://tc39.es/ecma402/#relativetimeformat-objects

NOTE: I do not understand how these comment tests work

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
   `AbortController`/`AbortSignal` primitive for cancellation (#596)
 * Wasm_of_ocaml: alternative effect implementation based on the Stack Switching proposal (#2189)
 * Lib: implement popover API (#1734)
+* Lib: add Intl.RelativeTimeFormat (#2070)
 
 ## Bug fixes
 * Compiler: bound the statement-nesting depth of generated functions by
@@ -201,12 +202,11 @@
 * Ppx: allow "function" in object literals (#1897)
 * Lib: add Dom_html.window.matchMedia & Dom_html.mediaQueryList (#2017)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
-* Lib: add Intl.RelativeTimeFormat (#2070)
 
 ## Bug fixes
 * Compiler: fix stack overflow issues with double translation (#1869)
 * Compiler: minifier fix (#1867)
-* Compiler: fix shortvar with --enable es6 (AssignTarget was not properly handled)  
+* Compiler: fix shortvar with --enable es6 (AssignTarget was not properly handled)
 * Compiler: fix assert failure with double translation (#1870)
 * Compiler: fix path rewriting of Wasm source maps (#1882)
 * Compiler: fix global dead code in presence of dead tailcall (#2010)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -201,6 +201,7 @@
 * Ppx: allow "function" in object literals (#1897)
 * Lib: add Dom_html.window.matchMedia & Dom_html.mediaQueryList (#2017)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
+* Lib: add Intl.RelativeTimeFormat (#2070)
 
 ## Bug fixes
 * Compiler: fix stack overflow issues with double translation (#1869)
--- a/lib/js_of_ocaml/intl.mli
+++ b/lib/js_of_ocaml/intl.mli
@@ -346,9 +346,16 @@ then (
       (intl##._PluralRules##supportedLocalesOf
          (jas [| "ban"; "id-u-co-pinyin"; "de-ID" |])
          (def options))
+
+    let options = Intl.RelativeTimeFormat.options () in
+    let () = options##.numeric := def (string "auto") in
+    let () = options##.style := def (string "short") in
+    let th_rtf = new%js Intl.relativeTimeFormat_constr (def (jas [| "th-TH" |])) (def options) in
+    fc (th_rtf##.format -1 "day");
   with Error err -> Console.console##debug (string (string_of_error err)))
 else Console.console##debug (string "Intl is not supported!")
 ]}
+
     @see <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl> for API documentation.
     @see <https://www.ecma-international.org/ecma-402/1.0/> for the ECMAScript specification. *)
 
--- a/lib/js_of_ocaml/intl.mli
+++ b/lib/js_of_ocaml/intl.mli
@@ -627,6 +627,47 @@ module PluralRules : sig
   end
 end
 
+module RelativeTimeFormat : sig
+  include Shared
+
+  class type resolved_options = object
+    method locale : Js.js_string Js.t Js.readonly_prop
+
+    method style : Js.js_string Js.t Js.readonly_prop
+
+    method numberingSystem : Js.js_string Js.t Js.readonly_prop
+
+    method numeric : Js.js_string Js.t Js.readonly_prop
+  end
+
+  class type options = object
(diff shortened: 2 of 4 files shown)


lib/js_of_ocaml/intl.mli
```diff
@@ -349,6 +349,13 @@ then (
   with Error err -> Console.console##debug (string (string_of_error err)))
 else Console.console##debug (string "Intl is not supported!")
 ]}
+
+    let options = Intl.RelativeTimeFormat.options () in
```
hhugo: This should move inside the `{[ ]}` block above

toastal: Moved it, but I still don’t understand how it works

hhugo: This comment is a doc comment.
It should follow the odoc syntax.
Code inside `{[ ... ]}` is a code block, see https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#code_blocks

toastal: So it’s not like… executed? Just a documentation comment.

lib/js_of_ocaml/intl.mli
```diff
@@ -627,6 +634,48 @@ module PluralRules : sig
...
+
+  class type format_part = object
+    method _type : Js.js_string Js.t Js.readonly_prop
+
+    method _value : Js.js_string Js.t Js.readonly_prop
+  end
+
+  class type t = object
+    method format :
+      (Js.number Js.t -> Js.js_string Js.t -> Js.js_string Js.t) Js.readonly_prop
```
hhugo: Why is this a readonly_prop and not a method  ?

lib/js_of_ocaml/intl.mli
```diff
@@ -627,6 +634,48 @@ module PluralRules : sig
...
+
+    method formatToParts :
+      Js.date Js.t Js.optdef -> format_part Js.t Js.js_array Js.t Js.meth
```
hhugo: This signature is not correct. this method takes a number and a string

lib/js_of_ocaml/intl.mli
```diff
@@ -346,9 +346,16 @@ then (
       (intl##._PluralRules##supportedLocalesOf
          (jas [| "ban"; "id-u-co-pinyin"; "de-ID" |])
          (def options))
+
+    let options = Intl.RelativeTimeFormat.options () in
+    let () = options##.numeric := "auto" in
+    let () = options##.style := "short" in
```
hhugo:
```suggestion
    let () = options##.numeric := def (string "auto") in
    let () = options##.style := def (string "short") in
```

lib/js_of_ocaml/intl.mli
```diff
@@ -346,9 +346,16 @@ then (
...
+    let th_rtf = new%js Intl.relativeTimeFormat_constr (def (jas [| "th-TH" |])) options in
```
hhugo:
```suggestion
    let th_rtf = new%js Intl.relativeTimeFormat_constr (def (jas [| "th-TH" |])) (def options) in
```

lib/js_of_ocaml/intl.mli
```diff
@@ -346,9 +346,16 @@ then (
...
+    fc (th_rtf.format -1 "day");
```
hhugo: This is not correct

lib/js_of_ocaml/intl.mli
toastal: I had read the (wrong) `DateTimeFormat.prototype.formatToParts` instead of `RelativeTime` from the spec.

lib/js_of_ocaml/intl.mli
toastal:
```console
$ rg "method format\s" -A 1 lib/js_of_ocaml/intl.ml
234:    method format : (Js.date Js.t -> Js.js_string Js.t) Js.readonly_prop
235-
--
366:    method format : (Js.number Js.t -> Js.js_string Js.t) Js.readonly_prop
367-
--
457:    method format :
458-      (Js.number Js.t -> Js.js_string Js.t -> Js.js_string Js.t) Js.readonly_prop
```

Why are all other `format`s `readonly_prop`?

lib/js_of_ocaml/intl.mli
toastal:
```console
> (new Intl.RelativeTimeFormat("th-TH", { numeric: "auto", style: "short" })).format(-1, "day")
"เมื่อวาน" 
```

Expected output for the given language in the browser console.

“This is not correct” isn’t a helpful comment for review. I think I need the `##` but I don’t know what that operator is doing. I believe `#` is used for OCaml objects (which are almost never talked about) so I would guess this `##` is something similar but specific to `js_of_ocaml`? I am just trying to match the patterns I see in surrounding code to get access to the missing `Intl.RelativeTimeFormat`.

vouillon: I think that should be:
```
fc (th_rtf##format (-1) (string "day"));
```

lib/js_of_ocaml/intl.mli
hhugo: It comes down to a quirk of the spec. For `Intl.NumberFormat` and `Intl.DateTimeFormat`, `format` is **not a method** — it's an accessor property (a getter) that returns a *bound* function (its `this` is already bound to the formatter instance), designed so you can detach it and still have it work, e.g. `numbers.map(nf.format)`. We model that "read a property to obtain a function value" shape with `readonly_prop` holding a function type — hence `(Js.number Js.t -> Js.js_string Js.t) Js.readonly_prop`.

`Intl.RelativeTimeFormat.prototype.format`, on the other hand, is an **ordinary method** (`format(value, unit)`) — there's no bound-function getter, and it takes two arguments. So here the right encoding is a plain method:

```ocaml
method format : Js.number Js.t -> Js.js_string Js.t -> Js.js_string Js.t Js.meth
```

which is what the current version of the PR has, so this is already resolved. The asymmetry with the other modules is intentional and matches the spec.

Refs: [`get Intl.NumberFormat.prototype.format`](https://tc39.es/ecma402/#sec-intl.numberformat.prototype.format) (accessor) vs. [`Intl.RelativeTimeFormat.prototype.format`](https://tc39.es/ecma402/#sec-Intl.RelativeTimeFormat.prototype.format) (method).

CHANGES.md
```diff
@@ -190,6 +163,7 @@
 * Ppx: allow "function" in object literals (#1897)
 * Lib: add Dom_html.window.matchMedia & Dom_html.mediaQueryList (#2017)
 * Lib: make the Wasm version of Json.output work with native ints and JavaScript objects (#1872)
+* Lib: add Intl.RelativeTimeFormat (#2070)
```
vouillon: This is not at the right place

lib/js_of_ocaml/intl.mli
```diff
@@ -627,6 +634,47 @@ module PluralRules : sig
...
+    method style : Js.js_string Js.t Js.optdef Js.prop
+
+    method numeric : Js.js_string Js.t Js.optdef Js.prop
+  end
+
+  val options : unit -> options Js.t
+
+  class type format_part = object
+    method _type : Js.js_string Js.t Js.readonly_prop
+
```
vouillon, line 664:
```ocaml
method unit : Js.js_string Js.t Js.optdef Js.readonly_prop
```

vouillon (changes requested):
This looks good. File `API-STATUS.md` should be updated as well.

PR #2075, opened 2025-07-07
Don't create loop-exiting edges if not necessary  

hhugo:


hhugo:
> By the way, this trick of adding loop-exiting edges does not work when the loop header has no predecessor.

I'm guessing this can only happen if the header is the first block of a function

hhugo:
> > By the way, this trick of adding loop-exiting edges does not work when the loop header has no predecessor.
> 
> I'm guessing this can only happen if the header is the first block of a function

We could prepend an empty block to functions that start with a loop header.

hhugo:
> We could prepend an empty block to functions that start with a loop header.

@vouillon, this is done in #2077

compiler/lib/structure.ml
```diff
@@ -125,7 +125,7 @@ let build_graph blocks pc =
           (Addr.Set.add enter_pc (Addr.Hashtbl.find preds leave_pc))));
   { succs; preds; reverse_post_order = !l; block_order }
 
-let dominator_tree g =
+let dominator_tree' g =
```
vouillon:
```suggestion
let reversed_dominator_tree g =
```

hhugo: done

compiler/lib/structure.ml
```diff
@@ -260,13 +268,15 @@ let shrink_loops blocks ({ succs; preds; reverse_post_order; _ } as g) =
         (* If we leave a loop, we add an edge from predecessors of
            the loop header to the current block, so that it is
            considered outside of the loop. *)
-        if not (Addr.Set.is_empty left_loops || is_small blocks g pc')
+        if not (Addr.Set.is_empty left_loops)
         then
```
vouillon:
```suggestion
```

PR #2077, opened 2025-07-07
Compiler: exit loop early, even leading loops

hhugo:


compiler/lib/structure.mli
```diff
@@ -24,3 +24,5 @@ val sort_in_post_order : t -> Addr.t list -> Addr.t list
 val blocks_in_reverse_post_order : t -> Code.Addr.t list
 
 val get_nodes : t -> Addr.Set.t
+
+val norm : program -> program
```
vouillon, line 28: A comment would be helpful. Something like:
> Ensure that all loops have a predecessor block. Function `Structure.shrink_loops` assumes this.

PR #2080, opened 2025-07-15
fs_fake: implement rename

jonludlam:


runtime/js/fs_fake.js
```diff
@@ -81,6 +81,17 @@ class MlFakeDevice {
     }
   }
 
+  rename(oldname, newname) {
+    if (this.exists(newname))
+      caml_raise_sys_error(this.nm(newname) + " : file already exists");
```
hhugo: This is not the expected behavior.
The doc says
> If a file named
    [newpath] already exists, its contents will be replaced with those of
    [oldpath].

runtime/js/fs_fake.js
```diff
@@ -81,6 +81,17 @@ class MlFakeDevice {
...
+    if (!this.exists(oldname))
+      caml_raise_sys_error(this.nm(oldname) + " : no such file or directory");
+    if (this.is_dir(oldname))
+      caml_raise_sys_error(this.nm(oldname) + " : not a file");
```
hhugo: The doc says:
> Rename a file or directory.  [rename oldpath newpath] renames the
    file or directory called [oldpath], giving it [newpath] as its new name,
    moving it between (parent) directories if needed. 

We shouldn't fail on directory.

You should probably add tests in `compiler/tests-compiler/sys_fs.ml`

compiler/tests-compiler/sys_fs.ml
```diff
@@ -59,6 +59,63 @@ f (); Sys.chdir "/static"; f ()
...
+in
+f (); Sys.chdir "/static"; f ()
+  |};
+  [%expect {|
+    new file contents: Hello world
+    new file contents: Hello world
+    |}]
+
+let%expect_test _ =
+  (* Check we can rename a file to a pre-existing file *)
```
hhugo: Can you add
- a test renaming a directory into an existing (not a director) file.
- a test renaming a file (not a directory) into an existing directory

runtime/js/fs_fake.js
```diff
@@ -81,6 +81,27 @@ class MlFakeDevice {
     }
   }
 
+  rename_dir(oldname, newname) {
+    var old_slash = this.slash(oldname);
+    var new_slash = this.slash(newname);
+    this.create_dir_if_needed(new_slash);
+    this.readdir(oldname).forEach((f) => {
+      this.rename(old_slash + f, new_slash + f);
+    });
```
smorimoto: Biome prefers `for...of` instead of `forEach`

compiler/tests-compiler/sys_fs.ml
jonludlam: done, plus one more - renaming a directory over a non-empty directory ought to fail.

runtime/js/fs_fake.js
jonludlam: Somehow I find that more surprising than some of the other js<->OCaml differences! Done though.

hhugo:
I might have a fix for the CI in https://github.com/ocsigen/js_of_ocaml/pull/2080/commits/ec6091331d2f49bd5cb8274e1fee6b8be5e621a6

PR #2081, opened 2025-07-17
Wasm/runtime: put back caml_string_of_array

vouillon:
Brr is using this function. This is an alias for `caml_string_of_uint8_array`.


dbuenzli:
We can also fix `brr`.

vouillon:
I was going to suggest a fix, indeed. `caml_string_of_uint8_array` was introduced in Js_of_ocaml 6.0.1.

dbuenzli:
I promise to do one until before next monday.

PR #2086, opened 2025-07-28
Runtime: allow one to directly access values exported in the runtime   

hhugo:
The goal is to address https://github.com/LexiFi/gen_js_api/issues/177
cc [mlasson](https://github.com/mlasson)

@vouillon, can this work with wasmoo ?

### DIFF
--- a/compiler/lib/specialize_js.ml
+++ b/compiler/lib/specialize_js.ml
@@ -288,6 +288,35 @@ let specialize_instrs ~target opt_count info l =
     match l with
     | [] -> List.rev acc
     | i :: r -> (
+        let i =
+          match i with
+          | Let (x, Apply { f; args; exact = false }) -> (
+              match Info.def info f with
+              | None -> i
+              | Some (Prim (Extern "caml_jsoo_runtime", [ name ])) -> (
+                  let name =
+                    match name with
+                    | Pc (String name) -> Some name
+                    | Pc _ -> None
+                    | Pv x -> (
+                        match Info.def info x with
+                        | Some (Constant (String name)) -> Some name
+                        | Some _ | None -> None)
+                  in
+                  match name with
+                  | None -> i
+                  | Some name -> (
+                      let name = Primitive.resolve name in
+                      match Primitive.arity name with
+                      | exception Not_found -> i
+                      | n ->
+                          if List.compare_length_with args ~len:n = 0
+                          then
+                            Let (x, Prim (Extern name, List.map args ~f:(fun x -> Pv x)))
+                          else i))
+              | Some _ -> i)
+          | _ -> i
+        in
         (* We make bound checking explicit. Then, we can remove duplicated
            bound checks. Also, it appears to be more efficient to inline
            the array access. The bound checking function returns the array,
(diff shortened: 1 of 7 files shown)


examples/boulderdash/boulderdash.ml
```diff
@@ -510,4 +510,12 @@ let start _ =
...
   Lwt.return ()
 
-let () = Lwt.async start
+let () =
+  let p : Js.js_string Js.t = Jsoo_runtime.Sys.external_ "process" in
+  let o : _ Js.t = Jsoo_runtime.Sys.external_ "obj" in
+  let del : 'a -> Jsoo_runtime.Js.t -> unit =
+    Jsoo_runtime.Sys.external_ "caml_js_delete"
+  in
+  del o (Jsoo_runtime.Js.string "process");
```
vouillon: `del` is a JavaScript function. You should not use it as if it was an OCaml function.
```suggestion
  ignore (Jsoo_runtime.Js.fun_call del [| o; Jsoo_runtime.Js.string "process" |]);
```

vouillon:
I have pushed something. It only works with whole-program compilation, though. With separate compilation, only the JavaScript primitives used in the Wasm runtime are linked-in. I'm not sure what to do about that.

compiler/lib/specialize_js.ml
```diff
@@ -288,6 +293,35 @@ let specialize_instrs ~target opt_count info l =
...
+                      match Primitive.arity name with
+                      | exception Not_found -> i
+                      | n ->
+                          if List.compare_length_with args ~len:n = 0
+                          then
+                            Let (x, Prim (Extern name, List.map args ~f:(fun x -> Pv x)))
+                          else i))
+              | Some _ -> i)
+          | _ -> i
+        in
```
vouillon: I think we should remove this change (and the associated change in [compiler/lib/flow.ml](https://github.com/ocsigen/js_of_ocaml/pull/2086/files#diff-78c5ff10d89005bd3e1bf13ec1d41119d5b30d73679d23b7ccbd7e006c798bd8).

hhugo: Sounds good. Can you push this change ? I'm away from keyboard

lib/runtime/jsoo_runtime.ml
```diff
@@ -124,6 +124,8 @@ module Sys = struct
   external restore_channel : out_channel -> redirection -> unit
     = "caml_ml_channel_restore"
 
+  external external_ : string -> 'a = "caml_jsoo_runtime"
```
hhugo: I'm not very happy with the name `Jsoo_runtime.Sys.external_` and its signature.

Maybe `val Jsoo_runtime.Js.runtime : string -> Jsoo_runtime.Js.t` would be a better fit. @vouillon, any thoughts ?

vouillon: What about `runtime_value`?

hhugo: You mean `Jsoo_runtime.Js.runtime_value` ?

PR #2090, opened 2025-08-26
Make test independent from build directory state

OlivierNicole:
The modified test could fail if the directory it's run in already contained a file named `append.txt`, which sometimes happened when the test has already been run previously by Dune, possibly with a different backend or profile.


OlivierNicole:
Since the compiler can be built with Dune 3.19 or newer, it can still be useful to correctly run the testsuite locally.

hhugo:
> I’m afraid I don’t know what they are

https://dune.readthedocs.io/en/stable/concepts/locks.html

PR #2092, opened 2025-08-27
Revert "Runtime: fix annotation of caml_compare"

### LINKED ISSUE #2089: [BUG] Change of behaviour in function comparison
The test `compiler/tests-ocaml/basic/equality.ml` now fails on master (dcec0ac3a1 at time of writing), although it was passing on the same commit last week.

The error:

```diff
 --- a/_build/default/compiler/tests-ocaml/basic/equality.expected
+++ b/_build/default/compiler/tests-ocaml/basic/equality.bc.js.output
@@ -47,9 +47,9 @@ Test 52 passed.
 Test 53 passed.
 Test 54 passed.
 Test 55 passed.
-Test 56 passed.
-Test 57 passed.
+Test 56 FAILED.
+Test 57 FAILED.
 Test 58 passed.
-Test 59 passed.
+Test 59 FAILED.
 Test 60 passed.
 Test 61 passed.
```

OlivierNicole:
This reverts commit 3ceafdf42b62c96225f0d482fa29ca0f1ab589f9: `caml_compare` can raise `Invalid_argument`.

Fixes #2089.

### DIFF
--- a/runtime/js/compare.js
+++ b/runtime/js/compare.js
@@ -251,7 +251,7 @@ function caml_compare_val(a, b, total) {
     b = b[i];
   }
 }
-//Provides: caml_compare mutable (const, const)
+//Provides: caml_compare (const, const)
 //Requires: caml_compare_val
 function caml_compare(a, b) {
   return caml_compare_val(a, b, true);


OlivierNicole:
Isn’t it the opposite? `mutable` seems to mean that there are no side effects.

vouillon:
We have three purity annotations:
- `mutator` (the default): the primitive may have side-effects
- `mutable`: the primitive does not have side-effects, but cannot be reordered with effectful operations (for instance, when one of the parameter is a mutable datastructure)
- `pure`: the primitive will always return the same value (given the same parameters).

OlivierNicole:
I understand now, thanks!

PR #2108, opened 2025-10-13
Wasm runtime: provide access to JavaScript eval function

vouillon:


runtime/wasm/runtime.js
```diff
@@ -211,6 +211,8 @@
     typeof: (x) => typeof x,
     // biome-ignore lint/suspicious/noDoubleEquals:
     equals: (x, y) => x == y,
+    // biome-ignore lint/security/noGlobalEval:
+    eval: (x) => globalThis.eval("(" + x + ")"),
```
hhugo: There is no additional parenthesis in the js runtime. The js and wasm runtime need to agree.

vouillon: I propose we modify the JS runtime, so that the primitives work for arbitrary JS expressions.

Before this was optimized with Wasm_of_ocaml as well, I have seen code that was working with Js_of_ocaml since the expression was inlined, but not with an `eval` without parentheses. Not sure what it was, but maybe an object literal.

runtime/js/jslib.js
```diff
@@ -479,23 +479,23 @@ function caml_js_strict_equals(x, y) {
 //Requires: caml_jsstring_of_string
 function caml_js_eval_string(s) {
   // biome-ignore lint/security/noGlobalEval:
-  return eval(caml_jsstring_of_string(s));
+  return eval?.('"use strict";' + caml_jsstring_of_string(s));
```
hhugo, line 487: I would duplicate part of the commit message as a comment here. Together with a link explaining the optinal chaining trick.

runtime/wasm/runtime.js
hhugo: Adding the extra parens means that we now accept invalid inputs (ie `"a)(b"`).
Maybe we can live with that but we need a comment explaining the trade-off

vouillon: I have added this to the comment in function `caml_js_expr` in `runtime/js/jslib.js`.

runtime/wasm/jslib.wat
```diff
@@ -128,6 +129,13 @@
       (ref.i31 (call $strict_equals
                   (call $unwrap (local.get 0)) (call $unwrap (local.get 1)))))
 
+   (func (export "caml_js_expr") (export "caml_pure_js_expr")
+         (export "caml_js_var") (export "caml_js_eval_string")
```
hhugo: `caml_js_eval_string` doesn't use extra parens in js but does in wasm.

vouillon: Oh, indeed! I missed that. The Wasm code was wrong.

hhugo:
I see the following in `gc_target.ml`
```ocaml
        | [ Pv _ ] ->
            let* () =
              register_fragment "eval" (fun () ->
                  let lex = Parse_js.Lexer.of_string {|(x)=>eval("("+x+")")|} in
                  Parse_js.parse_expr lex)
            in
```

Shouldn't we use an indirect call here as well ?

runtime/wasm/jslib.wat
```diff
@@ -128,6 +131,22 @@
...
+   (func $caml_js_eval_string (export "caml_js_eval_string")
+      (param $vs (ref eq)) (result (ref eq))
+      (local $s (ref $bytes))
+      (local.set $s (ref.cast (ref $bytes) (local.get $vs)))
+      (return_call $wrap (call $eval (call $jsstring_of_bytes (local.get $s)))))
+
+   (@string $rparen "(")
+   (@string $lparen ")")
+
+   (func (export "caml_js_expr") (export "caml_pure_js_expr")
```
hhugo: Is this ever going to be used ? I would expect gc_target to always translate theses primitives

hhugo:
@vouillon, there are few pending comments before we can merge.

runtime/wasm/jslib.wat
hhugo: ping @vouillon

vouillon: Indeed, it is simpler to remove this code.

PR #2111, opened 2025-10-13
Get magic number version from compiler-libs

vouillon:


compiler/lib/dune
```diff
@@ -3,8 +3,8 @@
  (public_name js_of_ocaml-compiler)
  (synopsis "Js_of_ocaml compiler library")
  (libraries
-  compiler-libs.common
-  compiler-libs.bytecomp
+  ocaml-compiler-libs.common
```
hhugo, line 6: Is this really needed ?

vouillon: `Config.exec_magic_number` is shadowed by our own module `Config`. Do you see another way to access it?
 I encountered the same issue in #1721 (see 4fbb798b9642e4d576b1d43d17ed9758ae07e128), since module `Primitive` is also shadowed.

`ocaml-compiler-libs.common` is a dependency of `ppxlib`, so this is not really a new dependency.

PR #2116, opened 2025-10-24
Primitive checks: remove no longer used output files

vouillon:


hhugo:
You need to update https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/tests-check-prim/gen_dune.ml

hhugo (approved):

PR #2117, opened 2025-10-24
ppx_optcomp_light improvements

vouillon:


compiler/ppx/ppx_optcomp_light.ml
```diff
@@ -148,67 +149,78 @@ let keep loc (attrs : attributes) =
...
+            match
+              (parse_res
+                 (pexp_ident (lident (string "ocaml_version"))
+                 >>| (fun () -> Version Version.current)
+                 ||| (pexp_ident (lident (string "ast_version"))
+                     >>| fun () -> Int Ppxlib.Selected_ast.version)
+                 ||| (pexp_construct (lident (string "true")) drop >>| fun () -> Bool true)
+                 ||| (pexp_constant (pconst_integer __ none)
+                     >>| fun () d -> Int (int_of_string d))
+                 ||| (pexp_construct (lident (string "false")) drop
```
hhugo, line 169: Move this branch after the `"true"` one

compiler/ppx/ppx_optcomp_light.ml
```diff
@@ -285,7 +297,21 @@ let traverse =
             | Some pattern -> Some { case with pc_lhs = pattern })
       in
       super#cases cases
+
+    method! signature_item item =
```
hhugo, line 312: What about matching on signature instead of individual items ? Similar to what we do for structures?

vouillon: OxCaml  defines signatures [using a record](https://github.com/oxcaml/oxcaml/blob/419927ca9e360aec416c1104c633061b8d554ac4/parsing/parsetree.mli#L1097-L1102):
```ocaml
and signature =
  {
    psg_modalities : modalities;
    psg_items : signature_item list;
    psg_loc : Location.t;
  }
```
while mainstream OCaml uses [a plain list](https://github.com/ocaml/ocaml/blob/e50c56797dee0f98e3a6d9b942a9b5385be9d75e/parsing/parsetree.mli#L904C1-L904C36):
```ocaml
and signature = signature_item list
```
I did not see any other way to write some code compatible with both.

hhugo: This is unfortunate. Do we really need conditional sig item ? Can you point me to the use cases ?

hhugo: An alternative would be emit a floating attribute for filtered out items.

hhugo: I've pushed a commit doing that, let me know what you think

vouillon: The use case is when we want to use different types depending on the OCaml version.
https://github.com/ocsigen/js_of_ocaml/blob/382a62bc058b25c5bcdbeb9d1493cd64881978b7/compiler/lib/ocaml_compiler.mli#L76-L88

vouillon: Your alternative is fine with me.

PR #2118, opened 2025-10-24
Cmdliner v2

smorimoto:


dbuenzli:
Also the changes have nothing to do with cmdliner. They propagate the exit code of `main` to let `exit` pick them up (which apparently was not done before). 

At that point you could also replace the `exit` calls in the error branches by the code number so that the program has a single place where it `exit`s.

PR #2124, opened 2025-11-03
Optimizations: string conversion and copy to and from arraybuffer

vouillon:


hhugo (approved):
I didn't review very carefully, but it looks good

PR #2140, opened 2026-01-13
Misc: install completion

hhugo:


dbuenzli:
The failure pattern is "interesting", but it seems the `dune` instructions trigger a bug in `dune` itself.

WardBrian:
> The failure pattern is "interesting", but it seems the `dune` instructions trigger a bug in `dune` itself.

The pattern used here is almost exactly what we [use in stanc3](https://github.com/stan-dev/stanc3/blob/master/src/stanc/dune). 

Requesting `dune build compiler/bin-js_of_ocaml/cmdliner-support compiler/bin-wasm_of_ocaml/cmdliner-support` generates the expected files in the _build directory.

So, it seems like the issue comes specifically from specifying two `share_root` install targets with the `(dir ...)` atom. 
Comment either one out and it will build, so at least we have a fairly specific bug to report...

nojb:
Unfortunately, this PR does not seem enough to be able to build `js_of_ocaml` in a workspace (in a monorepo, without OPAM): ocaml/dune#13500.

hhugo:
> > Can you check with https://github.com/ocsigen/js_of_ocaml/tree/fix-completion ? The downside is that it could break if cmdliner changes the set of files it generate.
> 
> I confirm it works, thanks!

Do you need the fix the upcomming release ?

PR #2151, opened 2026-01-29
Fix miscompilation of loops at start of exception handlers

vouillon:


compiler/tests-compiler/loop_in_handler.ml
```diff
@@ -0,0 +1,51 @@
+open Util
+
+let%expect_test "deadcode elimination of cyclic values" =
```
hhugo: The name should be updated

PR #2155, opened 2026-02-03
Doc: document API status in jsoo and brr

hhugo:


API-STATUS.md
```diff
@@ -0,0 +1,231 @@
...
+|-----|------|-----|---------------------|
+| FormData | Yes | Yes | `Form` · Brr: `Brr_io.Form.Data` |
+
+## Graphics
+
+| API | jsoo | Brr | jsoo Module / Notes |
+|-----|------|-----|---------------------|
+| Canvas 2D | Yes | Yes | `Dom_html` (canvasRenderingContext2D) · Brr: `Brr_canvas.C2d` |
+| WebGL | Yes | Partial | `WebGL` · Brr: `Brr_canvas.Gl` (WebGL 1 only) |
+| WebGL2 | No | No | [#1226](https://github.com/ocsigen/js_of_ocaml/issues/1226) |
```
dbuenzli, line 87: This is incorrect, `Brr` supports WebGL2 and `WebGL 1` in an usafe manner.

hhugo: Ok, will update

dbuenzli: Note I didn't check thoroughly because I suspect this document has been produced by an LLM and not been checked by human.

dbuenzli: > Note I didn't check thoroughly because I suspect this document has been produced by an LLM and not been checked by human.

Also if that's the case could you please indicate it at the start of the document, so that it can be adequately treated as disinformation.

hhugo: Correct, and I aim to have it used by LLM only. I will add note about this and its correctness implication.

PR #2157, opened 2026-02-03
Explicitly track generated names

vouillon:


hhugo:
This PR seems to have bumped the compilation memory usage for PRT quite a bit. @vouillon do you want to take a look ?

PR #2159, opened 2026-02-03
Lib: fix onbeforeunload handler breaking navigation (#1436) 

### LINKED ISSUE #1436: [BUG] semantics of (onbeforeunload, Dom_html.handler) incompatible with major browsers
**Describe the bug**
Adding/removing hooks for the beforeunload event with `Dom_html.window##.onbeforeunload := Dom_html.handler …` is not working out-of-the-box with both Chromium, Firefox, and Safari. So I had to devise the following hack (unsafe JS code):

```ocaml
  let prompt_before_unload () : unit =
    Js.Unsafe.js_expr "window.onbeforeunload = function(e) {e.preventDefault(); return false;}" in

  let resume_before_unload () : unit =
    Js.Unsafe.js_expr "window.onbeforeunload = null" in

  (…)
  if not sync then prompt_before_unload () else resume_before_unload ()
  (…)
```
— *Source code:* [https://github.com/ocaml-sf/learn-ocaml/blob/a6e4c5e6/src/app/learnocaml_exercise_main.ml#L292-L310](https://github.com/ocaml-sf/learn-ocaml/blob/a6e4c5e61362f569d6279d5d990705d3ac1d4f94/src/app/learnocaml_exercise_main.ml#L292-L310)

**Expected behavior**
I open this PR in case it is possible for you to tweak `Dom_html.handler`, and be able to directly use js_of_ocaml's abstractions (**not** `Js.Unsafe.js_expr` :) for this use case, **and** stay compatible with (Chromium, Firefox, Safari…)

FTR → I had sketched a possible explanation of why the expected code doesn't work in Firefox here:
https://github.com/ocaml-sf/learn-ocaml/issues/467#issuecomment-1059802295

**Versions**
Version of packages used to reproduce the bug:
- ocaml 4.12.1
- js_of_ocaml 3.9.0

(I'll be happy to provide more details on the versions used if need be)

Cc @yurug FYI

hhugo:
## Summary

Fixes #1436.

`Dom.handler` converts the OCaml `bool t` return value to a JavaScript return value. When a `beforeunload` handler returns `Js._true` (allow navigation), the JS value `true` is passed to the browser. Modern browsers interpret **any non-null, non-undefined return value** from a `beforeunload` handler as a request to show the "Leave page?" confirmation dialog. This means that even when the OCaml handler intended to allow navigation, the browser would show a blocking dialog.

## Changes

The fix detects `beforeunload` events at runtime in `Dom.handler` and `Dom.full_handler`:
- When the handler returns `Js._true` (allow navigation), return `undefined` to the browser instead of `true`, so no dialog is shown.
- When the handler returns `Js._false` (block navigation), also set `event.returnValue = ""` for legacy browser compatibility.
- In `Dom.invoke_handler`, normalize `undefined` back to `Js._true` so callers that chain handlers still see a proper `bool t` value.

Also adds `Dom.beforeUnloadEvent` and `Dom_html.beforeUnloadEvent` class types with the `returnValue` property, and types `onbeforeunload` to use `beforeUnloadEvent` instead of the generic `event` type.

### DIFF
--- a/lib/js_of_ocaml/dom_html.mli
+++ b/lib/js_of_ocaml/dom_html.mli
@@ -913,6 +913,12 @@ and eventTarget = object ('self)
   method dispatchEvent : event t -> bool t meth
 end
 
+and beforeUnloadEvent = object
+  inherit event
+
+  method returnValue : js_string t prop
+end
+
 and popStateEvent = object
   inherit event
 
@@ -3458,7 +3464,7 @@ class type window = object
 
   method onunload : (window t, event t) event_listener prop
 
-  method onbeforeunload : (window t, event t) event_listener prop
+  method onbeforeunload : (window t, beforeUnloadEvent t) event_listener prop
 
   method onblur : (window t, focusEvent t) event_listener prop
 
@@ -3660,7 +3666,7 @@ module Event : sig
 
   val unload : event t typ
 
-  val beforeunload : event t typ
+  val beforeunload : beforeUnloadEvent t typ
 
   val resize : event t typ
 
--- a/lib/js_of_ocaml/dom_html.mli
+++ b/lib/js_of_ocaml/dom_html.mli
@@ -913,12 +913,6 @@ and eventTarget = object ('self)
   method dispatchEvent : event t -> bool t meth
 end
 
-and beforeUnloadEvent = object
-  inherit event
-
-  method returnValue : js_string t prop
-end
-
 and popStateEvent = object
   inherit event
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -89,6 +89,7 @@
 * Runtime/wasm: fix unmarshalling of compressed data (#2141)
 * Runtime: fix compilation of loops at start of exception handlers (#2151)
 * Compiler: fix parallel renaming (#2156)
+* Lib: fix `onbeforeunload` handler breaking navigation (#1436)
 
 # 6.2.0 (2025-07-30) - Lille
 
--- a/lib/js_of_ocaml/dom.ml
+++ b/lib/js_of_ocaml/dom.ml
@@ -454,20 +454,27 @@ let no_handler : ('a, 'b) event_listener = Js.null
    the event), we must return [undefined] to the browser instead of [true]
    so that it does not trigger the dialog.  When the handler returns [false]
    (block the event), we also set [returnValue] for legacy browser compat. *)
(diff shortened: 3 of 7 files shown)


hhugo:
@vouillon, any though ?

lib/js_of_ocaml/dom_html.mli
```diff
@@ -913,6 +913,12 @@ and eventTarget = object ('self)
   method dispatchEvent : event t -> bool t meth
 end
 
+and beforeUnloadEvent = object
```
vouillon: This class type is already there.

hhugo:
@vouillon, I don't know if this goes in the right direction .. If you think it does, we'd have to fix the janestreet test

vouillon:
What about making these changes instead?
```ocaml
type ('a, 'b) event_listener = ('a, 'b -> bool t optdef) meth_callback opt

let handler f =
  Js.some
    (Js.Unsafe.callback (fun e ->
         let res = f e in
         if Js.to_bool res
         then Js.undefined
         else (e##preventDefault; Js.def Js._false)))

let invoke_handler (f : ('a, 'b) event_listener) (this : 'a) (event : 'b) :
      bool t =
  Js.Optdef.get (Js.Unsafe.call f this [| Js.Unsafe.inject event |])
    (fun () -> Js._true)

let beforeunload_handler (f : beforeUnloadEvent t -> bool) :
      ('a, beforeUnloadEvent t) Dom.event_listener =
  handler (fun e ->
      if f e then (e##.returnValue := Js.string "x"; Js._false)
      else Js._true)
```
It might also make sense to add these functions:
```ocaml
let listener f : ('a, 'b) event_listener =
  Js.some
    (Js.Unsafe.callback (fun e ->
         f e;
         if Js.to_bool e##.defaultPrevented
         then Js.def Js._false
         else Js.undefined))

let full_listener f : ('a, 'b) event_listener =
  Js.some
    (Js.Unsafe.meth_callback (fun this e ->
         f this e;
         if Js.to_bool e##.defaultPrevented
         then Js.def Js._false
         else Js.undefined))
```

hhugo:
Do we need beforeunload_handler if we have listener ?

vouillon:
> Do we need beforeunload_handler if we have listener ?

No, we don't really need it. It is defined in term of `Dom.handler`, after all.

hhugo:
closing in favor of #2251

Issue #1532, opened 2023-11-19
[FEATURE REQUEST] allow ECMAScript import and export when defining jsoo stubs


hhugo:
This this possible now that #1529 is here.

Fizzixnerd:
I'd like to see this feature implemented.  I have lots of functional programming experience, though I'm new-ish to ocaml (I come from a mostly Haskell and F# background).  I'd like to give this a shot, but would need some direction definitely to at least getting started.  If anyone is able to provide such direction, I would be greatly appreciative.  I cannot seem to find out where on Matrix/IRC I should go for a more realtime chat experience either, but that could be because no such channel exists.  I will think a bit about a more details spec on what I'd imagine this feature to look like.

Fizzixnerd:
One of the first things I think needs to be decided is: can we treat ES6 static module imports as OCaml modules?  Harder still: should we?

Ultimately, I currently think we shouldn't.  It is unclear if it's even possible today, and it's certainly unclear if it will remain possible indefinitely in the future as ES changes.  ES is fundamentally a more dynamic language than OCaml, and I think just accepting that and thinking of ES modules as more akin to (ocaml) objects than (ocaml) modules is probably for the best.

Instead, I think a general ES module should be represented as an OCaml object, and so the "types" of ES modules are just classes.

That is, binding to an ES module should be as easy as declaring an OCaml `class` (or perhaps `class type`) stub.

As an example taken from https://pixijs.com/8.x/playground?exampleId=basic.container of something I'd ultimately like to be able to do:

```ocaml
open Js_of_ocaml

(* For application below *)
type init_args = {
    background : js_string optdef;
    resizeTo : js_any optdef
}

(* Binding for the Application class *)
class%js.class type application = object 
    method init : init_args Js.t -> unit js_promise
    val stage : js_any (* etc. *)
    val ticker : js_any (* etc. *)
    val canvas : js_any (* etc. *)
    (* ... *)
end

class%js.class type assets = .. (* ... *)
class%js.class type container = .. (* ... *)
class%js.class type sprite = .. (* ... *)

(* Uses the same rules for underscores as described in https://ocsigen.org/js_of_ocaml/latest/manual/bindings *)
class%js.module type pixi = object
    val _Application : application
    val _Assets : assets
    val _Container : container
    val _Sprite : sprite
    (* Any object that extends this type would be okay too, which is nice for writing stubs! *)
end

[...]

PR #2163, opened 2026-02-04
Compiler: JS: initial tooling for ecmascript modules

hhugo:
This PR adds initial compiler tooling for working with ECMAScript modules.

**New compiler support:**
- `Esm` module (`compiler/lib/esm.ml`): parses ES modules, extracts imports/exports, builds a module dependency graph (with cycle support), resolves star re-exports, and topologically sorts modules.
- `Esm_bundle` module (`compiler/lib/esm_bundle.ml`): bundles a module graph into a single program — rewrites imports to direct references, generates namespace bindings, optimizes `ns.field` accesses into named imports, and supports tree shaking from entry-point exports.
- Script-to-module conversion.

**Supporting changes:**
- `js_assign.ml`: the program top level is now recorded as a `Params` block instead of `Var_scope`, so that under the `Preserve` naming strategy top-level `var` declarations (which ESM output places directly at the top level, e.g. vars hoisted by tree shaking) get names allocated.
- New test suite `compiler/tests-esm/` (38 tests) using the vendored `ppx_expect_light`.

Preliminary work for: #1532, #1161, #551

### DIFF
--- a/compiler/lib/esm.ml
+++ b/compiler/lib/esm.ml
@@ -1,0 +1,420 @@
+(* Js_of_ocaml compiler
+ * http://www.ocsigen.org/js_of_ocaml/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, with linking exception;
+ * either version 2.1 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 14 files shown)


hhugo:
Pushed four `Compiler: esm:` commits fixing bugs found while building the ESM-output experiment on top of this branch (#2402):

- **ES evaluation order in the bundler** — modules were emitted in an SCC topological order whose tie-break is the module id, so independent sibling modules came out in alphabetical order instead of import-declaration order (also the root cause of the platform-dependent ordering previously seen on Windows CI). New `Esm.evaluation_order`, used by `Esm_bundle.bundle`; two expect tests updated to the ES-correct order.
- **External imports** — unresolvable specifiers (e.g. `node:module`) failed `build_graph`; `resolve` returning `None` now marks the import as external and keeps it verbatim.
- **Tree shaking: entry side effects** — an entry module exporting nothing was shaken to an empty file; entry modules are now always marked reached.
- **Tree shaking vs namespace imports** — `import * as ns` kept every export of the source alive; the `ns.field` → named-import optimization now runs before shaking.

Issue #2122, opened 2025-10-28
[BUG] Stack overflow in some web browsers for an ocaml program that uses menhir to generate its parser


relokin:
Apologies in advance, I haven't been able to isolate the problem and create a minimal test case. I will attempt to explain how the bug manifests itself hoping that I can get some help to debug further.

**Describe the bug**
[herdtools7](https://github.com/herd/herdtools7/) uses `js_of_ocaml` to create a js interface for the herd7 memory model simulator. When running the js through certain browsers (for example, Chrome `141.0.7390.123 (Official Build) (arm64)` on MacOS 15.1), we run into a `Stack overflow` exception. The bug can be reproduced here: https://diy.inria.fr/www/ by clicking at the play button at the bottom of the page. The `Stack overflow` exception is generated when the litmus test is parsed (left hand side window) and more specifically when the body of the main body of the litmus test is parsed (first instruction of P0, in the case of the default litmus test that is `MOV W0,#1`). The parser is generated using `menhir` and the parsing rules can be found here: https://github.com/herd/herdtools7/blob/master/lib/AArch64Parser.mly. 

The `Stack overflow` exception is generated for `AArch64` (one of the architecture supported in herd7), if we switch to `x86` (using the drop down menu in the centre of the web page), the simulator works as expected. `x86` (https://github.com/herd/herdtools7/blob/master/lib/X86Parser.mly) has simpler parser than `AArch64`. I've tried using `--table` to enable menhir's table backend but that didn't help either.

**Expected behavior**
Normally we would expect the parser to correctly parse the default litmus test and herd7 would proceed as normally. I can get the correct output if I run the generated js with some browsers (for example Safari `Version 18.1 (20619.2.8.11.10)` on MacOS 15.1) or node.js.

**Versions**

    js_of_ocaml 5.5.0

I have found that this commit https://github.com/ocsigen/js_of_ocaml/commit/29c693ba166962725d209123bd80bd1793c7ff78 seems to be the culprit, prior to it the generated js works as expected.

hhugo:
You mention version `5.5.0` but your link uses version `5.8.2`. Theses versions are 1.5 / 2 years old, could you upgrade to the latest version `6.2.0` ?

hhugo:
Can you try with latest master. I wonder if #2167 helps ..

4y8:
[...]

PR #2166, opened 2026-02-04
Variable coalescing

vouillon:
Merge variables with disjoint lifespan in JavaScript code. This results in smaller and more efficient code. In particular, the V8 bytecode uses one slot on the stack per local variable which is not used in nested functions. For the toplevel function, this can yield a huge stack frame.

I have made some measurements of the optimization pass as well as the constant sinking pass (#2167) on two programs: ocamlc and PRT.

Metrics:
- **Register count**: Number of registers used by the largest (toplevel) function in V8
- **Uncompressed**: File size in bytes
- **Compressed**: Gzip-compressed file size in bytes

## OCaml (ocamlc)

| Optimization | Registers | Uncompressed | Compressed |
|--------------|-----------|--------------|------------|
| None         | 3005      | 2,351,542    | 606,480    |
| Sinking      | 1778 (-41%) | 2,326,301 (-1.1%) | 600,010 (-1.1%) |
| Coalescing   | 2173 (-28%) | 2,286,633 (-2.8%) | 550,096 (-9.3%) |
| Both         | 771 (-74%) | 2,259,902 (-3.9%) | 541,914 (-10.6%) |

## PRT (partial render table benchmark)

| Optimization | Registers | Uncompressed | Compressed |
|--------------|-----------|--------------|------------|
| None         | 13539     | 1,548,716    | 507,021    |
| Sinking      | 4957 (-63%) | 1,466,933 (-5.3%) | 461,435 (-9.0%) |
| Coalescing   | 9432 (-30%) | 1,504,402 (-2.9%) | 475,907 (-6.1%) |
| Both         | 1263 (-91%) | 1,421,410 (-8.2%) | 423,323 (-16.5%) |

## PRT (--disable use-js-string)

| Optimization | Registers | Uncompressed | Compressed |
|--------------|-----------|--------------|------------|
| None         | 13831     | 1,599,973    | 513,841    |
| Sinking      | 5503 (-60%) | 1,538,861 (-3.8%) | 482,321 (-6.1%) |
| Coalescing   | 9446 (-32%) | 1,563,006 (-2.3%) | 487,647 (-5.1%) |
| Both         | 1356 (-90%) | 1,492,182 (-6.7%) | 443,180 (-13.7%) |

## Observations

1. **Register reduction is dramatic** — the combined optimizations reduce registers by 74-91%, with the largest function going from 13,539 to just 1,263 registers in PRT.

2. **The optimizations are synergistic** — together they achieve much better results than either alone, especially for register count.

3. **Sinking excels at register reduction** — it's more effective at reducing register count than coalescing alone.

4. **Coalescing excels at compression** — it has a larger impact on compressed size (9.3% for OCaml, 6.1% for PRT) compared to sinking (1.1%, 9.0%), likely because it increases code repetition that compressors can exploit.

5. **PRT benefits more** — probably because it has larger functions with more optimization opportunities.

6. **use-js-string improves optimization effectiveness** — with `use-js-string` (the default), OCaml strings are implemented as JavaScript strings and can be inlined directly into functions. Without it, strings are wrapped and must remain defined at toplevel. This limits constant sinking opportunities, explaining the reduced improvements (16.5% vs 13.7% compressed size reduction, 9.0% vs 6.1% for sinking alone).

### DIFF
--- a/compiler/tests-compiler/gh1354.ml
+++ b/compiler/tests-compiler/gh1354.ml
@@ -49,7 +49,7 @@ with Exit ->
        var
         runtime = globalThis.jsoo_runtime,
         caml_maybe_attach_backtrace = runtime.caml_maybe_attach_backtrace,
-        _b_ = runtime.caml_wrap_exception;
+        caml_wrap_exception = runtime.caml_wrap_exception;
        function caml_call2(f, a0, a1){
         return (f.l >= 0 ? f.l : f.l = f.length) === 2
                 ? f(a0, a1)
@@ -60,14 +60,14 @@ with Exit ->
         Stdlib = _a_.Stdlib,
         Stdlib_Printf = _a_.Stdlib__Printf,
         _a_ = 0,
-        _c_ = _a_,
-        _d_ =
+        _b_ = _a_,
+        _c_ =
           [0, [4, 0, 0, 0, [12, 10, 0]], runtime.caml_string_of_jsbytes("%d\n")];
-       try{0; _c_ = _a_ + 1 | 0; throw caml_maybe_attach_backtrace(Stdlib[3], 1);}
-       catch(_e_){
-        _b_ = _b_(_e_);
-        if(_b_ !== Stdlib[3]) throw caml_maybe_attach_backtrace(_b_, 0);
-        caml_call2(Stdlib_Printf[3], _d_, _c_ | 0);
+       try{0; _b_ = _a_ + 1 | 0; throw caml_maybe_attach_backtrace(Stdlib[3], 1);}
+       catch(_d_){
+        _a_ = caml_wrap_exception(_d_);
+        if(_a_ !== Stdlib[3]) throw caml_maybe_attach_backtrace(_a_, 0);
+        caml_call2(Stdlib_Printf[3], _c_, _b_ | 0);
         runtime.caml_register_global(3, [0], "Test");
         0;
         return;
@@ -112,8 +112,8 @@ with Exit ->
        var
         runtime = globalThis.jsoo_runtime,
         caml_maybe_attach_backtrace = runtime.caml_maybe_attach_backtrace,
-        _b_ = runtime.caml_string_of_jsbytes,
-        _c_ = runtime.caml_wrap_exception;
+        caml_string_of_jsbytes = runtime.caml_string_of_jsbytes,
+        caml_wrap_exception = runtime.caml_wrap_exception;
        function caml_call3(f, a0, a1, a2){
         return (f.l >= 0 ? f.l : f.l = f.length) === 3
                 ? f(a0, a1, a2)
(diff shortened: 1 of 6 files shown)


hhugo:
Do you think this could help with https://github.com/ocsigen/js_of_ocaml/issues/2122

hhugo:
Can you rebase the branch ?

compiler/tests-compiler/var_coalescing.ml
```diff
@@ -0,0 +1,346 @@
...
+    if (a > 5) { r = a; break outer; }
+    b = x + 2;
+    r = b;
+  }
+  console.log(r);
+})(10)
+|};
+  [%expect {|
+    (function(x){
+       var a, a, a;
```
hhugo: This should be cleaned up.

compiler/tests-compiler/var_coalescing.ml
```diff
@@ -0,0 +1,346 @@
...
+})(1)
+|};
+  [%expect {|
+    (function(a){
+       var a;
+       switch(a){
+         case 0:
+          a = 10; a = a + 1; break;
+         case 1:
+          a = 20; a = a + 2; break;
```
hhugo: I though we had a simplification to turn `a = a + n` to `a += n`

hhugo:
I've added some tests and commented on the result

compiler/tests-compiler/var_coalescing.ml
```diff
@@ -0,0 +1,349 @@
...
+
+let%expect_test "overlapping lifespans — no coalescing" =
+  test {|
+(function(x) {
+  var a = x + 1;
+  var b = x + 2;
+  console.log(a + b);
+})(0)
+|};
+  [%expect {| (function(b){var a = b + 1, b = b + 2; console.log(a + b);}(0)); |}]
```
hhugo, line 81: Should we try to move `b = b + 2` outside the var declaration ?

vouillon: I'm not sure this is worth it. Especially since we are not generating this kind of code.

compiler/lib/js_traverse.ml
```diff
@@ -2003,8 +2026,29 @@ class simpl =
...
+      result
+
     method statement s =
-      let s = super#statement s in
+      let s =
+        match s with
+        | Variable_statement (Var, [ DeclIdent (V x, Some (EVar (V y), _)) ])
+          when Code.Var.equal x y && Code.Var.Tbl.get declared x ->
+            Empty_statement
+        | Expression_statement (EBin (Eq, EVar (V x), EVar (V y))) when Code.Var.equal x y
```
hhugo, line 2093: Should we look into ESeq ?

vouillon: I don't think we generate such code.

hhugo: Dont we emit ESeq in js_simpl ?

vouillon: The variable coalescing pass happens after that. And it will not coalesce `x` and `y` if `x = y` occurs deep inside an expression.

compiler/tests-compiler/gh1354.ml
```diff
@@ -49,25 +49,25 @@ with Exit ->
...
+        _c_ = _a_,
         _d_ =
           [0, [4, 0, 0, 0, [12, 10, 0]], runtime.caml_string_of_jsbytes("%d\n")];
-       try{0; _b_ = _a_ + 1 | 0; throw caml_maybe_attach_backtrace(Stdlib[3], 1);}
+       try{0; _c_ = _a_ + 1 | 0; throw caml_maybe_attach_backtrace(Stdlib[3], 1);}
        catch(_e_){
-        var _c_ = caml_wrap_exception(_e_);
-        if(_c_ !== Stdlib[3]) throw caml_maybe_attach_backtrace(_c_, 0);
-        caml_call2(Stdlib_Printf[3], _d_, _b_ | 0);
+        _b_ = _b_(_e_);
```
hhugo: maybe primitive alias should not be considered `generated` so that they are preserved here.

compiler/lib/js_variable_coalescing.ml
```diff
@@ -0,0 +1,1228 @@
...
+   could theoretically be merged with preceding variables, they are
+   typically allocated on the heap (e.g., in V8's context objects)
+   rather than the stack. Thus, they do not contribute to stack frame
+   size, but are likely slower to access.
+
+   Block-scoped variables (let, const) are excluded from this analysis.
+   Js_of_ocaml rarely uses them, except occasionally for captured
+   variables, so there is little benefit in extending the implementation
+   to support them. In pretty mode, only compiler-generated variables
+   are coalesced to preserve user-defined names.
```
hhugo, line 77: How do you feel about keeping all names in pretty mode ? and only coalescing variable without names..

vouillon: It seems more principled to keep only generated names.

And keeping all names defeat the goal of reducing the stack frame size. Here are the numbers for the toplevel function of ocamlc:
||Register count|
|-|-:|
|master | 1695
|keep all names | 1602
|keep generated names | 744

hhugo: The comment above should read, keep user-defined names.

PR #2167, opened 2026-02-04
Constant sinking

vouillon:
This optimization moves constant variable declarations closer to their usage sites, minimizing live ranges. Shorter live ranges reduce register pressure and make a subsequent variable coalescing pass more effective, since variables with non-overlapping live ranges can share the same name.
This pass is also able to inline many constants that are used once.

See #2166 for some numbers.

### DIFF
--- a/compiler/lib/js_constant_sinking.ml
+++ b/compiler/lib/js_constant_sinking.ml
@@ -1,0 +1,1019 @@
+(* Js_of_ocaml compiler
+ * http://www.ocsigen.org/js_of_ocaml/
+ * Copyright (C) 2026 Jérôme Vouillon
+ * Laboratoire PPS - CNRS Université Paris Diderot
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, with linking exception;
+ * either version 2.1 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+(** Constant sinking optimization pass.
+
+    This optimization moves constant variable declarations closer to their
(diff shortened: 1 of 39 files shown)


compiler/lib/js_output.ml
```diff
@@ -143,9 +143,7 @@ struct
   open D
 
   let nane_of_label = function
-    | Javascript.Label.L v ->
-        assert accept_unnamed_var;
-        Utf8_string.of_string_exn (Format.asprintf "<%a>" Code.Var.print v)
```
hhugo: Why was this reverted ?

compiler/lib/js_constant_sinking.ml
```diff
@@ -0,0 +1,966 @@
...
+              then Format.eprintf "Analysis: enter default case scope %d@." info.id;
+              List.iter
+                ~f:(fun (s, _) ->
+                  self#statement s;
+                  scopes#next_index ())
+                stmts;
+              scopes#leave_scope ()
+          | None -> ());
+          process_case_list cases2
+      | _ -> super#statement stmt
```
hhugo: Don't use catch all here.
- Should you handle ForAwait here ?

vouillon:
> Have you considered the interaction with share_constant ? It seems to work ok, we first lift some constant up to the toplevel and sink them down later. I wonder if we could benefit from doing the two pass at once, maybe sharing less in some scenarios.

I think the two passes are complementary. It seems simpler to keep them separate.

compiler/lib/js_constant_sinking.ml
```diff
@@ -0,0 +1,1022 @@
...
+   The walk processes scopes from innermost (lca) to outermost (toward root),
+   accumulating valid scopes in `acc`. When we hit a loop/function boundary,
+   we discard accumulated scopes (walk [] rem) since we can't cross it.
+   Returns the deepest (innermost) valid position. *)
+let find_valid_target info lca_stack =
+  if not (is_suffix lca_stack info.decl_scope)
+  then None
+  else
+    match info.kind with
+    | Primitive -> if is_suffix lca_stack info.decl_scope then Some lca_stack else None
```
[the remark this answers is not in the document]
hhugo: @vouillon, can you look into this review ?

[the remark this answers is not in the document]
vouillon: Basically, at some point, I pushed the "is_suffix" logic into each branch and forgot it remove it at the beginning of the function.

vouillon:
The JavaScript debuggers do not stop on constant values, so that should not make any difference.
I have updated the documentation.

hhugo (approved):

compiler/lib/js_output.ml
hhugo: I've reverted the revert

PR #2173, opened 2026-02-06
Keep track of generated names in a more space efficient way

vouillon:
Use a bitset instead of storing a pair (name, generated).

### DIFF
--- a/compiler/lib/code.ml
+++ b/compiler/lib/code.ml
@@ -125,14 +125,15 @@ end = struct
   include T
 
   module Name = struct
-    type t =
-      { name : string
-      ; generated : bool
-      }
-
     let names = Int.Hashtbl.create 100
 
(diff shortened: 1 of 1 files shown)


compiler/lib/code.ml
```diff
@@ -125,39 +125,48 @@ end = struct
...
-      { name : string
-      ; generated : bool
-      }
-
     let names = Int.Hashtbl.create 100
 
-    let reset () = Int.Hashtbl.clear names
+    let generated_set = ref (BitSet.create' 16)
+
+    let generated v = not (BitSet.mem !generated_set v)
```
hhugo: The names is confusing, generated_set is set for variable that are not generated

vouillon: Oh, right!

PR #2177, opened 2026-02-06
--build-config / --apply-build-config

vouillon:
--build-config prints build-relevant compiler settings as key=value pairs and exits. --apply-build-config overrides these settings, taking precedence over other flags. This enables coherent compilation when linking units that must share the same settings.

### DIFF
--- a/compiler/lib-cmdline/build_config.ml
+++ b/compiler/lib-cmdline/build_config.ml
@@ -23,29 +23,7 @@ let print_and_exit keys =
   Printf.printf "%s\n" (Build_info.to_config_string (Build_info.get_values keys));
   exit 0
 
-let validate_and_set keys (key, value) =
-  match
-    List.find_opt keys ~f:(fun k -> String.equal (Build_info.config_key_name k) key)
-  with
-  | None -> failwith (Printf.sprintf "unknown key %S" key)
-  | Some (Build_info.Bool_key { set; _ }) -> (
-      match value with
-      | "true" -> set true
-      | "false" -> set false
-      | _ -> failwith (Printf.sprintf "key %S expects true or false, got %S" key value))
-  | Some (Build_info.Enum_key { set; valid; _ }) ->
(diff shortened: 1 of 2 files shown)


compiler/lib-cmdline/build_config.ml
```diff
@@ -0,0 +1,55 @@
...
+          (Printf.sprintf
+             "key %S expects one of {%s}, got %S"
+             key
+             (String.concat ~sep:", " valid)
+             value)
+
+let parse keys input =
+  List.iter (Build_info.parse_config_string input) ~f:(validate_and_set keys)
+
+let process ?(extra_keys = []) target ~apply ~print_and_exit:do_print =
```
hhugo: extra_keys doesn't seem to be used

compiler/bin-js_of_ocaml/compile.ml
```diff
@@ -166,6 +168,12 @@ let run
   Config.set_target `JavaScript;
   Jsoo_cmdline.Arg.eval common;
   Config.set_effects_backend effects;
+  Config.Flag.set "toplevel" toplevel;
```
hhugo: Maybe toplevel should be moved out of the record and `Cmd_arg` should set Config.Flag when `--toplevel` is passed.

compiler/lib/build_info.ml
```diff
@@ -49,6 +49,95 @@ let effects_backend_of_string = function
...
+      }
+  ; Bool_key
+      { name = "use-js-string"
+      ; get = Config.Flag.use_js_string
+      ; set = Config.Flag.set "use-js-string"
+      }
+  ]
+  @
+  match target with
+  | `JavaScript ->
```
hhugo: Can't we unconditionally include "toplevel" ?

compiler/lib/build_info.ml
hhugo: @vouillon, the logic here needs to be updated.

PR #2186, opened 2026-02-14
Fix auto complete install again

hhugo:


hhugo:
@nojb, Can you test this PR ?

PR #2189, opened 2026-02-20
Wasm native effects (Stack Switching proposal)

vouillon:
Blocked by:
- [x] #2177 
- [x] https://github.com/ocaml/dune/pull/13613
- [x] #2231
- [x] and a new release of dune


tools/node_wrapper.ml
```diff
@@ -1,8 +1,4 @@
-let extra_args_for_wasoo =
-  [ "--experimental-wasm-imported-strings"
```
hhugo: Do we no longer need these flags

vouillon: Both flags have been removed in node 26 (the features are always enabled).
`--experimental-wasm-imported-strings` has been the default for quite some time.
`--experimental-wasm-stack-switching` ([JSPI](https://v8.dev/blog/jspi)) is enabled by default since node 25.

hhugo: Could we maintain a list of flags per node and feature used ? Same thing for chrome ? It could live in the doc

vouillon: I'm not sure this belongs to the doc if these flags are only used for testing.
We have a fallback polyfill, so `--experimental-wasm-imported-strings` is never necessary.
Maybe I can document that JSPI is enabled by default since Chrome 137 and node 25?

hhugo: > Maybe I can document that JSPI is enabled by default since Chrome 137 and node 25?

Yes.

> --experimental-wasm-imported-strings has been the default for quite some time.
> --experimental-wasm-stack-switching ([JSPI](https://v8.dev/blog/jspi)) is enabled by default since node 25.

But the readme says

> The generated code works with Chrome 119, Node.js 22, Firefox 122, and Safari 18.2 (or more recent versions of these applications).

I think we should try to be more precise maybe.

hhugo:
Is this ready to merge ? Can we just drop commit related to dune 3.23 ?

vouillon:
> Is this ready to merge ? Can we just drop commit related to dune 3.23 ?

We still need dune 3.23 for `--effects native` to work properly with separate compilation. Though I guess I could switch to whole program compilation for now for profile `with-native-effect`. Or keep just d546b3688962bce086a5687619f2c5fb42244fae.

hhugo:
> > Is this ready to merge ? Can we just drop commit related to dune 3.23 ?
> 
> We still need dune 3.23 for `--effects native` to work properly with separate compilation. Though I guess I could switch to whole program compilation for now for profile `with-native-effect`. Or keep just [d546b36](https://github.com/ocsigen/js_of_ocaml/commit/d546b3688962bce086a5687619f2c5fb42244fae).

My understanding is that we don't need `lang dune 3.23`, We just need dune version 3.23 with jsoo >= 6.4. See https://github.com/ocsigen/js_of_ocaml/pull/2250

vouillon:
> Is this ready to merge ? Can we just drop commit related to dune 3.23 ?

This is ready. I have dropped the unneeded commits.

compiler/bin-wasm_of_ocaml/link_wasm.ml
```diff
@@ -71,13 +72,30 @@ let options =
...
-      ; allowed_imports
-      ; binaryen_options = { common; opt; merge }
-      }
+    let effects_backend =
+      match
+        List.find_map
+          ~f:(fun (k, v) -> if String.equal k "effects" then Some v else None)
+          variables.Preprocess.set
+      with
+      | Some "native" -> Ok (`Native : Js_of_ocaml_compiler.Config.effects_backend)
```
hhugo: Can we use Build_info.effect_backend_of_string ?

compiler/bin-wasm_of_ocaml/compile.ml
```diff
@@ -83,6 +83,7 @@ let preprocessor_variables () =
         (match Config.effects () with
         | `Disabled | `Jspi -> "jspi"
```
hhugo: can we use string_of_effect_backend from Build_info ?

compiler/bin-wasm_of_ocaml/compile.ml
```diff
@@ -80,10 +80,11 @@ let preprocessor_variables () =
   (* Keep this variables in sync with gen/gen.ml *)
   [ ( "effects"
     , Wat_preprocess.String
-        (match Config.effects () with
-        | `Disabled | `Jspi -> "jspi"
-        | `Cps -> "cps"
-        | `Double_translation -> assert false) )
+        ((match Config.effects () with
+         | `Disabled -> `Jspi
```
hhugo: Why do we have this behavior ? Should it be documented ?

vouillon: We are using the same runtime `runtime-standard.wasm` for both `disabled` and `jspi`. I have added a comment. I'm not sure we need to document that.

hhugo (approved):

PR #2195, opened 2026-03-07
Compiler, Runtime: initial support for OCaml 5.5

hhugo:


compiler/lib/parse_bytecode.ml
```diff
@@ -240,11 +240,13 @@ end = struct
...
-          event.ev_compenv.ce_stack
-          []
-        |> List.sort ~cmp:(fun (i, _) (j, _) -> compare i j)
+        try
+          Ident.fold_name
+            (fun ident i acc -> (event.ev_stacksize - i, ident) :: acc)
+            event.ev_compenv.ce_stack
+            []
+          |> List.sort ~cmp:(fun (i, _) (j, _) -> compare i j)
+        with _ -> raise Not_found
```
vouillon: This looks wrong.

hhugo: this is temporary, see https://github.com/ocaml/ocaml/pull/13275#discussion_r2896203749

compiler/tests-ocaml/match-exception/streams.ml
```diff
@@ -29,11 +29,14 @@ let rec iter_stream_match f s =
...
+let test_iter_stream () =
   let limit = 10000000 in
   try
     iter_stream_match ignore (make_stream_up_to limit);
     print_endline "iter_stream with handler case (match) is tail recursive"
   with Stack_overflow ->
     assert false
 ;;
+
+let () = test_iter_stream ()
```
hhugo, line 41: What is this diff about ?

hhugo:
@kit-ty-kate, you can update your alpha opam repo with the tip of this PR.

compiler/tests-ocaml/match-exception/streams.ml
vouillon: We get a memory leak if we don't put the code in its own function since the head of the stream remains live.

compiler/lib/parse_bytecode.ml
hhugo: no longer need with alpha3

avsm:
I've tested the branch with 5.5 beta and it seems to work in my build smoke tests

hhugo:
> I've tested the branch with 5.5 beta and it seems to work in my build smoke tests

The support is now part of master, this PR just updates the testsuite to run with ocaml 5.5

PR #2211, opened 2026-04-08
Put more values into global variables

vouillon:
Put constant expressions at toplevel into global variables, in order to reduce the size of the toplevel function.

For the PRT benchmark, the size of the toplevel function is divided by 2.

### DIFF
--- a/compiler/lib-wasm/code_generation.ml
+++ b/compiler/lib-wasm/code_generation.ml
@@ -433,6 +433,46 @@ let is_small_constant e =
   | W.GlobalGet name -> global_is_constant name
   | _ -> return false
 
+let rec check_is_constant st (e : W.expression) =
+  match e with
+  | Const _ | RefFunc _ | RefNull _ -> true
+  | RefI31 e -> check_is_constant st e
+  | GlobalGet x -> (
+      match Var.Map.find_opt x st.context.constant_globals with
+      | Some { init = Some _; _ } -> true
+      | Some { init = None; _ } | None -> false)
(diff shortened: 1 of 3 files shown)


compiler/lib-wasm/globalize.ml
```diff
@@ -104,6 +106,32 @@ let traverse_block p st pc =
   let st = List.fold_left ~f:(fun st x -> declare x st) ~init:st b.Code.params in
   List.fold_left ~f:(fun st i -> traverse_instruction st i) ~init:st b.Code.body
 
+let available x st = Code.Var.Set.mem x st.globals || Code.Var.Set.mem x st.constants
+
+let max_non_available = 2
```
hhugo, line 111: Can you add a comment explaining this value ?

vouillon: I have changed the heuristics and added a comment.

compiler/lib-wasm/code_generation.ml
```diff
@@ -433,6 +433,74 @@ let is_small_constant e =
...
+            else
+              ( placeholder :: init_args
+              , W.ArraySet (ty, W.GlobalGet x, W.Const (I32 (Int32.of_int i)), a)
+                :: patches
+              , i + 1 ))
+          ~init:([], [], 0)
+          args
+      in
+      if List.is_empty patches
+      then None, st
```
hhugo: Shouldn't this be an assert false ?

hhugo (approved):

PR #2212, opened 2026-04-10
Extract loops from toplevel function into helper functions

vouillon:
Wasm engines tier up (compile to optimised native code) functions that contain loops. For the toplevel/init function, which is large and called only once, this is wasteful. Extract contained loops into small helper functions so only the helpers get tiered up.

### DIFF
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -2446,6 +2446,7 @@ module Generate (Target : Target_sig.S) = struct
         functions
     in
     global_context.init_code <- [];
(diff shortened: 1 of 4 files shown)


hhugo:
Couldn't it benefit js as well ? Allowing to optimize the loop ..

vouillon:
> Couldn't it benefit js as well ? Allowing to optimize the loop ..

That could benefit JS as well, indeed. This seems harder to implement, though.
Also, we have actually observed performance issues with the Wasm code (Turbofan, V8's optimizing compiler, has performance issues with large Wasm functions), but not with JS code.

vouillon:
Closed in favor of #2245 which works both for JavaScript and Wasm.

PR #2221, opened 2026-04-28
Dom updates

### LINKED ISSUE #453: Add selection* attributes to appropriate input elements
Following this patch for text areas https://github.com/ocsigen/js_of_ocaml/commit/4229e4b8178069d73cc646c7ce203b41736712b9, certain types of input elements (password, text, etc) have selection* attributes defined on them.

The trouble is how to apply those attributes to input elements in a type safe manner, while disallowing certain input types, following the spec (radio buttons and check boxes, etc).

A list of input elements and their allowed attributes is here: https://html.spec.whatwg.org/multipage/forms.html#do-not-apply

### LINKED ISSUE #679: [FEATURE REQUEST] window.performance.now is not provided
It would be nice to provide bindings to window.performance.now ( https://developer.mozilla.org/en-US/docs/Web/API/Performance/now ).

The precision has been reduced because of the concerns raised by spectre-class attacks and isn't much better than Date's but the API is really much nicer to work with.

Thanks.

vouillon:
Fixes #679.
Fixes #453.

Type corrections (old types did not match runtime behaviour):

| Member | Old type | New type | Why |
|---|---|---|---|
| `Dom_html.inputElement.defaultChecked` | `js_string t prop` | `bool t prop` | Spec: it's a boolean reflecting the `checked` content attribute. |
| `Dom_html.createVideo` (return) | `audioElement t` | `videoElement t` | Pre-existing typo: the runtime value is a video element. |
| `Dom_html.selection.collapse` | `bool t -> unit meth` | `Dom.node t opt -> unit meth` | The previous shape was that of `Range.collapse(toStart)`; the actual JS method is `Selection.collapse(node, offset?)` and accepts `null` to remove the selection. |

### DIFF
--- a/lib/js_of_ocaml/dom.ml
+++ b/lib/js_of_ocaml/dom.ml
@@ -88,6 +88,8 @@ class type node = object
 
   method parentNode : node t opt prop
 
+  method parentElement : element t opt readonly_prop
+
   method childNodes : node nodeList t prop
 
   method firstChild : node t opt prop
@@ -99,6 +101,8 @@ class type node = object
   method nextSibling : node t opt prop
 
   method namespaceURI : js_string t opt prop
+
+  method isConnected : bool t readonly_prop
 
   method insertBefore : node t -> node t opt -> node t meth
 
@@ -114,23 +118,25 @@ class type node = object
 
   method compareDocumentPosition : node t -> DocumentPosition.t meth
 
+  method contains : node t -> bool t meth
+
+  method getRootNode : node t meth
+
+  method getRootNode_options : getRootNodeOptions t -> node t meth
+
+  method isEqualNode : node t -> bool t meth
+
+  method isSameNode : node t -> bool t meth
+
+  method normalize : unit meth
+
   method lookupNamespaceURI : js_string t -> js_string t opt meth
 
   method lookupPrefix : js_string t -> js_string t opt meth
 end
 
-let appendChild (p : #node t) (n : #node t) = ignore (p##appendChild (n :> node t))
-
-let removeChild (p : #node t) (n : #node t) = ignore (p##removeChild (n :> node t))
-
-let replaceChild (p : #node t) (n : #node t) (o : #node t) =
-  ignore (p##replaceChild (n :> node t) (o :> node t))
-
-let insertBefore (p : #node t) (n : #node t) (o : #node t opt) =
-  ignore (p##insertBefore (n :> node t) (o :> node t opt))
-
 (** Specification of [Attr] objects. *)
-class type attr = object
+and attr = object
   inherit node
 
   method name : js_string t readonly_prop
@@ -161,6 +167,10 @@ and element = object
 
   method tagName : js_string t readonly_prop
 
+  method localName : js_string t readonly_prop
+
+  method prefix : js_string t opt readonly_prop
+
   method getAttribute : js_string t -> js_string t opt meth
 
   method setAttribute : js_string t -> js_string t -> unit meth
@@ -168,6 +178,14 @@ and element = object
   method removeAttribute : js_string t -> unit meth
 
   method hasAttribute : js_string t -> bool t meth
+
+  method hasAttributes : bool t meth
+
+  method toggleAttribute : js_string t -> bool t meth
+
+  method toggleAttribute_force : js_string t -> bool t -> bool t meth
+
+  method getAttributeNames : js_string t js_array t meth
 
   method getAttributeNS : js_string t -> js_string t -> js_string t opt meth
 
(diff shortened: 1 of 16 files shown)


lib/js_of_ocaml/dom_html.mli
```diff
@@ -2241,10 +3405,20 @@ class type window = object
...
+
   method frameElement : element t opt readonly_prop
 
   method open_ : js_string t -> js_string t -> js_string t opt -> window t opt meth
 
+  method postMessage : 'a. 'a -> js_string t -> unit meth
+
+  method postMessage_transfer : 'a 'b. 'a -> js_string t -> 'b js_array t -> unit meth
+
+  method requestIdleCallback : (idleDeadline t -> unit) Js.callback -> int meth
```
hhugo: Here is what mdn says:
> Note: A timeout option is strongly recommended for required work, as otherwise it's possible multiple seconds will elapse before the callback is fired.

Should we add it ?

lib/js_of_ocaml/dom_html.mli
```diff
@@ -2241,10 +3405,20 @@ class type window = object
...
 
+  method opener : window t opt prop
+
   method frameElement : element t opt readonly_prop
 
   method open_ : js_string t -> js_string t -> js_string t opt -> window t opt meth
 
+  method postMessage : 'a. 'a -> js_string t -> unit meth
+
+  method postMessage_transfer : 'a 'b. 'a -> js_string t -> 'b js_array t -> unit meth
```
hhugo, line 3411: Mdn says:
> postMessage(message)
postMessage(message, targetOrigin)
postMessage(message, targetOrigin, transfer)
postMessage(message, options)

lib/js_of_ocaml/dom.mli
```diff
@@ -206,12 +260,37 @@ class type text = characterData
...
+class type documentType = object
+  inherit node
+
+  method name : js_string t readonly_prop
+
+  method publicId : js_string t readonly_prop
+
+  method systemId : js_string t readonly_prop
+
+  method internalSubset : js_string t opt readonly_prop
```
hhugo: internalSubset doesn't seem to be standard

lib/js_of_ocaml/dom_html.ml
```diff
@@ -438,6 +756,14 @@ and touch = object
...
 
   method pageY : number_t readonly_prop
+
+  method radiusX : number_t optdef readonly_prop
+
+  method radiusY : number_t optdef readonly_prop
+
+  method rotationAngle : number_t optdef readonly_prop
+
+  method force : number_t optdef readonly_prop
```
hhugo: Consider adding a comment about why it is optdef

lib/js_of_ocaml/dom_html.ml
```diff
@@ -723,6 +1341,34 @@ and element = object
 
   method innerText : js_string t prop
 
+  method dataset : domStringMap t readonly_prop
+
+  method tabIndex : int prop
+
+  method hidden : bool t prop
```
hhugo: Mdn says:
```
This attribute may have one of three values:
[true](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#true)
The element is hidden.

[false](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#false)
The element is not hidden. This is the default value for the attribute.

["until-found"](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#until-found)
The element is hidden until found, meaning that it is hidden but will be revealed if found through in page search or reached through fragment navigation.
```

lib/js_of_ocaml/performance.ml
```diff
@@ -0,0 +1,125 @@
...
+
+let is_supported () = Js.Optdef.test Js.Unsafe.global##.performance
+
+let mark ?detail ?startTime (perf : performance Js.t) name =
+  let opts : _ performanceMarkOptions Js.t = Js.Unsafe.obj [||] in
+  Option.iter (fun v -> opts##.detail := v) detail;
+  Option.iter (fun v -> opts##.startTime := v) startTime;
+  perf##mark_options name opts
+
+let makeMeasureOptions ?detail ?start ?start_time ?_end ?end_time ?duration () =
```
hhugo, line 117: Should we really have both start and start_time ? _end and end_time ?

lib/js_of_ocaml/dom.ml
```diff
@@ -189,9 +207,77 @@ and element = object
 
   method getElementsByTagName : js_string t -> element nodeList t meth
 
+  method getElementsByClassName : js_string t -> element nodeList t meth
+
+  method matches : js_string t -> bool t meth
+
   method attributes : attr namedNodeMap t readonly_prop
+
+  method children : element nodeList t readonly_prop
```
hhugo: AI tells me it should be a collection.

hhugo: as well as getElementsByClassName, getElementsByTagName, getElementsByTagNameNS

lib/js_of_ocaml/dom_html.ml
```diff
@@ -694,6 +1094,216 @@ and tokenList = object
...
+and shadowRoot = object
+  inherit Dom.documentFragment
+
+  method mode : js_string t readonly_prop
+
+  method host : element t readonly_prop
+
+  method innerHTML : js_string t prop
+
+  method activeElement : element t opt readonly_prop
```
hhugo, line 1108: https://dom.spec.whatwg.org/#interface-shadowroot mention many more fields.

lib/js_of_ocaml/dom_html.mli
```diff
@@ -932,196 +1690,470 @@ class type optionElement = object
...
+  method src : js_string t prop
 
-  method htmlFor : js_string t prop
+  method srclang : js_string t prop
+
+  method label : js_string t prop
+
+  method default : bool t prop
+
+  method readyState : int readonly_prop
```
hhugo: mediaElement used:
```
  method readyState_int : int readonly_prop

  method readyState : readyState readonly_prop
```

lib/js_of_ocaml/dom_html.mli
vouillon: Actually, this method is not available on Safari. I'm just going to remove it.

lib/js_of_ocaml/dom_html.ml
vouillon: These properties should not be optdef, actually.

lib/js_of_ocaml/dom_html.ml
vouillon: I'm not sure we want to implement the Shadow DOM API for now.

hhugo: Fine

lib/js_of_ocaml/dom_html.ml
vouillon: `"until-found"` is recent. I'm going to make it write-only for now.

lib/js_of_ocaml/performance.ml
vouillon: Do you think we should just have `start` and `end` with a variant to select between a timestamp and a performance mark?

hhugo: I'm not sure. either that or an error if both are provided.

hhugo:
Why did you decide to revert some of the change in https://github.com/ocsigen/js_of_ocaml/pull/2221/changes/06c43dd6f2d294f44a7eb7e41af7ab395fffae36 ?

vouillon:
> Why did you decide to revert some of the change in [06c43dd](https://github.com/ocsigen/js_of_ocaml/commit/06c43dd6f2d294f44a7eb7e41af7ab395fffae36) ?

These are breaking changes. I'm not sure we want to make them now.

hhugo (approved):

hhugo:
I did not review the whole diff carefully but relied on AI to point me to interesting sections of the diff.

PR #2223, opened 2026-04-29
Fix Int64.of_string

vouillon:
The empty-input check in `$caml_i64_of_digits` only fires when the whole string is empty; for inputs like `"+"`, `"-"` or `"0x"` the digit cursor sits past the end of the string, and the next `array.get_u traps`. Guard on `$i >= $len` so `caml_failwith` fires cleanly.


hhugo:
@copilot resolve the merge conflicts in this pull request

hhugo (approved):

PR #2224, opened 2026-04-29
Fix caml_oo_cache_id

vouillon:
I'm not sure how to test this.


hhugo:
We could implement a check ensuring we always return or raise in js primitives.

Issue #2125, opened 2025-11-04
[BUG] object%js with references to self can trigger false unused variable warnings


johnridesabike:
**Describe the bug**

If using references to "self"/"this" with `object%js`, and any methods do not use that variable, it triggers an unused variable warning. Minimal example:

```
let () =
  Js_of_ocaml.Js.export_all
    object%js (self)
      method foo = self
      method bar = ()
    end
```

```
3 |     object%js (self)
                  ^^^^^^
Error (warning 27 [unused-var-strict]): unused variable self.
```

**Expected behavior**
The above example should not trigger the warning.

Looking at the PPX output, it seems that `self` gets added as an argument to each method function, so the presence of `bar` is what triggers the warning in this case. Adding more methods triggers the warning more times.

**Versions**
Version of packages used to reproduce the bug
```
$ opam list --search js_of_ocaml --installed
# Packages matching: match(*js_of_ocaml*) & installed
# Name               # Installed # Synopsis
brr                  0.0.8       Browser programming toolkit for OCaml
js_of_ocaml          6.2.0       Compiler from OCaml bytecode to JavaScript
js_of_ocaml-compiler 6.2.0       Compiler from OCaml bytecode to JavaScript
js_of_ocaml-ppx      6.2.0       Compiler from OCaml bytecode to JavaScript
js_of_ocaml-toplevel 6.2.0       Compiler from OCaml bytecode to JavaScript
tyxml                4.6.0       A library for building correct HTML and SVG documents
```

hhugo:
I had a look at the generated code. We emit one function per method with "self" as the first argument. A warning is emitted anytime a method doesn't use "self" (the first argument of the generated function). I think the best we can do is to disable the warning entirely for "self" by prefixing the body by a dummy `let _ = "self_id"`.

johnridesabike:
My current workaround is prefixing it with an underscore (`_self`) to suppress the warning, so that solution would work just as well for me.

hhugo:
fixed by #2128

PR #2225, opened 2026-04-29
Update to Oxcaml 5.2.0minus31

vouillon:


hhugo:
Does this PR mean we won't support the previous version of oxcaml ?

vouillon:
> Does this PR mean we won't support the previous version of oxcaml ?

OxCaml is an experimental compiler. Users will always be on the latest version. So, supporting multiple versions doesn't seem worth the effort.

compiler/tests-check-prim/unix-Unix.5.2+ox.output
```diff
@@ -245,6 +338,7 @@ caml_bigstring_memcmp
 caml_hash_mix_bigstring
 
 From +effect.js:
+caml_continuation_use_and_update_handler_noexc
```
hhugo, line 339: You could update the `//Version` in effect.js to reflect that it is not used with oxcaml

lib/tests/test_poly_equal.ml
```diff
@@ -21,10 +21,9 @@ open Js_of_ocaml
 let%expect_test "poly equal" =
   let obj1 = Js.Unsafe.obj [||] in
   let obj2 = Js.Unsafe.obj [||] in
-  assert (obj1 = obj2);
+  assert (obj1 = obj1);
```
hhugo, line 24: Wny do we have this diff ?

ppx/ppx_js/lib_internal/ppx_js_internal.ml
```diff
@@ -745,12 +745,15 @@ let literal_object self_id (fields : field_desc list) =
...
+        let ghost_self_id =
+          { self_id with ppat_loc = { self_id.ppat_loc with loc_ghost = true } }
+        in
         lift_function_body_constraint
           (Ast_builder.Default.pexp_fun
              ~loc:{ body.pexp_loc with loc_ghost = true }
              Nolabel
              None
-             self_id
+             ghost_self_id
```
hhugo: Why this diff ?

compiler/tests-check-prim/unix-Unix.5.2+ox.output
hhugo: I realize now that it was added in this PR. Is it expected that it shows as unused here ?

lib/tests/test_poly_equal.ml
vouillon: This fixes the test. It makes sense to check that `obj1 = obj1`, and the value of `obj1 = obj2` is tested by the next line.
I can put it in a different PR if you prefer.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
vouillon: This comes from a [Jane street patch](https://github.com/oxcaml/opam-repository/blob/main/packages/js_of_ocaml-compiler/js_of_ocaml-compiler.6.0.1%2Box1/files/js_of_ocaml-ppx-js-ghost-location.patch).
I guess OxCaml can produce a spurious warning there.

lib/tests/test_poly_equal.ml
hhugo: ok

hhugo: keep it here

ppx/ppx_js/lib_internal/ppx_js_internal.ml
hhugo: This kind of changes can affect error messages and merlin behavior. I'd like to understand what's going on here .

compiler/tests-check-prim/unix-Unix.5.2+ox.output
vouillon: Primitive `caml_continuation_use_and_update_handler_noexc` has been removed from the OxCaml runtime, which provides `caml_continuation_update_handler_noexc` instead.

> You could update the //Version in effect.js to reflect that it is not used with oxcaml

I don't see how to do that.

ppx/ppx_js/lib_internal/ppx_js_internal.ml
vouillon: When you have a literal object:
```ocaml
object%js (self)
  method meth x = e
end
```
The generated code for the method will be
```ocaml
fun self -> let _ = self in e
```
This is affecting the parameter `self` in `fun self -> ...`. I'm not sure how, though. I'm going to investigate.

vouillon: I suspect this was an independent attempt at fixing #2125. I have removed it.

runtime/js/float32.js
```diff
@@ -424,14 +424,14 @@ function caml_fma_float32_bytecode(x, y, z) {
 //Version: >= 5.2, < 5.3
 //OxCaml
 function caml_simd_float32_min_bytecode(x, y) {
-  return Math.min(x, y);
+  return x < y ? x : y;
```
hhugo, line 429: Why this diff ?

vouillon: `caml_simd_float32_min_bytecode` mirrors the x86 instruction `MINSS`, which selects the second operand when the comparison fails (NaN or equal, including ±0), while `Math.min` propagates NaN and `Math.min(-0, 0) = -0`.

hhugo: Do we have a test for this ? Maybe a comment would help future review .

vouillon: I have added tests and comments.

hhugo (approved):

PR #2227, opened 2026-04-30
Initial support for OCaml 5.6

vouillon:


hhugo:
@vouillon, now that OCaml 5.5 is release as well as jsoo 6.4, I think it would make sense to merge this PR so that one can more easily test the next ocaml version

kit-ty-kate:
i think it would be nice to have a preview package that works with 5.6 in opam-repository. Would any of you two be able to do that? (if you don't have the time i'm happy to do it)

hhugo:
> i think it would be nice to have a preview package that works with 5.6 in opam-repository. Would any of you two be able to do that? (if you don't have the time i'm happy to do it)

Can one use a commit ref instead of a tag for such package ?

kit-ty-kate:
of course

PR #2229, opened 2026-05-02
Add QuickJS-NG support

hhugo:
## Summary

- Adds a `quickjs` dune profile that swaps the `node` binary on PATH for `tools/quickjs_wrapper.exe`, a thin wrapper around QuickJS-NG (`qjs`) that forwards `--std`/`--stack-size`.
- Adds a `process.env` fallback in `runtime/js/sys.js` so `Sys.getenv` reads `std.getenviron()` under QuickJS, plus a `caml_strerror` fallback for when `node:util` is unavailable.
- **Real-disk filesystem support under qjs (opt-in).** New `runtime/js/fs_quickjs.js` provides an `MlQuickJSDevice` (and matching `MlQuickJSFd`) backed by qjs's `os` / `std` modules — `os.open`/`read`/`write`/`stat`/`readdir`/`rename`/etc. The default jsoo runtime ships only weakdef stubs (`MlQuickJSDevice = class {}`, `fs_quickjs_supported() => false`), so users opt in by passing `+fs_quickjs.js` on the jsoo command line. The bundled `quickjs` dune profile passes that automatically, so the test suite gets real-fs without any per-test plumbing. Primitives qjs doesn't expose (`access`, `chmod`, `truncate`, `link`) raise "not implemented".
- **Byte-faithful stdout / stderr under qjs.** `MlFakeFd_out.write` for fd 1 / 2 prefers `os.write(fd, buffer, off, len)` over `puts(string)`, so non-UTF-8 bytes (e.g. raw bytes in `Str` regex output) survive instead of becoming U+FFFD. The pure-JS UTF-8 fallback for `jsoo_text_decoder` (used when `TextDecoder` isn't a global) was also tightened to validate continuation bytes per the WHATWG default error-handling rule, instead of greedily consuming the next bytes on any high-bit lead.
- Refactors `MlFakeFd_out` so `console.log`/`console.error` sinks strip the trailing `\n` via a wrapper installed at construction time; byte-faithful sinks (`os.write`, `std.out.puts`) get bytes / strings unchanged.
- Fixes a handful of latent host-portability bugs in the runtime (`fs_fake.js`, `mlBytes.js`, `sys.js`, `lib/js_of_ocaml/intl.ml`) that QuickJS surfaced but were already wrong on engines other than Node.
- Guards tests that genuinely cannot run under QuickJS (stdin pipe `EBADF`, missing `caml_unix_*` primitives, CommonJS `require`, `Sys.command`, stdout/stderr flush-on-exit ordering) with `(<> %{profile} quickjs)`. Snapshot-coupled tests in `tests-compiler` are gated via a `not_quickjs` case in `gen-rules/gen.ml`. Tests that previously needed FS guards (`lib-channels/{close_in,in_channel_length,seek_in}`, `lib-sys/rename`, `lib-unix/common/append`, `tests-io/non_ascii_filenames`, `lib-str/t01`) now run under qjs unmodified.
- Under the `quickjs` profile, jsoo warnings are non-fatal (`(env (quickjs) ...)` drops `--Werror`), since the suite trips a few engine-specific warnings that don't matter for a non-default target.
- Adds a `make tests-quickjs` target and a CI job (`.github/workflows/quickjs.yml`) that runs `dune build @runtest @runtest-js --profile=quickjs` against QuickJS-NG.
- Switches the documented browser-bundling recipe from `--target-env=browser` to `--external:node:*` in esbuild, so the same compiled `.js` works for both Node and browser bundling.

Includes the docs-browser-compat commits from the still-open #2216.

## Test plan
- [ ] `make tests` (Node) still passes
- [ ] `make tests-quickjs` passes locally with QuickJS-NG
- [ ] CI QuickJS-NG job is green


hhugo:
@glondu this PR contains a few change to improve quickjs support. Do you have time to give it a try ?

glondu:
This is impressive, I wasn't expecting as much!

In general, when I target jsoo, I usually avoid io/fs/sys... Therefore, I didn't face most of the issues this PR tries to solve... The most relevant fix for me is the fallback implementation of TextEncoder/TextDecoder.

As said elsewhere, I already managed to have a working workflow with jsoo 6.3.2, babeljs, esbuild and quickjs, you can see it at play there: https://salsa.debian.org/ocaml-team/dh-ocaml/-/commit/56a3b546239b80022d75471cae4a340175e17e87

Note that at the moment, I'm using quickjs (not quickjs-ng), since it's the only JS runtime that I found available on all Debian architectures (quickjs-ng FTBFS on Hurd at the moment).

Actually, I find it a bit odd to put quickjs-specific (or node-specific, or even wasi-specific) stuff in jsoo itself... I like to think of JS (and WASM) as a generic (and OS-neutral) runtime environment, and jsoo as a compiler for that. IMHO, it should be the responsibility of the application developer (or a separate library) to handle runtime-specific stuff. See for example how I isolated OS-specific stuff in the commit linked above.

Concerning your question:

> Do you have time to give it a try ?

I don't have time to try it right now, but I definitely will (and report issues if any) when it flows down to Debian.

PR #2231, opened 2026-05-05
Switch to dune 3.23

vouillon:
It uses `js_of_ocaml --build-config` to manage config details.

Blocked on:
- [x] https://github.com/ocaml/dune/issues/14441
- [x] https://github.com/ocaml/dune/issues/14438


compiler/tests-ocaml/basic-io/dune
```diff
@@ -1,5 +1,6 @@
 (tests
  (names wc)
  (modes js wasm)
+ (deps wc.ml)
  (action
   (run node %{test} wc.ml)))
```
hhugo: We could use %{dep:wc.ml} here instead.

hhugo:
dune 3.23.1 will hopefully restore compatibility with ocaml 4.13. See https://github.com/ocaml/dune/pull/14443

hhugo:
Can this be merged ?

vouillon:
> Can this be merged ?

I think it is ready. But I'm wondering whether this should go in Js_of_ocaml 9.4. Maybe I can make a separate PR with just the last commit, which ensures a consistent behavior with all versions of dune?

hhugo:
> > Can this be merged ?
> 
> I think it is ready. But I'm wondering whether this should go in Js_of_ocaml 9.4. Maybe I can make a separate PR with just the last commit, which ensures a consistent behavior with all versions of dune?

I don't understand what you mean. With this PR, one now depend on dune.3.23.1.

vouillon:
We don't really need dune 3.23.1 for the next release. So, maybe we should wait a bit before merging this PR?

hhugo:
> We don't really need dune 3.23.1 for the next release. So, maybe we should wait a bit before merging this PR?

I see what you mean. Changes in this PR are about `lang dune 3.23` which changes sandboxing behavior for user's rules. We can hold on this PR for now. But we still need a fix, CI is currently broken on master with dune 3.23.1.

hhugo:
@vouillon, should we push the other commits now ? (gen-rules + updated deps in tests)

PR #2232, opened 2026-05-06
Wasm: dispatch lifecycle events when async instantiation completes

### LINKED ISSUE #11: [BUG] wasm is not ready when DOMContentLoaded fires
**Describe the bug**
In 
[fiat-html.tar.gz](https://github.com/ocaml-wasm/wasm_of_ocaml/files/13446001/fiat-html.tar.gz) (unpack the tar.gz, run `python -m http.server 8000` in the `fiat-html` directory, and then navigate to http://localhost:8000/?argv=%5B%22-h%22%5D&interactive&wasm in Chrome), the code fails with "ReferenceError: synthesize is not defined".  But if you change the `setTimeout` in `main.js` to be 1000 instead of 1, so that it waits a full second, then the code works fine.

The .tar.gz contains the standalone wasm_fiat_crypto.ml used to generate the wasm with `-package unix -w -20 -g -package js_of_ocaml -linkpkg` and then `wasm_of_ocaml --source-map --no-inline`.

**Expected behavior**
All exported functions should be available by the time DOMContentLoaded fires

**Versions**
Chrome Version 119.0.6045.160 (Official Build) (64-bit)
ocaml-option-fp.1,ocaml-variants.4.14.1+options
```
$ wasm_of_ocaml --version
5.3.0+git-f517bbca-dirty
$ ocamlc --version
4.14.1
```

JasonGross:
## Summary
- Compiling and instantiating the generated WebAssembly is asynchronous, but the launcher only returns a Promise that callers typically don't await. Surrounding JavaScript that runs on `DOMContentLoaded` therefore has no way to tell when the OCaml exports are actually ready.
- This PR dispatches `wasmoocaml:loaded` (and `wasmoocaml:error` on failure) `CustomEvent`s on `globalThis` when the asynchronous Wasm instantiation completes, so external code can wait for the runtime to be ready.
- Also adds `CustomEvent` to `compiler/lib/reserved.ml` so the free-variable check in `compiler/bin-wasm_of_ocaml/gen/gen.ml` accepts the new reference (`CustomEvent` joins existing standard DOM globals like `XMLHttpRequest`/`DOMException`).
- Ported from ocaml-wasm/wasm_of_ocaml#143 (which fixes ocaml-wasm/wasm_of_ocaml#11) to the merged js_of_ocaml repository.

## Test plan
- [x] `make lint-js` passes (only an unrelated biome.json schema-version notice)
- [x] `dune build runtime/wasm` succeeds
- [x] `dune build compiler/lib compiler/lib-wasm compiler/bin-wasm_of_ocaml compiler/bin-js_of_ocaml runtime` succeeds (verifies the free-variable check accepts `CustomEvent`)
- [ ] `make tests` — could not run locally; this switch lacks `ppx_expect` and other test-only dependencies. CI should cover it.
- [ ] In a browser, a `wasmoocaml:loaded` event fires on `globalThis` after the Wasm runtime finishes initializing
- [ ] When instantiation throws, a `wasmoocaml:error` event fires before the error propagates

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,9 @@
   API, plus a `fonts` property on `Dom_html.document` (#2255)
 * Lib: `Lwt_js_events` `load`/`error`/`abort` and their `seq_loop` variants now
   accept any element, not only images (#2404)
+* Runtime/wasm: dispatch `wasmoocaml:loaded` and `wasmoocaml:error`
+  `CustomEvent`s on `globalThis` so that surrounding JavaScript can wait
+  for the asynchronous Wasm instantiation to complete
 
 ## Bug fixes
 * Compiler/Wasm runtime: fix toplevels built on Windows — the embedded cmi
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,10 @@
 * Runtime/wasm: dispatch `wasmoocaml:loaded` and `wasmoocaml:error`
   `CustomEvent`s on `globalThis` so that surrounding JavaScript can wait
   for the asynchronous Wasm instantiation to complete
+* Compiler: standalone executables now dispatch `jsoocaml:loaded` and
+  `jsoocaml:error` `CustomEvent`s on `globalThis`, mirroring the Wasm
+  lifecycle events, so that surrounding JavaScript can wait for the
+  program to run the same way with either backend
 
 ## Bug fixes
 * Compiler/Wasm runtime: fix toplevels built on Windows — the embedded cmi
--- a/compiler/bin-js_of_ocaml/compile.ml
+++ b/compiler/bin-js_of_ocaml/compile.ml
@@ -263,6 +263,7 @@ let run
       (one : Parse_bytecode.one)
       ~check_sourcemap
       ~standalone
+      ~lifecycle_events
       ~shapes
       ~(source_map : Source_map.Encoding_spec.t option)
       ~link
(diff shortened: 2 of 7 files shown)


JasonGross:
I assume all of the CI failures are spurious?

hhugo:
re-looking at this, I wonder if we should have a symmetric behavior in the js_of_ocaml side so that it's easier to switch between backend

JasonGross:
> re-looking at this, I wonder if we should have a symmetric behavior in the js_of_ocaml side so that it's easier to switch between backend

Done

JasonGross:
CI failure seems like a flake:
[ERROR] Failed to get sources of xtmpl.1.2.0: curl failed

#=== ERROR while fetching sources for xtmpl.1.2.0 =============================#
OpamSolution.Fetch_fail("https://www.good-eris.net/xtmpl/releases/xtmpl-1.2.0.tar.bz2 (curl failed: \"/usr/bin/curl --write-out %{http_code}\\\\n --retry 10 --retry-delay 2 --user-agent opam/2.5.2 -L -o /tmp/opam-53840-0f5ff8/xtmpl-1.2.0.tar.bz2.part -- [https://www.good-eris.net/xtmpl/releases/xtmpl-1.2.0.tar.bz2\](https://www.good-eris.net/xtmpl/releases/xtmpl-1.2.0.tar.bz2/)" exited with code 28)")

PR #2233, opened 2026-05-06
Variable coalescing: fix handling of break instruction

vouillon:
This PR also provides a better fix for empty try bodies (#2201).

### DIFF
--- a/compiler/lib/js_variable_coalescing.ml
+++ b/compiler/lib/js_variable_coalescing.ml
@@ -419,9 +419,15 @@ let build_cfg stmts candidates param_vars =
      - [exit] is the node id of the statement following the current one.
      - [context] contains the targets for break and continue statements.
 
-     The function returns the node id of the first statement of the visited block
-     (which acts as the entry point for that block). *)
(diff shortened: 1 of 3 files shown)


compiler/lib/js_variable_coalescing.ml
```diff
@@ -594,7 +613,8 @@ let build_cfg stmts candidates param_vars =
         then visit_stmt ~labels:(labels @ all_labels) context exit inner_stmt
         else
           let context =
-            { labels = all_labels; break = exit; continue = None } :: context
```
hhugo, line 597: Unrelated to the diff but, `labels` is ignored here but used on line 594 above. Maybe we could have 2 mutually recursive functions `visit_stmt` and `visit_stmt_own_context labels:...`

vouillon: I have pushed a commit. Is it what you had in mind?

hhugo: Perfect, thanks

hhugo (approved):

PR #2236, opened 2026-05-06
CI fix with dune 3.23

vouillon:
dune 3.23 adds a lower bound to the menhir dependency (ocaml/dune#14168).


hhugo:
Don't you need to add the constraint to the dune-project file as well ?

Issue #596, opened 2017-06-15
Provide bindings to the Fetch API


vouillon:
https://fetch.spec.whatwg.org/

hhugo:
closed by https://github.com/ocsigen/js_of_ocaml/pull/2239

PR #2239, opened 2026-05-12
Lib: add Promise + Fetch bindings (#2031, #596)

### LINKED ISSUE #2031: [FEATURE REQUEST] Integrate a Promise API
The folks at Funkyworker have an API that nicely sidesteps the type unsafety of the JS Promise API: https://github.com/funkywork/nightmare/blob/main/lib/js/promise.mli (see also https://github.com/funkywork/nightmare/blob/main/lib/js/stubs/caml_promise_stubs.js)

Should we consider including it in Js_of_ocaml? I hear that they would be fine with this copyright-wise.

Having a Promise API would also unlock #596.

hhugo:
This is following #2221

## Summary

Closes [#2031](https://github.com/ocsigen/js_of_ocaml/issues/2031) and addresses the long-standing Fetch request ([#596](https://github.com/ocsigen/js_of_ocaml/issues/596)).

- **`Promise` module** in `lib/js_of_ocaml/`: bindings to JS Promise with an indirection wrapper that prevents auto-flattening of thenables, so `'a Promise.t Promise.t` stays sound at the OCaml type level. Wrap helpers live in `runtime/{js,wasm}/promise.{js,wat}` and are exposed via `Jsoo_runtime.Promise`. Wrapping is conditional on the value being thenable, so non-thenable resolves pay no allocation and foreign promises handed in via `of_any` flow through unchanged. `then_` takes an optional `?on_error` for the two-callback `.then(f, g)` form.
- **`Js_of_ocaml_lwt.Promise.{to_lwt,of_lwt}`** for Lwt interop, with a `Rejected of error` exception for non-OCaml rejection reasons.
- **`Abort` module**: general-purpose `AbortController` / `AbortSignal` primitive, separate from `Fetch` so it can be reused by future bindings (Streams, custom cancellation, …).
- **`Fetch` module**: `Headers`, `Request`, `RequestInit`, `Response`, `ResponseInit` class types and the `fetch` entry point. Body-reader methods (`text`, `json`, `blob`, `arrayBuffer`, `formData`) typed with `Promise.t`. Uses `Abort.signal` for the `signal` field.
- **More Promise-typed `Dom_html` bindings**: `requestFullscreen` / `exitFullscreen` / `requestPointerLock` now return `unit Promise.t Js.meth` (legacy fire-and-forget kept under `_`-suffixed names); `mediaElement.play` returns `unit Promise.t Js.meth` (legacy `play_` kept); `Animation.finished` and `Animation.ready` exposed as `animation t Promise.t readonly_prop`.
- **`CONTRIBUTING.md`**: documents the existing leading-`_` (mangling) vs trailing-`_` (alternate form / legacy stub) split for new bindings.
- **`API-STATUS.md`**: updated to mark Promise, Fetch, AbortController, Fullscreen, Pointer Lock as bound; collapsed Tier 1.

Out of scope (left for follow-ups): `ReadableStream` / response body streaming, `Request` clone-from-Request constructor, richer `FormData` integration, the wider modern surface on `navigator` (mediaDevices, clipboard, serviceWorker, permissions, credentials, …).

## Test plan

- [x] `dune build` (jsoo)
- [x] `WASM_OF_OCAML=true dune build` (wasm)
- [x] `dune runtest lib/tests` — sync sanity tests for `Promise` (constructor / instanceof / make sync body)
- [x] `dune runtest compiler/tests-jsoo` — end-to-end async `test_promise` covering resolve / then_ / catch / finally / all / race / nested promise / `of_any` of a foreign promise / reject-with-promise (verifies that JS doesn't auto-follow on rejection)
- [x] `dune runtest compiler/tests-jsoo` — `test_lwt_promise` round-trips Promise ↔ Lwt for both fulfilled and rejected paths

### DIFF
--- a/lib/js_of_ocaml/dom_html.mli
+++ b/lib/js_of_ocaml/dom_html.mli
@@ -1163,6 +1163,14 @@ and animation = object
 
   method updatePlaybackRate : number_t -> unit meth
 
+  method finished : animation t Promise.t readonly_prop
+  (** Resolves when the animation reaches its end, or rejects when it is
+      cancelled. *)
+
+  method ready : animation t Promise.t readonly_prop
+  (** Resolves when the animation is ready to play (i.e. the user agent
+      has finished any pending changes to its state). *)
+
   method oncancel : (animation t, animationPlaybackEvent t) event_listener writeonly_prop
 
   method onfinish : (animation t, animationPlaybackEvent t) event_listener writeonly_prop
@@ -1426,14 +1434,9 @@ and element = object
 
   method blur : unit meth
 
-  method requestFullscreen_ : unit meth
-  (** Returns a [Promise] in JavaScript. Bound as [unit meth] to avoid pulling
-      in a Promise type; the proper version can later be added under
-      [requestFullscreen]. *)
+  method requestFullscreen : unit Promise.t meth
 
-  method requestPointerLock_ : unit meth
-  (** Returns a [Promise] in JavaScript (since 2024). Bound as [unit meth];
-      the proper version can later be added under [requestPointerLock]. *)
+  method requestPointerLock : unit Promise.t meth
 
   method animate : 'a 'b. 'a -> 'b -> animation t meth
 
@@ -2275,6 +2278,8 @@ class type imageElement = object ('self)
 
   method complete : bool t prop
 
+  method decode : unit Promise.t meth
+
   method onload : ('self t, event t) event_listener prop
 
   method onerror : ('self t, event t) event_listener prop
@@ -2555,7 +2560,7 @@ class type mediaElement = object
 
   method load : unit meth
 
-  method play : unit meth
+  method play : unit Promise.t meth
 
   method pause : unit meth
 
@@ -3139,9 +3144,7 @@ class type document = object
 
   method timeline : documentTimeline t readonly_prop
 
-  method exitFullscreen_ : unit meth
-  (** Returns a [Promise] in JavaScript. Bound as [unit meth]; the proper
-      version can later be added under [exitFullscreen]. *)
+  method exitFullscreen : unit Promise.t meth
 
   method exitPointerLock : unit meth
 
--- a/lib/js_of_ocaml/dom_html.mli
+++ b/lib/js_of_ocaml/dom_html.mli
@@ -2684,7 +2684,7 @@ class type canvasElement = object
 
   method toDataURL_type : js_string t -> js_string t meth
 
-  method toDataURL_type_compression : js_string t -> number_t -> js_string t meth
+  method toDataURL_compression : js_string t -> number_t -> js_string t meth
 
   method getContext : context -> canvasRenderingContext2D t meth
 end
--- a/lib/js_of_ocaml/fetch.ml
+++ b/lib/js_of_ocaml/fetch.ml
@@ -1,0 +1,174 @@
+(* Js_of_ocaml library
+ * http://www.ocsigen.org/js_of_ocaml/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, with linking exception;
+ * either version 2.1 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+open! Import
+
+class type headers = object
+  method append : Js.js_string Js.t -> Js.js_string Js.t -> unit Js.meth
+
+  method delete : Js.js_string Js.t -> unit Js.meth
+
+  method get : Js.js_string Js.t -> Js.js_string Js.t Js.opt Js.meth
+
+  method has : Js.js_string Js.t -> bool Js.t Js.meth
+
+  method set : Js.js_string Js.t -> Js.js_string Js.t -> unit Js.meth
+
+  method forEach :
+       (Js.js_string Js.t -> Js.js_string Js.t -> headers Js.t -> unit) Js.callback
+    -> unit Js.meth
+end
+
+let headers : headers Js.t Js.constr = Js.Unsafe.global##._Headers
+
+let headers_of_list (l : (string * string) list) : headers Js.t =
+  let h : headers Js.t =
+    Js.Unsafe.new_obj (Js.Unsafe.global##._Headers : _ Js.constr) [||]
+  in
+  List.iter (fun (k, v) -> h##append (Js.string k) (Js.string v)) l;
+  h
+
+class type body = object
+  method bodyUsed : bool Js.t Js.readonly_prop
+
+  method arrayBuffer : Typed_array.arrayBuffer Js.t Promise.t Js.meth
+
+  method blob : File.blob Js.t Promise.t Js.meth
+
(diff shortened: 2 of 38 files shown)


lib/js_of_ocaml/fetch.ml
```diff
@@ -0,0 +1,171 @@
...
+  method forEach :
+       (Js.js_string Js.t -> Js.js_string Js.t -> headers Js.t -> unit) Js.callback
+    -> unit Js.meth
+end
+
+let headers : (unit -> headers Js.t) Js.constr = Js.Unsafe.global##._Headers
+
+let headers_of_array (arr : (Js.js_string Js.t * Js.js_string Js.t) Js.js_array Js.t) :
+    headers Js.t =
+  Js.Unsafe.new_obj (Js.Unsafe.global##._Headers : _ Js.constr) [| Js.Unsafe.inject arr |]
```
vouillon: `arr` is a JS array of OCaml tuples `[0, k, v]` while the Headers constructor expects an array of arrays `[k, v]`.

lib/lwt/lwt_promise.mli
```diff
@@ -0,0 +1,36 @@
...
+
+exception Rejected of Promise.error
+(** Raised by the [Lwt.t] returned by {!to_lwt} when the underlying
+    [Promise] rejects with a non-exception value. If the promise was
+    rejected with an OCaml exception (e.g. via {!Promise.error_of_exn})
+    that exception is raised directly instead. *)
+
+val to_lwt : 'a Promise.t -> 'a Lwt.t
+(** [to_lwt p] is an Lwt thread that fulfils with the value [p] resolves
+    with, or fails with {!Rejected} (or the original exception) if [p]
```
vouillon: This always fails with `Rejected` if `p` is rejected.

vouillon (changes requested):

lib/js_of_ocaml/fetch.ml
hhugo: replaced by `headers_of_list`

runtime/js/promise.js
```diff
@@ -0,0 +1,45 @@
...
+function caml_jsoo_promise_wrapper(x) {
+  this.wrapped = x;
+}
+
+//Provides: caml_jsoo_promise_wrap
+//Requires: caml_jsoo_promise_wrapper
+function caml_jsoo_promise_wrap(x) {
+  return x != null && typeof x.then === "function"
+    ? new caml_jsoo_promise_wrapper(x)
+    : x;
```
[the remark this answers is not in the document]
hhugo: Won't fix

lib/js_of_ocaml/dom_html.mli
```diff
@@ -1427,13 +1435,16 @@ and element = object
   method blur : unit meth
 
   method requestFullscreen_ : unit meth
-  (** Returns a [Promise] in JavaScript. Bound as [unit meth] to avoid pulling
-      in a Promise type; the proper version can later be added under
-      [requestFullscreen]. *)
+  (** Fire-and-forget binding. See {!requestFullscreen} for the version that
+      exposes the returned [Promise]. *)
+
+  method requestFullscreen : unit Promise.t meth
```
vouillon: I think we can just remove the variants with an underscore, since they were introduced after the last release.

vouillon (commented):
We could also add `HTMLImageElement.decode()`.

lib/js_of_ocaml/fetch.mli
```diff
@@ -0,0 +1,177 @@
...
+  method keepalive : bool t readonly_prop
+
+  method signal : Abort.signal t readonly_prop
+
+  method clone : request t meth
+end
+
+val request : (js_string t -> request t) constr
+
+val request_with_init : (js_string t -> requestInit t -> request t) constr
```
vouillon, line 130: The first argument to `new Request(...)` can also be an existing `Request`.

lib/js_of_ocaml/promise.ml
```diff
@@ -0,0 +1,100 @@
...
+
+let bind f p = then_ f p
+
+let all (ps : 'a t list) : 'a list t =
+  let arr = Js.array (Array.of_list ps) in
+  let raw = Js.Unsafe.meth_call promise_global "all" [| Js.Unsafe.inject arr |] in
+  let cb =
+    Js.wrap_callback (fun (a : Js.Unsafe.any Js.js_array Js.t) ->
+        let lst = List.map unwrap (Array.to_list (Js.to_array a)) in
+        resolve lst)
```
vouillon: We could just return the list, since `lst` is not a promise.

lib/js_of_ocaml/dom_html.mli
hhugo: done

lib/js_of_ocaml/promise.mli
```diff
@@ -0,0 +1,126 @@
...
+(** [finally f p] returns a new promise that settles the same way as [p],
+    after invoking [f] for its side effect. *)
+
+val map : ('a -> 'b) -> 'a t -> 'b t
+(** [map f p] is [then_ (fun x -> resolve (f x)) p]. *)
+
+val bind : ('a -> 'b t) -> 'a t -> 'b t
+(** Alias for {!then_}. *)
+
+(** {1 Combinators} *)
```
vouillon, line 105: What about `allSettled`, `any`, and `withResolvers`?

lib/js_of_ocaml/promise.ml
hhugo: I think I prefer a stronger type-safety at the cost of a conditional no-op wrapping. I've added a bunch of type annotations.

lib/js_of_ocaml/promise.mli
hhugo: added

lib/js_of_ocaml/fetch.mli
hhugo: added

lib/js_of_ocaml/promise.ml
vouillon: The cost is that `resolve` always allocates a promise.

hhugo: Indeed

hhugo: Wait, but `.then` will need to allocate a Promise if the result of the cb is not a Promise  anyway

vouillon: `.then` always allocates a promise.

hhugo: Ive pushed a change

hhugo:
@vouillon, any other review on this PR ?

lib/lwt/lwt_promise.ml
```diff
@@ -0,0 +1,39 @@
...
+open Js_of_ocaml
+
+exception Rejected of Promise.error
+
+let to_lwt (p : 'a Promise.t) : 'a Lwt.t =
+  let t, w = Lwt.task () in
+  let _ : unit Promise.t =
+    Promise.then_
+      ~on_error:(fun e ->
+        Lwt.wakeup_later_exn w (Rejected e);
```
vouillon, line 28: This makes it more robust against cancellation:
```suggestion
        if Lwt.is_sleeping t then Lwt.wakeup_later_exn w (Rejected e);
```

lib/lwt/lwt_promise.ml
```diff
@@ -0,0 +1,39 @@
...
+        Promise.resolve ())
+      (fun x ->
+        Lwt.wakeup_later w x;
```
vouillon, line 31:
```suggestion
        if Lwt.is_sleeping t then Lwt.wakeup_later w x;
```

lib/lwt/lwt_promise.ml
hhugo: Here is what the lwt doc says
```
    If the promise is not pending, [Lwt.wakeup_later] raises
    {!Stdlib.Invalid_argument}, unless the promise is {{!Lwt.cancel} canceled}.
    If the promise is canceled, [Lwt.wakeup_later] has no effect.
```

vouillon: Oh, right!

vouillon (approved):

PR #2243, opened 2026-05-21
Avoid stack overflow on deep mutual recursion under --effects=double-translation

hhugo:
## Summary

- Under `--effects=double-translation`, cps_needed mutually recursive functions get both a direct and a CPS version (paired via `caml_cps_closure`). The CPS side trampolines on `caml_stack_check_depth`, but the direct side called its siblings via ordinary JS calls, so a sufficiently deep direct-style mutual recursion blew the JS stack.
- `Generate_closure` now runs under double-translation. For each SCC of paired closures it renames the direct half, gates every recursive tail call on `caml_stack_check_depth` (mirroring the CPS-side pattern), and replaces the direct slot of `caml_cps_closure` with a wrapper that drives a runtime `caml_direct_trampoline` loop. When the guard fires the call bounces to the same inner direct closure via `caml_trampoline_return`, and the wrapper's loop reapplies it with a fresh stack budget — no CPS involvement. An alternative would be to force cps for mutually recursive calls at the cost of slower runtime. 
- New regression test `compiler/tests-jsoo/lib-effects/mutual_recursion_deep.ml` runs `ping`/`pong` mutual recursion at depth 1,000,000.
- New expect test `compiler/tests-compiler/double-translation/mutual_recursion.ml` snapshots the emitted JS for paired mutual recursion (the `caml_cps_closure` pairing, the `caml_direct_trampoline` wrapper on the direct half, and the `caml_stack_check_depth`-gated tail calls that bounce via `caml_trampoline_return`).
- Refactors the `Generate_closure` paired-closure machinery into single-purpose helpers and effects-mode-specific dispatch (`dispatch_component_disabled` / `dispatch_component_dt`). No behavior change — generated output verified byte-identical at each step.

## Test plan

- [x] `dune build @compiler/tests-compiler/double-translation/runtest` clean (verified after the refactor commits)

Verified on the original fix but **not re-run after the follow-up refactor commits** — re-confirm before merge:

- [ ] `dune runtest compiler/tests-jsoo/lib-effects/` under `default`, `with-effects`, and `with-effects-double-translation`
- [ ] `dune build @compiler/tests-jsoo/lib-effects/runtest-js` (all three profiles)
- [ ] `dune runtest compiler/tests-jsoo --profile=with-effects-double-translation`
- [ ] `dune build @compiler/tests-jsoo/runtest-js --profile=with-effects-double-translation`
- [ ] `dune build @compiler/tests-ocaml/effects/runtest --profile=with-effects-double-translation`
- [ ] `dune build @lib/tests/runtest-js --profile=with-effects-double-translation`
- [ ] Pre-existing `js_parser_printer.ml` (`using` / `html comments`) failures unrelated to this change (reproduce on master under the same Node version)

### DIFF
--- a/compiler/lib/generate_closure.ml
+++ b/compiler/lib/generate_closure.ml
@@ -127,6 +127,24 @@ type w =
 
 let wrapper_closure pc args cloc = Closure (args, (pc, []), cloc)
 
+(* Source location of a closure's entry block, used to tag the wrapper. *)
+let start_loc blocks ci =
+  let block = Addr.Map.find (fst ci.cont) blocks in
+  match block.body with
+  | Event loc :: _ -> loc
+  | _ -> Parse_info.zero
+
+let debug_cycle msg all =
+  if debug_tc ()
+  then (
+    Format.eprintf "%s of size (%d).\n%!" msg (List.length all);
+    Format.eprintf
+      "%a\n%!"
+      (Format.pp_print_list
+         ~pp_sep:(fun fmt () -> Format.pp_print_string fmt ", ")
+         Var.print)
+      all)
+
 module Trampoline = struct
   let direct_call_block ~counter ~x ~f ~args =
     let return = Code.Var.fork x in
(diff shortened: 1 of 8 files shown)


hhugo:
@4y8 do you want to try this PR with `--effects=double-translation`.

4y8:
That worked, thanks!

compiler/lib/generate_closure.ml
```diff
@@ -136,149 +183,302 @@ module Trampoline = struct
...
+    }
+
+  let wrapper_block inner ~args loc =
+    let args_arr = Code.Var.fresh () in
+    let result = Code.Var.fresh () in
+    { params = []
+    ; body =
+        [ Event loc
+        ; Let (args_arr, Prim (Extern "%js_array", List.map args ~f:(fun x -> Pv x)))
+        ; Let (result, Prim (Extern "caml_direct_trampoline", [ Pv inner; Pv args_arr ]))
```
vouillon, line 349:
```suggestion
        ; Event Parse_info.zero
        ; Let (result, Prim (Extern "caml_direct_trampoline", [ Pv inner; Pv args_arr ]))
```

vouillon (approved):

PR #2245, opened 2026-05-21
Extract loops from toplevel function into helper functions

vouillon:
Extract outermost loops of the entry function into helper closures, so engines tier up the small helpers instead of the giant initialization code. The pass benefits both JavaScript and WebAssembly output.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
   asynchronous toplevel originally contributed in the unmerged #435 (#66, #833)
 * Lib: add `FontFace` module — partial binding to the CSS Font Loading
   API, plus a `fonts` property on `Dom_html.document` (#2255)
+* Extract loops from toplevel function into helper functions (#2245)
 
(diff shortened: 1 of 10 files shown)


hhugo:
Looking quickly at the benchmarks, it seems to be within the noise.
I would expect this optim to mostly benefit short lived programs. @vouillon, do you have any example that see significant improvement with this PR ?

vouillon:
The main issue is that V8 can be very slow, or even crash, when optimizing a large Wasm function. Extracting loops help a bit since it make it less likely for the function to get optimized (loops can make it look hot). But it does not actually really help. If we are loading the same code several time (for instance when reloading the page, or opening it in another tab), a large function will eventually get optimized. So, the right fix is #2423.

Regarding JavaScript, this might help if the toplevel function is quite large and spends a significant amount of time in a loop: the loop can get optimized without having to optimize the whole toplevel function.

PR #2247, opened 2026-05-22
Shorten --build-config output

vouillon:
Use a compact encoding for the `--build-config` output: a key at its default is omitted; all other keys still emit `name=value`. Restructure `set_values` to iterate over the known keys rather than the input entries, so that an omitted bool is reset to its default (instead of silently keeping whatever value the caller left it at) and a missing enum is reported as an error.

For the typical wasm builds this reduces e.g.
```
  effects=jspi+toplevel=false+use-js-string=true  (45 chars)
```
to
```
  effects=jspi                                    (12 chars),
```
which keeps the build-config directory that dune derives from this output well within Windows' MAX_PATH limit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -233,6 +233,10 @@
   translation; this fixes `ba_to_ba` (where the read position was
   translated but the write position was not) as well as `string_to_ba`,
   `bytes_to_ba` and `ba_to_bytes` for fortran_layout bigarrays
+* Runtime: bigarray hashing now zero-extends tail elements as the C
+  runtime does; signed 8/16-bit bigarrays whose size is not a multiple
+  of the hashed word size hashed differently from native (and the byte
+  case trapped under WASI on the Wasm runtime)
 * Runtime: the `#` flag no longer adds a base prefix to zero;
   `Printf.printf "%#x" 0` printed `0x0` where native prints `0`
 * Runtime: the fake filesystem no longer ignores `Open_append`; writes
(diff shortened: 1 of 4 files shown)


hhugo:
What about using a more cryptic but compact encoing that would reduce the path size of all configs, event non default ones ?

vouillon:
> What about using a more cryptic but compact encoding that would reduce the path size of all configs, event non default ones ?

Do you have anything in mind?

For context, here is the failure:
```
../../../../../install/default/bin/wasm_of_ocaml.exe: Error: Sys_error("../../../../.js/effects=jspi+toplevel=false+use-js-string=true+wasi=false/.runtime/afa295cfa152c532f6a91c8d8d067377/runtime.bc.runtime.wasma: No such file or directory")
```

Maybe we can shorten the path on dune's side as well? There are three occurrences of `runtime` in this path.

hhugo:
> Maybe we can shorten the path on dune's side as well? There are three occurrences of `runtime` in this path.

On the dune side, we could do something similar to ppx-es, where we map a payload (ie. jsoo config) to its digest in a table and have the digest in the path.

hhugo:
> Do you have anything in mind?

- We could omit boolean values and use `!` prefix for negation. `effects=jspi+!toplevel+use-js-string+!wasi`
- Or use 1/0 for true/false
- We could use field order and omit fieldname. `jspi,0,1,0`

vouillon:
So, maybe we can change dune to put runtimes in `.js/afa295cfa152c532f6a91c8d8d067377/runtime.bc.runtime.wasma`. And keep this PR to fix the issue I have with #1831 in the meantime.

hhugo:
> So, maybe we can change dune to put runtimes in `.js/afa295cfa152c532f6a91c8d8d067377/runtime.bc.runtime.wasma`. And keep this PR to fix the issue I have with #1831 in the meantime.

Ok, should we merge after the release ?

Issue #1438, opened 2023-03-23
[FEATURE REQUEST] Explain how to use third party libraries through wasm.


hhugo:
and change jsoo to ease initialization of wasm files

hhugo:
x-link https://gitlab.com/tezos/tezos/-/merge_requests/13440

PR #2249, opened 2026-05-26
Wasm implementations of zstd / blake2

vouillon:


hhugo:
Maybe it's a good time to start documenting how to rely on C code with wasm_of_ocaml (see https://github.com/ocsigen/js_of_ocaml/issues/1438)

hhugo:
Can you move the docker dep out of default alias ?

hhugo (approved):

hhugo:
### CI: `WASI / 5.3 / Ubuntu` failure — wasm linker crash on `zstd-impl.wasm`

The only failing check is **WASI / 5.3 / Ubuntu**. The `wasm_of_ocaml link` step raises `Invalid_argument("index out of bounds")` for **every** test that links under the `effects=native+wasi=true` profile (digests, blake2, atomic, …), so it's not test-specific. Reproduced locally:

```
dune build --profile wasi-with-native-effects compiler/tests-ocaml/lib-digest/digests.bc.wasm.js
→ wasm_of_ocaml: Error: Invalid_argument("index out of bounds")
```

**Backtrace** (with `OCAMLRUNPARAM=b`) points at the linker's data-section scanner:

```
Wasm_link.Scan.scanner.mem_map   compiler/lib-wasm/wasm_link.ml:1032   (maps.mem.(idx))
  ← rewrite (996) ← memidx (1033) ← data (1503) ← data_section (1512) ← write_section_with_scan ← Link.link_to_module
```

`wasm_link.ml:1503` is the **active data segment with explicit memory index** form (`0x02`): it reads a `memidx` and remaps it via `maps.mem.(idx)`, which overflows.

**Root cause.** The precompiled `runtime/wasm/zstd-impl.wasm` added here (the C zstd from `zstddeclib.c`) is a full module — unlike the hand-written `.wat` fragments it carries its own `memory` section (5) and a `data` section (11, ~2.3 KB) with active segments in the explicit-memidx form. When the linker merges it, remapping a data segment's memory index hits `maps.mem.(idx)` out of bounds — the per-module `mem` map (`wasm_link.ml:859/870`, default `[||]`) has no entry for that index. It only surfaces under WASI because that's where this C zstd module is linked into the program (elsewhere zstd comes from the JS runtime, so the blob isn't scanned).

**Suggested fix** (in `compiler/lib-wasm/wasm_link.ml`): handle a linked module that defines its own memory with explicit-memidx (`0x02`) data segments — populate `maps.mem` with that memory so the remap resolves (or normalize those segments to memory 0).

This is deterministic, not flaky — re-running won't help.

PR #2251, opened 2026-05-27
Lib: fix onbeforeunload handler breaking navigation (#1436)

### LINKED ISSUE #1436: [BUG] semantics of (onbeforeunload, Dom_html.handler) incompatible with major browsers

hhugo:
## Summary

Alternative to #2159, following @vouillon's [suggestion](https://github.com/ocsigen/js_of_ocaml/pull/2159#issuecomment-4555792459).

Fixes #1436.

Change `event_listener` to return `bool t optdef` instead of `bool t`, so a handler can return `undefined` ("no opinion") in addition to a boolean. The browser interprets any non-undefined return value from a `beforeunload` listener as a request to show the confirmation dialog — so under the old type, returning `Js._true` to allow navigation surfaced a spurious dialog. With the new shape, "allow" maps to `Js.undefined` and the dialog only appears when the handler returns `Js._false` (which also calls `preventDefault`).

Unlike #2159 this avoids the runtime `event.type === "beforeunload"` sniffing in `Dom.handler`.

## Changes

- `Dom.event_listener`: now `('a, 'b -> bool t optdef) meth_callback opt`.
- `Dom.handler` / `full_handler`: return `Js.undefined` on `Js._true`; return `Js.def Js._false` (+ `preventDefault`) on `Js._false`.
- `Dom.invoke_handler`: normalize `undefined` → `Js._true` via `Js.Optdef.get`.
- Add `Dom.listener` / `Dom.full_listener` for void-returning listeners (return `Js.def Js._false` only if the user code already called `preventDefault`).
- Retype `Dom_html.window##.onbeforeunload` and `Dom_html.Event.beforeunload` against the existing `beforeUnloadEvent` class; corresponding update in `lwt_js_events.mli`.
- Refresh the `lib/tests-browser/test_beforeunload.ml` strategies to use the new shape.

## Test plan

- [ ] `dune build`
- [ ] `dune build @lib/runtest`
- [ ] `dune build @compiler/tests-jsoo/runtest`
- [ ] Manually exercise `lib/tests-browser/test_beforeunload.ml` in a current browser (preventDefault path) — the dialog still appears for "block" strategies and is absent for "allow".
- [ ] Confirm Jane Street virtual_dom tests pass (the previous PR broke them).

### DIFF
--- a/lib/js_of_ocaml/dom.mli
+++ b/lib/js_of_ocaml/dom.mli
@@ -463,20 +463,32 @@ val no_handler : ('a, 'b) event_listener
 
 val handler : (('e #event t as 'b) -> bool t) -> ('a, 'b) event_listener
 (** Create an event handler that invokes the provided function.
-      If the handler returns false, the default action is prevented. *)
+    If the handler returns false, the default action is prevented. *)
 
 val full_handler : ('a -> ('e #event t as 'b) -> bool t) -> ('a, 'b) event_listener
 (** Create an event handler that invokes the provided function.
-      The event target (implicit parameter [this]) is also passed as
-      argument to the function.  *)
+    The event target (implicit parameter [this]) is also passed as
+    argument to the function.  *)
+
+val listener : (('e #event t as 'b) -> unit) -> ('a, 'b) event_listener
+(** Create an event listener that invokes the provided function and lets
+    the browser decide what to do based on whether [preventDefault] was
+    called.  Unlike [handler], the listener returns no opinion when
+    [preventDefault] was not called — which is what events like
+    [beforeunload] require to avoid spurious confirmation dialogs. *)
+
+val full_listener : ('a -> ('e #event t as 'b) -> unit) -> ('a, 'b) event_listener
+(** Same as [listener] but also passes the event target (implicit
+    parameter [this]) to the function. *)
 
 val invoke_handler : ('a, 'b) event_listener -> 'a -> 'b -> bool t
-(** Invoke an existing handler.  Useful to chain event handlers. *)
+(** Invoke an existing handler.  Useful to chain event handlers.
+    Listeners that return no opinion are normalized to [Js._true]. *)
 
 val eventTarget : (< .. > as 'a) #event t -> 'a t
 (** Returns which object is the target of this event.
-      It raises [Not_found] in case there is no target
-      (if the event has not been triggered yet) *)
+    It raises [Not_found] in case there is no target
+    (if the event has not been triggered yet) *)
 
 type event_listener_id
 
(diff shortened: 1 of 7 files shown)


lib/js_of_ocaml/dom.mli
```diff
@@ -463,20 +463,32 @@ val no_handler : ('a, 'b) event_listener
...
-      argument to the function.  *)
+    The event target (implicit parameter [this]) is also passed as
+    argument to the function.  *)
+
+val listener : (('e #event t as 'b) -> unit) -> ('a, 'b) event_listener
+(** Create an event listener that invokes the provided function and lets
+    the browser decide what to do based on whether [preventDefault] was
+    called.  Unlike [handler], the listener returns no opinion when
+    [preventDefault] was not called — which is what events like
+    [beforeunload] require to avoid spurious confirmation dialogs. *)
```
vouillon:
```suggestion
  (** Create an event listener from a [unit]-returning function.  Unlike
      [handler], which signals "prevent default" via a [false] return,
      the callback here must call [preventDefault] explicitly on the event.
      Convenient for events like [beforeunload] where there is no natural
      boolean to return. *)
```

vouillon (approved):

PR #2252, opened 2026-05-28
Lib: implement the Popover API

hhugo:
## Summary
- Add `Dom_html` bindings for the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API):
  - `element##.popover` attribute, and the `showPopover` / `hidePopover` / `togglePopover` methods, with `showPopover ?source`, `hidePopover`, and `togglePopover ?force ?source` wrapper functions taking labeled arguments.
  - `popoverTargetElement` / `popoverTargetAction` on `inputElement` and `buttonElement`.
  - `beforetoggle` / `toggle` events (`Event.beforetoggle`, `Event.toggle`), `onbeforetoggle` / `ontoggle` handlers on `eventTarget`, and `ToggleEvent.source`.
- Update `API-STATUS.md` (new Popover API row) and `CHANGES.md`.

Replaces #1734, rebased onto current `master` with the merge conflicts resolved. Original work by @SylvainBoilard.

### Note
`ontoggle` was previously declared on `detailsElement` as a read/write `prop`; it is now inherited from `eventTarget` as a `writeonly_prop` (consistent with the other `on*` handlers). Code that *read* the handler back will need to adjust; setting it is unchanged.

## Test plan
- [x] `dune build lib ppx examples toplevel` (the `Dom_html` library and its in-repo consumers)
- [x] `make tests`

### DIFF
--- a/lib/js_of_ocaml/dom_html.ml
+++ b/lib/js_of_ocaml/dom_html.ml
@@ -786,6 +786,8 @@ and toggleEvent = object
   method newState : js_string t readonly_prop
 
   method oldState : js_string t readonly_prop
+
+  method source : element t opt readonly_prop
 end
 
 and mediaQueryListEvent = object
@@ -892,6 +894,10 @@ and eventTarget = object ('self)
   method onpointerover : ('self t, pointerEvent t) event_listener writeonly_prop
 
   method onpointerup : ('self t, pointerEvent t) event_listener writeonly_prop
+
+  method onbeforetoggle : ('self t, toggleEvent t) event_listener writeonly_prop
+
+  method ontoggle : ('self t, toggleEvent t) event_listener writeonly_prop
 
   method dispatchEvent : event t -> bool t meth
 end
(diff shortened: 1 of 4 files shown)


lib/js_of_ocaml/dom_html.ml
```diff
@@ -1296,6 +1302,16 @@ and focusOptions = object
   method preventScroll : bool t writeonly_prop
 end
 
+and showPopover_options = object
+  method source : element t writeonly_prop
```
vouillon: This avoids the need for a coercion:
```suggestion
  method source : #element t writeonly_prop
```

lib/js_of_ocaml/dom_html.ml
```diff
@@ -1296,6 +1302,16 @@ and focusOptions = object
...
+end
+
+and togglePopover_options = object
+  method force : bool t writeonly_prop
+
+  method source : element t writeonly_prop
```
vouillon:
```suggestion
  method source : #element t writeonly_prop
```

vouillon (changes requested):

vouillon (approved):

PR #2253, opened 2026-05-28
Compiler: fix UGEINT bytecode lowering

vouillon:
The lowering used `[Ult (z, y)] = (z < y)` instead of `[not (Ult (y, z))] = (y >= z)`, giving the wrong result when the operands are equal.


hhugo:
Can the second commit contain a change to the test ?

hhugo (approved):

PR #1833, opened 2025-02-11
implement a wrapper of CSS Font Loading API

mesmerizingwaft:
# Summary
This PR partially adds a wrapper of  [CSS Font Loading API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API)

This PR includes:
- `css_font_load.ml(i)` : actual implementation of the wrapper
- adding a property `fonts` to `Dom_html.document` to return `FontFaceSet`
- `examples/css_font_load`

# Note
1. Hi. This is my first PR on this repo. Let me know if anything is missing in this PR.

2. I didn't include a test for this feature because it seems the font loading API is not working on test codes in a straightforward manner (which may require stub or something of FontFaceSet)
Do you guys have any idea to handle this?

PR #2255, opened 2026-05-28
Lib: implement the CSS Font Loading API

hhugo:
## Summary

Alternative to #1833, rebased onto current `master`, building on the original work by @mesmerizingwaft.

Adds a partial `FontFace` binding to the [CSS Font Loading API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API):
- `FontFace.fontFace` / `fontFaceSet` class types and `FontFace.create`.
- A `fonts` property on `Dom_html.document` returning the `FontFaceSet`.

On top of the original commit:
- Named the module `FontFace`, after its primary JS constructor — consistent with the other single-interface bindings (`IntersectionObserver`, `MutationObserver`, `MessageChannel`, `ServiceWorker`, …), and discoverable by the API name users reach for. The `FontFaceSet` type rides along in the same module, as `MessagePort` does under `MessageChannel`. `FontFace.create` drops the redundant `_font_face` suffix.
- Tightened binding type safety; descriptors (`family`, `style`, `weight`, …) are writable `prop`s and `delete` returns `bool t`.
- Obtain the `FontFace` constructor via `Js.Unsafe.global##._FontFace`, matching the convention used by every other binding (a bare global reference would `ReferenceError` where `FontFace` is absent).
- Added a browser test under `lib/tests-browser/` (`test_font_face`) covering construction, descriptor defaults/writes, load success/failure, `document.fonts.{check,load,ready,delete}`. The test chain ends in a terminal `Promise.catch` so an unexpected rejection records a failed check and still renders the summary.

## Test plan

- [x] `dune build lib/js_of_ocaml/`
- [ ] `make tests`
- [ ] Manually open `lib/tests-browser/test_font_face.html` in a browser and confirm all checks pass


lib/js_of_ocaml/css_font_loading.ml
```diff
@@ -0,0 +1,68 @@
...
+
+  method display : js_string t prop
+
+  method status : js_string t readonly_prop
+
+  method load : fontFace t Promise.t meth
+end
+
+class type fontFaceSet = object
+  method add : fontFace Js.t -> unit meth
```
vouillon:
```suggestion
  method add : fontFace Js.t -> fontFaceSet Js.t meth
```

hhugo: fixed

PR #2261, opened 2026-06-06
doc: migrate the documentation to odoc + wodoc

balat:
Migrates the Js_of_ocaml documentation (manual + API) to a single `dune build @doc` (odoc), themed with the Ocsigen site chrome by [wodoc](https://github.com/ocsigen/wodoc), and published to this repo's `gh-pages` (served at https://ocsigen.org/js_of_ocaml/) — on the model now shared by the other Ocsigen projects. The same odoc sources are also what ocaml.org renders.

## What changes
- `manual/*.wiki` → `manual/*.mld` (`wodoc convert`), with `manual/index.mld` + `manual/api.mld`; the manual lives in the `js_of_ocaml` package (`(documentation (package js_of_ocaml))`).
- All the per-project doc machinery is now **one declarative config**, [`doc/wodoc`](doc/wodoc): project name, packages, the left navigation, the sibling-package link table, etc. There is **no build script, no template/menu HTML, no Python and no highlight.js** in the repo any more — `wodoc build` generates the page template, the shared top menu, the version selector, the left nav, the cross-package reference resolution and the syntax-highlight starter.
- `.github/workflows/doc.yml`: a single `wodoc build` call (build + assemble), deployed to `gh-pages` (push master → `dev`; manual dispatch for releases with `set_latest`; overlays doc sources from master for pre-migration tags). The main CI keeps a `dune build @doc @doc-manual` compile check.
- `tools/check_manual_links.sh` is repurposed to check **external** links only — internal references are now verified by odoc at build time (`dune build @doc` fails on an unresolved `{!…}`).
- retires the wiki pipeline (removes `manual/*.wiki`, `manual/SYNTAX.md`, the `installdoc`/wikidoc rsync); keeps the interactive examples (`gen_dune_inc`, `@doc-manual`).

## Presentational markup (invisible on ocaml.org)
The `.mld` manual can carry CSS classes and layout containers via `{%wodoc:…%}` markers (e.g. `{%wodoc:div class=…%}`, or `{%wodoc:@ class=…%}` to put a class on the next element). Stock odoc doesn't know the `wodoc` target and simply drops it, so the **same sources** render as plain semantic docs on ocaml.org and as the themed website here — no fork, nothing extra to maintain.

## Build & preview
```
wodoc build --config doc/wodoc --label dev --out _doc-site/dev \
  --menu https://ocsigen.org/wodoc/menu.html [--local]
```
`--local` also fetches the shared `/css//img/` assets so the build is viewable offline (`python3 -m http.server`). See [`doc/README.md`](doc/README.md).

## Invariants
Links version-relative via `{{base}}`; version `<select>` absolute via `{{pub}}=/js_of_ocaml`; themed CSS at `/css/ocsigen-odoc.css`; no `/wodoc/` in output (bar the Wodoc tool's own menu link); cross-project content links relative.

Draft pending review. Merging deploys `dev/`; a manual `workflow_dispatch` on the latest release tag with `set_latest` flips the `/js_of_ocaml/` landing.


hhugo:
Can you also update the script in https://github.com/ocsigen/js_of_ocaml/blob/master/tools/check_manual_links.sh. or is it already handled by odoc ?

hhugo:
I gave the build script a try. It would be nice to have script download extra resources (such as svg and css) so that one can visualize the website locally.

balat:
Thanks @hhugo! Both addressed, and I pushed a few related cleanups:

1. check_manual_links.sh: internal references (`{!…}` / `{{!…}}` / manual pages) are now checked by odoc itself: dune `build @doc` fails on any unresolved ref. So I've repurposed the script to check external links only (the one thing odoc doesn't do), scanning the .mld manual for `{{:http…}` and checking them.

2. Local preview: good point. wodoc build now has a `--local` flag that fetches the shared assets (/css/…, /img/…) next to the output so the site renders offline:

```
wodoc build --config doc/wodoc --label dev --out _doc-site/dev \
  --menu https://ocsigen.org/wodoc/menu.html --local
cd _doc-site && python3 -m http.server   # http://localhost:8000/dev/js_of_ocaml/
```

Other changes since your review:

 - The per-project doc machinery is gone: no more build.sh, template.html, leftnav.html, gen-manual-nav.py, menu.wiki, resolve-siblings.py or *-highlight.js. It's all driven by one declarative config, doc/wodoc, consumed by wodoc build (which runs dune build @doc @doc-manual, themes every page, generates the nav, resolves cross-package refs, and ships a default syntax-highlight starter covering the eliom/lwt/js_of_ocaml ppx syntaxes).
 - The CI (doc.yml) now just calls wodoc build directly (no wrapper script), fetching the shared menu from its single canonical copy in ocsigen.github.io.
 - The PR description is updated to match.

One thing worth mentioning: in the .mld manual you can use `{%wodoc:…%}` markers to add CSS classes/containers; stock odoc drops them, so the same sources stay plain on ocaml.org and render themed here.

hhugo:
```sh
wodoc build --config doc/wodoc --label dev --out _doc-site/dev   --menu https://ocsigen.org/wodoc/menu.html --local
Fatal error: exception Sys_error("https://ocsigen.org/wodoc/menu.html: No such file or directory")
```

hhugo:
> ```shell
> wodoc build --config doc/wodoc --label dev --out _doc-site/dev   --menu https://ocsigen.org/wodoc/menu.html --local
> Fatal error: exception Sys_error("https://ocsigen.org/wodoc/menu.html: No such file or directory")
> ```

wodoc was not properly updated, It now works

hhugo:
image

In the menu of api pages, I think we need visual marker for individual sections

hhugo:
Style for tables needs to be fixed

image

vs what's online 

image

hhugo:
- What's the story for API/Manual versioning ?
- Do we need to regenerate all pages any time the template changes ?

balat:
Thanks @hhugo! Both fixed in the shared theme (`/css/ocsigen-odoc.css`), so a
fresh `--local` build picks them up:
- tables now have borders + a header row;
- the menu sections and the "On this page" outline have separators / more spacing.

Versioning: manual + API are one odoc tree, versioned together. CI rebuilds only `dev` (from `master`); a stable release is a frozen snapshot of `dev` on `gh-pages` (no rebuild). Let's keep only two versions as discussed.

Template/chrome changes need a rebuild (automatic for `dev`); CSS tweaks like these don't — they're served from the one shared stylesheet.

balat:
If you agree, I will merge by myself and check everything at the same time.
The CI will build only the dev version of the doc, and I need to synchronize with the stable version.

hhugo:
Reviewed the `.wiki` → `.mld` conversion for content loss and fixed a batch of issues in 09e1803. Several look like **systematic `wodoc convert` artifacts**, not one-offs — worth grepping for in the other Ocsigen projects going through the same migration:

### 1. Bare URLs: `//` eaten as wiki italic → odoc `{e `
The converter treats `//` in prose as wiki emphasis, so bare links get mangled:
```
See https://dune.readthedocs.io/...   →   See https:{e dune.readthedocs.io/...
```
This broke 3 external links (debug, overview, wasm_overview). The same URL written with link markup converts fine, so it only hits *bare* URLs.
**Grep:** `grep -rn 'https\?:{e' *.mld` (and more broadly `:{e ` / `://` survivors).

### 2. Nested lists silently flatten
`wodoc convert` emits nested wiki lists as indented light-syntax (`-`/`+`), but **odoc light syntax does not nest by indentation at all** (tested 2/3/4-space indents on odoc 3.2.1 — all render as flat sibling lists). Hierarchy is lost without any warning. Fix is heavy `{ul {li … {ul …}}}`. Hit `errors.mld` (AND/OR condition grouping) and `contribute.mld` (dir tree under `compiler/`).
**Grep:** `grep -rnE '^[[:space:]]+[-+] ' *.mld` to find indented (intended-nested) items.

### 3. Inline code containing braces
`{{{--opt {1,2,3}}}}` → `[--opt {1,2,3]}` (brace landed outside the code span).
**Grep:** look for unbalanced `[` … `]` / stray `}` in converted inline code.

### 4. Cross-package `<<a_api>>` → dead `{!…}` links
`<<a_api>>` refs to modules **outside the published doc package set** become `{!…}` that don't resolve — and odoc treats unresolved `{!…}` as a **warning, not an error**, so `dune build @doc` stays green and the broken link ships silently. Here it was `Jsoo_runtime.*` (`js_of_ocaml-compiler.runtime`) and `Wasm_of_ocaml_compiler_dynlink.*` (`wasm_of_ocaml-compiler`) — neither is in the doc set / `(sibling …)` list, so I demoted them to inline code.
**Worth doing per project:** build docs and scan the full output for `Failed to resolve reference`; any root module not in your wodoc `packages`/`sibling` set is a dead link.

### 5. Duplicate auto-derived heading anchors
Two headings with the same text (`Optimizations` as both an `{1:…}` section and a plain `{2 …}` subsection) produce a duplicate HTML `id` / "Label … is ambiguous" warning; `#anchor` links then resolve arbitrarily. Gave the second one an explicit distinct label.
**Grep:** check the build log for `Label '…' is ambiguous`.

---

Also a few non-content notes on the pipeline itself (left for you, didn't touch):
- `doc.yml` has no `concurrency:` guard — a master push (deploy `dev/`) and a release `workflow_dispatch` can race on `gh-pages`; the release job's plain `git push` would then fail non-fast-forward.
- `${{ github.event.inputs.version }}` is interpolated directly into `run:` shells (maintainer-only, low blast radius, but the textbook injection footgun).
- The `set_latest` toggle mentioned in the PR description isn't an input — release runs repoint `latest` unconditionally. (A routine master push correctly *cannot* flip `latest`, so the core safety property holds.)
- 5 stub pages (`bindings`, `camlp4`, `environment-variable`, `library`, `runtime-files`) are built but unreachable from the nav/index — fine if kept for URL stability, just flagging it's intentional.

hhugo:
Update on point #4 above: rather than demoting those two refs to inline code, I kept them as `{!…}` links and instead declared the owning packages as wodoc siblings in `doc/wodoc`:

```
(sibling Jsoo_runtime                   js_of_ocaml-compiler/Jsoo_runtime)
(sibling Wasm_of_ocaml_compiler_dynlink wasm_of_ocaml-compiler/Wasm_of_ocaml_compiler_dynlink)
```

Both target pages (and the `val-assume_no_perform` / `val-loadfile` anchors) exist in the `dune build @doc` output, so the cross-package links should resolve in the assembled site. I couldn't run `wodoc` locally to confirm the final stitched-up output — worth a check on your end.

The general warning in #4 still stands for the other migrations: any `{!Mod…}` ref to a package not in your wodoc `packages`/`sibling` set is a silent dead link (warning, not build error).

hhugo:
Verified now with an actual `wodoc build` (had to `opam pin wodoc` first). Correcting my previous follow-up:

- **`Jsoo_runtime.Effect.assume_no_perform`** (effects.mld) resolves **natively** — `js_of_ocaml-compiler` is in the dependency closure, so no sibling is needed. Adding one actually produced a broken **nested `<a><a>`** in the output.
- **`Wasm_of_ocaml_compiler_dynlink.loadfile`** (build-toplevel.mld) is the only one that needs a sibling: `wasm_of_ocaml-compiler` is *not* a dependency of `js_of_ocaml`, so without the sibling it renders as plain text (dead link).

Final state: one sibling added for `Wasm_of_ocaml_compiler_dynlink`; both refs confirmed as single, correct `<a href>`s in the assembled `_doc-site/`, and a full scan of the manual pages shows no other nested/broken anchors.

Refinement of the takeaway for the other migrations: a cross-package `{!Mod…}` ref only resolves natively if that package is in the project's **dependency closure**. Refs to *sibling-but-not-dependency* packages need an explicit `(sibling …)`; adding a sibling for a package that's *already* a dependency double-links it. Easiest check is to grep the assembled HTML for both unresolved refs (plain text) and `<a [^>]*>\s*<a ` (double anchors).

hhugo:
I've made some small fixes. You should take a look at https://github.com/ocsigen/js_of_ocaml/pull/2261#issuecomment-4701022496 as it might affect other project. I happy to let you merge this PR when you see fit.

balat:
Thank you Hugo and Cloclo ! I fixed the same issues on other projects.
Merging this.

balat:
Merged. Let me know if you see issues.

PR #2262, opened 2026-06-11
Wasm runtime clean up

vouillon:
On top of branch `wasi` for now.

- Preprocessor now uses `$wasi` instead of `wasi` for variables (like all other Wasm identifiers)
- Remove Binaryen's tuples and `pop` instruction
- Type fixes (Binaryen's validator accepts some invalid code)
-  Add a few missing function names (the preprocessor uses the export name when no explicit name is given, but this is non standard)

With these changes, [`wax`](https://github.com/vouillon/wax) is able to parse all the runtime files.


hhugo:
Can we land this before the release ?

vouillon:
As you like

hhugo:
> As you like

Can this break external wasm runtime files ?

vouillon:
> > As you like
> 
> Can this break external wasm runtime files ?

I don't think any external wasm runtime file uses conditionals at the moment.

hhugo:
Can we add a CI linting job that check wasm files can be converted to wax ?

Issue #2263, opened 2026-06-12
Wasm runtime: correctness bugs found in a review of the .wat files


hhugo:
This is the raw results of reviewing the wasm runtime using the claude fable5 model by instructing it to review wax formatted runtime files.
----------

This is the result of a review of the Wasm runtime (`runtime/wasm/*.wat`), performed on the `wasm-runtime-clean-up` branch, reading the files formatted with `wax` (Rust-like syntax) and cross-checking suspicious spots against the original `.wat`, the parallel JS runtime in `runtime/js/`, and the OCaml C runtime. Findings in the numeric files were additionally verified by compiling test programs with the in-tree `wasm_of_ocaml` and diffing against native. All of these pre-date the cleanup branch (none were introduced by it).

## Uncatchable traps where an OCaml exception (or a result) is expected

- [ ] #2280 **`float.wat` — `caml_parse_float` traps on `"nin"`** (verified by execution). The "nan" branch advances `i` before failing, and the "inf" check is not an `else`: for input `"nin"` it matches `i`/`n` then does an out-of-bounds `array.get` on `s[3]`. `float_of_string "nin"` kills the process with an uncatchable `WebAssembly.Exception`; native raises `Failure`.
- [x] #2271 **`bigarray.wat` — `bigarray_hash` traps under WASI on byte-kind bigarrays.** The int8/uint8/char word loop reads via `$dv_get_i32`, whose WASI implementation `ref.cast`s the backing array to `$i32_array`, but byte-kind bigarrays are backed by `$i8_array`. Hashing any such bigarray of ≥ 4 elements traps. Should use `$dv_get_i32_unaligned` (which casts to `$bytes`).
- [x] #2308 **`str.wat:646-649, 699-702` — `re_replacement_text` traps instead of raising `Failure`.** The group bound check uses `i32.gt_u` where it needs `i32.ge_u` (in both loops): for `\n` with `n = numgroups` the check passes and the following `array.get` is out of bounds. `Str.global_replace (Str.regexp "a") "\\1" s` traps; the JS runtime raises `Failure "Str.replace: reference to unmatched group"`.
[...]

PR #2271, opened 2026-06-12
Runtime: zero-extend bigarray hash tail elements (JS and Wasm)

hhugo:
The C runtime (`caml_ba_hash` in hash.c) reads tail bytes as `unsigned char` and 16-bit tail elements as `uint16`. Both jsoo runtimes got this wrong, in different ways:

- **JS** read tail elements through the typed array, sign-extending for the int8/int16 kinds. The sign-extended last byte ORs over the other tail bytes, so `int8_signed` bigarrays of sizes 5, 6 and 7 (same leading bytes, high tail bytes) all hash to the same value — and none of them match native.
- **Wasm** sign-extended the last tail byte (`dv_get_i8`) for every byte kind (int8/uint8/char). Additionally, its byte-kind word loop used the aligned `dv_get_i32`, which under WASI casts the backing array to `$i32_array` and traps — byte-kind bigarrays are backed by `$i8_array`. It now uses `dv_get_i32_unaligned` (identical behavior on the JS-bindings path, byte-wise reads under WASI).

These are the bigarray-hash items from #2263 and the follow-up JS runtime review.

Commits:
1. **Revive the test_bigarray expect tests** — `test_bigarray.ml` was sitting in the `test_float16` `(test)` stanza, which has no ppx_expect preprocessing and no inline-test runner: the module didn't compile (`Uninterpreted extension 'expect_test'`) and its tests never ran. It now has its own inline-tests library (gated on OCaml ≥ 5.2 for the float16 kind), taking `bigarray_stubs.c` with it. This is also why the bug survived: the pre-existing hash test only used sizes 20 and 300, which never leave a tail.
2. **Regression test** with tail-exercising sizes and high tail values, expectations recording the buggy JS values (including the size-5/6/7 collision).
3. **The fix** in both runtimes, expectations flipped to the native ground truth — the test passes identically under js and native (`best`); native passing confirms the values, and the unsigned kinds (which were already correct in JS) are unchanged.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -233,6 +233,10 @@
   translation; this fixes `ba_to_ba` (where the read position was
   translated but the write position was not) as well as `string_to_ba`,
   `bytes_to_ba` and `ba_to_bytes` for fortran_layout bigarrays
+* Runtime: bigarray hashing now zero-extends tail elements as the C
+  runtime does; signed 8/16-bit bigarrays whose size is not a multiple
+  of the hashed word size hashed differently from native (and the byte
+  case trapped under WASI on the Wasm runtime)
 * Runtime: the `#` flag no longer adds a base prefix to zero;
   `Printf.printf "%#x" 0` printed `0x0` where native prints `0`
 * Runtime: the fake filesystem no longer ignores `Open_append`; writes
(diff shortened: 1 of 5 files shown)


hhugo:
@vouillon, Can you check the wasm diff ?

vouillon:
This looks good to me.

Issue #2270, opened 2026-06-12
JS runtime: correctness bugs found in a review of runtime/js


hhugo:
This is the JS-runtime sibling of the Wasm runtime review in #2263. Every file in `runtime/js/*.js` was reviewed function by function, cross-checking suspicious spots against the OCaml C runtime and against the parallel Wasm runtime in `runtime/wasm/*.wat` (the two runtimes are meant to be behaviorally equivalent). Findings were verified by executing the actual runtime code under node v22 wherever practical: int64 and nat arithmetic were differentially tested against BigInt oracles (200k/20k random cases), the LXM PRNG against a BigInt transcription of `prng.c`, MD5/BLAKE2 against RFC test vectors, the zstd decoder against node's `zlib.zstdCompressSync` (300 cases incl. multi-block/RLE/repeat-offset/checksummed frames), mlBytes and the UTF-8 codecs against fuzzed models, and the Str engine against hand-assembled bytecode programs. Items already filed in #2263 (`Marshal.to_buffer` returning 0, the BLOCK32 *writer* truncation, JS-string hashing divergence, `Condition.wait` parity) and the already-fixed #2264 are not repeated.

## Uncatchable crashes / hangs where an OCaml exception (or a result) is expected

- [x] #2277 **`str.js:249-265` — `re_match`'s SIMPLEOPT/SIMPLESTAR/SIMPLEPLUS miss the end-of-string check** (verified by execution). The C engine guards each with `txt < end` (so does `str.wat`); the JS versions read `s[pos]` past the end, getting `undefined`, and `in_bitset(set, undefined)` tests **bit 0 of byte 0**, i.e. whether `\000` is in the class. Negated classes (`[^;]`) contain `\000`, and `str.ml` emits the simple opcodes for bare `[^x]*` / `[^x]+` when disjoint from what follows — so `Str.string_match (Str.regexp "[^;]*;") "no semicolon" 0` **loops forever** (SIMPLESTAR/SIMPLEPLUS), and SIMPLEOPT advances `pos` past the string end, corrupting match positions. (jsoo's own lib-str tests only use negated classes inside groups, which don't take this code path.)
[...]

PR #2280, opened 2026-06-12
Runtime: float correctness fixes across the JS and Wasm runtimes (#2263, #2270)

hhugo:
Consolidates the **float-related fixes from both runtime reviews** — `runtime/js`
(#2270) and the Wasm `.wat` files (#2263) — into one PR. Each fix is test-first,
with expectations recorded buggy-first then flipped to native-verified output.

All new tests live in `compiler/tests-jsoo/test_floats.ml` and run on the **js,
wasm, and native (`best`)** backends, with native as the oracle.

## JS runtime (`ieee_754.js`, from #2270)

**`caml_ldexp_float` double-rounded into the subnormal range** (`ieee_754.js:177-195`):
scaling through `2^-1023` rounds once on the intermediate and again on the final
multiply — `ldexp (1. +. epsilon_float) (-1075)` returned `0.` instead of
`2^-1074`. Ported musl's scalbn scaling (single final rounding), which is what
`float.wat` already uses.

**Hex float literals parsed inexactly** (`ieee_754.js:524-531` + the regex item):
`mantissa * Math.pow(2, exponent)` underflowed to 0 for representable values
(`0x1.123456789abcdp-1023` → `0.`) and double-rounded mantissas wider than 53
bits. The hex path now parses the digits into a BigInt and rounds once to the
nearest double (ties-to-even, correct overflow/underflow/subnormal boundaries),
and the regex accepts digits on either side of the point (`0x.8p1`), as C99 and
native do.

**`caml_format_float` beyond the `toFixed`/`toExponential` limits** (`ieee_754.js:468,475`
+ the ≥1e21 item): precisions above 100 threw an uncatchable `RangeError`, and
`%f` of values ≥ 1e21 printed a rescaled value instead of the exact expansion.
Both now go through an exact BigInt path (decompose the double into `m × 2^e`,
scaled division with round-half-even, matching C printf). The wasm-on-JS
`format_float` binding in `runtime/wasm/runtime.js` shares this code.

## Wasm runtime (`float.wat`, from #2263)

**`caml_parse_float` trapped on `"nin"` and accepted non-decimal literals**
(was #2307, folded in here):

- The `inf` check was a sibling of (not an `else` of) the `nan` check, so after
  `nan` matched its `n` and advanced the cursor, `inf` read past the end:
  `float_of_string "nin"` died with an uncatchable trap where native raises
  `Failure`. The `inf` check is now an `else` of the `nan` check.
- The non-WASI fallback coerced with JavaScript number parsing, which accepts
  binary/octal literals (`"0b101"` → `5.`, `"0o17"` → `15.`); it now rejects any
  character that cannot appear in a decimal float first, mirroring the JS
  runtime's `r_float` guard.

**`caml_float_of_hex` wrapped a huge exponent modulo 2^32** (`float.wat`, newly
fixed here): the exponent was accumulated in a 32-bit int and overflow was only
flagged when `exp*10+d` wrapped below the last digit, so an exponent whose value
mod 2^32 landed back in range slipped through — `float_of_string
"0x1p12884901890"` returned `4.0` instead of `infinity`. It now saturates to
overflow as soon as the exponent reaches `0x10000000` (far beyond any
representable double exponent, and small enough that `exp*10` can no longer
wrap), yielding `infinity` (or `0` for a negative exponent). The JS hex parser
above already handles this via BigInt.

---

Supersedes #2307 (its commit is included here). Remaining float items from the
reviews are left for follow-ups because they don't fit a js+wasm test (each would
need a toolchain or workload this suite can't run): `float32_hash` -0.0/NaN
normalization (OxCaml-only `float32.wat`/`float32.js`), `parse_format` `%+f`/`% f`
(unreachable from `Printf`), and `caml_make_vect`'s float-branch size limit
(needs a multi-GB allocation to observe).


vouillon:
Bigint raises the requirement from ECMAScript 6 to ECMAScript 2020.

hhugo:
> Bigint raises the requirement from ECMAScript 6 to ECMAScript 2020.

We already rely on optional chaining which is ecmascriot 2020. We re now documenting the babel workflow to downgrade ecmascript requirement. I would say it's fine to depend on it. What do you think

vouillon:
> > Bigint raises the requirement from ECMAScript 6 to ECMAScript 2020.
> 
> We already rely on optional chaining which is ecmascriot 2020. We re now documenting the babel workflow to downgrade ecmascript requirement. I would say it's fine to depend on it. What do you think

That's fine. But README.md still mentions ECMAScript 6.

runtime/wasm/float.wat
```diff
@@ -739,6 +746,33 @@
               (i32.add (local.get $buf) (local.get $len))))
 )
 (@else
+         ;; The JS number coercion below accepts things OCaml does not,
+         ;; e.g. binary/octal literals ("0b101", "0o17"); only a plain
+         ;; decimal float can reach this point, so reject any character
+         ;; that cannot appear in one.
+         (local.set $i (i32.const 0))
```
vouillon: We should start after the leading spaces.

hhugo: fixed

PR #2307, opened 2026-06-15
Runtime/wasm: fix caml_parse_float trap and reject non-decimal literals

hhugo:
`caml_parse_float` (wasm runtime, `runtime/wasm/float.wat`) had two bugs, both from the Wasm runtime review (#2263).

### 1. Uncatchable trap on inputs like `"nin"`

The `"inf"` check was a **sibling** of the `"nan"` check rather than an `else` of it, so both ran on the same 3-character window. For `"nin"`, the `"nan"` branch matched the leading `n` and advanced the cursor; the `"inf"` check then resumed from the moved cursor (`i`), matched `i`…`n`, and read one byte **past the end** of the string:

```
RuntimeError: array element access out of bounds
    at caml_parse_float (...)
```

`float_of_string "nin"` killed the process with an uncatchable `WebAssembly.Exception` where native raises `Failure`. Fixed by making the `"inf"` check an `else` of the `"nan"` check.

### 2. Accepts JavaScript binary/octal literals

The non-WASI fallback parsed the decimal path with JS number coercion, which accepts more than OCaml does:

- `float_of_string "0b101"` → `5.`
- `float_of_string "0o17"` → `15.`

native raises `Failure` for both. The fix scans the (underscore-stripped) string and rejects any character that cannot appear in a decimal float before coercing — mirroring the `r_float` guard in the JS runtime (`ieee_754.js`). The WASI path already rejects these via `strtod`'s end-pointer check, so only the `@else` branch changed.

### Test

Adds `compiler/tests-wasm_of_ocaml/parse_float.ml`, covering the trap (`"nin"`, `"nif"`, `"ina"`), the wrongly-accepted literals (`"0b101"`, `"0o17"`, `"1d0"`, `""`), and a range of valid values that must still parse (`"1.5"`, `"1e3"`, `".5"`, `"5."`, `"-2.5"`, `"1_000.5"`, `"nan"`, `"-nan"`, `"inf"`, `"-inf"`, `"infinity"`, `"0x1p4"`). It reproduces the out-of-bounds trap against the unpatched runtime and passes with the fix.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,10 @@
   Pretty-printed output (`--pretty`) is unchanged. (#1117)
 
 ## Bug fixes
+* Runtime/wasm: `float_of_string` no longer traps on inputs like `"nin"`
+  (the `nan`/`inf` detection read past the end of the string) and no
+  longer accepts JavaScript binary/octal literals such as `"0b101"` or
+  `"0o17"`; both now raise `Failure` like the native runtime (#2263)
 * Runtime: the QuickJS standard file descriptors raise on a write or
   read error instead of returning 0; an error on stdout (e.g. EPIPE)
   used to make the flush/write loop spin forever, and a read error was
--- a/compiler/tests-wasm_of_ocaml/dune
+++ b/compiler/tests-wasm_of_ocaml/dune
@@ -11,7 +11,7 @@
  (modes js wasm))
 
 (tests
- (names marshal_regressions)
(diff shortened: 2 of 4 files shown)


hhugo:
Folded into #2280, which now consolidates all the float-related fixes from both runtime reviews (#2263 Wasm + #2270 JS) into a single PR. The `caml_parse_float` commit from this branch is included there unchanged.

PR #2320, opened 2026-06-17
Doc: use conventional comment style in JavaScript examples

smorimoto:
The manual consistently wrote runtime annotation comments without a space after `//` (e.g. `//Provides:`), and a couple of JavaScript examples used `var`. This updates the documentation to a more conventional style:

- Use the spaced form `// Provides:` (and `// Requires:`, `// Version:`, etc.) for runtime annotations. Both forms are accepted by the linker; the spaced form is more idiomatic for JavaScript comments.
- Use `const` instead of `var` in the JavaScript examples.

Only the documentation is touched (`manual/*.mld`, the `Js.Unsafe.runtime_value` doc comment, and a `CHANGES.md` entry). The runtime sources themselves are left unchanged.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,9 @@
   from `0.x` fractions, normalise exponent forms (`1e+05` → `1e5`), and
   pick backticks for string literals when they reduce escape count.
   Pretty-printed output (`--pretty`) is unchanged. (#1117)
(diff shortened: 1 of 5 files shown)


smorimoto:
Related: https://github.com/ocamllabs/vscode-ocaml-platform/pull/2162

PR #2332, opened 2026-06-18
Runtime/wasm: normalize -0.0 and NaN in float32_hash

vouillon:


hhugo:
Can we have a test for NaN

PR #2341, opened 2026-06-18
Runtime/wasm (WASI): fix preopen prefix matching and empty-path resolution

hhugo:
Two WASI path-resolution fixes in `runtime/wasm/fs.wat` from the review (#2263):

**Preopen prefix with a trailing slash** — `prefix_match` required a path to start
with the preopen prefix followed by `/` (or equal it). If the prefix itself ended
in a slash (e.g. `wasmtime --dir=tmp/`), it looked for `/` at the wrong offset and
never matched, so the first filesystem operation failed. A single trailing slash
on the prefix is now ignored (only affects the previously-broken case).

**Empty path** — `make_absolute` turned an empty path into the current directory,
so `stat ""` / `open ""` etc. silently operated on the cwd. The empty path is now
left empty, matches no preopen, and `unix_resolve_path` / `caml_sys_resolve_path`
raise `ENOENT`, like native.

### Tests

`compiler/tests-jsoo/test_unix.ml` gains `stat "" → ENOENT` (js, wasm, native).
The trailing-slash case is WASI-config-specific (a `--dir=x/` preopen) so it has
no portable test.

### Verification

js + native pass; `fs.wat` assembles under the `wasi` profile. The WASI backend
can't be run here — wasmtime 45.0.2 panics on jsoo's WasmGC output
(`gc_runtime.rs:481`) — so the WASI paths need CI.

### DIFF
--- a/runtime/wasm/fs.wat
+++ b/runtime/wasm/fs.wat
@@ -200,8 +200,10 @@
       (param $path (ref $bytes)) (result (ref $bytes))
       (local $need_slash i32) (local $i i32) (local $abs_path (ref $bytes))
       (if (i32.eqz (array.len (local.get $path)))
-         (then ;; empty path
-            (return (global.get $current_dir))))
+         (then ;; empty path: leave it empty so it matches no preopen and the
+               ;; operation fails with ENOENT, like native (open "" etc.),
+               ;; rather than silently resolving to the current directory
+            (return (local.get $path))))
       (if (i32.eq (i32.const 47) ;; '/'
              (array.get_u $bytes (local.get $path) (i32.const 0)))
          (then ;; absolute path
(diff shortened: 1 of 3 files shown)


runtime/wasm/fs.wat
```diff
@@ -283,6 +285,13 @@
       (param $prefix (ref $bytes)) (param $path (ref $bytes)) (result i32)
       (local $i i32) (local $len i32)
       (local.set $len (array.len (local.get $prefix)))
+      ;; A preopen prefix may carry a trailing slash (e.g. "--dir=tmp/");
+      ;; ignore it so it still matches "<prefix>/<rest>".
+      (if (i32.and (i32.gt_u (local.get $len) (i32.const 0))
+                   (i32.eq (array.get_u $bytes (local.get $prefix)
+                              (i32.sub (local.get $len) (i32.const 1)))
+                           (i32.const 47))) ;; '/'
```
vouillon, line 293: There is a bug here: both operands of `i32.and` are evaluated.

hhugo: fixed in #2352

PR #2359, opened 2026-06-20
doc: order the manual sidebar with @children_order

balat:
Part of an effort to make the Ocsigen documentation well-formed on ocaml.org. `manual/index.mld` already is the package home page, but without `@children_order` odoc lists the manual pages in the sidebar alphabetically. Add `@children_order` so the sidebar follows the logical reading order already used in the page body (Overview, Installation, Quick Start, …, then the API reference).

Standalone `odoc compile` of `manual/index.mld` succeeds; every page name in the list maps to an existing `manual/*.mld`.

### DIFF
--- a/manual/index.mld
+++ b/manual/index.mld
@@ -1,3 +1,5 @@
+@children_order overview install quickstart javascript-interop ppx ppx-deriving rev-bindings errors lwt options compilation-modes linker tailcall effects runtime-representation browser-compat wasm_overview wasm_runtime build-toplevel debug performances contribute examples bindings library environment-variable runtime-files camlp4 api
+
 {0 Js_of_ocaml}
 
 Js_of_ocaml is a compiler from OCaml bytecode programs to JavaScript. It makes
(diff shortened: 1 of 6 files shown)


manual/index.mld
```diff
@@ -1,3 +1,5 @@
+@children_order overview install quickstart javascript-interop ppx ppx-deriving rev-bindings errors lwt options compilation-modes linker tailcall effects runtime-representation browser-compat wasm_overview wasm_runtime build-toplevel debug performances contribute examples bindings library environment-variable runtime-files camlp4 api
```
hhugo: The last items are not real pages, they should probably not be there. 
- bindings
- library
- environment-variable
- runtime-files
- camlp4

hhugo:
In ocaml.org, I see no doc, do I need to do something special ?
https://ocaml.org/p/js_of_ocaml-compiler/6.4.1
https://ocaml.org/p/js_of_ocaml/6.4.1

balat:
No, nothing to do on our side. The ocaml.org docs pipeline is broken. no package released since end of May has its docs built (see for ex dune 3.24.0). Likely still recovering from the April incident ([discuss thread](https://discuss.ocaml.org/t/missing-ocaml-org-package-documentation/17977)).

manual/index.mld
balat: you're right they're leftovers from the wiki conversion.
the legacy URLs they were meant to preserve changed scheme during the migration anyway, so they serve 
no purpose. Removed.
I fixed all the links towards these pages in tuto: https://github.com/ocsigen/tuto/pull/50/changes

PR #2362, opened 2026-06-20
doc: fix broken odoc cross-references, markup, and stale doc URLs

hhugo:
Release-prep review of the manual and READMEs. Doc-only — no `CHANGES.md` entry.

## 1. Broken cross-page section links (manual)
Four `{{!page-"X".label}…}` references to **hyphenated** section labels silently dropped their anchor — readers landed on the page top instead of the section. odoc reads the text before the last hyphen as a *kind qualifier*, so a label name containing `-` is unreferenceable (hyphen-free labels like `.conversions` work fine; verified in the generated HTML, which produced `linker.html` with no `#…`).

Fixed by renaming the four referenced labels to use `_`, so the anchors resolve:
- `linker.html#finding_runtime` (from compilation-modes ×2)
- `linker.html#writing_primitives` (from javascript-interop ×2)
- `errors.html#stack_traces` (from debug)
- `options.html#env_vars` (from environment-variable)

## 2. Invalid markup in `wasm_runtime.mld`
`{{{$}}}`, `{{{$ocaml_version}}}` etc. raised "bad markup" / "unpaired `}`". Changed to inline-code `[$]`, `[$ocaml_version]`, … matching the surrounding style.

## 3. Stale version in `install.mld`
Said "OCaml 4.13 to **5.4**"; `dune-project` pins `(>= 4.13) (< 5.6)`, so the range is **5.5**.

## 4. Stale published doc URLs (READMEs + package metadata)
The odoc/wodoc migration changed the manual URL layout from `…/<label>/manual/<page>` to `…/<label>/js_of_ocaml/<page>.html`, but the READMEs and package metadata still used the old path, which now **404s**. Repointed to the new structure, standardising on `https` and the `latest` label:
- `README.md` (tailcall), `README_wasm_of_ocaml.md` (wasm_overview), `benchmarks/README.md` (performances)
- `dune-project` `homepage`/`documentation` and the generated `*.opam` (were pointing at the dead `latest/manual/overview`)

`dune build @doc` is clean (no manual-authored warnings).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ optimized:
 - self recursive functions (when the tail calls are the function itself) are
   compiled using a loop.
 - trampolines are used otherwise.
-  [More](http://ocsigen.org/js_of_ocaml/dev/manual/tailcall) about tail call
+  [More](https://ocsigen.org/js_of_ocaml/latest/js_of_ocaml/tailcall.html) about tail call
   optimization.
 
 Effect handlers are supported with the `--effects={cps,double-translation}`
--- a/README_wasm_of_ocaml.md
+++ b/README_wasm_of_ocaml.md
@@ -28,7 +28,7 @@ OCaml 5.x code using effect handlers can be compiled in three different ways:
 
 ## Installation and usage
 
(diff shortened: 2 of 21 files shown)


hhugo:
@balat, here are some other changes for the doc.

PR #2373, opened 2026-06-25
Use Wax for runtime files

vouillon:


compiler/lib-wasm/wax_link.ml
```diff
@@ -0,0 +1,112 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ *)
+
+open Stdlib
+
+let ocaml_version =
```
hhugo: Can we move that in Ocaml_version.ml maybe

compiler/lib-wasm/wax_link.ml
```diff
@@ -0,0 +1,112 @@
...
+let with_preprocessed_files ~variables ~inputs action =
+  (* [name-wasm-functions] (default true) toggles the export-to-id naming pass
+     rather than acting as a conditional-compilation variable. *)
+  let name_functions =
+    match
+      List.find_map variables ~f:(fun (k, v) ->
+          if String.equal k "name-wasm-functions" then Some v else None)
+    with
+    | Some (Bool b) -> b
+    | Some _ | None -> true
```
hhugo: shouldn't we fail if we get `Some (String _ | Version _)` ?

hhugo (approved):
LGTM with few comments

PR #2375, opened 2026-06-25
doc: add TESTING.md for the inline-test infrastructure

hhugo:
Document the in-tree ppx_expect_light / ppx_optcomp_light / ppx_light_predicate toolchain: declaring an inline-test library, writing let%expect_test/[%expect] tests, the two gating mechanisms ([@@if]/[@@@if] compile-time vs [@when]/[@tags] run-time) and when to use which, the predicate grammar and its compile-time vs run-time atoms, running tests (js/wasm/quickjs, dune promote), and running a single file or test via the inline-test runner.


hhugo:
@vouillon, just FYI, in case you've missed by recent push

PR #2385, opened 2026-06-30
Release 6.4.1 (dune 3.24 compatibility)

hhugo:
## 6.4.1 — bugfix release

A point release on top of **6.4.0** containing **only** the dune 3.24 compatibility fixes, so 6.4.0 keeps building/testing against the new dune without pulling in the rest of `# dev`.

Branched from the `6.4.0` tag (same workflow as the 6.3.1/6.3.2 point releases — a standalone branch to be **tagged, not merged**). The PR's merge-base is `6.4.0`, so the diff below is exactly the three release commits.

### Included
- **#2371** — Runtime/wasm: fix path handling from dune variable in `args.ml` (derive wat-module name from basename, so dune 3.24's leading-`./` path form keeps producing valid `module:path` lines).
- **#2384** — Tests: pass file paths as literal args (with explicit `(deps ...)`) instead of `%{dep:...}` in the `md5`/`md5_nat` and `dump_sourcemap` tests, which echo the path verbatim.

### Release prep
- `CHANGES.md`: `# dev` → `# 6.4.1 (2026-06-30) - Lille`, with both bug-fix entries.
- `VERSION`: `6.4.0` → `6.4.1`.

### Verification
Reproduced and verified with **dune 3.24.0** (the version that broke CI on master): the `tests-io` (md5/md5_nat, all modes) and `tests-sourcemap` runtest aliases pass on this branch.

> ⚠️ Do not "Merge" — this is a release branch. Tag `6.4.1` on it (`dune-release`) and close the PR.


hhugo:
Released on opam

PR #2387, opened 2026-07-02
Doc: check documentation examples (#1020)

### LINKED ISSUE #1020: Improve API documentation with small examples
I'd like to supply the API docs with example snippets to show how the module is used. Just reading function listings is not always the best way to learn. Such snippets will have to be maintained, though, so it is a commitment to include it. What do you think?

hhugo:
Closes #1020.

Type-check the code examples embedded in the js_of_ocaml documentation — the manual **and** the library `.mli` doc-comments — so they can't silently rot as the API evolves.

## The convention (checked by default, opt out with a marker)

An `{@ocaml[ … ]}` code block in a `.mld` manual page or a `.mli` doc-comment is checked by default:

| Block | Behaviour |
|---|---|
| `{@ocaml[ … ]}` | **type-checked** against the real library |
| `{@ocaml parse[ … ]}` | parse-only (syntax) — for hypothetical identifiers / placeholder types |
| `{@ocaml skip[ … ]}` | ignored — pseudo-code, elisions, non-parsing fragments |
| `{@ocaml prelude[ … ]}` | shared setup prepended to the type unit (hide it in a plain `(* … *)` comment in a `.mli`) |
| `{[ … ]}` | never checked |

The markers are odoc metadata, stripped from the rendered output, so **readers see no difference** (all render as `<pre class="language-ocaml">`).

## How it works

`manual/examples-check/` extracts the blocks and, for the type tier, compiles them as a **library** — building the `.cma` type-checks every module but skips the final link, so JS-only `external`s still check — against the relevant library + ppx. The parse tier runs `ocamlc -stop-after parsing`. Everything runs under `@runtest`, is **compile-only** (never executed — the examples touch the DOM), and errors are reported at the original source line. Sources needing a different library/ppx get their own unit (`js_of_ocaml`, `js_of_ocaml-tyxml`, `js_of_ocaml-lwt` + `lwt_ppx`, `ppx_deriving_json`); the `js_of_ocaml` interface is globbed, so future examples are covered automatically.

## Coverage

**61 type-checked, 21 parse-only** across:

- the whole manual (`manual/*.mld`) — 50 type / 21 parse / 2 skip;
- the whole `js_of_ocaml` library interface — **9 type / 0 parse** (every example type-checked; only a commented method and a signature sketch stay `{[ ]}`);
- `js_of_ocaml-tyxml` and `js_of_ocaml-lwt` — 1 type each.

## Bugs the checker caught

Most sources only gained invisible markers, but type-checking surfaced real latent bugs, fixed here:

- **javascript-interop** — canvas `class type` used unqualified `imageElement`/`canvasElement` (which don't resolve) → qualified to `Dom_html.*`.
- **ppx page** — a property-attributes `object%js` literal named two fields that mangle to the same JS name (`also_readonly`/`also_readwrite` → `also`) → disambiguated with a trailing underscore.
- **MutationObserver / ResizeObserver** — `observe` (which returns the observer) was called in `if … then` with no `else`, i.e. in `unit` position → wrapped in `ignore`.
- **PerformanceObserver** — the example passed an extra `()` to `observe`.
- **Geolocation** — a stray `;` before an `in`.
- **Performance** — two `perf##mark` results discarded in `;` position (an error under `-strict-sequence`); `perf##measure`'s optional 2nd/3rd args passed as plain strings instead of `Js.def`.
- **Intl** — `letterSort` discarded the result of an in-place `##sort` (which returns the array) in `;` position, forcing a `unit` return that clashed with the real signature (and the `collator` was allocated per-comparison inside the callback → lifted out).

### DIFF
--- a/.github/workflows/wax.yml
+++ b/.github/workflows/wax.yml
@@ -31,7 +31,9 @@ jobs:
           ocaml-compiler: "5.4.0"
 
       - name: Install wax
-        run: opam pin add --yes wax https://github.com/vouillon/wax.git
+        run: |
+          opam pin add --yes wax-lib https://github.com/vouillon/wax.git
+          opam pin add --yes wax https://github.com/vouillon/wax.git
 
       - name: Convert Wasm runtime files to wax
         run: opam exec -- ./tools/wat2wax.sh
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -101,6 +101,14 @@
   `taggedEvent` variant, `CoerceTo.mouseScrollEvent` and
   `Event._DOMMouseScroll` (#2350)
 
+## Documentation
+* Doc: check documentation examples. An `{@ocaml[ … ]}` code block in an `.mli`
+  doc-comment or an `.mld` manual page is type-checked against the library by
+  `dune build @runtest`, so examples that no longer match the API fail the build.
+  The `{@ocaml parse[ … ]}` and `{@ocaml skip[ … ]}` markers opt out of typing
+  (parse-only) and of checking entirely; plain `{[ … ]}` blocks are never
+  checked. See `manual/examples-check/` (#1020)
+
 ## Bug fixes
 * Compiler: don't rewrite `x = e + x` into `x += e` for `+` (not
   commutative on strings), which reversed `Filename.concat` operands and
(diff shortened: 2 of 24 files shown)


hhugo:
cc @balat

balat:
The documentation code checking is interesting. I'm wondering if it could be included in wodoc, or even odoc.

hhugo:
> The documentation code checking is interesting. I'm wondering if it could be included in wodoc, or even odoc.

#1020 mentions https://github.com/ocaml/odoc/pull/303

PR #2389, opened 2026-07-03
Wasm: resolve assets relative to the runtime file, not require.main

vouillon:
The node runtime loaded the separately-compiled `.wasm` assets from `path.dirname(require.main.filename)` — the process entry point — so a program failed to find them when loaded through `require` (e.g. behind a CLI wrapper) or a symlink, and crashed outright when `require.main` was undefined (worker threads, `node -e`). Use `module.filename` instead, locating the assets relative to the runtime file itself, matching the browser branch which already self-locates via `document.currentScript.src`. `module` is already an allowed runtime global, so nothing else changes. Same fix in the WASI runtime.

### DIFF
--- a/compiler/tests-wasm_of_ocaml/wasm_asset_resolution.t
+++ b/compiler/tests-wasm_of_ocaml/wasm_asset_resolution.t
@@ -1,0 +1,20 @@
+Wasm assets are located relative to the runtime file, not the process entry
+point, so a program still finds them when it is loaded through `require` (e.g.
+behind a CLI wrapper) rather than run directly.
+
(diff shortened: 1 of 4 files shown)


compiler/tests-wasm_of_ocaml/wasm_asset_resolution.t
```diff
@@ -0,0 +1,20 @@
+Wasm assets are located relative to the runtime file, not the process entry
```
hhugo, line 1: Maybe move this to tests-wasm_of_ocaml if we're only testing wasm behavior

hhugo (approved):

PR #2391, opened 2026-07-03
Runtime: fix comparison of immediate vs custom block with a compare op

vouillon:
#2270 dropped the immediate-vs-custom path in caml_compare_val, always ordering the immediate before the block. That matches the native runtime for the built-in custom blocks (int32/int64/nativeint/bigarray), which define no compare_ext, but breaks blocks that do provide one: zarith represents small values as immediates and large ones as custom blocks, so polymorphic comparison across that boundary must call the custom op.

Restore the path, driven by a compare_ext operation (falling back to compare for runtimes like zarith_stubs_js that only register it), and mark the built-in blocks with compare_ext: null so they keep ordering the immediate before the block.


hhugo:
- why wasn't it caught ? Should we run janestreet tests in js ?
- do you plan to add compare_ext to zarith_js_stubs ? We could drop the compare fallback eventually 
- It seems to be a good time to document custom implementation in jsoo.
- how does it work in wasm ?

vouillon:
Maybe we should run the janestreet tests in js, indeed.
I can add compare_ext to zarith_js_stubs, indeed.
The Wasm runtime follows the OCaml runtime more closely and already has both compare and compare_ext.

hhugo:
I wonder if the janestreet tests should be run in a dedicated job.

PR #2394, opened 2026-07-07
CI: dedup the Jane Street test suite into a janestreet workflow

hhugo:
Follow-up to the `jane_street` job added in #2391.

The Jane Street suite was run from two places: the `jane_street` job in `js_of_ocaml.yml` and the `jane_street_tests` rows of the `wasm_of_ocaml.yml` build job, duplicating the opam-repository checkouts, the OxCaml pinning dance and the per-library test steps.

Move all of it to a dedicated `janestreet.yml` workflow with a single job parameterized by backend (js / wasm) and compiler (5.3 / OxCaml). The `JS_OF_OCAML` / `WASM_OF_OCAML` variables consumed by `ci_setup` and its generated dune-workspace are derived from the backend matrix key. The release-profile Bonsai suites — previously a dedicated whole-program wasm row that repeated the full set-up — now run unconditionally in every job, so all four configurations also exercise whole-program compilation. A failing suite no longer skips the remaining ones: the test steps run under `!cancelled()`, gated on the install step's outcome, and the job still fails if any suite fails.

`wasm_of_ocaml.yml` keeps the regular test coverage those rows provided: Ubuntu 5.3 joins the base matrix and OxCaml gets a plain include row. All remaining rows use separate compilation, so that matrix axis is gone.

Also drop the ocaml-wasm fork pins (`base`/`time_now`/`ppx_inline_test`/`ppx_expect`) entirely: they date from when the test suite was built with Jane Street's ppx_expect and needed wasm-capable forks. Since 6d6642633 the tests use the in-tree `ppx_expect_light` and no opam package in the repo depends on any of these, so the pins registered but nothing ever installed them. This is also what lets the plain OxCaml wasm row work without the Jane Street opam repository: its test dependencies resolve from the ox repository, as in the js_of_ocaml workflow.


hhugo:
@vouillon, is this based on your #2391 PR. It moves (and dedup) all janestreet jobs into a separate file. Can you take a look.

PR #2409, opened 2026-08-05
Build with Introcaml

let-def:
[Introcaml](https://discuss.ocaml.org/t/ann-introcaml-alpha-polymorphic-printing-and-introspection-for-ocaml/18407) is an extension of the OCaml compiler that instruments values to make them printable at runtime.

Currently, it is available via a dedicated opam switch:
```bash
opam switch create 5.5.0+introcaml
```

Because Introcaml makes slight modifications to the compiler and the runtime, `js_of_ocaml` requires a few adaptations (primarily stubs to ignore the new metadata):
- Ignoring a new field in `Const_block`.
- Ignoring a new `NEXT_RESERVED_BITS` instruction.
- Providing stubs for a few new runtime primitives.

This PR contains the fixes necessary to allow `js_of_ocaml` to build with Introcaml. I am not entirely sure what the ideal deployment story is here; I understand you may be hesitant to merge these changes until a path toward upstreaming Introcaml is established, but it would be great if users of the instrumented switch could still utilize `js_of_ocaml` in the meantime.


runtime/js/obj.js
```diff
@@ -340,3 +340,32 @@ function caml_int_as_pointer(i) {
   if (i === 0) return null;
   caml_failwith("%int_as_pointer is not supported in javascript.");
 }
+
+//Provides: caml_compiler_block_descs_ref
+var caml_compiler_block_descs_ref = [0];
```
hhugo, line 345: I would annotate the primitives with //If: introspect to avoid bloating the runtime. toplevel and separate compilation currently include all primitives.

hhugo:
Is there any plan to make jsoo support introspection and be able to pretty print values ?

let-def:
I did not have this in mind, but I think it can be done.
The current compilation scheme is optimized to be extremely cheap with the native backend, the same can be achieved (to a reasonable degree) in JavaScript.

The idea would be to turn metadata into array sub-classes:

```js
class MetaArrayXX extends Array {
    static desc = {name: "Some", tag: 0, args: ["Any"]};
}
```

Or (closer to the native description):

```js
class MetaArrayXX extends Array {
    static desc = 1234;
}

global_descs[1234] = {name: "Some", tag: 0, args: ["Any"]};
```

Then allocations on the JS side are updated from `[0, x]` to `new MetaArrayXX(0, x)`.
This can be quite verbose, so we can abstract both metadata injection and object allocation:

```js
function metadataFactory(desc) {
    return class extends Array {
        static desc = desc;
        static n(...args) {
            return new this(...args);
        }
    };
}
```
```js
// Module-wide definition
const metadata = metadataFactory({name: "Some", tag: 0, args: ["Any"]});
...
// Allocation: from [a, b, c] to:
metadata.n(a, b, c);
```

which can further be minified. Runtime lookup is done by inspecting `obj.constructor.desc`, if it exists.

By using a prototype, we preserve most existing JS engine optimizations, not touching objects themselves, and with a factory we avoid bloating the generated code.

hhugo:
Thanks! I rebased the branch on master and pushed two extra commits on top:

- **Compiler: expose build-time flags in Config, gate the Introcaml stubs** — adds `Config.oxcaml` / `Config.introspect` (build-time constants, replacing the ad-hoc `[@@if ...]` definitions in `Instr` and `Wat_preprocess`), and exposes `introspect` as a `//If:` flag for the JS runtime and as `$introspect` for the wasm preprocessor. The new primitive stubs in `obj.js` / `obj.wat` are gated on it, so they are only linked when the compiler was built with Introcaml. Also ran ocamlformat and added a CHANGES entry.
- **Runtime: skip Introcaml reserved header bits when unmarshaling** — Introcaml's `extern.c` emits `CODE_RESERVED_BITS` (0x1a) followed by a 32-bit value before a block whose header has reserved bits set. Both unmarshalers (`marshal.js`, `marshal.wat`) now skip that prefix the same way `intern.c` does. Without this, `compiler/tests-dynlink` failed with `input_value: ill-formed message` on the `5.5.0+introcaml` switch.

With that, `dune build @all` passes on `5.5.0+introcaml`, and the plain 5.5.0 build is unaffected. The only remaining `@runtest` diffs on the Introcaml switch are the `ppx_deriving_json` `.mlt` tests, where Introcaml's toplevel prints abstract values polymorphically (`<abstr> {write = <fun>; read = <fun>}`), which seems expected.

https://claude.ai/code/session_01BhkxFhBPTqwxXVnyXaUHbP

PR #2421, opened 2026-09-06
Keep `static` on the same line as the class element it modifies

### LINKED ISSUE #2420: [BUG] Line break after `static` in a class body breaks the generated code on Safari 17
**Describe the bug**

Generated code that links `runtime/js/int64.js` fails to run on Safari 17.x. `Int64.of_string` throws:

```
TypeError: undefined is not an object (evaluating 'MlInt64.UNSIGNED_MAX.udivmod')
```

The printer emits a breakable space after `static`, so a long line can be wrapped between `static` and the name that follows it:

```js
class bh{constructor(a,b,c){/* ... */}static
UNSIGNED_MAX=new
bh(cE,cE,cZ);static
SIGNED_MAX=new
bh(cE,cE,0x7fff);static
SIGNED_MIN=new
bh(al,al,hK);
```

JavaScriptCore 17.x does not read `static`, a line break, then `NAME = ...` as a static field: it takes `static` as a field of its own and `NAME` as an instance field. `MlInt64.UNSIGNED_MAX` is then `undefined`, and the `udivmod` call in `caml_int64_of_string` throws.

JSC 18.5 parses it correctly, so this is a JavaScriptCore bug rather than a spec violation on your side -- Node parses the same snippet as a static field. But [`ECMASCRIPT.md`](https://github.com/ocsigen/js_of_ocaml/blob/master/ECMASCRIPT.md) lists Safari 13.4+ among the supported engines, and Safari 17 is inside that range.

Rewriting `static` + newline to `static` + space in the affected bundle was enough to make the program run on iOS 17.5, and the numerical result agreed with a desktop run.

I do not have an OCaml toolchain on this machine, so I am reporting from a shipped artifact rather than from a build of my own: the `stanc.js` released by [stan-dev/stanc3](https://github.com/stan-dev/stanc3) has three occurrences, the `UNSIGNED_MAX` / `SIGNED_MAX` / `SIGNED_MIN` fields of `runtime/js/int64.js`. Any output that links int64.js and happens to wrap at that column should show the same thing.

**Expected behavior**

`static` stays on the same line as the class element that follows it.

`js_output.ml` already uses `PP.non_breaking_space` for this class of hazard -- `return`, `throw`, `async`, `using`, `await using`. The four class-element cases use `PP.space` instead:

- `CEMethod` -- https://github.com/ocsigen/js_of_ocaml/blob/27e40ddaacc2a065ae2dfd880482b753a442d45d/compiler/lib/js_output.ml#L2125-L2128
- `CEField` -- https://github.com/ocsigen/js_of_ocaml/blob/27e40ddaacc2a065ae2dfd880482b753a442d45d/compiler/lib/js_output.ml#L2134-L2137
- `CEAccessor` -- https://github.com/ocsigen/js_of_ocaml/blob/27e40ddaacc2a065ae2dfd880482b753a442d45d/compiler/lib/js_output.ml#L2152-L2155
- `CEStaticBLock` -- https://github.com/ocsigen/js_of_ocaml/blob/27e40ddaacc2a065ae2dfd880482b753a442d45d/compiler/lib/js_output.ml#L2171-L2172

Switching those to `PP.non_breaking_space` looks like it would be enough, if you think it is worth guarding against. Happy to send a PR if you would like one.

**Versions**

Reported from `stanc.js` as released by stan-dev/stanc3; the artifact does not record which js_of_ocaml built it. The `PP.space` pattern is present on master (`27e40dd`).

Possibly related: #1906 covered the ES6 compatibility of these same three fields. This is a different problem -- Safari 17 does support ES2022 static fields, and only the line break defeats it.

habakan:
Fixes #2420.

One correction to the issue: I described this as a line that "can be wrapped". It is not occasional. In compact mode `PP.space` sets the pending separator to `"\n"` while `PP.non_breaking_space` sets it to `" "`, and a separator is always required between `static` and an identifier -- so every static class element is emitted with `static` alone on its line.

Safari 17 reads that as a field named `static` followed by a separate instance field, so `MlInt64.UNSIGNED_MAX` is undefined and `Int64.of_string` throws. Safari 18.5 parses it correctly, so this is a JavaScriptCore bug rather than a spec violation -- but `ECMASCRIPT.md` lists Safari 13.4+ as supported.

This adds `static` to the set of keywords `js_output.ml` already prints with `non_breaking_space` (`return`, `throw`, `async`, `using`, `await using`), in all four class-element cases.

One expectation moved: in pretty mode a static *method* was being split across two lines, so `js_parser_printer.ml` now reads `static bar(){`. The static field and static block cases in that same block already fitted on one line and are unchanged.

Checked with `let () = print_endline (Int64.to_string (Int64.of_string "0x7fffffffffffffff"))`: before, the generated JS has three `static`-then-newline fields; after, none, and the program prints the same value under node.

On `make tests` I get the same set of failures before and after this change, so it introduces none -- but I should say that my environment cannot run the whole suite (`graphics` wants xquartz here, so some packages are missing and a few cram tests cannot resolve `js_of_ocaml`). CI will be the real check. `make fmt` reports nothing.

First contribution here -- happy to close this if you would rather not guard against a browser bug that Apple has already fixed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,11 @@
   for the static evaluator and for `Sys.getenv` lookups at runtime (#2415)
 
 ## Bug fixes
+* Compiler: keep `static` on the same line as the class element it modifies.
+  In compact mode the separator was a newline, and Safari 17 reads `static`
+  alone on a line as a field name rather than a modifier, so every static
+  class field became an instance field; `Int64.of_string` then threw on
+  `MlInt64.UNSIGNED_MAX` (#2420)
 * Compiler/Wasm: sourcemaps were silently disabled on Windows: the detection
   of Binaryen's sourcemap support used Unix redirection syntax, so it always
   concluded that sourcemaps were unsupported (#2418)
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -2125,7 +2125,7 @@ struct
             if static
             then (
               PP.string f "static";
-              PP.space f);
+              PP.non_breaking_space f);
             method_ f class_element_name n m;
             PP.end_group f
         | CEField (decorators, static, n, i) ->
@@ -2134,7 +2134,7 @@ struct
             if static
             then (
               PP.string f "static";
-              PP.space f);
+              PP.non_breaking_space f);
             class_element_name f n;
             (match i with
             | None -> ()
@@ -2152,7 +2152,7 @@ struct
             if static
             then (
               PP.string f "static";
-              PP.space f);
+              PP.non_breaking_space f);
             PP.string f "accessor";
             PP.space f;
             class_element_name f n;
@@ -2169,7 +2169,7 @@ struct
         | CEStaticBLock l ->
             PP.start_group f 0;
             PP.string f "static";
-            PP.space f;
+            PP.non_breaking_space f;
             block f l;
             PP.end_group f);
         if not last then PP.break f);
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -2095,6 +2095,14 @@ struct
             PP.space f);
         PP.end_group f
 
+  and static_keyword f static =
+    if static
+    then (
+      PP.string f "static";
+      (* [static] is not a restricted token, but JavaScriptCore up to Safari 17
+         reads a line break after it as a field named [static] *)
+      PP.non_breaking_space f)
+
   and class_declaration f i x =
     PP.start_group f 0;
     decorator_list f x.decorators;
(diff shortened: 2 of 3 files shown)


hhugo:
Thanks for the report and fix.

- I think this is a reasonable change. It took me a while to figure out what the spec says in that case, but I'm now mostly convinced it's a bug in jsc.
- Can you add a comment in the code explaining why we use `non_breaking_space` even though `static` is not a restricted token.

habakan:
Thanks. Added the comment next to the `non_breaking_space` in the method case, explaining that `static` is not a restricted token but JavaScriptCore up to Safari 17 reads a line break after it as a field named `static`. I kept it in one place, like the Chrome comment in `last_semi`, rather than repeating it in all four class-element cases — happy to put it on each if you would prefer.

https://github.com/ocsigen/js_of_ocaml/blob/2a8d9e7806f4a8bcc4cb8d4e1b1337935ddbdf07/compiler/lib/js_output.ml#L1474-L1479

PR #2430, opened 2026-09-11
Introcaml: support runtime introspection in JavaScript

hhugo:
Prototype of the design sketched by @let-def in https://github.com/ocsigen/js_of_ocaml/pull/2409#issuecomment-5234576840. Built on top of #2409: only the last commit is new, the others are that PR (rebased on master).

@let-def, could you take a look?

## What it does

- **IR**: `Block` allocations and `Tuple` constants carry the descriptor index (the reserved header bits). `NEXT_RESERVED_BITS` sets it for the following `MAKEBLOCK*`/`MAKEFLOATBLOCK`; constants read it with `Obj.get_reserved` from the DATA section / `Const_block`.
- **JS codegen**: a block with a descriptor is allocated as `new desc_xx(tag, ...)`, an instance of an `Array` subclass created by `caml_block_class(desc)` and declaring `static desc = desc`. Objects themselves are untouched (plain arrays with a different prototype), and `caml_obj_get_reserved` reads `obj.constructor.desc`. One class variable per descriptor, shared across the compilation unit and initialized at its start. The constructor always gets at least two arguments (tag + a field), so it cannot be mistaken for `new Array(length)`.
- **Descriptors**: the `BDSC` section (whole program) or `cu_block_descs` (separate compilation, one per unit) are embedded as a constant and registered with `caml_register_block_descs`, which prepends them to the list returned by `caml_compiler_block_descs`, so `Introspect.Index.self_index` picks them up incrementally. `caml_read_bdsc_section` returns `[]`.
- **Flag**: everything is behind `--enable/--disable introspection`, enabled by default only when the compiler is built with Introcaml. The compiler test harness disables it so snapshots do not depend on the switch; `compiler/tests-compiler/introspect.ml` and `compiler/tests-jsoo/test_introspect.ml` (both gated by `[@@@if introspect]`) cover it.
- **Wasm**: no support, stubs kept. The `caml_compiler_block_descs` stub returned a block of size 0, which trapped as soon as `Introspect` dereferenced it; it is now a proper reference cell.

## Result

With the `5.5.0+introcaml` switch, `Introspect.Print.print_any_endline` on options, lists, records, variants, polymorphic variants, results, refs and closures prints the same thing under node as the native binary. The descriptor hashes match, so `Hashtbl.seeded_hash_param` is compatible between the two runtimes for these values.

Two differences remain, both on the Introcaml side (worth fixing there):

- **Float arrays / all-float records print half their elements**: `Introspect.Dyn` computes `double_to_wo_shift` from `Sys.word_size` (32 in JS) and halves `Obj.size`, but in JS `Obj.size` of a float array is already the number of floats. Something like `Obj.size (Obj.repr [|0.|]) - 1` would be portable.
- **Whole floats print as ints** (`0.` → `0`): JS cannot distinguish them, and descriptors only carry approximations for immediates. A `Float` approximation for float fields would fix it for records/tuples.

Also noticed while testing on wasm: `Introspect` calls `Obj.size` on floats (in `find_tag`), which traps in wasm_of_ocaml; not addressed here.

## Testing

On `5.5.0+introcaml`: the compiler, jsoo, ocaml, full, dynlink and lib test suites pass in JS; the jsoo suite passes on wasm. On plain 5.5.0: compiler and jsoo suites pass, generated code unchanged.

https://claude.ai/code/session_01BhkxFhBPTqwxXVnyXaUHbP

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@
   extended `Const_block`, skip the reserved header bits in marshaled data, and
   provide stubs for the new primitives, gated by the `introspect` flag in
   `//If:` directives and wasm `(@if $introspect ...)` blocks (#2409)
+* Compiler/Runtime: with Introcaml, support runtime introspection
+  (`Introspect.Print` and friends) in JavaScript: blocks carrying a
+  descriptor are allocated as instances of an `Array` subclass holding the
+  descriptor index, and the block descriptors of the program are embedded;
+  can be disabled with `--disable introspection` (#2409)
 * Lib: add `WebGL2` — bindings to the WebGL2 rendering context. The context
   inherits every method and constant of `WebGL`, and adds the WebGL2 objects
   (vertex array objects, queries, samplers, syncs, transform feedback), 3D and
--- a/compiler/lib-wasm/gc_target.ml
+++ b/compiler/lib-wasm/gc_target.ml
@@ -1008,7 +1008,7 @@ module Constant = struct
   let rec translate_rec c =
     match c with
     | Code.Int i -> return (Const, W.RefI31 (Const (I32 (Targetint.to_int32 i))))
-    | Tuple (tag, a, _) ->
+    | Tuple (tag, a, _, _) ->
         let* ty = Type.block_type in
         let* l =
           Array.fold_left
@@ -1097,7 +1097,7 @@ module Constant = struct
     | Float32 f ->
         let* e = Memory.make_float32 (return (W.Const (F32 (Int64.float_of_bits f)))) in
         return (Const, e)
-    | Float_array l ->
+    | Float_array (l, _) ->
         let l = Array.to_list l in
         let* ty = Type.float_array_type in
         (*ZZZ Boxed array? *)
--- a/compiler/lib-wasm/generate.ml
+++ b/compiler/lib-wasm/generate.ml
@@ -1498,7 +1498,7 @@ module Generate (Target : Target_sig.S) = struct
           | Pv indices -> Some (indices, ctx.global_flow_info.info_defs.(Var.idx indices))
           | Pc _ -> None )
       with
-      | Bigarray { kind; layout; _ }, Some (indices, Expr (Block (_, l, _, _))) ->
+      | Bigarray { kind; layout; _ }, Some (indices, Expr (Block (_, l, _, _, _))) ->
           Some
             ( kind
             , layout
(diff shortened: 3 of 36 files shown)


let-def (approved):
The code looks good to me.

let-def:
A few things I think are worth looking for before considering this for merge:

1) Performance impact. The compilation scheme we choose should lend itself nicely to VM optimizations, but it is worth checking that this is true in practice.

2) Increase in generated code size. The scheme as it is implemented makes object allocation more verbose; since this operation is very common, I am afraid it could lead to an observable bloat of generated code. Luckily, it is easy to tweak, so I don't think we should worry about this.

3) Integration with JS environments: should we add whatever is necessary to `devtoolsFormatters` to get introspection-driven pretty-printed objects in the browser?

hhugo:
### Upstream (Introcaml) patches needed for float blocks

Two limitations found while testing the DevTools formatter are in Introcaml itself, not in js_of_ocaml. Until they land, float arrays and all-float records print incompletely (`[|1.5; 2.5; 3.5|]` shows as `[|1.5|]`, and `{fa = 1.5; fb = 2.5}` is not recognized as a record).

**1. `Introspect` halves the size of float blocks on 32-bit word size** (`stdlib/introspect.ml`, `Dyn.double_to_wo_shift`, used by `fields_of_block` and `find_tag`). It assumes a float takes two words, but js_of_ocaml (and wasm_of_ocaml) report a 32-bit `Sys.word_size` while `Obj.size` already counts elements. Suggested fix:

```ocaml
let double_to_wo_shift =
  match Sys.backend_type, Sys.word_size with
  | Other ("js_of_ocaml" | "wasm_of_ocaml"), _ -> 0
  | _, 64 -> 0
  | _, _ -> 1
```

(or, backend-agnostic, count with `Array.length (Obj.obj obj : float array)` in both places.)

**2. `Lambda.Const_float_array` carries no `block_desc`**, unlike `Const_block`. A constant all-float record (`{fa = 1.5; fb = 2.5}`, all fields literals) is compiled to that constant, so it is emitted without reserved bits and no descriptor can be recovered for it; float blocks allocated by `MAKEFLOATBLOCK` are fine. js_of_ocaml is already plumbed for it: `Code.Float_array` constants carry a descriptor and are allocated through the descriptor class when the reserved bits are present (see the `float blocks` test in `compiler/tests-jsoo/test_introspect.ml`, which documents the current behavior).

PR #2437, opened 2026-09-18
Runtime: specialize direct partial applications in double-translation mode

vouillon:
Avoid an extra `caml_call_gen_direct` frame for one- and two-argument partial applications in double-translation mode.


hhugo:
@vouillon, we should add some benchmark with the different strategies we have so that we can spot regression and improvements.

